PR #28, opened 2013-06-18
Emotiv builds on linux and mac

fraimondo:
Adds support for building in 32 and 64 bits linux and 64 bits mac.

### DIFF
--- a/realtime/src/acquisition/emotiv/Makefile
+++ b/realtime/src/acquisition/emotiv/Makefile
@@ -12,7 +12,6 @@ MACHINE = $(shell uname -m)
 
 FIELDTRIP = ../../../..
 FTBUFFER = $(FIELDTRIP)/realtime/src/buffer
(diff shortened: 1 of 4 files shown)


robertoostenveld:
sorry for this taking ages to get resolved. At the time you suggested it, we were not using git yet (but only maintaining a copy of the code on github). As of a year or so, we are using git, branches, PRs, etc. But I don't have an emotiv hardware (nor the SDK) and did not have much time to look into compilation on  other platforms. However, right now I am revisiting the build process of all realtime code, including this. 

The updated code is presently in https://github.com/robertoostenveld/fieldtrip/tree/bug3265-win64 (and will be merged into master soon) and discussed on http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3265. Would you still be interested to have a look at the code and see whether we can improve on the emotiv support?

PR #33, opened 2014-02-19
git pull request test, some groundwork for bug2478

arbackus:
Code does not work yet. Problem: When calling the "return" function in
ft_preamble, the caller is not aborted - may have to do with the nested
structure of ft_preamble.


oostenveld:
this has been resolved through svn/bugzilla, see http://bugzilla.fcdonders.nl/show_bug.cgi?id=2478

PR #34, opened 2014-02-26
Plexonv2

iandol:
Nothing much seems to be happening on http://bugzilla.fcdonders.nl/show_bug.cgi?id=1795 so I'll make this an official pull. At the moment .plx files use the old inefficient plx reading code, only the new .pl2 format which FT can't support. But even for .plx files, the new code using the official Plexon SDK is many orders of magnitude faster and feature rich. This code is slightly modified from Thomas' original patch, in that currently FT cannot load most modern PLX/PL2 files as they potentially contain both 40kHz sampled spike data and 1kHz sample LFPs in the same file. Part of this problem is a simple bug in where the `if any(fsample~=fsample(1))` gets checked, in the original patch it checks all channels even if they're empty, in my version it waits and only checks the non-empty channels with data. Pulling this shouldn't cause any problems for those who are still using the legacy code, you have to specify `plexon_plx_v2` before any of this gets used...

I really wish there was more robust support for files with more than 1 sampling frequency, I've had to build my own plexon file support, and this seems an unnecessary limitation for FT (why can't you have a cfg setting for this?). But that is another matter...

### DIFF
--- a/fileio/ft_read_header.m
+++ b/fileio/ft_read_header.m
@@ -1657,7 +1657,7 @@ switch headerformat
     end
     fsample = [orig.SlowChannelHeader.ADFreq];
     if any(fsample~=fsample(1))
-      warning('different sampling rates in continuous data not supported, please select channels carefully');
+      error('different sampling rates in continuous data not supported');
     end
     for i=1:length(orig.SlowChannelHeader)
       label{i} = deblank(orig.SlowChannelHeader(i).Name);
@@ -1720,7 +1720,7 @@ switch headerformat
 	fsample = [orig.SlowChannelHeader.ADFreq];
 	fsample = fsample(chansel); %select non-empty channels only
 	if any(fsample~=fsample(1))
-      warning('different sampling rates in continuous data not supported, please select channels carefully');
+      error('different sampling rates in continuous data not supported');
     end
     % only the continuous channels are returned as visible
     hdr.nChans      = length(label);


jens-k:
What is the status on this one? I tried to follow http://www.fieldtriptoolbox.org/getting_started/plexon/ and read in Plexon pl2 (!) files using plexon_plx_v2.

Using ft_preprocessing, ft_read_spike, ft_read_header with
```
cfg.dataformat = 'plexon_plx_v2';
cfg.headerformat = 'plexon_plx_v2';
```
got me: 
`unsupported header format "plexon_plx_v2"`

Reading it in without specifying the format yields
`unsupported header format "unknown"`

Defining the format as 'plexon_plx' works but takes forever and requires more than two orders of magnitude more memory than the original file size (my 64 GB RAM wasn't enought to read in a 400 MB file). I didnt really see a difference whether it was spiking or LFP data.

What should I do? Are pl2 files generally not supported? Should I use the plexon sdk to read in the data and then turn it into a Matlab structure by hand? In that case, I could amend the tutorial and add some code snippet to help people get going.

schoffelen:
Hi Jens, I guess that nobody ever took this PR to the finish line. Thanks for checking in. I don't know whether pl2 files generally are (or once upon a time were) supported. 
In reading the history and the associated bug (now hosted on bugzilla.fieldtriptoolbox.org) it seems that the idea was to have people indeed have the sdk installed, and then with the code in this PR merged, they could use 'plexon_plx_v2' as a fileformat. Would you be interested in following this up?

jens-k:
Will give it a try!

jens-k:
That brings us to the question, should I attempt to support more than one sampling frequency in one ft structure? 

Or should I discard spike data if LFP data is present (since that is more fieldtrip-like data) and only load spike data if its the only data present?

robertoostenveld:
ft_read_data should only return LFPs, and only of a single sampling frequency (if there are multiple, you'll have to pick one; see e.g. EDF for an implementation of that). 

spikes from the same file could in principle be read in by https://github.com/fieldtrip/fieldtrip/blob/master/fileio/ft_read_spike.m which returns a totally different data representation.

jens-k:
Ok, I only had to make minimal adjustments to iandol's code to make it work.
https://github.com/jens-k/fieldtrip/tree/plexon_pl2

ft_read_data only reads lfps, ft_read_spike only read spikes. For now, I changed ft_read_header in a way that LFPs have precedence and *only* info on LFPs is returned. Only if there are no analog channels, header info on spikes is returned.

One implication of this is that if someone has data containing lfps and spikes and calls ft_read_spike, that function calls ft_read_header and only gets the lfp header. For now, this is only a cosmetic problem since ft_read_spike only uses the original plexon header (hdr.orig) and discards the rest. Not elegant, though.

Also ft_read_events calls ft_read_header and uses the returned hdr.Fs to reconstruct samples from timestamps. If LFP data is present, events are therefore given with the LFP sampling frequency, otherwise assuming spike sampling frequency. That might be intended or not (?).

Alternatively, 
a) ft_read_header could always return info on both LFP and spike data (if present) and the calling function has to deal with it, or 
b) we introduce datatypes 'plexon_plx_v2_ad' and 'plexon_plx_v2_spike' (or something like that) and allow/force the user to choose.

What do you think?

jens-k:
Thats the current output for my test data.
- 'lfp' has **both** lfp and spikes (but spike channels are empty)
- 'ua' has only spikes

I can write a proper test once I find the time to reduce these data sets - or is there a small dataset with both spikes and LFP that I could use?

By the way, this is orders of magnitude faster and less memory consuming than before. I could not load one dataset on my office pc before (64 GB RAM), I am now handling several datasets on my laptop.

```
hdr = ft_read_header(lfp);
% hdr = 
% 
%   struct with fields:
% 
%                     Fs: 1000
%               nSamples: 6630563
%     TimeStampPerSample: 40
%                  label: {32×1 cell}
%                 nChans: 32
%            nSamplesPre: 0
%                nTrials: 1
%                   orig: [1×1 struct]
%               chantype: {32×1 cell}
%               chanunit: {32×1 cell}

hdr = ft_read_header(ua); 
% hdr = 
% 
%   struct with fields:
% 
%              Fs: 40000
%        nSamples: 265224166
%           label: {128×1 cell}
%          nChans: 128
%     nSamplesPre: 0
%         nTrials: 1
%            orig: [1×1 struct]
%        chantype: {128×1 cell}
%        chanunit: {128×1 cell}

dat = ft_read_data(lfp);
% dat =
% 	32x6630563 double
	
spk = ft_read_spike(ua);
% spk = 
% 
%   struct with fields:
% 
%         label: {1×128 cell}
%     timestamp: {1×128 cell}
%          unit: {1×128 cell}
%      waveform: {1×127 cell}
%        dimord: '{chan}_lead_time_spike'

events = ft_read_event(ua);
% events = 
% 
%   1×16671 struct array with fields:
% 
%     type
%     sample
%     timestamp
%     value
%     offset
%     duration
%
% events(10).sample
% 
% ans =
% 
%       361776

events = ft_read_event(lfp);
% events = 
% 
%   1×16671 struct array with fields:
% 
%     type
%     sample
%     timestamp
%     value
%     offset
%     duration
% 
% events(10).sample
% 
% ans =
% 
%         9045
```

robertoostenveld:
Let me give a quick answer. The user should never see spike details in hdr, although ft_read_spike can call ft_read_header under the hood. It would be fine if spike details (channel names etc) remain hidden in the hdr.orig. The design is that ft_read_header/data/event are a triplet for continuous data, and ft_read_spike is separate of those and represents point-process data. The two are linked through timestamps.   

In the example with `ua`, the hdr should have nChans=0 and nSamples=0.

iandol:
Hi, I can try to give you a test PL2 file with LFP and spike data if it would help? I haven't used the code in this branch for a couple of years, but I'm happy to help if I can.

robertoostenveld:
@iandol yes, some data would help. Please see http://www.fieldtriptoolbox.org/faq/how_should_i_send_example_data_to_the_developers/

jens-k:
@iandol Yes that'd be good. Thanks!

jens-k:
> In the example with `ua`, the hdr should have nChans=0 and nSamples=0.

@robertoostenveld 
Ok, the result is a bit unelegant but it works (other parts of the code require or add a few dummy fields):

```
hdr = ft_read_header(ua); 
% hdr = 
% 
%   struct with fields:
% 
%            orig: [1×1 struct]
%           label: {0×1 cell}
%          nChans: 0
%              Fs: []
%        nSamples: 0
%     nSamplesPre: 0
%         nTrials: 0
%        chantype: {0×1 cell}
%        chanunit: {0×1 cell}
```

That also works well for the higher-level functions. The only oddity is that if ft_read_event is called on pure spike data it returns proper events but only as timestamps without samples (that seems ok). Also, events are sorted by their type in that case (maybe because it cannot sort them by sample?).

jens-k:
@iandol  Hey Ian, any progress on that data? I would test it right away. Thanks!

iandol:
@jens-k — sorry my data was on a backup disk and took a bit of searching to find and verify:

https://www.dropbox.com/s/1mhbl4wvb7zhm6r/test01.pl2?dl=0

This is the PL2 file, the channel mapping is as follows:

Channels

The data was recorded with a tetrode, and so there are 4 channels for each data source, and there are SPK (online sorted spikes, only CH3 TETSPK09 has data I think), WB (TETWB09 on CH6 wideband for the tetrode only), LFP (TETFP09 on CH6, again from the tetrode) sources, plus some extra empty FP channels and the event markers. All my data is from tetrodes, I don't have single electrode data sorry, but this is what I used to use my forked code on (I haven't tried to load PL2s into Matlab for ages)...

jens-k:
Hey @iandol , thanks for sharing your data! I looked into it and yeah, this is a complex dataset. Reading it in with my code fails and I believe it should fail when following Robert's design guidelines. There are some LFP channels with 40k Hz sampling rate and others with 1k.

One could try to read in only one type of channel (e.g. only the ones with the lowest sampling frequency since these are most likely to be traditional LFP channels) but it feels like the user should take care of this before reading this in instead.

When I fake identical sampling rates I can read the data in, so that seems to be the only problem.

I'll ask people in my lab to also test the code with their single-sampling-rate data and will comment here again once it seems to work.

iandol:
All the LFP data **is** 1000Hz (FPxxx), the 40KHz channel is the "source" wideband data (WB) and the spike waveform data (SPK, discontinuous data samples taken around the spike time at 40KHz).

This data is really pretty standard for Plexon recordings apart from the tetrode input. Plexon recording systems are _expensive_ precisely because they are designed to record wideband, spike waveform, spike event, LFP and event data simultaneously. I'd be surprised that there were any datasets where a user threw away the original data and only stored downsampled LFP data and online sorted spike events, you want to be able to re-sort your spikes and that means you should always keep your WB data.

My solution in 2014 was to write my own handling code, and it seems that this is still going to be necessary, but perhaps @robertoostenveld has a more elegant solution?

jens-k:
I think the idea here is that you do spike sorting first, get rid of all high sampling rate data and only then read it into fieldtrip. Of course Fieldtrip could do the last step for the user too. I am surprised the 40k Hz channels are categorized as 'slow channels' but nevertheless one could sort them out based on their Fs,, e.g. by keeping only channels with the lowest sampling frequency.

If I do so I can read in your data @iandol , yielding 16 channels sampled with 1000 Hz ('lfp_TETFP09'-'lfp_TETFP12' and 'lfp_FP21'-'lfp_FP32'). 
Reading in spikes yields 20 spike channels ('mua_TETSPK01'-'mua_TETSPK20').
Does that make sense?

(code at: https://github.com/jens-k/fieldtrip/tree/plexon_pl2

iandol:
Yes, if a user manually preprocesses the PL2 they can remove the wideband channel and any spike waveform data, re-sort the spikes using e.g. offline sorter. It would nevertheless be nice to not have to do this manual step if you just want to access the LFP, and that requires being able to select the lower sample rate continuous channels during loading.

schoffelen:
I haven't followed in detail, but following up on Ian's last comment: the EDF file format may serve as an example here. EDF files may contain channels sampled at different rates, and if the user does not specify which channels are to be read, the reading functions return (in the header information, as per ft_read_header) only the channels which are sampled with the majority vote sampling rate. Perhaps you could build in some glue-code in ft_read_header that 1) only selects the 'cont' type channels for further processing, and 2) then uses the majority Fs (by default). This would of course only work if the 'proper' LFPs are in the majority (as compared to the 40kHz cont-channels)

jens-k:
As part of next week's NWB hackathon, Tara van Viegen, Stefen Buergers and I are gonna attempt first steps on integrating NWB into Fieldtrip.

If anyone has any comments, tips, etc. please let us know. We have a preliminary plan on our project description page:
https://neurodatawithoutborders.github.io/nwb_hackathons/HCK08_2020_Remote/projects/fieldtrip_integration/

schoffelen:
see #721

jens-k:
Ah sorry, I had two tabs open and posted this in the wrong one. Nevermind.

I am still on this one, too. Still waiting for a colleague to run some tests on their data.

jens-k:
Just to clarify, in @iandol 's  code the field .unit is used to assign spikes for each channel a unit. The field is a cell array with as many entries as there are labels (= channels).

That is not the way .unit is used in the tutorials or how it is defined in ft_datatype_spike, right? While I find it a good solution, another would be to have a field .channelmapping that maps the labels (here: units) onto channels. This might be even better because it allows - in addition to several units per channel - also several channels per units (which might become more common since probes are becoming more dense.

I assume this should not break any of the spike-related functions because they operate on the basis of the labels (= units). Is that correct?

(This came up while working on https://github.com/fieldtrip/fieldtrip/issues/721 with @sbuergers and  @TaravanViegen .)

jens-k:
@schoffelen 
>  Feel free to reopen the issue and self assign + change its label.

It seems I don't have the rights to do that.

There will be code coming soon for both issues, though. We will probably go forward with the channelmapping field for now. Please let us know if that's ok.

schoffelen:
I don't have too detailed historical insights into the what and why of spike data. This was mostly @martinvinck Martin's input, in combination with @robertoostenveld

PR #44, opened 2014-06-25
ft_trialfun now exits if ft_read_event fails and rethrows the error.

thht:
if ft_read_event would fail with an error, ft_trialfun_general would continue running without even showing the error. i suggest that in the case of ft_read_event throwing an error, the execution of the trialfun should be stopped and the error propagated to the user.


robertoostenveld:
Hi Thomas, This has been waiting for a long time. 

I solved it slightly differently, see https://github.com/fieldtrip/fieldtrip/commit/c9f852f4791a69ef6b807d6deaff3969683f1aa3. 

We just switched over completely to git, so pull requests are now easy to handle and should be merged quickly

PR #55, opened 2015-02-05
bugfix: ft_prepare_sourcemodel does not update 'inside' field with tight grid

nno:
The order of updating grid.dim and grid.inside seems wrong, because booleanfields does not just return the fieldnames of elements with boolean datatype, but also requires that these elements have prod(grid.dim) elements.

### DIFF
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -669,12 +669,13 @@ if strcmp(cfg.grid.tight, 'yes')
   grid.xgrid   = grid.xgrid(xmin_indx:xmax_indx);
   grid.ygrid   = grid.ygrid(ymin_indx:ymax_indx);
   grid.zgrid   = grid.zgrid(zmin_indx:zmax_indx);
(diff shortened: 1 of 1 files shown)


oostenveld:
Hi Nikolaas,

This issue was due to commit 245ce19474385a1e0435f1baf896c12d5d453a2e that I did yesterday. I was aware that it would leave some things in a broken state and disabled the automatic in-house and ftp release mechanism.  

I have worked on this (and other) issues in the mean time. Can you check whether my commit 75f7493ef8257104d894e8b86e7497dc34eb8368 solves it for you?

oostenveld:
see also http://bugzilla.fcdonders.nl/show_bug.cgi?id=2796

nno:
Hi Robert,

indeed your commit https://github.com/fieldtrip/fieldtrip/commit/75f7493ef8257104d894e8b86e7497dc34eb8368 resolves it, thanks.
I will close the associated bug report at http://bugzilla.fcdonders.nl/show_bug.cgi?id=2827

thanks,
Nick

PR #56, opened 2015-04-30
Add support for AnyWave HDF5 format

bcolombet:
Hello all,
I'm the developer of AnyWave (http://meg.univ-amu.fr/wiki/AnyWave).

This software is designed to visualize and proceed multi modal data (EEG, SEEG, MEG).
AnyWave allows exporting data to HDF5 format which we would like FT to be able to read.

A  common use could be to mark events in AnyWave (seizures, spikes, ...) and then import data and events in Fieldtrip for further processing.

### DIFF
--- a/fileio/ft_filetype.m
+++ b/fileio/ft_filetype.m
@@ -1198,6 +1198,10 @@ elseif filetype_check_extension(filename, '.csv')
   type = 'csv';
   manufacturer = 'Generic';
   content = 'Comma-separated values, see http://en.wikipedia.org/wiki/Comma-separated_values';
+elseif filetype_check_extension(filename, '.ah5')
+    type = 'AnyWave';
+    manufacturer = 'AnyWave, http://meg.univ-amu.fr/wiki/AnyWave';
+    content = 'MEG/SEEG/EEG data';
 end
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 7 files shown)


oostenveld:
Hi Bruno,

Thanks for the clean patch, no problem adding this!

I fetched the patch and merged it into our primary SVN version of the code. The github version is only a read-only copy, hence the detour. It now shows up in github as 6d006aab3a4d76da2e635114617d4a05d30e87ea. 

See http://www.fieldtriptoolbox.org/faq/why_is_fieldtrip_maintained_in_svn_and_not_in_git in case you wonder.

Could I get some test files (not too large) for reference and to add to our testing framework?
 
Is it an idea/possibility to also interface the two the other way around, i.e. use ft_write_data to export to *.ah5 files?

cheers.
Robert

bcolombet:
Hi Robert,

note that the March 2015 version of AnyWave does not fully support .ah5 format yet.
The incoming May 2015 will do.
I've tested the FT functions with my current developer version and also with the help of a regular FT user.

Next week, Mac, Windows and Linux binary versions will be available, so it will be possible to export data to .ah5 for testing. Meanwhile I could provide some already exported data in .ah5 for testing too.

I'm ok to write export support from FT to .ah5, no problem but maybe I will need your help!
We also lack the full support for MEG data (Gradians, and physical parameters),

Thanks again.

oostenveld:
no problem to wait a bit and then export some of my own data to ah5. I downloaded AnyWave and was able to look at some CTF data just fine. Doing a comparison of the original data (the tutorial Subject01.ds) and the converted format would actually be ideal for testing, as it would also pick up hard-to-detect errors (such as 0/1 sample offset and calibration errors).

Regarding export: we don't write MEG data with ft_write_data, only EEG (and similar) with minimal header details (and also relatively little event info at the moment).  So the ambitious for the writer would not have to be very high. If you look at the ft_write_data function, you will see how simple/limited it actually is.

PR #57, opened 2015-05-19
Fixed bug in ft_read_event. This function would fail for neuromag fif files without event channels

thht:
Does what the title says. The patch looks quite messy, but all it does is adding an if statement in line 1313 that checks, whether trigger is empty. Only if it is not empty, the following code is executed. The mess comes from the added indentation.


oostenveld:
Thanks. I copied the change over in SVN and committed it.

mac011> svn commit 
Sending        fileio/ft_read_event.m
Transmitting file data .
Committed revision 10414.

PR #58, opened 2015-06-06
nlx event file reader bug

sin-mike:
I found a mismatch between timestamps and sample number while reading nlx event file in case of non-continuous (e.g. with pauses) data acquisition. The 8724696 commit is the issue. I think I managed to solve it using mapping from timestamp to sample number acquired from data file.
BR

Mike from KFU

### DIFF
--- a/fileio/ft_read_event.m
+++ b/fileio/ft_read_event.m
@@ -1497,7 +1497,7 @@ switch eventformat
       hdr = ft_read_header(filename);
     end
     % the event file is contained in the dataset directory
-    if     exist(fullfile(filename, 'Events.Nev'))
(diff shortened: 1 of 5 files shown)


oostenveld:
Hi Mike,

Gaps in nlx files are not supported by design in the high level code. The mapping of timestamps to samples becomes too inaccurate or too slow. Furthermore, different channels can have different gaps, which was especially observed in 256 channel systems. 

Did you look at http://www.fieldtriptoolbox.org/getting_started/neuralynx ? 
Please give that a try first. For now I won't incorporate any of your suggestions.

best
Robert

PR #59, opened 2015-06-17
provide uniform "ft_platform_supports" function

nno:
- add function "ft_platform_supports" that returns whether the current platform supports a certain feature / capability
- remove duplicates of "matlabversion.m"
- instead of using "matlabversion" in functions, the new function "ft_platform_supports" is used. The aim is to provide more readable code
- code is Octave compatible


oostenveld:
I just started looking at this. Note to self: the following occurrences need to be checked to ensure that the sub-toolboxes remain stand-alone:

 delete mode 100644 engine/private/matlabversion.m
 delete mode 100644 fileio/private/matlabversion.m
 create mode 100644 ft_platform_supports.m
 delete mode 100644 peer/private/matlabversion.m
 delete mode 100644 plotting/private/matlabversion.m
 delete mode 100644 qsub/private/matlabversion.m
 delete mode 100644 test/private/matlabversion.m
 delete mode 100644 utilities/private/matlabversion.m

nno:
I was not aware that the sub toolboxes should remain stand-alone.

In that case, the only solution I see is having a copy of ft_platform_supports (or whatever name is preferable) in all sub-directories, though it is not ideal from a maintenance perspective. Do you agree with that observation? (I would be happy to update the PR to do so).

oostenveld:
I merged your branch and just kicked off the regression test battery on MATLAB2012b (which is the default for testing). It will take some time for the results to be in. I will compare those against the results of the latest svn version (which are also visible on http://www.fieldtriptoolbox.org/development/dashboard).

oostenveld:
regarding the copying of functions: we have a trick for that (i.e. an svn hook running on the server), which is one of the reasons for not abandoning svn yet. See http://www.fieldtriptoolbox.org/development/svn#how_to_maintain_two_copies_of_the_same_file 

Don't worry, I will set that up.

oostenveld:
what do you mean with RF, BF, DOC, OCTV?

oostenveld:
I had to make some changes to ft_version to make it git compatible. I found that test_randomseed started failing (on 2012b) due to the output variable not being assigned in case an input variable is specified.

oostenveld:
I have committed all changes to a personal branch on https://github.com/oostenveld/fieldtrip/tree/pull59. The test battery now runs fine, so I'll copy the changes over to SVN.

oostenveld:
All changes are now in SVN. See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2828#c10 for details.

nno:
Thanks for committing the changes, copying the functions, and running the test battery. I'm happy to hear that the tests run fine.

The tags RF, BF, DOC, OCTV are used in PyMVPA / CoSMoMVPA, and it has become a habit of me to use them. They mean "refactoring", "bug fix", "documentation change", and "improved Octave compatibility". A full list is here:

http://cosmomvpa.org/contribute.html#notes-on-committing

In CoSMoMVPA they are used to generate a summary of recent changes:

http://cosmomvpa.org/#changes-since-last-month

Looking at FT's commit history, I saw that full words are used. I will try to do so as well in future PRs.

nno:
Commit 0939bf90f456ece35b4a96b1f09bdb049382f3ad causes an issue due to missing "matlabversion", details here: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2828#c12

nno:
Bug report here: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2911

nno:
The matlabversion issue was fixed by 6df47c5ee826443e2404d68834085edde282909d

PR #60, opened 2015-06-17
OCTV: let size_equal support an arbitrary number of arguments

nno:
GNU Octave has a builtin function with the same name with the same function signature. The original  size_equal in FT makes cause Octave to give unexpected errors.


oostenveld:
external/stats should probably not be added to the path on octave. 

-----
mac011> svn commit 
Adding         private
Adding         private/size_equal.m
Deleting       size_equal.m
Transmitting file data .
Committed revision 10468.

enhancement - allow more than two input arguments, hide size_equal from general users by putting it in private. It is only used here as subfunction, it does not replace a Mathworks stats function.

PR #61, opened 2015-06-17
in ft_senstype, use a recursive call only when the input is a cell string

nno:
this improves compatibility with GNU Octave


oostenveld:
is the iscellstr function available in all relevant MATLAB versions?

see http://www.fieldtriptoolbox.org/development/guidelines/code which includes guidelines on older matlab versions.

nno:
It is present in 2009b, and the copyright information is:

"%   Copyright 1984-2003 The MathWorks, Inc.
%   $Revision: 1.17.4.2 $  $Date: 2004/04/10 23:32:38 $"

This suggests it was made available in either version 6.5.1 or 7.0.0. Is that old enough? 
- The guidelines mention "we try to support it for previous MATLAB versions up to five years old."; iscellstr is definitely supported for those versions.
- But the guidelines also mention "the nargout function in Matlab 6.5 and older does not work on function handles",  which may suggest that support for 6.5 is also desirable.

oostenveld:
That is old enough.

we don't support 6.5 actively any more. There is still some code in FT that deals with 6.5, but no new code needs to be added. There are multiple issues with it.  I recall a recent workshop where someone showed up with 7.3 (2006b) that it did not work with FT. I told the participant to team up with someone else for the hands-on sessions.

PR #62, opened 2015-06-17
improve Octave support for griddata

nno:
Note: this PR is branched from #59, so it may be better to consider the latter first.

This PR aims to make topoplot more compatible with GNU Octave


nno:
The PR has been updated to reflect the use of ft_platform_supports in master

PR #65, opened 2015-07-09
Added reading with precompiled Nlx2matCSC

sin-mike:
I found that in case of nxl files there is a huge advantage using native mex-libs provided by Neuralynx for reading data. One could get 2-10 fold decrease in importing time.

BR,
Mike from KSU

### DIFF
--- a/fileio/private/read_neuralynx_ncs.m
+++ b/fileio/private/read_neuralynx_ncs.m
@@ -33,6 +33,36 @@ if nargin<3
   endrecord = inf;
 end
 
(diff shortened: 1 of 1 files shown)


oostenveld:
Hi Mike, 

Could we please follow this up at http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2924  ?

I will post some files there for you to test. 

Robert

oostenveld:
I have attached two files to http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2924

Could you get them from there and check whether they still work as desired?

PR #66, opened 2015-07-10
Nlx mex read

sin-mike:
I followed the bugzilla issue here http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2924
and managed to get it work together. 
Added neuralynx toolbox to ft_hadtoolbox 

I'm not sure about try-catch, especially in case Nuralynx will try to change lib interfaces, but the merge works now.

BR,
Mike

### DIFF
--- a/fileio/private/ft_hastoolbox.m
+++ b/fileio/private/ft_hastoolbox.m
@@ -33,7 +33,7 @@ function [status] = ft_hastoolbox(toolbox, autoadd, silent)
 %    You should have received a copy of the GNU General Public License
 %    along with FieldTrip. If not, see <http://www.gnu.org/licenses/>.
 %
-% $Id$
+% $Id: ft_hastoolbox.m 10418 2015-05-22 10:19:38Z roboos $
 
 % this function is called many times in FieldTrip and associated toolboxes
 % use efficient handling if the same toolbox has been investigated before
@@ -104,6 +104,7 @@ url = {
   'DENOISE'       'see http://lumiere.ens.fr/Audition/adc/meg, or contact Alain de Cheveigne'
   'BCI2000'       'see http://bci2000.org'
   'NLXNETCOM'     'see http://www.neuralynx.com'
+  'NEURALYNX'     'see http://neuralynx.com/research_software/file_converters_and_utilities/' 
   'DIPOLI'        'see ftp://ftp.fcdonders.nl/pub/fieldtrip/external'
   'MNE'           'see http://www.nmr.mgh.harvard.edu/martinos/userInfo/data/sofMNE.php'
   'TCP_UDP_IP'    'see http://www.mathworks.com/matlabcentral/fileexchange/345, or contact Peter Rydesaeter'
@@ -254,6 +255,8 @@ switch toolbox
     status  = exist('load_bcidat', 'file');
   case 'NLXNETCOM'
     status  = (exist('MatlabNetComClient', 'file') && exist('NlxConnectToServer', 'file') && exist('NlxGetNewCSCData', 'file'));
+  case 'NEURALYNX'
+    status = (exist('Nlx2MatCSC', 'file')) % we use only this func at the moment
   case 'DIPOLI'
     status  = exist('dipoli.maci', 'file');
   case 'MNE'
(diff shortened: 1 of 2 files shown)


oostenveld:
Hi Mike,

thanks for testing. The reason for me preferring not to use try-catch is that I rather be informed about incompatibilities with future code (allowing to fix them) rather than them remaining undetected. If future versions of the mex files don't work, the quick solution is just to delete the mex files.

I am back in my office where I was able to do some testing myself (on my 64 bit OSX computer). According to me the record numbering in the next files starts with 0 rather than 1. Could you please test the latest version of the fieldtrip SVN/GIT release and compare whether you get the same results with and without mex files?

Robert

sin-mike:
Dear Robert,

I used one file of mine with the following paramenters (based on the file header inf)
-FileType CSC
-FileVersion 3.3.0
-RecordSize 1044
-CheetahRev 5.6.3 

The file has 250 records
With mex-enabled ft reads only first 249 records and returns double warning:
"The beginning of the record range occurs before the beginning of the file. The start of the range was changed from 0 to 1."
Comparing to mex-disabled reading, no warnings in reading, 250 records read, first 249 timestamps are identical to mex-enabled, for which I conclude that the record count in nlx mex lib starts with 1. I suggest leaving read_neuralynx_ncs.m 2aaf250 commit with
[1, NRecords_to_read] 
rather than
[0, NRecords_to_read-1]

BR,
Mike

oostenveld:
For me it was just the other way around, i.e. without the "-1" I get one record less than expected with my test file (which has 191 records).

On http://neuralynx.com/research_software/file_converters_and_utilities/ there are two sets of mex files, one for Neuralynx which is windows only and one from Ueli Rutishauser which includes all platforms. 

Might it be that one of them is using one-offset and the other is zero-offset?

Could you try the mex files from Ueli, removing the "_v3" part from the name? I cannot do it the other way around, since Neuralynx has not shared the source code for their "6.0.0" version of the mex files, which means that I cannot recompile it on maci64.

oostenveld:
if it is a mex file version difference (as I suspect), we have to extend the code with a detection of v3 (from Ueli) versus v6 (from Neuralynx). The v3 would work for all computers, the v6 only for windows. 

... I wish that companies like Neuralynx would realise that researchers are not only using Windows computers nowadays...

sin-mike:
Dear Robert,

I know the pain, I do have windows in my lab now. I could check this issue when get home and I could try it on OSX and Ubuntu. 

For now I would suggest using ispc() function to discover if the host computer is windows-based, then the shifts would be correct. What do you think?

Mike

oostenveld:
also with ispc there are still two (possible incompatible) versions of the mex files. 

I have disabled the mex for now (see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2924#c4), let's first sort out what the difference between the mex files is.

sin-mike:
So, I checked the working of Ueli's v3 libs on OSX. True, they have different interfaces of lib functions and they require -1 shift. In addition, they don't make any warnings or whatsoever in reading data with improper record bounds, it just reads less. So one record, the first or the last one, could easily be lost in reading. I suppose that Ueli's libs are based on older sources than that of v6 from NLX.

Still, these libs have different names, and using your marvellous ft_hastoolbox function we could account for these names. Will this solution work for ft?

sin-mike:
Dear Robert,

I introduced Ueli's nlx libs to ft_hastoolbox function and made necessary changes in read_neuralynx_ncs function.

I've run some tests on my PC machine and observed complete match between data read manually and data read with NLX. When I'm at my mac, I'll run tests with Ueli's libs. But according to my yesterday's experience it must work now.

Best regards,
Mike from KFU

oostenveld:
I made some small changes, they are now called neuralynx_v3 and neuralynx_v6. For the rest I think this is the perfect solution.

mac011> svn commit utilities/ft_hastoolbox.m  fileio/private/read_neuralynx_ncs.m
Sending        fileio/private/read_neuralynx_ncs.m
Sending        utilities/ft_hastoolbox.m
Transmitting file data ..
Committed revision 10552.

oostenveld:
if you check the present release code once more (from svn or github) and agree to the present implementation, then please close this issue.

thanks, Robert

sin-mike:
Dear Robert, 
I absolutely agree with your changes
Sincerely,
Mike

PR #67, opened 2015-10-14
Add decimation saving option to biosemi2ft

phpvdb:
Changes to the code involve biosemi2ft and OnlineDataManager.h.
biosemi2ft:
Adds 6th input argument to biosemi2ft that allows to specify a decimation factor for saving data. Incoming triggers (STATUS channel) will be delayed if more than one is detected within the decimation period. People should realise and understand how this (delaying triggers) possibly can have an effect on their recordings. The default decimation factor is set to 1 (no decimation), the program then behaves the same as before. Also, printed information to console has been extended and improved a bit.
OnlineDataManager.h:
Adds the option to change the saving sample frequency


robertoostenveld:
thanks, I included your suggestion. I also copied the compiled mack binary in the release, and recompiled on maci64. I did not recompile on any other platform.

mac011> svn commit src/buffer/cpp/odmTest.cc src/buffer/cpp/OnlineDataManager.h src/acquisition/biosemi/biosemi2ft.cc bin/maci/biosemi2ft bin/maci64/biosemi2ft 
Sending        bin/maci/biosemi2ft
Sending        bin/maci64/biosemi2ft
Sending        src/acquisition/biosemi/biosemi2ft.cc
Sending        src/buffer/cpp/OnlineDataManager.h
Sending        src/buffer/cpp/odmTest.cc
Transmitting file data .....
Committed revision 10777.

PR #69, opened 2015-11-04
numeric bit arrays are properly cast as logical bit arrays in ft_eloreta

grahamfindlay:


robertoostenveld:
thanks, well spotted. I hope that we don't have the same error elsewhere. 

Since we still use SVN for the main version, I copied your patch over and committed it there. It will automatically propagate towards the git/github version.

mac011> svn commit inverse/ft_eloreta.m
Sending        inverse/ft_eloreta.m
Transmitting file data .
Committed revision 10856.

PR #70, opened 2015-11-06
Fix 'function not found' bug in beamformer_sam.m

dtxe:
The SAM_costfun function is inside the inverse/private directory, so that it can't be called by fminsearch by function name. This proposed bugfix passes the function handle instead. The function handle is available to inverse/beamformer_sam.m, and makes the SAM_costfun function subsequently available to fminsearch.


robertoostenveld:
hmm, I cannot really make sense out of the pull request. There are 96 files that are different, a lot of them not related to the suggested change. Did you make the change to a recent version of FT?

dtxe:
My apologies, I must have made the pull request from a different branch or before the current commits were rolled in. I will reapply the patch to the current master branch then recreate the pull request.

robertoostenveld:
sorry, I already clarified it: it seemed to be due to me messing up a commit this morning. So it is my fault. 

I now have the isolated patch.

robertoostenveld:
Since the main development is still in subversion, I will copy the patch over from git to svn and apply it there.

robertoostenveld:
mac011> svn commit inverse/
Sending        inverse/beamformer_sam.m
Transmitting file data .
Committed revision 10855.

I copied it over and applied it. The change will also automatically be propagated to the fieldtrip version on github and bitbucket. See http://www.fieldtriptoolbox.org/faq/why_is_fieldtrip_maintained_in_svn_and_not_in_git for an explanation. We are planning to move completely to git.

thanks,
Robert

PR #72, opened 2015-12-09
Split read_besa_besa into two functions

serk1:
read_besa_besa_header reads and returns the header information from the file (ignoring the data).

read_besa_besa behaves as it did before, and calls read_besa_besa_header to get the header info.

I also removed the debugging messages and implemented the filter information tag BCAL:CHFI


robertoostenveld:
mac011> svn commit  
Sending        fileio/private/read_besa_besa.m
Adding         fileio/private/read_besa_besa_header.m
Transmitting file data ..
Committed revision 10987.

I merged the pull request into the original SVN copy, it will automatically end up in git.

PR #76, opened 2016-02-03
bugfix - previous commit did not change sel+1 to 2*sel for case inter…

roemervandermeij:
…sect

@schoffelen 
See commit title ;).

Github!


roemervandermeij:
Oh shoot. I had thought pull request were commit specific, or at least at the time of generating the pull request. I made some commits afterwards, and they were automatically added to the request. 

Always use a different branch for working separate things, which I should have done :o ;)

roemervandermeij:
I totally did not do that again.

robertoostenveld:
Ehm, I am a bit confused. Ideally I would like now to review the suggested changes, but it seems to me that this is just a bunch of random commits... which is of course fine, since that is how we used to work ;-)

But you are right, doing the work on separate feature branches is important. Then also the new feature does not have to be committed (and shared with the whole world) immediately, but only when sufficiently finished and tested. We once discussed this one http://nvie.com/posts/a-successful-git-branching-model. Although it is more elaborate than what we need, it is interesting to consider. Especially, we should not continue the "push everything immediately to the one and only release/master branch" style that we had with svn.

roemervandermeij:
Sorry for the late reply, I hadn't realized my email filters placed these emails in the 'ignore' list...

Yeah, sorry about this. I was working from the SVN model in my mind, and thought each pull request just encapsulated all commits up until the time of the request.

I think it's a good idea to create a branch for every fix or 'set of commits'. It could be cumbersome when we suddenly find an error unrelated to a piece of code we're working on at a time, but, creating a branch and checking it out is not difficult either.

PR #78, opened 2016-02-05
improved compatibility with GNU Octave

nno:
This PR proposes several changes to improve GNU Octave compatibility. These changes were proposed earlier in #62, #63 and #64 (now closed).

Overview of changes:
- `ft_platform_supports.m` supports several new options, including `alim`, `stats`, and `uimenu`, which may give different results for Octave and Matlab.
- bug fixes in `fileio/matlablt2010b/@uint64` classes which can cause a hard crash in Matlab (see #64). Note that incorporating these fixes would require a recompile using `mex` of the `@uint64` `.c` files, which is not part of this PR.
- `ft_senstype.m` only does a recursive call on cellstrings. Improves Octave compatibility.
- the `stats` external directory is only added in `ft_defaults.m` if necessary (i.e. if the stat functions are not available). This avoids many warnings in Octave saying that the stats functions shadow builtin functions.
- `ft_prepare_layout.m` uses a different pattern to read the files. This makes this function work in Octave.
- `topoplot_common.m` falls back on linear interpolation if more advanced interpolation options (such as `v4`) are not available, such as in Octave.

I would also be happy to split up this PR in multiple pieces, though changes of different lines in   `ft_platform_supports.m` affect different pieces.


robertoostenveld:
thanks, looks good to me. I don't have time to test and am about to leave for a wek. I hope that @schoffelen can keep an eye on things to make sure that nothing breaks (for too long).

nno:
Thanks @robertoostenveld, that's much appreciated. Hopefully not too many things will break :-)

schoffelen:
shoot, Robert just beat me to approving the pull request.

PR #89, opened 2016-02-18
ft_channelrepair outputs all of the fields that were given in data struct

kdoelling1919:
Previously, ft_channelrepair only returns sampleinfo and trialinfo as additional fields to the output struct. The more modern versions of fieldtrip have added fields in the data struct. This code fix adds in all additional fields so that the output and input will have the same fields even if more are added (or removed) in future versions.


robertoostenveld:
thanks, useful contribution!

robertoostenveld:
Just to be sure that you got it: The important deviation from your suggested code is that I called it cfg.keepchannel='repair' instead of 'neighbours'

PR #91, opened 2016-02-19
Kdoelling1919 rejectviz

robertoostenveld:
I improved the handling of the provenance and history, did some general cleanups, and renamed the cfg.keepchannel='neighbours' option into cfg.keepchannel='repair'. I also added an early check on the presence of cfg.neighbours.


robertoostenveld:
note that this replaces pull request #87

PR #93, opened 2016-02-19
DOC - refer to the new www.fieldtriptoolbox.org website URL ...

robertoostenveld:
... instead of to the old F.C. Donders Centre web address. I have excluded fieldtrip/external on purpose, since I did not want to make any changes there together with this main batch of changes. It might be that in fieldtrip/external there are still some fcdonders or neuroimaging URLs.


schoffelen:
shall I check all 1178 files first?

PR #96, opened 2016-02-19
make the explicit allowed option checking transparent for progress

schoffelen:
this addresses bug 3072


schoffelen:
sorry, I meant bug 3062

PR #105, opened 2016-02-24
FIX - resolve bug3077 regarding nargin

robertoostenveld:
furthermore, I made the nargin handling more consistent throughout. See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3077


robertoostenveld:
This involves a lot of (small) changes to a large number of files, so I hope there is not a typo somewhere.

PR #107, opened 2016-02-25
Bug 2811

vlitvak:
enhancement - return also fiducials for BESA sfp files

### DIFF
--- a/fileio/ft_read_headshape.m
+++ b/fileio/ft_read_headshape.m
@@ -880,13 +880,11 @@ switch fileformat
 	case 'besa_sfp'
 		[lab, pos] = read_besa_sfp(filename, 0);
 		shape.pos = pos;
-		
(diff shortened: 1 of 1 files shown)


schoffelen:
Hi Vladimir, apologies for not communicating this sooner, but this morning I already incorporated your change into ft_read_headshape (as per pull request 106). Sorry for the double work, but at least it got you started on git :o).

vlitvak:
Yes, I did it more to try how it all works. Still haven't figured out how to see what's been committed already.

schoffelen:
well, in principle you could do a git diff on the relevant file. Yet, somehow in the process the formatting of ft_read_headshape got screwed up, so git considered the text in the whole file to be replaced by something else. Typically, I would have expected only the small changes highllghted.

PR #110, opened 2016-02-28
FIX - replace pnt by pos in the data structures used in the code, 

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3081 and thanks to Lau


robertoostenveld:
Not all test scripts completed running and there are some that return errors. But I don't see errors related to the original commit, so I will go ahead and merge this in the master.

PR #112, opened 2016-03-01
option of including a noise covariance structure in the low-level pa…t of dipole fitting is removed

ualsbombe:
Removed the low-level option of having a noise-covariance structure


robertoostenveld:
as discussed with Lau on Skype.

PR #114, opened 2016-03-02
bugfix - npos should be the number of dippos and not the number of ve…

schoffelen:
…rtices in the mesh


schoffelen:
This should now finally allow me to close bug 3041, about the very important ft_sourcedepth function

PR #115, opened 2016-03-03
ENH - added color code for .obj for ft_read_headshape

simonhomoelle:


fileio/ft_read_headshape.m
```diff
@@ -81,7 +83,7 @@
 coordsys       = ft_getopt(varargin, 'coordsys', 'head');    % for ctf or neuromag_mne coil positions, the alternative is dewar
 fileformat     = ft_getopt(varargin, 'format');
 unit           = ft_getopt(varargin, 'unit');
-
+image          = ft_getopt(varargin, 'image');               % path to .jpeg file
```
robertoostenveld: could you add some default value for "image" when it is not specified by the user, e.g. if the input file is FILENAME.obj, that it checks whether FILENAME.jpg exists and if so that it is used as "image"?

fileio/private/refine.m
```diff
@@ -0,0 +1,183 @@
+function [pntr, trir, texr] = refine(pnt, tri,texture, method, varargin)
+
+% REFINE a 3D surface that is described by a triangulation
+%
+% Use as
+%   [pnt, tri] = refine(pnt, tri)
+%   [pnt, tri] = refine(pnt, tri, 'banks', selection)
```
robertoostenveld: please document in the help what the intended use is of the "selection" option.

fileio/ft_read_headshape.m
```diff
@@ -52,6 +53,7 @@
 %   'caret_spec'
 %   'brainvisa_mesh'
 %   'brainsuite_dfs'
+%   'obj'           Wavefront .obj file obtained with the structure.io
 %
```
simonhomoelle: @robertoostenveld Testcomment

PR #118, opened 2016-03-08
Sphering

ualsbombe:
As discussed the implementation of using a noise covariance has been implemented in ft_dipolefitting

See also: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3089


robertoostenveld:
please fix the white space in the code

robertoostenveld:
note to self: this uses the test data from http://www.fieldtriptoolbox.org/tutorial/natmeg/dipolefitting. I should copy part of it to a test directory.

robertoostenveld:
looks good!

PR #119, opened 2016-03-08
fixed a bug introduced by bugfix3089 related to resolution being defi…

roemervandermeij:
…ned after its needed to determine dointerp

Commit a559f38c05b71573cf62cb685e275cb49eef568e introduced a bug where default resolution was defined after it was necessary for determining 'dointerp'. I thought it was related to bug3089, hence the name of this branch, but not sure it's actually related to bug3089.

Anyway, I moved the dointerp determination down.

### DIFF
--- a/plotting/ft_plot_slice.m
+++ b/plotting/ft_plot_slice.m
@@ -152,17 +152,6 @@ if domask
   end
 end
 
-% determine whether interpolation is needed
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Sorry, my bad. Thanks for fixing.

PR #125, opened 2016-03-15
Minor change of ft_read_mri documentation

forodin23:
Added comment saying that in case of DICOM files the first file name should be given.

### DIFF
--- a/fileio/ft_read_mri.m
+++ b/fileio/ft_read_mri.m
@@ -45,6 +45,9 @@ function [mri] = ft_read_mri(filename, varargin)
 %   ANT - Advanced Neuro Technology (*.mri)
 %   Yokogawa (*.mrk, incomplete)
 %
+% If you have a series of files in DICOM format please only provide name of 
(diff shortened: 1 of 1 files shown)


robertoostenveld:
should it be the first? I recall that any one from the sequence will do. But selecting the first is not a bad idea, and it is good to explicitly document that a single file (from the sequence of files) should be specified.

forodin23:
Sorry, I didn't check that before. Yes, it can be any file, I tried several and the result was always the same. Are there other file formats in which this is the case or is it DICOM-exclusive?

robertoostenveld:
Thanks for checking. Dicom is the only format that I know that has a long list of nearly identical files. In some other data formats (not MRI per see) there are also "constellations" of files, but then typically with clearly different names and/or extensions (like analyze HDR+IMG).

For electrophys files (not MRI) we have the fieldtrip/utilities/private/dataset2files.m helper function. 

I just made a small documentation change directly in the master.

PR #128, opened 2016-03-16
Compatibility with Bandpass Butterworth filter

gllmflndn:
Add 'bandpass' as filter type for compatibility with MATLAB:
http://www.mathworks.com/help/signal/ref/butter.html?refresh=true#inputarg_ftype

Would it also be possible to _only_ add alternative implementation functions to the MATLAB path when MathWorks' toolboxes are missing:
https://github.com/fieldtrip/fieldtrip/blob/master/ft_defaults.m#L134

### DIFF
--- a/external/signal/butter.m
+++ b/external/signal/butter.m
@@ -49,7 +49,7 @@
 function [a, b, c, d] = butter (n, W, varargin)
 
 if (nargin>4 || nargin<2) || (nargout>4 || nargout<2)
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Regarding ft_defaults and drop-in-functions. Those are installed by default to deal with network license limitations (which might not apply at the FIL). See http://www.fieldtriptoolbox.org/faq/can_i_prevent_external_toolboxes_from_being_added_to_my_matlab_path for the solution.

PR #135, opened 2016-04-19
FIX - renamed pnt to pos in prepare_mesh_segmentation

thht:
there are still some .pnt fields left in prepare_mesh_segmentation. thus pull request fixes this


thht:
this refers to bug 3107

robertoostenveld:
thanks

PR #144, opened 2016-05-06
Enable passing freq data structures to ft_freqgrandaverage in a cell

janfreyberg:
Hi fieldtrippers,

To make it easier to use ft_freqgrandaverage programmatically (without
saving data to file), this change allows users to pass a cell array of
standard freq data to ft_freqgrandaverage. It checks if a cell array was
in argument one, and then "unpacks" it.

Previously it was very hard to pass data to this function in a script if you didn't want to save it to file, since you'd have to manually pass in every single freq data structure.

This could also easily be added to the other grandaverage functions (happy to do so) - as a suggestion!

Also fully backwards compatible.

Cheers,
Jan

### DIFF
--- a/ft_freqgrandaverage.m
+++ b/ft_freqgrandaverage.m
@@ -5,6 +5,8 @@ function [grandavg] = ft_freqgrandaverage(cfg, varargin)
 %
 % Use as
 %   [grandavg] = ft_freqgrandaverage(cfg, freq1, freq2, freq3...)
+%   or
+%   [grandavg] = ft_freqgrandaverage(cfg, {freq1, freq2, freq3...})
 %
 % The input data freq1..N are obtained from either FT_FREQANALYSIS with
 % keeptrials=no or from FT_FREQDESCRIPTIVES. The configuration structure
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Hi Jan,

I am not sure I see the reason for this. You are probably thinking of
freqall = {freq1, freq2, freq3, ....}
when preparing all freq structures, and then you can already do (in standard MATLAB syntax)
ft_freqgrandaverage(cfg, freqall{:})

This unpacks the cell array and all elements are passed as subsequent input arguments without having to specify individual variables or even the number of arguments. 

I found this on the mathworks site explaining it
http://nl.mathworks.com/help/matlab/matlab_prog/pass-contents-of-cell-arrays-to-functions.html#zmw57dd0e25506

Does this not serve the purpose you have in mind?

janfreyberg:
Hi Robert,

it does indeed! Yep - I think I just didn't know about this quirk in matlab syntax. That really does solve the problems I had in the past.

Thanks - this can be ignored!

Jan

PR #150, opened 2016-05-10
fix for proper inversion of Gram matrix in ft_sloreta; major nutmegtrip updates for source-time-freq plotting

sarangdalal:
ft_sloreta: fix for proper inversion of Gram matrix

nutmegtrip: major updates for source-time-freq plotting

ft_electroderealign: fixed help comments so that they display properly


robertoostenveld:
looks ok to me. 

thanks,
Robert

PR #151, opened 2016-05-10
bugfix3109

weiyongxu:
fix Bug 3109 ft_combineplanar does not work for trials containing one bad planar gradient channel.
As Jan-Mathijs suggested, the unnpaired channels are identified and removed.So there is no need to have error message anymore.
Also, in the previous version,  `length(sel_dH)==length(sel_dV)`  does not guarantee that the horizon and vertical gradiometers are combined correctly, it is possible that horizon and vertical gradiometers have the same number of bad channels. So the combination of horizon and vertical gradiometers maybe not from the same senser.


robertoostenveld:
thanks, much appreciated!

PR #152, opened 2016-05-11
bugfix - allow for correct handling of mouse click/button press

schoffelen:
@robertoostenveld could you review this suggested fix please? It's basically reverting some low-level gui related data handling, that currently causes mous clicks not to work in ft_artifact_zvalue (as per bug 3118).
With this change in place, it at least works again in matlab2013b.

Thanks.


robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3118

PR #153, opened 2016-05-11
ENH: added method 'project' to ft_electrodeplacment, that allows to project the electrodes onto the head surface

simonhomoelle:


robertoostenveld:
you should not project the chanpos, but the elecpos. The chanpos is to be removed. If needed, it will be regenerated elsewhere.

E.g. the chanpos for a bipolar electrode pair is in between the two electrodes.

simonhomoelle:
So now it only projects the elecpos.

robertoostenveld:
I should have read this with greater detail. ft_electrodeplacement is for manually locating individual electrodes, ft_electroderealign for modifying a set of electrodes. I will move it over to the other function.

robertoostenveld:
mbp> git push upstream master
Counting objects: 8, done.
Delta compression using up to 2 threads.
Compressing objects: 100% (8/8), done.
Writing objects: 100% (8/8), 2.02 KiB | 0 bytes/s, done.
Total 8 (delta 5), reused 0 (delta 0)
To git@github.com:fieldtrip/fieldtrip.git
   6e16dbc..1657b21  master -> master

I have moved the code over, and renamed chanpos into elecpos. In case the difference between the two is NOT clear, please let me know so that I can explain. Otherwise you'll be making wrong assumptions for the next 2,5 years.

PR #156, opened 2016-05-12
Fix Bug 2978 

weiyongxu:
Fix to Bug 2978 - "Plot Trial" button on GUI for ft_rejectvisual (summary mode) plots average of all trials instead of the single trial requested

Hi,

This bug is originated from ft_multiplotER：
line 327      [cfg, varargin{:}] = rollback_provenance(cfg, varargin{:});

This lead to cfg.trials='all',because the data structure in varargin{:} which was passed from rejectvisual_summary does not have any trial plot index from the user(by typing in the textbox in GUI).  This overwrite: 

line 158      cfg.trials         = ft_getopt(cfg, 'trials', 'all', 1);

which recieve the correct trial plot index from the user(by typing in the textbox in GUI).So before plotting ft_timelockanalysis averages all the trials for all the single trial plot.

It maybe not a good idea to change code in ft_multiplotER, so I would suggest only pass a single trial data to ft_multiplotER when it is called in rejectvisual_summary to plot specific single trial.I made a few line code change in rejectvisual_summary and it seems to work well.


robertoostenveld:
although the problem is visible in ft_rejectvisual, that is not where the problem originates, nor should be solved. 

I loaded a dataset (Subject01.ds) and with

cfg = [];
cfg.layout = 'CTF151.lay';
cfg.trials = 255; % this has a jump
ft_multiplotER(cfg, data)

I am also getting the average. So ft_multiplotER is not behaving as it should and not consistent with its  own documentation.

robertoostenveld:
I will make another pull request with (hopefully) the correct change.

PR #158, opened 2016-05-12
OpenMEEG pipeline update for non-magnetometer data (fix for Bug 3106)

sarangdalal:
Updated forward/private/leadfield_openmeeg.m according to our latest practice.

Notably:
- Elekta data now supported and confirmed working, addressing Bug 3106
- CTF data confirmed working, with and without reference channels, with and without synthetic 3rd gradient
- 4D/BTi MEG data confirmed working
- EEG data not re-tested, but "should" still work
- Matlab's multithreaded inverse is now chosen over om_minverser when 4+ cores are available, vastly speeding up computation
- added provision for saving OpenMEEG's intermediate files as mat-files or native binary


agramfort:
@sarangdalal can you also send us a PR to https://github.com/openmeeg/openmeeg/ so we keep the om_* matlab files in sync? thanks

sarangdalal:
@agramfort  I don't think there are any modifications to the om_* files from OpenMEEG's repository, I just added all the ones that were missing (as our revised pipeline relied on a couple that were previously not included in fieldtrip/external/openmeeg). Is there a way to keep these permanently in sync via github?

@robertoostenveld  There are some other files in external/openmeeg called icosahedron*.m that it seems you authored, and are used by openmeeg_eeg_leadfield_example.m and testOpenMEEGeeg.m. Are these still needed?

agramfort:
ok thanks for letting me know

cc @papadop

PR #160, opened 2016-05-13
bugfix: in ft_convert_units, fix issue when .orig.pos present

nno:
The original code threw and error when obj.orig field has field ".pos" but not a field ".pnt".


robertoostenveld:
thanks!

nno:
That was /very/ quick - thanks!

PR #164, opened 2016-05-18
ENH: hascolor only check if method 'headshape' specified

simonhomoelle:


robertoostenveld:
this way the code is broken and logic is not clear, creating risk in future code maintenance. Better would be to replace the boolean has color at the (only) line where it is actually needed by "isfield(headshape, 'color')" and avoid using the extra variable.

simonhomoelle:
Changed code as suggested

PR #174, opened 2016-06-10
vlim and hlim in multiplot are translated to ylim and xlim.

smkia:
vlim and hlim in multiplot are translated to ylim and xlim in singleplot when the interactive option is used. This could be advantageous when the multiplotER is used interactively.
In the previous version even if zlim, vlim, and hlim are set in the call of multiplotER, they are not forwarded correctly to the interactively called singleplotER and the subsequent topoplotER. For example, this is problematic especially when you want to have a 'maxabs' limited topoplot. Even if the vlim and zlim are set to 'maxabs' before calling multiplot, in the subsequent topoplot which is interactively plotted,  the zlim is still 'minmax'. The simple change proposed by this pull request solves the problem.

### DIFF
--- a/ft_multiplotER.m
+++ b/ft_multiplotER.m
@@ -866,6 +866,8 @@ if ~isempty(label)
     cfg = rmfield(cfg, 'inputfile');
   end
   cfg.channel = label;
+  cfg.xlim = cfg.hlim;
+  cfg.ylim = cfg.vlim;
   % put data name in here, this cannot be resolved by other means
   info = guidata(gcf);
   cfg.dataname = info.dataname;


robertoostenveld:
hmm, although I understand why you want this change, I have the feeling that it would move us in the wrong direction. The reason for moving away from xlim/ylim/zlim is that x/y/z are too confusing to label 3D data, and especially when the code needs to support 2D and 3D. 

The multiplot and singleplot figures technically have a horizontal and vertical axis. And we use color (in case of 3D). I remember that we tried at a certain point to move away from x/y/z, but the code is not consistent in this respect.

Let me move this over to bugzilla and get some other old-time developers to think along... see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3146 and please sign up on bugzilla and CC you to that bug.

robertoostenveld:
I have implemented another solution as discussed on http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3146

robertoostenveld:
@smkia I hope you are also happy with #219

smkia:
@robertoostenveld It works here perfectly. Thanks!

PR #184, opened 2016-06-24
critical bugfix for beamformer_lcmv nai/unitnoisegain orientation selection

sarangnemo:
critical bugfix!  the new nai/unitnoisegain orientation selection formula was incorrectly flipped (denominator vs numerator), now fixed and tested as providing correct orientation using a simulation.


sarangnemo:
sorry :-)

robertoostenveld:
thanks for detecting it so quickly!

PR #189, opened 2016-07-05
enhancement - ft_prepare_layout.m saves to & reads .mat layout files

TEMadsen:
These are pretty subtle edits, but it's my first pull request, so I thought I'd start simple.  :blush: 

I noticed the documentation of ft_prepare_layout.m says "an ASCII *.lay file...is no longer recommended."  However, I got errors when trying to use the layouts I had saved using .mat suffixes, and it turned out that the code to save cfg.output (line 882) was in text format after all.  So I changed ft_prepare_layout.m to save in matlab format if a .mat suffix is detected, but the default text file format otherwise, assuming some people still like to use .lay.  I edited the documentation accordingly.  I also added one line that may edit the layout variable back before the test display and saving steps.

Also, the code for loading a .mat layout file (line 379) only worked for variables named "lay" but not "layout," so I fixed that with an asterisk.

Hope this works and is helpful!
~Teresa

### DIFF
--- a/ft_prepare_layout.m
+++ b/ft_prepare_layout.m
@@ -1,9 +1,9 @@
 function [layout, cfg] = ft_prepare_layout(cfg, data)
 
-% FT_PREPARE_LAYOUT loads or creates a 2-D layout of the channel locations. This layout
-% is required for plotting the topographical distribution of the potential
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks, looking forward to future contributions!

PR #193, opened 2016-07-13
ENH - add new options for cfg.toi & cfg.pad in ft_freqanalysis

TEMadsen:
The default settings in ft_freqanalysis could take an unnecessarily long time to process, so I tried to improve them in two subtle ways:

1. The guess at what was meant by cfg.toi = 'all' was one spectral estimate per data sample, which seemed pretty excessive.  I think the minimum smoothing window from cfg.t_ftimwin makes more sense as a default step size, since the spectral estimate shouldn't change much in less time than that.
2. As noted in the documentation, the default cfg.pad = 'maxperlen' could run very slowly because of inefficient FFT computation on samples with a large prime factor sum.  A simple solution to avoid that is just rounding up to the next higher power of 2, so I added that option ('nextpow2') and made it the default.

Not sure if I'm overstepping my bounds by changing defaults, but I'm pretty sure these won't have an adverse effect on anyone.  Let me know what you think.

Thanks,
Teresa

### DIFF
--- a/ft_freqanalysis.m
+++ b/ft_freqanalysis.m
@@ -39,18 +39,17 @@ function [freq] = ft_freqanalysis(cfg, data)
 %   cfg.trials     = 'all' or a selection given as a 1xN vector (default = 'all')
 %   cfg.keeptrials = 'yes' or 'no', return individual trials or average (default = 'no')
 %   cfg.keeptapers = 'yes' or 'no', return individual tapers or average (default = 'no')
-%   cfg.pad        = number, 'maxperlen', or 'nextpow2' (default), length
+%   cfg.pad        = number, 'nextpow2', or 'maxperlen' (default), length
 %                   in seconds to which the data can be padded out. The
 %                   padding will determine your spectral resolution. If you
 %                   want to compare spectra from data pieces of different
 %                   lengths, you should use the same cfg.pad for both, in
 %                   order to spectrally interpolate them to the same
-%                   spectral resolution.  The new default option,
-%                   'nextpow2' uses the built-in Matlab function of the
-%                   same name to round the maximum trial length up to the
-%                   next power of 2.  By using that amount of padding, the
-%                   FFT can be computed more efficiently in case
-%                   'maxperlen' has a large prime factor sum.
+%                   spectral resolution.  The new option 'nextpow2' rounds
+%                   the maximum trial length up to the next power of 2.  By
+%                   using that amount of padding, the FFT can be computed
+%                   more efficiently in case 'maxperlen' has a large prime
+%                   factor sum.
 %   cfg.padtype     = string, type of padding (default 'zero', see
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Hi Theresa,

Re 1; I agree that the default cfg.toi option should be reasonably efficient. But some overlap between timwins would be good to be able to estimate the latency at which a peak happens. I would say that that requires at least 50% overlap. And the 50% overlap should apply to the shortest (probably highest frequency) timwin.

Could you implement this as cfg.toi='50%' i.e. as a string?

Re 2; Once upon a time we looked at the fft speed for different sample numbers, and back then the difference between powers of 2 and any other length was not that large. I don't recall whether we looked at specifically difficult lengths (like prime factors). 

If I try

```
n = 1:10000;
m = 2.^(0:log(10000)/log(2));
t = nan(size(n));
for i=n
  x = randn(i,1);
  for j=1:5
    tic;
    f = fft(x);
    tim(j) = toc;
  end
  t(i) = min(tim);
  plot(n(1:i), t(1:i)); hold on
  plot(m(m<i), t(m(m<i)), 'r*'); hold off
  title(i);
  drawnow
end
```` 
I indeed see quite some variance and the 2^x values as the fastest. Could you change the code first such that it is an option, but not the default? 

 Re 1 and 2; Could you for now change it such that the actual standing defaults are not changed, but the new options do become available (and documented)? Then we could start by doing something like

cfg.pad = ft_getopt(cfg, pad);
if isempty(pad)
  % give warning, explain that new default is better
  % set to old default 
end

TEMadsen:
Sure, I'll make those changes now.  For future reference, is a pull request the right way to start such a conversation, or would it be better to discuss on Bugzilla or the listserv first?
Thanks,
Teresa

robertoostenveld:
To save you double work (i.e. reverting part of the changes) in case of code suggestions that are might be disputed, I think it is more efficient to first initiate discussion, then implement. That means that a PR is not the way to start, as that starts with changing code. 
I am not constantly keeping track of the mailing list, although I believe @schoffelen is keeping a closer watch. Hence mailing list is also not ideal. 
That leaves either bugzilla, or to open a github issue (without PR. The advantage of bugzilla  is that more people get informed about the initial commit (through the fieldtrip-bugs mailing list) and therefore able to chime in. We still should change the way we are using github to make it more inclusive and to work better with teams. So for now bugzilla has my preference for discussions.

TEMadsen:
Sounds good - I'll start with Bugzilla next time.  

What do you think of the changes now?

robertoostenveld:
Looks good (at first glance) but I don't have time any more before holiday to test. @schoffelen will test and then merge.

schoffelen:
Hi Teresa, in general it looks good to me. Thanks for your contribution!

TEMadsen:
Great, thanks!  :smile:

PR #197, opened 2016-07-21
fixed problem with label, bug 3164

stephenwhitmarsh:


schoffelen:
I am not sure whether I understand everything, but your fix seems reasonable for the problem at hand for bug 3154. Yet, could you change the any back into all for neuromag122? I think the particular problem with neuromag 306 is that it's a combined planar gradiometer magnetometer system (hence resulting in a larger number of columns upon calling ft_senslabel) than the other systems.

schoffelen:
For the reason mentioned in my previous comment (i.e. the neuromag306 being an outlier system) the suggested fix for ft_combineplanar will break all other channel systems (although it might fix for neuromag306). Could you perhaps change the '4' into 'end'?

stephenwhitmarsh:
Yes, very good points I overlooked. Fixed as suggested.

PR #201, opened 2016-07-28
adding support for  Nihon Koden data format

dieloz:
This implementation is because Guangzhou workshop next August


schoffelen:
Looks clean enough to me :o). Gracias!

PR #202, opened 2016-08-02
fix: ensure spm is on the path

StolkArjen:


schoffelen:
Hi Arjen,

Could you undo this change (i.e. move the ft_hastoolbox call to after the if-statement), and add a call to ft_hastoolbox to after the if opt==2 statement. This should have the same effect to your change, apart from the fact that the toolbox is not unnecessarily added if opt==0

Then create a new PR.

StolkArjen:
Will do

StolkArjen:
Can you pull "fix update: ensure spm is on the path" and discard the other? It has the recommended fixes.

PR #205, opened 2016-08-09
Avoid potential error when assigning fieldname from filename

jens-k:
The filename of the events file is used as fieldname for the structure xml. This will lead to an error if the file name contains blanks (e.g. on EGI EEG systems). 
Replaced blanks in string with underscores before using it as a fieldname.


schoffelen:
Thanks!

PR #208, opened 2016-08-17
ENH - introduced cfg.plotevents option to ft_databrowser.

thht:
if a lot of events are present in the data, displaying them is a huge performance hit. i propose an option to disable displaying them.


robertoostenveld:
thanks!

PR #210, opened 2016-08-23
Solves problem with bnd.pnt, now replaced by .pos

jens-k:
Please read http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3176 before merging.


robertoostenveld:
this PR can be ignored, as it is solved differently (see bugzilla).

PR #214, opened 2016-09-05
Fixed bug leading to error if 'compile' = 'auto'

jens-k:
If variable 'compile' is set to 'auto', istrue(compile) will lead to an error. 
Checking for strcmp(compile, 'auto') first circumvents that issue.


robertoostenveld:
thanks!

PR #215, opened 2016-09-06
Fixes bug 3177 by calling poly2tri if necessary

jens-k:
Robert wrote poly2tri for this case but it wasn't called in the code yet. I hope I used it as intended. 
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3177


robertoostenveld:
hmm, it is called on line 274 in my code. Let me check.

robertoostenveld:
oh bummer, I realize I forgot to commit that change.

ip-145-116-191-159> git commit ft_electroderealign.m
[master 42bcde2] FIX - also call poly2tri, see https://github.com/fieldtrip/fieldtrip/pull/215
 1 file changed, 42 insertions(+), 39 deletions(-)

The reason for the poly2tri being only in the specific if-hexahedral statement is that I want to keep the exception (or "unusual data") handling close to whether it happens.  

Thanks for the suggested fix. I'll ignore it, but do let me know if it still is not 100% correct.

jens-k:
Seems to work over here. Thanks!

PR #221, opened 2016-09-16
added sleep option to control time between job check completion check…

roemervandermeij:
…s inside qsubcellfun

I added a sleep key-value option to allow for a bigger time between completion checks than the default 100ms. For some reason a couple of simultaneous qsubcellfun runs, with about 50 jobs each, stressed the scheduler too much at the super computing center here in San Diego. 

Created the pull request in case you also consider it a useful minor addition. Otherwise also okay. Right now the sleep time doesn't get forwarded to qsubget (currently a different default compared to qsubcellfun, 10ms instead of 100ms).


robertoostenveld:
I will change the default from 0.1 to 0.5 in a separate commit.

robertoostenveld:
[master 576a267] ENH - increased default sleep from 0.1 to 0.5, see also https://github.com/fieldtrip/fieldtrip/pull/221
 6 files changed, 8 insertions(+), 8 deletions(-)

mac011> git push upstream master
Counting objects: 9, done.
Compressing objects: 100% (9/9), done.
Writing objects: 100% (9/9), 937 bytes | 0 bytes/s, done.
Total 9 (delta 7), reused 0 (delta 0)
remote: Resolving deltas: 100% (7/7), completed with 7 local objects.
To git@github.com:fieldtrip/fieldtrip.git
   ca9c02e..576a267  master -> master

PR #222, opened 2016-09-19
enhancement - added neuromag306_combined to the list of meg sensor types

vlitvak:


robertoostenveld:
for reference: this was actually a new commit related to the point size in ft_plot_sens

PR #223, opened 2016-09-22
ENH: adapted qsub memory reservation for SGE systems

StolkArjen:
Since SGE wants memory to reserved according to -l memfree=XXG, where XX is amount of Gb, the code has been adjusted from

submitoptions = [submitoptions sprintf('-l mem_free=%.0f ', memreq+memoverhead)];

into

submitoptions = [submitoptions sprintf('-l mem_free=%.0fG ', round((memreq+memoverhead)/1024^3))];

at line 300 of qsubfeval.m.

### DIFF
--- a/qsub/qsubfeval.m
+++ b/qsub/qsubfeval.m
@@ -297,7 +297,7 @@ switch backend
     end
     
     if ~isempty(memreq) && ~isnan(memreq) && ~isinf(memreq)
-      submitoptions = [submitoptions sprintf('-l h_vmem=%.0f ', memreq+memoverhead)];
+      submitoptions = [submitoptions sprintf('-l mem_free=%.0fG ', round((memreq+memoverhead)/1024^3))];
     end
     
     if compiled


robertoostenveld:
I hope this is not SGE version specific, since the old code worked on some SGE clusters.

StolkArjen:
Apologies for the late response. Yes, this is SGE specific. It seemed that changing the "h_vmem" variable did not affect the actual memory reserved (despite this change being logged with qstat). Changing the "mem_free" variable does. At least for the SGE system at the Helen Wills Neuroscience Institute of UC Berkeley, and judging from their websites also the SGE systems at Stanford and Duke. I was therefore assuming that it'll have the same effect across the board of SGE systems. If that assumption proves to be a wrong for other institutes using SGE systems, we'll have to think of a way to automatically segregate the call under the different SGE environments out there, or simply change both the "h_vmem" and "mem_free" variables.

PR #227, opened 2016-09-27
lcmv: some updates with respect to subspace and regularization + invCy^2 computed only once

schoffelen:
@sarangnemo could you check whether you agree with the computation of invCy^2 only once, for the 'nai', and 'unitnoisegain' weights normalization?


sarangnemo:
Looks good to me!

PR #228, opened 2016-09-27
do not break in equals case

jjnurminen:
Sorry @robertoostenveld , I think this is how it should actually be.


robertoostenveld:
thanks for following up

PR #234, opened 2016-10-13
Bug3184

schoffelen:


schoffelen:
PS: I ran test_ft_selectdata prior to pulling, so I think we are safe...

PR #237, opened 2016-10-20
ft_sloreta critical bugfix (see Bug 2979)

sarangnemo:
ft_sloreta critical bugfix: the Gram matrix was improperly computed voxel-by-voxel, now fixed by computing it before the for-loop. (See Bug 2979)

Furthermore, orientation optimization was validated by confirming that the selected orientation indeed maximizes equation 13.23 from Sekihara & Nagarajan (2008) for a real dataset.

### DIFF
--- a/inverse/ft_sloreta.m
+++ b/inverse/ft_sloreta.m
@@ -208,6 +208,10 @@ elseif ~isempty(subspace)
   end
 end
 
+L = cell2mat(dip.leadfield);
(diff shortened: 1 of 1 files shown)


sarangnemo:
@schoffelen @robertoostenveld Shall I merge this, or would one of you like to take a look? (It's only a few lines of code that have changed, but the output changes dramatically.)

PR #241, opened 2016-10-26
Proposal for bug 3192 (ft_plot_mesh color management)

jens-k:
Color definitions are evaluated if possible ('brain', 'red', etc.). If this is suitable I can extend the solution to other plotting functions.

### DIFF
--- a/plotting/ft_plot_mesh.m
+++ b/plotting/ft_plot_mesh.m
@@ -11,8 +11,8 @@ function [hs] = ft_plot_mesh(mesh, varargin)
 %   ft_plot_mesh(pos, ...)
 %
 % Optional arguments should come in key-value pairs and can include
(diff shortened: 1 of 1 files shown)


plotting/ft_plot_mesh.m
```diff
@@ -139,11 +139,14 @@
 end
 
 % new colors management
-if strcmpi(vertexcolor,'skin') || strcmpi(vertexcolor,'brain') || strcmpi(vertexcolor,'cortex')
-  vertexcolor = eval(vertexcolor);
+if ischar(vertexcolor) && numel(vertexcolor) <= 63 % 63 is the max evaluable string length
+	try vertexcolor = eval(vertexcolor); catch, end
```
robertoostenveld: try-catch is expensive, i.e. it takes quite some time. Could you change it into the following?

if ischar(vertexcolor) && exist([vertexcolor '.m'], 'file')
  vertexcolor = eval(vertexcolor);
end

PR #245, opened 2016-10-27
improve the interpolating of an atlas onto a sourcemodel

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3190


robertoostenveld:
as suggested by @jens-k

PR #248, opened 2016-10-27
Proposal 2 for bug 3192

jens-k:
If colors are provided using 'brain', 'red', etc. they are now evaluated
if possible.


jens-k:
Here is a script for all altered functions.

[test bug3192.txt](https://github.com/fieldtrip/fieldtrip/files/556304/test.bug3192.txt)

PR #249, opened 2016-10-31
improve inside/outside detection of source points for FEM models

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3195

Please note that this chain of commits is not ready yet. 
It also needs to be reviewed by @simonhomoelle

### DIFF
--- a/forward/ft_inside_vol.m
+++ b/forward/ft_inside_vol.m
@@ -121,49 +121,63 @@ switch ft_voltype(headmodel)
     
   case {'simbio'}
     
+    numhex = size(headmodel.hex,1);
+    numpos = size(headmodel.pos,1);
+    numdip = size(dippos,1);
+    
     % FIXME we have to rethink which tissue types should be flagged as inside
     tissue = intersect({'gray', 'white', 'csf', 'brain'}, headmodel.tissuelabel);
-    brain = false(size(headmodel.tissue));
+
+    % determine all hexaheders that are labeled as "brain"
+    insidehex = false(size(headmodel.tissue));
     for i=1:numel(tissue)
-      fprintf('selecting source positions inside "%s"\n', tissue{i});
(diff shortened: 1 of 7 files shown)


forward/ft_inside_vol.m
```diff
@@ -120,59 +120,74 @@
...
-    for i=1:numdip
-      hexindx = find(hex2pos(:,fullindx(subindx(i))));
+    for i=1:numel(insidedip)
+      hexindx = find(hex2pos(:,insidepos(subindx(i))));
       for j=1:numel(hexindx)
         posx = headmodel.pos(headmodel.hex(hexindx(j),:),1);
         posy = headmodel.pos(headmodel.hex(hexindx(j),:),2);
         posz = headmodel.pos(headmodel.hex(hexindx(j),:),3);
         pos = [posx posy posz];
         if isequal(convhull(pos), convhull([pos; dippos(i,:)]))
```
robertoostenveld: By default all dipoles are tagged as outside as it initializes with inside=false(1,numdip).

Being inside one volume element is enough to show that it is inside brain... So I don't understand your comment yet.

simonhomoelle: The problem occurs if a source falls into an non-brain element that have a sharing face with a brain element.  I.e. A source point is inside such a non-brain element. We found the nearest neighbouring vertices and look up all elements that share this vertex. After that we only checked if it inside one of this 8 element not if this element is part of the brain.

robertoostenveld: Ah, the non-brain elements were pruned away in the past, but not any more.

So the check
if isequal(convhull(pos), convhull([pos; dippos(i,:)]))
should only be done for brain hexahedra.

robertoostenveld: or better (more efficient): we should prune the mesh both for hexaheders and for vertices. The remaining elements in that case are all brain elements.

PR #261, opened 2016-11-16
Bug3207 - improve support for EDF annotation channel

robertoostenveld:
as discussed by email and in http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3207


robertoostenveld:
note that commit 9ebbecc91c5701acddb9f5f8373e2062c4ad5741 should not be part of this PR. But it should be harmless, so I'll leave it in.

PR #263, opened 2016-11-16
bugfix - fixed 'siz' variable namechange

mrkdsmith:
Fixed a big where the variable 'siz' was not recognised. Replaced instances of 'siz' with 'size_avg'.
Thanks
Mark


robertoostenveld:
thanks!

PR #266, opened 2016-11-22
Improvements for Nicolet

janbrogger:
This is an improvement in reading Nicolet (Nervus) files. There is one technical fix, related to handling of TSINFO packets. The other is that we can read the montages from the files.


robertoostenveld:
thanks

PR #267, opened 2016-11-23
groundwork for neuronal variance

StolkArjen:


StolkArjen:
Not yet implemented, but ideally cfg.method = 'neuvar' also performs a tfr/mtmconvol-like analysis when cfg.toi is specified. Now it does the same as mtmfft, assigning resulting variance to freq = 0.

PR #273, opened 2016-12-09
Frebase

schoffelen:


schoffelen:
Hmmm, this PR seems to have a few strange commits, which however seem to pertain to me merging the master branch

PR #274, opened 2016-12-12
Read BioSemi status channel functionality added

edden-gerber:
This branch adds functionality to the ft_read_hdr function, which parses the BioSemi status channel. This channel is appended as the last data channel, and contains multiple (nested) fields within each sample's 24 bits, the most important of which is the 8-bit trigger channel (but other fields may also be useful, such as the StartEpoch flag). Attempting to read the trigger values directly from the status channel may lead to various problems: the quick and dirty fix is just subtracting the fixed value which the other fields add to the 24-bit numeric value, but it doesn't always work since the other fields can also change during the recording). 

The trigger channel is parsed in a way which removes repeated values and spurious values which can appear if the trigger value is changed halfway within a sample. 

The changes include an additional private function read_biosemi_status_channel, and additional lines of code in ft_read_hdr. 

The output status channels fields are stored in hdr.statusChannels. 

Example: 

>> hdr = ft_read_header('my_EEG_data.bdf');
>> hdr.statusChannels

ans = 

        Triggers: [3184640x1 double]
     ExtTriggers: [3184640x1 double]
      StartEpoch: [3184640x1 double]
           Speed: [3184640x1 double]
      CmsInRange: [3184640x1 double]
      LowBattery: [3184640x1 double]
    ActiveTwoMk2: [3184640x1 double]

### DIFF
--- a/fileio/ft_read_header.m
+++ b/fileio/ft_read_header.m
@@ -31,6 +31,9 @@ function [hdr] = ft_read_header(filename, varargin)
 %   hdr.FirstTimeStamp      number, represented as 32 bit or 64 bit unsigned integer
 %   hdr.TimeStampPerSample  number, represented in double precision
 %
+% For the BioSemi BDF format, an additional statusChannels field is
+% returned containing the BioSemi Trigger channel and control fields. 
+%
 % For continuously recorded data, nSamplesPre=0 and nTrials=1.
 %
 % Depending on the file format, additional header information can be
@@ -407,6 +410,10 @@ switch headerformat
         hdr.label(chan) = labelnew(chan);
       end
     end
+    % Read the BioSemi status channel and parse it
+    statChanData = read_biosemi_bdf(filename, hdr, 1, hdr.nSamples, hdr.nChans); % Read last channel
+    statusChannels = read_biosemi_status_channel(statChanData);
+    hdr.statusChannels = statusChannels;
     
   case {'biosemi_old'}
     % this uses the openbdf and readbdf functions that were copied from EEGLAB
--- a/fileio/private/read_biosemi_status_channel.m
+++ b/fileio/private/read_biosemi_status_channel.m
@@ -1,0 +1,148 @@
+function [ data ] = read_biosemi_status_channel( statChan )
+% This function recieves the BioSemi status channel (the last channel in 
+% the data matrix read from the BDF file), and extracts the individual 
+% control channels multiplexed within it. See the biosemi website (http://
+% www.biosemi.com/faq/trigger_signals.htm) for documentation of this 
+% channel structure. 
+% 
+% The following control channels are read:
+% - Trigger Value
+% - Extended Trigger Value
+% - Start Epoch Flag
+% - Speed Value
+% - CMS Within Range Flag
+% - Battery Low Flag
+% - ActiveTwo MK2 Flag
+%
+% For the trigger channel, the function treats a sequence of idential 
+% triggers is read as a single trigger (the first one).
+% The function also takes into account the possibility that the status 
+% message will change in the middle the sample, resulting in a spurrious 
+% intermediate sample where only some of the bits have been updated (for 
+% instance, when a trigger value of 10 is sent through a parallel port, the 
+% recorded values could be [0 0 0 8 10 10 10], if during the first sample 
+% the "2" bit is read before the value was changed but the "8" bit is read 
+% after. In such cases, the intermediate value is ignored. Therefore, a
+% valid trigger is recorded when: 
+% 1) There is a change to a non-zero value followed by zero or more 
+% identical values (e.g. the third value in [0 0 10 10 10] or 
+% [0 0 10 0 0]). 
+% 2) If there are two consecutive non-zero values, the second one is the
+% valid trigger (e.g. the third value in [0 8 10 10 10] or [0 8 10 0 0]). 
+%
+%
+% Written by Edden Gerber, lab of Leon Y. Deouell, June 2013
+
+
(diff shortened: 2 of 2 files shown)


robertoostenveld:
ft_read_header, data and event are being used by many Biosemi users, that apparently don't have problems getting their data segmented. I am in general not in favor of extending a generic interface with system specific details, as that tends to be hard to document, propagate in unexpected ways, and causes maintenance problems. 

Can you start by describing what the problem is that you are actually trying to solve? Might it be that a sequence of functions such as 

hdr = ft_read_header(filename)
status = ft_read_data(filename, 'chanindx', find(strcmp(hdr.label, 'STATUS'))
biosemidetails = status2details(status)

would also do it? If so, the status2details function should be added to the getting started with biosemi page on the wiki.

robertoostenveld:
now having seen the emails... did you look at the fileio/private/read_trigger code? That is the generic place to parse and convert a continuously sampled trigger/status channel into FT like events. Not all details of the status channel are supported yet, but there is a hint at it with "if this code ever needs to be enabled ...". Perhaps this is the time. 

Please look into that code first and consider how ft_read_event is called in the standard pipeline
ft_definetrial -> ft_preprocessing -> etc.
Adding the proposed details to the header field does not help the generic user, as you still need a custom trialfun. So better would be to
1) extend the FT events with battery, cmsinrange etc., or to
2) add a fieldtrip/trialfun/ft_trialfun_example_biosemi.m that contains the parsing

edden-gerber:
* About the problem itself: I'm not sure, buy it may be that this never came up as an issue because it's easy to deal with it in a "quick and dirty" way - just subtract a constant from the channel value (and sometimes maybe remove some spurious values caused by changes in the other fields of the status channel). But for anyone using BioSemi+FT the proper solution definitely can't be found on the wiki page (http://www.fieldtriptoolbox.org/getting_started/bdf). 

* status2details: I could not find this function anywhere...? 

* fileio/private/read_trigger: This function is called by ft_read_event which is the function I was looking for - it can read a BioSemi data file and generate an event list. - however it doesn't do it 100% right, there is some issue with how the 16 bits are read from the 24 bit sample (lines 344-366). So for my data at least, I needed to subtract 65280 from the trigger value to get the correct value. 

So I believe the best solution would be: 
1) Change the code section of ft_read_event for BioSemi so it handles the data correctly. It shouldn't be too hard, it's a matter of reading the sample so that the 8 unused upper bits of the 16-bit trigger field are indeed read as 0's (lines 344-363; as they are read now, they add 65280 to the trigger value encoded in the lower 8 bits). I can't say right now how exactly to change the code as I didn't go into the precise logic of your implementation. 
2) It think it would be beneficial to mention ft_read_event in your BioSemi wiki page, since otherwise I assume it's easy to not miss the fact that this is a very relevant function for any BioSemi+FT user.

StolkArjen:
Sorry for my curtness, recording ecog;

Is this reason why people first send out a zero trigger before sending the actual desired code, Edden?

Function doesnt exist: its a suggestion by Robert for adding your func.

1) cannot zoom into the details right now

2) read event is a general workhorse, called by read hdr for some formats. But apparently not thr case forbbiosemi. Would be good to have an additional trigger related section on that getting started page. Can you give it a try, edden?

robertoostenveld:
and the documentation on http://www.fieldtriptoolbox.org/getting_started/bdf also does not exist yet, it was also my suggestion to add new documentation there. A change in low-level code will not be found by other users, whereas the documentation will.

edden-gerber:
* I'm not sure about the reason for sending a 0 before the desired code. 

* status2details - yes, if so then my read_biosemi_status_channel function could work this way. 
It could be the simplest solution - just add it as a standalone function and refer to it in the wiki. 

* Actually ft_read_hdr never calls ft_read_event directly. And ft_read_event _does_ have a biosemi-specific code section, it's just not implemented 100% correctly. So if I run 
>> events = ft_read_event(my_data.bdf);
I get an events array, but for example, this trigger is read as 65482 instead of 202: 
>> events(10)
ans = 
        type: 'STATUS'
      sample: 8335
       value: 65482
      offset: []
    duration: []

robertoostenveld:
I think the existing code is correct (unless proven otherwise), but it is not complete.

edden-gerber:
So what would you suggest is best - add a standalone _read_biosemi_status_channel_ function, or look at _ft_read_event_ to make sure it reads the biosemi triggers correctly?

StolkArjen:
Hey Edden,
Do you think you could add your thoughts regarding resolving those trigger values, and possible example scripting, as a subsection to http://www.fieldtriptoolbox.org/getting_started/bdf? We could then take it from there (since the issue is not entirely clear to me), in terms of additional code?
Yours, Arjen

edden-gerber:
Hi Arjen, 
sure, I could do that. I'll send you the text in the near future.

robertoostenveld:
At the toolkit course this week I looked along with a participant who brought BDF files. They worked fine, i.e. after ft_read_event all trigger codes were there. The normal pipeline ft_definetrial->ft_preprocessing->etc worked for her. 

So it is unclear what this PR tries to solve. Note however that @dmatthes1982 in PR #395 also made a biosemi related suggestion. I actually do not understand why Biosemi users appear to be constantly trying to circumvent the ft_read_event function. I have just added a section to http://www.fieldtriptoolbox.org/getting_started/bdf to demonstrate how to do the preprocessing.

robertoostenveld:
Due to the lack of progress on this, I am closing this PR. Feel free to reopen and clarify what needs to be fixed or improved.

PR #275, opened 2016-12-13
Estimate noise properly with rank-deficient covariance matrices (Bug 3162)

sarangnemo:
Estimate noise using the singular value corresponding to the rank of the covariance matrix; ensures that rank-deficient covariance matrices no longer yield a noise estimate of 0. Refers to Bug 3162:
http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3162


sarangnemo:
I'm not sure why the ft_sloreta commits are showing up -- I believe they were already merged in a separate pull request at the time:
https://github.com/fieldtrip/fieldtrip/pull/237

schoffelen:
Hi Sarang, the other commits showing up don't seem to be related to code changes. I came across this also recently, but didn't understand it either.

Regarding your code changes, I don't fully understand them. First, you explicitly compute the rank of the Cy/Cf matrix. Then, this variable is only used if the data are NOT rankdeficient. Isn't this defying the purpose? (I may miss something here)...

sarangnemo:
Hi Jan Mathijs, on the contrary, the variable is now *always* used, since I removed the explicit "if isrankdeficient" condition.

In the case that the covariance matrix is full-rank, then the last singular value is selected as the noise estimate, as before. If it is rank-deficient, the singular value now corresponds to the last significantly non-zero singular value.

If the user has manually defined a lambda value, then the noise is chosen to be the larger of the selected singular value or the lambda. This is the same behavior as before for a full-rank covariance; this may change output for a rank-deficient covariance but makes the behavior consistent with what happens with full-rank covariance.

schoffelen:
My bad. Thanks for the clarification. It seems my brain is fried, I didn't read the changes correctly...

PR #276, opened 2016-12-15
FIX: bug3221, error reporting in ft_definetrial and warning reporting in ft_getuserfun

lennartverhagen:
See bug 3221, concerning ft_definetrial.

When a function specified by cfg.trialfun does not exist ft_getuserfun
replaces cfg.trialfun by an empty double (and gives a false double
negative warning while doing so), yet the subsequent error message
assumes cfg.trialfun is still a char.


schoffelen:
Thanks Lennart! Keep the contributions coming.

PR #279, opened 2016-12-19
Estimate noise properly with rank-deficient cov matrices (Bug 3162)

jens-k:
...for PCC, analogous to recent changes to beamformer_lcmv and _dics in commit 473c9d1ea14bee63f258479eee34cd122e4518da / pull #275

Runs for me without errors, please double-check anyway.


schoffelen:
Looks good to me! Thanks.

PR #282, opened 2016-12-20
Databrowser

schoffelen:
Does anybody know how to get rid of all these old commits being mentioned. It looks as if I need to do some rebasing (and that my 'master' (which is a fork from fieldtrip/fieldtrip) contains these commits in its history, although they are not associated with any code changes.


robertoostenveld:
To prevent them you would do a git pull (or merge from master, when working on a branch)  prior to making code changes.

I think that https://git-scm.com/docs/git-rebase would remove the "empty" commits.

PR #283, opened 2016-12-20
Databrowser fix

schoffelen:
I tried to do various combinations of git reset / revert / rebase, but to no avail.

Would it be OK to merge this one, with the superfluous commits in the commit history,

and my promise to not let it happen again?

@robertoostenveld


schoffelen:
I followed the following recipe: http://stackoverflow.com/questions/6102297/how-do-i-clean-up-my-github-fork-so-i-can-make-clean-pull-requests

but am still a bit hesitant to squash and merge, although it indeed seems that in this PR only 2 files are changed, although the commit report of the 'test something' commit looks scary

robertoostenveld:
It is now well documented, so please go ahead and merge.

PR #287, opened 2016-12-28
using external reading functions in fileio

roemervandermeij:
I have a proposition. I'm writing reading functions for some dataset, in its own particular non-recording-equipment related dataformat. I don't think it's a good idea to add such idiosyncratic datasets to ft_read_data/event/header. So, how about the possibility of using external unsupported reading functions? 

In this pull request is an easy and safe way to allow for this. In side the 'otherwise' of the main case-switch, dataformat/headerformat/eventformat is attempted to be executed as a function in a try-catch. If it fails, the regular failure message is given. Reasonably safe I would say. (Initially I thought of feeding xxxformat as a function handle, to prevent rare unintended executions, but the switch-case cannot handle that type). 

What do you think?

If you agree, I'd also happily remove the 'nmc_archive_k''s (idiosyncratic data like the above), which I added years ago for my work with some datasets. They are no longer used at the Donders, and the external people that depend on them will send me an email if they still use them.

### DIFF
--- a/fileio/ft_read_data.m
+++ b/fileio/ft_read_data.m
@@ -31,6 +31,9 @@ function [dat] = ft_read_data(filename, varargin)
 %
 % The list of supported file formats can be found in FT_READ_HEADER.
 %
+% To use an external reading function, use key-value pair: 'dataformat', FUNCTION_NAME.
+% (Function needs to be on the path, and take as input: filename, hdr, begsample, endsample, chanindx.)
+%
 % See also FT_READ_HEADER, FT_READ_EVENT, FT_WRITE_DATA, FT_WRITE_EVENT
 
 % Copyright (C) 2003-2016 Robert Oostenveld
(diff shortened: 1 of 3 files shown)


roemervandermeij:
I put some documentation on the top of each of the functions, but I assume you mean something higher level? I'm a bit hesitant to make it obvious to the naive user (by e.g. a wiki tutorial mention), as the low-level reading functions they have to use can be tricky and need to be done well. How about a short FAQ?

schoffelen:
I think that that's fine.

roemervandermeij:
Info has been added to:
http://www.fieldtriptoolbox.org/faq/how_can_i_import_my_own_dataformat?&#extend_the_fieldtrip_reading_functions

Pull away!

roemervandermeij:
nudge nudge ;)

robertoostenveld:
hé, wat ontdek ik hier nu in de code... dat is mooi!  In de kerstvakantie in 2016 is me dit blijkbaar niet opgevallen ;-) 

Zie https://gist.github.com/robertoostenveld/bf05b9ba0f7e33f14df9f6ae753de66c waar ik een toepassing hiervoor aan het uitproberen ben.

PR #288, opened 2017-01-03
bugfix: casesensitivity did not work when cfg.channel was specified

StolkArjen:


StolkArjen:
Actually the other way around: cfg.channel produced no result when casesensitivity = 'no' (default), or the original labels were not used (producing lowercase labels).

PR #290, opened 2017-01-03
again getting stuck with these empty commits, just PR'ing it, to get rid of them

schoffelen:


roemervandermeij:
Hmmmm. How do you update your own branch before committing to it?

I do:
1a) git checkout -B branchwithchanges
1b) git pull fieldtripmain master (or some other up to date copy)
2) commit changes
3) git push fieldtripme branchwithchanges
4) pull request on github
The order of 1a/b typically doesn't matter, AFAIK. 

Updating my working branch I don't do on github. Whenever I do, I get 'empty commits' such as the one you mention. Did you do this perhaps?

Or, an alternative reason. When the same function is edited/committed/merged by someone else before I merge my changes, even if the changes don't conflict, I cannot do a 'clean pull' (IIRC). I assume this is because there is ambiguity on which commit comes first in the tree, yours or the other person's, so there has to be some record of adding the commits to the commit tree in a particular order. I have the feeling that such a situation is more complicated though.

roemervandermeij:
That's how I've been doing my PRs. It's feels a bit cumbersome, but I don't see any other way...

schoffelen:
Oh, and in case your master gets messed up, you can get it cleanly in sync with the upstream master, by following the recipe outlined here: http://stackoverflow.com/questions/5916329/cleanup-git-master-branch-and-move-some-commit-to-new-branch

PR #294, opened 2017-01-08
ENH: output labels for a point of interest (cfg.roi)

StolkArjen:
Note that an extra cfg option needed to be created (cfg.output = 'label') to discriminate between the existing mask2label and the new roi2label functionalities. In future, cfg.output could be extended to support both 'mask' and 'label', and the function re-organized accordingly.


StolkArjen:
tested with test_ft_volumelookup and runs fine

PR #296, opened 2017-01-17
Bug3027 homer

robertoostenveld:
last week Monday I spent some time with Jenny to implement the Homer reading support. This commit includes the basic reading  support. The "opto" structure might not be 100% perfect yet.


robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3027 for details.

PR #302, opened 2017-01-19
Fixes read of Nervus data with mixed sampling rates

janbrogger:
- Fixes read of data with mixed sampling rates - only the channels with the most prevalent sampling rate is read.
- Added duration of gaps between EEG segments as boundary events
- Add more annotations


robertoostenveld:
thanks, I think that picking the most prevalent sampling rate is also consistent with EDF. 

Can you provide us with an example data set and a test script? Preferably one file with constant, and one file with different sampling rates. Right now we don't have any data with which we can check that it works as expected.

janbrogger:
Sorry for the delay. I love the idea of a test script. MATLAB is kind of new to me. Can you provide an example of a test script and ideally an idea of file layout and file naming. E.g. does my test have it's own folder, sane naming of files. That way it will be easier for other developers if the test does crash.

Also, note that as of now this code only reads Nervus files from about 2012 and onwards. Older files crash. I tried to parse the file format but they did some major changes to the file format so I gave up for now. Where do I put that kind of documentation?

robertoostenveld:
in fieldtrip/test there are many examples. A simple test script would look like this

```
function test_name

% MEM 3gb
% WALLTIME 00:10:00
% the mem and wall time are for our compute cluster, where we run the tests regularly. 

filename = 'your_file.ext' 
% I will change this once I receive the file through dropbox, filesharing, or so
% the binary file is not part of the git commit, but I'll put it somewhere on our network drive

hdr = ft_read_header(filename)
evt = ft_read_event(filename)
dat = ft_read_data(filename)

assert(1==1) % this passes
assert(1==2) % this gives an error

assert(size(dat,1)==hdr.nChans)
assert(size(dat,2)==hdr.nSamples)

etc.
```
there can of course be multiple files to test. 

See for example 

```
mac011> grep -l edf *.m
ref_datasets.m
test_bug1254.m
test_bug1924.m
test_bug2887.m
test_bug2954.m
test_bug3207.m
```

robertoostenveld:
Hi @janbrogger, there is a conflict in the code of ft_read_data. It does not look serious, but I think that it is easiest if you fix it.

robertoostenveld:
ping @janbrogger, can you have a look at this conflict? You know the code, so it should not be hard to solve.

robertoostenveld:
ping @janbrogger, can you have a look at this conflict? You know the code, so it should not be hard to solve.

janbrogger:
Will do it on Friday. Thanks for the pingback.

robertoostenveld:
There are a lot of changes here. Without a test script and test data, it is not possible to see whether it all works as expected. 

Can you please share at least one Nervus dataset in which all channels have the same sampling rate and one dataset in which they have different sampling rates? And then a test script that calls ft_read_header/data/event on both of them, plus calls ft_databrowser on both of them.

janbrogger:
Sorry that I re-activated this pull request. Working on a test script. I have some test data in Nervus, EDF and ASCII. Should this test data be committed? In its own folder??

janbrogger:
I added the tests you requested. The databrowser opens correctly. I also added two tests that compare the fieldtrip-read data with equivalent data that have been ASCII exported from Nicolet.

I don't have any data easily available  in which all channels have the same sampling rate - all our files have this problem. 

Note that I had to hardcode a local path in the test script test_nicolet_reading (''C:\Midlertidig_Lagring\fieldtrip\test\private\test_nicolet_reading\') which should be replaced with your "DCCN" path. The committ has a zip archive attached with the test data. I don't know how you do it.

robertoostenveld:
This has been replaced by pull request #1166

PR #303, opened 2017-01-23
TST: minor fixes in tests

nno:
For the tests in the 'test' directory:
- renamed some functions so that the filename matches the function name
- fixed some syntax errors when using dccnpath


robertoostenveld:
thanks

PR #306, opened 2017-01-24
ENH: use nanmean/nanmedian in common avg/median rereferencing to supp…

StolkArjen:
…ort ft_channelrepair method nan

### DIFF
--- a/preproc/ft_preproc_rereference.m
+++ b/preproc/ft_preproc_rereference.m
@@ -54,9 +54,9 @@ end
 
 % compute the average value over the reference channels
 if strcmp(method, 'avg')
-  ref = mean(dat(refchan,:), 1);
+  ref = nanmean(dat(refchan,:), 1);
 elseif strcmp(method, 'median')
-  ref = median(dat(refchan,:), 1);
+  ref = nanmedian(dat(refchan,:), 1);
 end
 
 % apply the new reference to the data


schoffelen:
Would this be a safe change? In itself, it leads to numerically stable (or in some cases predictable) behavior. Yet, what happens if: 

- not all time points are nans for a given channel?
- the location of the nans are not consistent across trials (which the function is unaware of)?

Each of these cases would break the correctness of the balancing for potential subsequent forward modelling (as, by the way, would the 'median' as method do). Obviously, if the referencing is done in the context of ECoG analysis, which is typically not subsequently subjected to source reconstruction, the consequences are probably minor. 

Yet, we may need to keep in mind that EEG people may also run into this functionality.

@robertoostenveld : should we allow this function to behave transparently for nans, or should we prevent the naive user for incorrect behaviour. I suspect that just throwing a warning will go unnoticed.

robertoostenveld:
warnings would indeed not be noticed. I am fine with supporting this behavior. In most cases I think it is appropriate. 

If we would want to do something with it, we might want to make it a global option to allow or disallow nans in these kind of computations and add a nancheck like this

assert(istrue(cfg.allownan) || ~any(isnan(dat(:))), 'nan is not allowed in the data')

prior to each call of nanmean, nanvar, nanstd, etc (which is either ours, or the one from mathworks). 

but that goes beyond the scope if this suggestion.

schoffelen:
…or incorporate this nan-checking business into a ft_mean and ft_median etc function ...

StolkArjen:
Just wanted to add that this is - personally - not necessarily a highly desired feature. Simply realized that since ft_channelrepair as of recent supports nan'ing channel data, issues will occur when the user tries to re-reference the data afterwards.

The thing boils down to how normal it is for nans to be distributed inconsistently across channels and time, as this will indeed lead to unequal estimation of the common avg/median over time. I could not imagine this to happen. But then again, I have seen interesting pipeline creatures lately hehe.

robertoostenveld:
I added an option "handlenan" and kept the default behavior identical. See also test_bug2732

PR #309, opened 2017-01-26
adds the choice of whether the source plot should be visible or not

ualsbombe:
is useful if you just want to save

### DIFF
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -45,6 +45,7 @@ function ft_sourceplot(cfg, functional, anatomical)
 %   cfg.downsample    = downsampling for resolution reduction, integer value (default = 1) (orig: from surface)
 %   cfg.atlas         = string, filename of atlas to use (default = []) see FT_READ_ATLAS
 %                        for ROI masking (see "masking" below) or in "ortho-plotting" mode (see "ortho-plotting" below)
+%   cfg.visible       = 'on' or 'off' (default = 'on'): whether figure will be visible or not
 %
 % The following parameters can be used for the functional data:
 %   cfg.funcolormap   = colormap for functional data, see COLORMAP (default = 'auto')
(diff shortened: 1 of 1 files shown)


roemervandermeij:
In that case, make sure the properties are set upon generation figure(key,value). In you suggestion, it will be visible momentarily.

robertoostenveld:
sorry for my ignorance, but why would you want to do that? Why did we have that option (with a fixed value) in the first place? Probably/supposedly there is a good reason for it, does that reason also apply to other figures?

robertoostenveld:
cool, this is funny. Ok, I might get it...

```
>> h = figure('visible', 'off')
>> plot(randn(10))
Warning: MATLAB has disabled some advanced graphics rendering features by switching to software OpenGL. For more information, click here. 
>> set(h, 'visible', 'on')
```

@ualsbombe, are you switching it to visible at a later stage, or are you saving it to disk without looking at it first?

ualsbombe:
roemer: I may have to retract my pull request, because I realized that if it is not visible, nothing is actually saved.
robert: I wanted to do that because I wanted to run a loop creating lots of figures to save them but not see them. The reason is that it is annoying to have them pop out.
robert second comment: I have not switched it back, no. I don't know if it makes difference.

robertoostenveld:
please test a bit further, it would be cool if this works since I see some use for it elsewhere.

ualsbombe:
Ok, will try a bit more.

ualsbombe:
Ok, I followed @roemervandermeij 's suggestion

And it is possible to load invisible figures and make them visible afterwards

`%% create invisible figure
h = figure('visible', 'off');
plot(randn(10));

%% save invisible figure
savefig(h, 'lala.fig');

%% load invisible figure
h2 = open('lala.fig');

%% make figure visible againg
set(h2, 'visible', 'on');`

robertoostenveld:
thanks, I'll add this. For consistency I will also add it to other plotting functions that open their own figure.

PR #315, opened 2017-02-01
Bug3235 ft_volumereslice support for flip and permute

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3235


robertoostenveld:
@schoffelen I have followed http://www.fieldtriptoolbox.org/development/git?&#development_workflow, which means that it should now be ready to merge into master. Agreed?

PR #317, opened 2017-02-03
ENH+RF: make webwrite throw an error in case of non-success and other minor changes

nno:
- make webwrite throw an error in case of non-success
- display he error message if it fails
- added documentation explaining the command and differences from builtin Matlab
- mark third argument as unused
- some refactoring making the code shorter and reducing code duplication

### DIFF
--- a/compat/matlablt2015a/webwrite.m
+++ b/compat/matlablt2015a/webwrite.m
@@ -1,12 +1,26 @@
-function status = webwrite(varargin)
+function status = webwrite(url, data, ignored_options)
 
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Thanks. Does this also work for octave?

robertoostenveld:
sorry, I pressed the wrong button. It is now merged.

nno:
Thanks for the merge. It runs for Octave on my machine, I will now try and see if travis can use it too.

PR #318, opened 2017-02-04
BLD+OCTAVE: do not include matrix.h if mex.h is included

nno:
mex.h includes matrix.h, at least on Octave and recent versions of Matlab. Octave complains if matrix.h is included due to a naming conflict.
I have not been able to test this on old versions of Matlab (2009).

Files that include matrix.h but not mex.h (and are therefore not changed):

```
    ./peer/src/peerslave.c
    ./peer/src/test_engine.c
    ./peer/src/test_peerslave.c
    ./src/geometry.c
```

Should those include mex.h? It would probably make them more Octave compatible.

For reference, the following bash script was used:
```
TMP=/tmp/foo

for i in `find . -name '*.c'`; do 
  if [ `grep "#include \"matrix.h\"" $i | wc -l` -gt 0 ]; then 
    if [ `grep "#include \"mex.h\"" $i | wc -l` -gt 0 ]; then
      echo "Fix: $i"
      grep --invert-match  "#include \"matrix.h\"" $i > $TMP
      cat $TMP > $i
    else
      echo "Skip: $i"
    fi
  fi
done
```


nno:
I removed the import in src/geometry.c because it seems unnecessary there as well.

nno:
I did not manage to compile code in ``peer/src``, see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3243.

Therefore I don't know if the import of matrix.h can be removed safely.

nno:
There were also some files using <matrix.h> instead of "matrix.h", these have been identified and fixed now as well.

robertoostenveld:
I am not maintaining the code in external and not willing to spend time on it. Especially the code in external/dmlt is a collection of mex files from all over the place. I don't even know if recompile scripts are available. Compiling on 5 platforms (osx64, win32, win64, linux32, linux64) with a consistent MATLAB version of a few years old is a lot of work.

Can you revert all changes that are in non-core fieldtrip code?

Changes to main FT mex code are welcome if you prepare them as a clean PR, but it will take some time before the corresponding mex files get recompiled.

robertoostenveld:
see also http://www.fieldtriptoolbox.org/development/guidelines/code#compiling_mex_files and http://www.fieldtriptoolbox.org/?do=search&id=mex. The documentation on the wiki should be updated, feel free to do so.

nno:
@robertoostenveld I would be happy to make revert changes to non-core fieldtrip code. What is non-core, however? According to bugzilla, "The core consists of the *.m files in the following directories: fieldtrip, private and public.". A literal interpretation would mean that no files should be changes, as all changes here are *.c files, which I assume is not what you meant. 

Files that I would consider ``core`` would include these directories:
- @config
- compat
- src

Not in core:
- contrib
- external
- peer
- realtime

Is that correct?

nno:
ping @robertoostenveld, which file locations would you consider 'non-core'?

robertoostenveld:
all directories that can (theoretically) serve as a stand-alone toolbox are not "core". Those are also the ones shared with eeglab and spm.

But the distinction is not so clear and changed over time (as requirements changed, and since different people have different opinions on how stuff should be organized). Let me try to split it 

```
./connectivity
./fileio
./forward
./inverse
./plotting
./preproc
./specest

./contrib
./external
./realtime

./engine
./peer
./qsub

./compat (should hardly ever need changes, only when end-user oriented functions are renamed)

./bin
./src
./test

./*.m (i.e. top level m-files)
./@config
./private
./statfun
./template
./trialfun
./utilities
```

The src directory and the handling of mex files at one location is primarily for convenience, and not consistent with the further organization.

nno:
Thanks @robertoostenveld, in the list above, where is the split? Is ``compat`` and everything below it 'core', and everything above ``compat`` is 'non-core'?

robertoostenveld:
There is no 100% clear divide. But the last part is certainly core. Bin/src/test also do not belong to anything else, so are also core.

compat is along a completely different dimension. It only serves to keep backward compatibility when we change function names that the user might have included in their scripts, and when Mathworks changes stuff. I don't see a reason why you would make changes there (unless it pertains to octave/matlab compatibility and matlab version compatibility). But the ideal user would not have compat on his/her path.

nno:
Thanks @robertoostenveld, I created a fresh PR at #324 and will close this one.

PR #319, opened 2017-02-07
enhancement - support varying frequency resolution for the sine taper

vlitvak:


robertoostenveld:
I reviewed the code, it looks fine. There is no test script that involves the sine taper so I did not perform any regression tests.

PR #320, opened 2017-02-07
ENH - improved coordsys and units in atlas and surface files used in ft_sourceplot

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2032

@schoffelen can you please review and merge?


schoffelen:
Reading through the changes all looks fine to me. Given that it is explicitly stated that the testscript currently fails, I don't see much need to pull the branch and to test it myself (apart from the fact that this would be useful as a git-exercise). 

I guess that one obvious way to move forward would be to add a coordsys field to the cortical sheets that are in fieldtrip/template/anatomy before merging this. As far as I can see, this would allow the failed_test_function to be upgraded to a proper test function

robertoostenveld:
So you agree that coordsys should be added to the folded surfaces in the mat files? If so, I'll add that before the merge. I presume that all are in MNI coordinates.

That makes the test script run, but does not resolve the original issue yet: that requires that ft_sourceplot checks the units and coordsys. I can do that shortly after ft_read_headshape.

schoffelen:
Yes, I think indeed it's all 'spm', or 'mni'. FT uses this interchangebly.

May I suggest you to add the coordsys to the folded surfaces, so that this can be merged. We can look into the original bug at a later moment (I can do that as well).

robertoostenveld:
Since by now this issue involves changes to core functions such as ft_checkdata, ft_datatype and ft_sourceplot, I think it is better to do a full test run on the branch. I have just started the regression testing batch on commit 8a31d96. It should be done by tomorrow morning.

robertoostenveld:
hmm, glad that I did the test. I already started getting these:

test_bug2096.txt:=== test_bug2096 FAILED in 6 seconds
test_bug2460.txt:=== test_bug2460 FAILED in 0 seconds
test_bug2556.txt:=== test_bug2556 FAILED in 3 seconds
test_bug2623.txt:=== test_bug2623 FAILED in 1 seconds
test_bug3190.txt:=== test_bug3190 FAILED in 9 seconds

robertoostenveld:
I started a new test batch for commit fd533bf.

robertoostenveld:
I started a test batch for commit 2f5fea6.

schoffelen:
Just for my education. When you say: "I started a test batch" do you run a full schedule-tests.sh? Or a subset pertaining to the possible affected functions?

robertoostenveld:
I reviewed "test_tutorial_beamformingextended" that failed once with 2f5fea6 (due to coordsys mismatch). It does not fail any more, even though I did not change the code. I don't understand the cause of the failure, but think the code is fine.

PR #321, opened 2017-02-07
bugfix - moved the code extracting direction for Neuromag

vlitvak:
This code previously ran even when not required and broke plotting for
any balanced neuromag sensors.

### DIFF
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -209,26 +209,6 @@ if istrue(orientation)
   end
 end
 
-% determine the rotation-around-the-axis of each sensor
-% only applicable for neuromag planar gradiometers
-if ft_senstype(sens, 'neuromag')
-  [nchan, ncoil] = size(sens.tra);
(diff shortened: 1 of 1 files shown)


schoffelen:
Looks like a clean fix to me. I guess that this patch solves http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3231

schoffelen:
I have pulled Vladimir's branch and am currently testing it. As it turns out, it still throws an error when trying to plot squares for the planar gradiometers with a balanced sensor array in the input.
Using 'coilshape', 'point' turns out OK. I tried to fix this by including an explicit call to unbalancing, which crashes due to some unit conversion issues in ft_apply_montage. This latter problem could be a genuine issue, or caused by my quick and dirty construction of a balanced sensor array (basically by hand)

Bottom line: this patch fixes an issue but my looking into this potentially uncovered another issue, which should be followed up somewhere else

schoffelen:
Thanks for the fix Vladimir!

PR #323, opened 2017-02-10
ft_volumenormalizes now properly restores warning settings

thht:
hi,
ft_volumenormalize sets warnings to off but never restores it.


robertoostenveld:
thanks!

PR #324, opened 2017-02-12
OCTAVE: in "core" functions, do not include matrix.h if mex.h is included

nno:
mex.h includes matrix.h, at least on Octave and recent versions of Matlab. Octave complains if matrix.h is included due to a naming conflict.

Adapted from #318, but this PR does not touch non-core files. I will close #318 shortly.

For reference, the following bash script was used to make the changes:
```
TMP=/tmp/foo

CORE="./@config
./private
./statfun
./template
./trialfun
./utilities
./compat
./bin
./src
./test"


for subdir in $CORE; do 
  for i in `find ${subdir} -name '*.c'`; do
    for matq in \" \<; do
      if [ $matq = "<" ]; then
        matc=">"
      else
        matc=$matq
      fi
      for mexq in  \" \<; do
        if [ $mexq = "<" ]; then
          mexc=">"
        else
          mexc=$mexq
        fi
        mat="#include ${matq}matrix.h${matc}"
        mex="#include ${mexq}mex.h${mexc}" 
        if [ `grep "$mat" $i | wc -l` -gt 0 ]; then 
          if [ `grep "$mex" $i | wc -l` -gt 0 ]; then
            grep --invert-match "${mat}" $i > $TMP
            cat $TMP > $i
            echo "Fixed: $i"
          else
            echo "Skip: $i"
          fi
        fi
      done
    done
  done
done
```

### DIFF
--- a/@config/private/deepcopy.c
+++ b/@config/private/deepcopy.c
@@ -11,7 +11,6 @@
 
 #include <math.h>
 #include "mex.h"
-#include "matrix.h"
 
 void
 mexFunction (int nlhs, mxArray * plhs[], int nrhs, const mxArray * prhs[])
--- a/@config/private/increment.c
+++ b/@config/private/increment.c
@@ -8,7 +8,6 @@
 
 #include <math.h>
 #include "mex.h"
-#include "matrix.h"
 
 void
 mexFunction (int nlhs, mxArray * plhs[], int nrhs, const mxArray * prhs[])
--- a/@config/private/setzero.c
+++ b/@config/private/setzero.c
@@ -8,7 +8,6 @@
 
 #include <math.h>
 #include "mex.h"
-#include "matrix.h"
 
 void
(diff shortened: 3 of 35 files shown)


robertoostenveld:
I tested compilation with MATLAB 2012b, 2015b and 2016b on OS X 10.12 (El Captain). This was not 100% trivial, since the older versions are not designed for the newer OS X version. After those changes compilation went fine.

Some changes to ~/.matlab/R2012b/mexopts.sh were needed, idem for 2015b.

I noticed this warning
```
Error: source file for abs cannot be found.
Error: source file for min cannot be found.
Error: source file for max cannot be found.
Error: source file for plus cannot be found.
Error: source file for minus cannot be found.
Error: source file for times cannot be found.
Error: source file for rdivide cannot be found.
``` 
which is because the uint64 class support files moved from fileio to compat/matlablt2010b. I have changed that in ft_compile_mex.

nno:
Thanks @robertoostenveld.

PR #325, opened 2017-02-12
FIX - revert correct detection of source+mesh, hopefully fixing test …

schoffelen:
…error in tutorial_networkanalysis

-> creating PR for reference. I will run the test batch first before merging.

### DIFF
--- a/utilities/ft_datatype.m
+++ b/utilities/ft_datatype.m
@@ -108,7 +108,7 @@ elseif ismesh && isparcellation
 elseif issource && isparcellation
   type = 'source+label';
 elseif issource && ismesh
-  type = 'mesh';
+  type = 'source+mesh';
 elseif ismesh
   type = 'mesh';
 elseif issource


robertoostenveld:
looks good, thanks

schoffelen:
However good it looks, test_tutorial_networkanalysis_new is still failing (but now at a different location)

Also, I ran 2 batches of tests, one on branch 'master', the other on 'tutorial_nwa':

The former returned 4 FAILED tests, the latter 11 :o(

schoffelen:
I restored my branch, to follow up on the newly failing functions

test_bug103.txt
test_bug1082.txt
test_bug1166.txt
test_bug1368.txt
test_bug2879.txt
test_bug2986.txt
test_bug3042.txt
test_bug70.txt

This one is still failing:

test_tutorial_networkanalysis_new.txt

robertoostenveld:
The original change (only) meant to be an improvement in the order of detection the various objects. The error in  test_tutorial_networkanalysis_new was a relatively obscure one, for data with source-level CSD matrices. How about the others?

schoffelen:
will look into it

schoffelen:
Looks like datatype detection stuff indeed.

robertoostenveld:
I just looked at ft_datatype. I think we have a conceptual problem there: "mesh" is getting more important, but does not distinguish sufficiently from "source". And then there is also "source+mesh".

The newer data representations (which getdimord likes) are improvements over the old, but is incompatible with the old style of handling: detecting the data type and then selecting the code to run.

If I compare "source" to other stuff, the problem is that source data is functional data on a geometry, whereas timelock and freq are functional data on a geometry (channels) PLUS time/freq info. For source the time/freq info is usually not explicit, and we currently do not detect the functional field itself, only the geometry.

Would it not be better to do something like this and look for the functional fields instead?

fn = fieldnames(data)
for i=1:numel(fn)
  dimord{i} = getdimord(data, fn{i})
end

istimelock = any(strcmp(dimord, 'chan_time') | strcmp(dimord, 'rpt_chan_time')) 
isfreq = any(strcmp(dimord, 'chan_freq') | strcmp(dimord, 'rpt_chan_freq')) ...
insource = here also a test for some functional data, even "inside" would be OK

robertoostenveld:
perhaps good to look into these

```
test/failed_ft_datatype.m
test/failed_ft_datatype_sens.m
test/failed_ft_datatype_source.m

test/test_datatype_parcellation.m
test/test_datatype_segmentation.m
test/test_datatype_source.m
```

schoffelen:
Regarding source/mesh/source+mesh: I suspect (but need to check) that the newly failing stuff is due to ft_checkdata failing, because it is called with sth. like ft_checkdata(data, 'datatype', 'source'), or ft_checkdata(data, 'datatype', 'mesh'). Is it possible that it fails because it does not correctly detect 'source+mesh'?

robertoostenveld:
yes, that seems to be the case. See below 

```
%%

data = [];
data.pos = randn(10,3);
data.pow = randn(10,1);
assert(ft_datatype(data, 'source'));

%%

data = [];
data.pos = randn(10,3);
assert(ft_datatype(data, 'source'));

%%
% this fails

data = [];
data.pos = randn(10,3);
data.tri = randn(16,3);
assert(ft_datatype(data, 'mesh'));

%%

data = [];
data.pos = randn(10,3);
data.tri = randn(16,3);
data.pow = randn(10,1);
assert(ft_datatype(data, 'source+mesh'));
```

bedtime for me now...

schoffelen:
In analogy to one of the comments above, if I do:

data = [];
data.trial = cell(1,2);
data.time = cell(1,2);
data.label = {'a';'b'};
data.topolabel = {'c';'d'};
data.topo = randn(2);

and then

assert(ft_datatype(data,'raw'))
assert(ft_datatype(data,'comp'))
assert(ft_datatype(data,'raw+comp'))

all return true, which is inconsistent with the source+mesh type.

This can be easily accommodated in ft_datatype

PR #328, opened 2017-02-13
Layouthorizontal

roemervandermeij:
@robertoostenveld 
Are you aware of vertical being used at all? Right now, it creates boxes that span the entire width of the screen (height in horizontal), which doesn't make sense to me in case Nchan > 3 or so. They just become stripes. 

Am I missing something? It's not like it's a matter of resizing the figure window, we usually enforce 'axis equal' (e.g. in ft_plot_lay). 

If it's just because it is, I propose we do something like:
if nchan<XX
   stretch width/height to max possible
else
   enforce square/nearly-rectangular boxes
end

Or something like cfg.boxshape = auto/square (not my preference, I suspect it's catering to nobody?)

PS: I thought I'd do a unplanned quick PR, which is why the ft_platform_support commit sneaked in

### DIFF
--- a/ft_prepare_layout.m
+++ b/ft_prepare_layout.m
@@ -51,10 +51,11 @@ function [layout, cfg] = ft_prepare_layout(cfg, data)
 % Alternatively you can specify the following layouts which will be
 % generated for all channels present in the data. Note that these layouts
 % are suitable for multiplotting, but not for topoplotting.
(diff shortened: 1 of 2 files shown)


robertoostenveld:
ft_databrowser uses the vertical layout. Works fine if you ask me, also with large Nchan.

PR #329, opened 2017-02-13
OCTAVE: some changes in includes for .mex compilation

nno:
The following changes fix errors in Octave when trying to compile .mex files.

PR #330, opened 2017-02-13
NF: add Makefile to build mex files

nno:
This PR adds a Makefile with support for compiling .mex* functions. In particular the 'on-the-fly' compilation strategy used in some functions seems not to work in Octave. 

I am aware of utilities/ft_compile_mex.m, and this PR does not aim to replace it. Instead the aim is to provide a unix-friendly way to build the .mex files. 

Future additions to the Makefile may include installation and running tests from the terminal; for example see https://github.com/CoSMoMVPA/CoSMoMVPA/blob/master/Makefile


nno:
Note that the travis build currently fails due to the issue reported in #329:

https://travis-ci.org/nno/fieldtrip/jobs/201229836

robertoostenveld:
what is the reason for replicating ft_compile_mex in a Makefile? Is it required for the travis test procedure? I am slightly worried about long term maintenance.

nno:
@robertoostenveld you're right that there is some duplication here. I only found out about ft_compile_mex when I wrote most of the Makefile. 

Let's explore the alternative, adjusting ft_compile_mex to be able to build under Octave:
- Main difference: the Makefile contains the locations of all .mex output files not in src and properly deals with dependencies, i.e. if necessary it compiles things in src if necessary, and copies mex files from src to other locations if necessary. ``ft_compile_mex`` does not do this. The documentation mentions a batch script which I did not find, but even if it was there it would require a two-step building process, i.e. first compile and then copy over all the files. This seems like a tedious process that can be automated quite trivially. (The downside of ``make`` may be limited support on Windows platforms)
- ``ft_compile_mex`` does not provide a "clean" option, which would be very useful for providing source-only releases and for testing whether compilation works . This maybe useful in ft_mex_compile. 
- ``ft_compile_mex`` compiles files in non-core, i.e. in realtime and a function from the fileexchange. In #318 @robertoostenveld wrote "Can you revert all changes that are in non-core fieldtrip code? Changes to main FT mex code are welcome", so I wonder whether these non-core functions should still be included or not?
- Some functions do not compile in Octave (such as realtime stuff), so these would have to be excluded.  
- For Octave compatibility it would require some changes with compiling the files in ``@uint64`` since Octave uses "max" internally when showing messages. It seems much cleaner to me to remove the .m files with 'magic compilation' in that directory anyway. What do you think about this?
- Re maintainability: ``ft_compile_mex`` could do with some refactoring: there are unused statements, there is considerable code duplication; cells grow unnecessarily after adding each file; functionality could be compartmentalised by using more subfunctions; some code is very convoluted and could be much shorter (e.g. ``add_mex_source``); it should use onCleanup when using ``cd``. (For now I'll leave difficulties to test this function as a future topic).

Bottom line: an alternative would be a substantial rewrite of ``ft_compile_mex`` to address these issues. ``ft_compile_mex`` could then be called from a drastically shortened Makefile, so that one can still run "make mex". Removing 'magic compilation' from ``@uint64``.m files could potentially be addressed in a separate PR. 

@robertoostenveld what are your thoughts about this?

nno:
Please note that this build currently fails on travis due to #329 not being merged.

robertoostenveld:
@nno, could you also add a paragraph to the general mex section at http://www.fieldtriptoolbox.org/development/guidelines/code#compiling_mex_files to explain the alternative compilation procedure?

nno:
Thanks @robertoostenveld, I added a paragraph to the general mex section on the wiki.

PR #333, opened 2017-02-16
Interactive subplots

roemervandermeij:
These commits allow for ft_single/topo/multiplotER/TFR in subplots to be interactive. 

It was actually surprisingly simple. All that was needed 'under the hood' was an addition to ft_select_channel. The trick is to save the (subplotted) cfg/data in a subfield of the same guidata, with a unique identifier for the current axes, which is also added as the gca's 'tag'. When ft_select_channel is then called from within this subplot, it searches for a tag, and if successful, used the appropriate data for selection. 

I've tested it in all the applicable functions (which are only the above plotting functions), and tested with test_tutorial_clusterpermutation, which generates several topo's in subplots, which are automatically interactive now!

### DIFF
--- a/ft_multiplotER.m
+++ b/ft_multiplotER.m
@@ -747,7 +747,7 @@ end
 % Make the figure interactive:
 if strcmp(cfg.interactive, 'yes')
   
-  % add the channel information to the figure
+  % add the cfg/data/channel information to the figure under identifier linked to this axis
   ident                 = ['axh' num2str(round(sum(clock.*1e6)))]; % unique identifier for this axis
   set(gca,'tag',ident);
   info                  = guidata(gcf);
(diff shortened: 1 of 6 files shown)


roemervandermeij:
BTW, I was originally focusing on something similar, using 'subplots' via nested layouts. Using layouts there is a 'snugger' fit between 'subplots', which I like more. However, it's more complicated to implement, requiring e.g. cfg.hpos/vpos/width/height to be added to high level plotting functions (and code dealing with the nesting). W.r.t. the ft_select_channel the changes would have been similar (but different ;)). 
_(This is separate from the current issues with nested layouts and ft_plot_lay, see bug3252)_

robertoostenveld:
I trust you on this, I did not test it myself. I will keep an eye on it when the situation arises in some analysis.

PR #334, opened 2017-02-16
CLN: remove unused src/nannumel.c

nno:


nno:
I did not find there any nannumel.mex* files included in FieldTrip. This function does not seems to be used anywhere.

robertoostenveld:
agreed, that was related to an older implementation of nanmean etc.

PR #336, opened 2017-02-16
FIX: align_ctf2spm crashed with spm12 on the path

StolkArjen:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3255


StolkArjen:
The issue is though that spm12 might be set to the path at one point, eg when fusing ct and mr. Meaning the next time ft convert coordsys is called, it will fail because of missing spm normalise and the likes. The warning would thus need to suggest a workaround, until the spm12 equivalent of spm normalise is implemented.

robertoostenveld:
Either way it fails. I have it rather failing for someone who added spm12 to the path than for someone who has added spm2 to the path. Why not have it fail explicitly in the code where spm12 is not supported by doing 

if ft_hastoolbox(‘spm12’)
 error(‘spm12 is not yet supported’)
end

There should never be the need to add a specific version to the path that was not explicitly requested by a user.

StolkArjen:
Ok, agreed. An error is now thrown, at switch case spm12.

PR #337, opened 2017-02-17
ENH - represent the output of ft_componentanalysis as timelock+comp

robertoostenveld:
when the input is timelock. See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3256


robertoostenveld:
I ran the test batch: one test is still running bug 549 passed without problems.

PR #340, opened 2017-02-19
Refixed mysteriously disappeared bugfix regarding LCMV NAI orientation selection (previously submitted as PR 184)

sarangnemo:
Refixed bugfix on NAI orientation, originally contributed in PR 184 and mysteriously disappeared with PR 227


agramfort:
@sarangnemo does that improve the perf on the phantom data?

sarangnemo:
@agramfort Yes, our testing last week uncovered this problem. It explains the bizarre results we were getting with the phantom, and it also appears stable with a single sphere lead field. Now the phantom results look great for the dipoles that were 200nAm or less in strength. Performance chart still forthcoming :-)

agramfort:
great.

looking forward to see the charts.

don't hesitate to share a dropbox folder with scripts
so I can also play :)

PR #341, opened 2017-02-20
Fix failed sourceanalysis

schoffelen:
This should fix it. I ran the dashboard batch. An earlier version still had some failures due to some old trailing test data (which I computed fresh in the mean time). 

Also, test_bug3205 failed, something to with ft_electroderealign, this could be caused by my branch being created before this issue was fixed.

I am running another batch now, so far all looks OK. Will notify here about the outcome in a few minutes...


schoffelen:
Hmmm, all runs through fine now, apart from test_ft_sourceanalysis 

I cannot reproduce this when running it locally.

robertoostenveld:
thanks. I'll delete the old test results, lets see what result is tomorrow morning.

PR #345, opened 2017-02-22
FIX - made raw2timelock more robust, removed sampleinfo

schoffelen:
zoiets?

I have run 'ft_test run test_ft_checkdata' without problems


robertoostenveld:
looks good to me

PR #346, opened 2017-02-22
bugfix - missing bracket

vlitvak:


robertoostenveld:
Hi @vlitvak, you have three commits in this PR. One of them (in ft_plot_sens) was already merged earlier. This suggests that you are working on an outdated copy. I also see that you are requesting the pull from your master branch. It is better practice to make a development branch and implement+commit the changes on that, rather than in master. That keeps the master clean. Perhaps @agramfort can recommend you a graphical git client for windows that makes it easier to work with branches and multiple remotes.

PR #347, opened 2017-02-22
Pull343 - improve ft_sourcemovie

robertoostenveld:
@StolkArjen had some suggestions for ft_sourcemovie. When I tested them, I realized that the test script was incomplete. I fixed that. I also made some other improvements to ft_sourcemovie along the way. 

The test script takes a long time to run. The final test data is now in /home/common/matlab/fieldtrip/data/test/pull343. 

@schoffelen, could you review and merge the changes in this PR? After that, PR #343 will automatically be closed (as its commits are contained in this PR).

PS it would be good if the hidden functionality with source2 (either as 2nd source, or as parcellation) would be documented.

### DIFF
--- a/ft_sourcemovie.m
+++ b/ft_sourcemovie.m
@@ -8,7 +8,7 @@ function [cfg, M] = ft_sourcemovie(cfg, source, source2)
 % where the input source data is obtained from FT_SOURCEANALYSIS and cfg is
 % a configuratioun structure that should contain
 %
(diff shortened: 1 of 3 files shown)


robertoostenveld:
although I discussed with @schoffelen that there would be more work to do, I think it is better to merge this for now. Otherwise it will remain here too long, whereas the current changes are useful.

PR #348, opened 2017-02-24
Bsscca

schoffelen:
I have cleaned up the stuff related to bsscca + implemented some new functionality for this method


robertoostenveld:
could you add it to one of the test scripts? Or make a new test script "test_pull348.m"?

robertoostenveld:
... I realize that a test script will fail due to "external/cellfunction" missing. Do you want to add that as well, or include it in private?

schoffelen:
OK, zoiets?

robertoostenveld:
"made small fix" --> duim omhoog!

PR #352, opened 2017-03-01
enh: re-organized surface coloring and added new colors, see bug 3260

StolkArjen:


schoffelen:
may I suggest to make the default vertexcolor in ft_sourceplot cortex_light? curv is not guaranteed to be present (only if well-defined fs-based meshes have been loaded). I am not sure whether for example the template meshes have this info.

also, I would throw a warning in ft_plot_mesh if 'vertexcolor'= 'curv', and the curv field is absent

StolkArjen:
I'm fine with that but it's worth mentioning that changing the default will make current figures not immediately 'backward-compatible' (well, by specifying cfg.vertexcolor = 'curv' they are).

schoffelen:
I see, does this mean that the templates actually do have a 'curv' field?

StolkArjen:
load('/Users/arjsto/Documents/MATLAB/fieldtrip/template/anatomy/surface_pial_left.mat');
>> mesh

mesh = 

         tri: [350814x3 double]
        sulc: [175409x1 double]
        curv: [175409x1 double]
        unit: 'mm'
    coordsys: 'mni'
         pos: [175409x3 double]


yes

StolkArjen:
In fact, even single subjects processed with FreeSurfer:

> mesh = ft_read_headshape('freesurfer/surf/lh.pial')

mesh = 

          pos: [115828x3 double]
          tri: [231652x3 double]
         sulc: [115828x1 double]
         curv: [115828x1 double]
         area: [115828x1 double]
    thickness: [115828x1 double]
         unit: 'mm'

schoffelen:
alright, keep curv as default then, but please include the warning...

StolkArjen:
Good point - added.

schoffelen:
Sorry Arjen, here I am again, but wouldn't it be a bit cleaner if all the vertexcolor handling in ft_plot_mesh is done at a single location in the function? There's now also an additional chunk around line 164, besides the section you added in lines 135-146. Would it be possible to move said section to a bit below?

StolkArjen:
Moved the code down, while abiding the original order of vertexcolor-related code (might break functionalities/dependencies otherwise).

schoffelen:
Thanks, could you however change lines 163-165 in an appropriate elseif statement? Rather than making it two separate if-ends?

StolkArjen:
Aye, captain

schoffelen:
👍

PR #353, opened 2017-03-02
enh, bugfix (3248): added sphere option to ft_plot_sens, other updates

smg211:
ft_write_sens: added missing semicolons to a few lines
read_bioimage_mgrid: fixed issue where electrodes not presented in bioimage suite were given actual coordinates in ft
write_bioimage_mgrid: fixed electrode numbering
ft_plot_sens: fixed bug 3248 by changing default style to [], setting individual default inputs to plot3 (Marker, MarkerSize, Color, and Linestyle), and only using style when explicitly specified.  Added sphere option for plotting electrodes (related to bug 2386).  Changed coilsize and coilshape to sensize and shape, respectively, rather than having separate inputs for coils and electrodes (related to bug 2438).

### DIFF
--- a/fileio/ft_write_sens.m
+++ b/fileio/ft_write_sens.m
@@ -1,4 +1,4 @@
-function ft_write_sens(filename, sens)
+function ft_write_sens(filename, sens, varargin)
 
 % FT_WRITE_SENS writes electrode information to an external file for further processing in external software.
 %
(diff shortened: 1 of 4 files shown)


smg211:
Forgot to change 4 instances of "both" back to the original variable name "coil" (lines 65, 202, 232)

smg211:
Ah okay, that makes sense.  I just added elecsize/shape to the help and included a check to make sure size/shape are not specified for both elec and coil, and then assigned coil/elec shape and size to common variables so that they are treated the same throughout the code.

Thanks,
Sandon

PR #354, opened 2017-03-03
fix - empty oxy3 header info handled properly

Horschig:
When an oxy3 header was empty, e.g. no description
was given, there were special instances where this
led to a crash in str2num (namely, using Matlab2016b
and adding FT with all subfolders).

An explicit check whether header information is empty
is added to prevent a crash here.

The user should now never experience such crashes.

For more information, send a mail to jorn@artinis.com


Horschig:
btw, commit message according to our internal git message style:
image

schoffelen:
There's no code to review here, so I have to trust Jorn on his blue eyes.

Horschig:
I do not have blue eyes, but thanks anyway ;)

PR #355, opened 2017-03-03
FIX: use uimagesc so that data with uneven axis is shown correctly.

nno:
Fixes issue reported at http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3261


schoffelen:
thanks

PR #357, opened 2017-03-04
new: bis2fieldtrip elec conversion and vice versa

StolkArjen:


schoffelen:
Where are these files located, and by which function(s) are they called?

StolkArjen:
These wrapper functions rely on changes/additions to ft read and write sens proposed by sandon (smg201) in another pr. Given that transformation fields of the scans are required for correctly reading in the elec positions (localized with bioimage suite software), we decided not to molest ft read/write sens with that input but to create these wrapper functions. If not desired to have them in the main folder (there are other ft2xx functions), we could also put the code on the wiki.

schoffelen:
shouldn't they then be in fileio/private?

I think that the ft2xxx function in main pertain to more full-fledged data, i.e. data with time courses etc.

robertoostenveld:
I have not reviewed the final changes to ft_read_sens and ft_read_mri yet, but presume that both return the correct value for unit and coordsys (ijk is NOT a coordinate system).

PR #358, opened 2017-03-05
Master

sappelhoff:
Layout of 64 channel actiCap is based on the general easy cap electrode
positions. Compared to the easycapM11 template, however it has fewer
electrodes. See this link for comparison:
http://www.brainproducts.com/files/public/downloads/actiCAP-64-channel-Standard-2_1201.pdf


robertoostenveld:
The layout should be called "acticap-64ch-standard2.mat" to be consistent with the description in the PDF. Given the description, there is apparently also a standard-1 version. Can you please rename teh file?

Can you please add some documentation to http://www.fieldtriptoolbox.org/template/layout#acticap including a screenshot/bitmap version of the original layout (from the PDF) and a link to the PDF?

sappelhoff:
Hi Robert, 
I renamed the file to its proper name. The name does suggest a 1st version, however it is no longer used by Brain Products and refers to an outdated layout (to my knowledge).

I also made an entry to the fieldtrip wiki that you linked in your comment.

Best,
Stefan

robertoostenveld:
thanks, much appreciated!

PR #359, opened 2017-03-06
regressconfound cleanup

ilinabg:
I cleaned up ft_regressconfound, and deleted redundant code (some variables were always being set to false so all the code that used those variables was never being executed). I also added the cfg.beta field to the output so that users can specify if they want to keep the beta values that are computed in the regression model.

### DIFF
--- a/ft_regressconfound.m
+++ b/ft_regressconfound.m
@@ -1,11 +1,11 @@
-function [data] = ft_regressconfound(cfg, datain)
+function [data] = regressconfound_tcg(cfg, datain)
 
 % FT_REGRESSCONFOUND estimates the regression weight of a set of confounds
 % using a General Linear Model (GLM) and removes the estimated contribution
 % from the single-trial data.
 %
 % Use as
-%   timelock = ft_regressconfound(cfg, timelock)
+%   timelock = ft_regressconfound(scfg, timelock)
 % or as
 %   freq     = ft_regressconfound(cfg, freq)
 % or as
@@ -24,6 +24,8 @@ function [data] = ft_regressconfound(cfg, datain)
 %                     make the confounds orthogonal (default = 'yes')
 %   cfg.statistics  = string, 'yes' or 'no', whether to add the statistics
 %                     on the regression weights to the output (default = 'no')
+%   cfg.beta        = string, 'yes' or 'no', whether to add the beta
+%                     values as a field to the output (default = 'no')
 %   cfg.model       = string, 'yes' or 'no', whether to add the model to
 %                     the output (default = 'no')
 %   cfg.ftest       = string array, {N X Nconfounds}, to F-test whether
@@ -64,6 +66,9 @@ function [data] = ft_regressconfound(cfg, datain)
 %
 % $Id$
 
+addpath ~/Documents/MATLAB/fieldtrip;
+ft_defaults;
+
 % these are used by the ft_preamble/ft_postamble function and scripts
 ft_revision = '$Id$';
 ft_nargin   = nargin;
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Thanks. Is there a reason why the beta's should *not* be kept in the output?

ft_regressconfound.m
```diff
@@ -94,6 +96,7 @@
 cfg.normalize  = ft_getopt(cfg, 'normalize', 'yes');
 cfg.model      = ft_getopt(cfg, 'model', 'no');
 cfg.statistics = ft_getopt(cfg, 'statistics', 'no');
+cfg.beta       = ft_getopt(cfg, 'beta', 'no');
```
robertoostenveld: The default used to be that beta was kept. Why change that default?

I propose to call the option cfg.keepbeta, in line with other cfg.keepxxx options elsewhere in FT.

robertoostenveld:
hi @ilinabg, correct me if I am wrong: I think it is your turn to act upon this PR.

StolkArjen:
Thanks for the reminder. Will discuss your suggestion with Ilina after spring break.

ilinabg:
Hi, I just pushed the changes that you suggested, the beta's are now kept in the default case.

robertoostenveld:
@StolkArjen do you want to do a final review before this gets merged into the master branch?

StolkArjen:
Looks good to me. @robertoostenveld: can you check and pull?

StolkArjen:
This is a relatively safe pull. All the datatype specific handling (eg issource, isfreq, istimelock) were already deactivated (set to false) in the previous code. Ilina removed that redundant code and added the ability to save out beta values, a feature that seemingly got lost some time in the past.

StolkArjen:
I found that now the beta is added to the output by default, this throws an issue when calling ft_selectdata with cfg.trials afterwards. 

The error is located at line 529 in ft_selectdata:
data.(datfield) = cellmatselect(data.(datfield), seldim, selindx, numel(dimtok)==1);
where data.(datfield) = data.beta

In effect, the trials are also attempted to be selected from the beta field, who of course doesn't have a trial dimension. I guess we'll have to make a simple test case producing the issue, and think of a solution.

StolkArjen:
This is still an issue:

Dear Fieldtrip-users,
 
We are not able to use function ft_selectdata after applying head movement correction using ft_regressconfound. The error that we get is the well-known index exceeds matrix dimensions.
It is due to the fact that the size of x is incorrect, it doesn’t correspond to the number of trials but is 7 which is the number of regressors in the beta variable that is created by using ft_regressconfound.
There is a simple way to solve it and this is deleting the field ‘beta’ before running ft_select data.
 
My question is, just to be sure, the variance is regressed out by using ft_regressconfound so the beta variable is not needed in any further step right?
Maybe this is also something for the debug team? Thus that ft_selectdata runs after running ft_regressconfound.
 
Best
Dr Conny Quaedflieg

schoffelen:
I think that, on a meta-level, the issue may arise from the fact that the 'dimord' of the betas (and associated measures, e.g. F, prob, etc) is different from the 'dimord' of the cleaned data. 

To accommodate this, at some point we came up with having a numeric data field specific dimord, e.g. posdimord, leadfielddimord etc.

Would a robust solution lie in the direction of making the output of ft_regressconfound less ambiguous?

Alternatively, is there any reason why the betas should be kept at the highest level of the structure? Are there any FT functions that use this info at a later stage? Would it be worthwile considering to make the function's output mutually exclusive, i.e. either the betas, or the cleaned data?

For instance, ft_mvaranalysis has an (undocumented) option cfg.output, which determines whether the output of the function will be 'parameters' (i.e. model coefficients etc.), the 'model', or the 'residual'

StolkArjen:
That seems to be solution that'd work here too. I feel that, similarly to ft_mvar, people would use ft_regress to either analyze the residuals, or the beta values. We could add a similar cfg.output option.

robertoostenveld:
it is unclear to me what the status is of this. It has been merged, but there still seems an ongoing discussion. @StolkArjen if you want to follow this up, please make a new issue or pull request.

StolkArjen:
Will do, haven't had the time yet to update ft regresconf. Will do so in the form of a new PR

PR #360, opened 2017-03-07
FIX - in some cases .inside was still created as double

jens-k:
...instead of logicals. This lead to an error later on.

### DIFF
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -778,7 +778,7 @@ if isempty(sel)
   inside = mask(sub2ind(dim, pos(:,1), pos(:,2), pos(:,3)));
 else
   % only loop over the points that can be dealt with
-  inside = zeros(size(pos,1), 1);
+  inside = false(size(pos,1), 1);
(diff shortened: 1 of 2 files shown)


schoffelen:
Would you mind writing a little test function (and committing it to fieldtrip/test/...) that would reproduce this, had you not fixed it? E.g. you can call it test_pull360

jens-k:
I added a test_pull360.m to this branch.

schoffelen:
Great thanks! almost there :o)
Convention has it that the test 'scripts' are actually functions, which start like

function test_pull360

% TEST test_pull_360
% TEST ft_prepare_sourcemodel

% WALLTIME 00:10:00
% MEM 500mb

++++++++++++++
The lines starting with % are needed to tell torque what resources are needed to run this function, and the lines starting with % TEST serve to automatically trigger a run of this test function once somebody commits a change in the functions mentioned

jens-k:
Mmmh, now maybe?

schoffelen:
:1st_place_medal:

schoffelen:
Thanks!

PR #361, opened 2017-03-07
ENH - update the channel position, weighted with the montage (it used…

robertoostenveld:
… to be changed to nan). See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3229

### DIFF
--- a/fileio/private/ft_apply_montage.m
+++ b/fileio/private/ft_apply_montage.m
@@ -18,7 +18,7 @@ function [input] = ft_apply_montage(input, montage, varargin)
 %   montage.labelnew = Mx1 cell-array
 %
 % As an example, a bipolar montage could look like this
-%   bipolar.labelold  = {'1', '2', '3', '4'}
+%   bipolar.labelold  = {'1',   '2',   '3',   '4'}
 %   bipolar.labelnew  = {'1-2', '2-3', '3-4'}
 %   bipolar.tra       = [
 %     +1 -1  0  0
@@ -259,7 +259,7 @@ if k > 0 && isfield(input, 'trial') % check for raw data now only
   % use an anonymous function to test for the presence of NaNs in the input data
   hasnan = @(x) any(isnan(x(:)));
   if any(cellfun(hasnan, data_unused.trial))
-      error('FieldTrip:NaNsinInputData', ['Your input data contains NaNs in channels that are unused '...
+    error('FieldTrip:NaNsinInputData', ['Your input data contains NaNs in channels that are unused '...
       'in the supplied montage. This would result in undesired NaNs in the '...
       'output data. Please remove these channels from the input data (using '...
       'ft_selectdata) before attempting to apply the montage.']);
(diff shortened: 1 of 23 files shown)


robertoostenveld:
@schoffelen can you review this one?

schoffelen:
oh nee, moet 'k 11 files doorploegen :o(

robertoostenveld:
misschien will @StolkArjen hem wel even reviewen. Die is toch klaarwakker...

schoffelen:
nee hoor, ik kijk wel even.

schoffelen:
Just a naive question; as per the bug the averaging of positions makes sense, but in the case of more funky tra-matrices it doesn't (e.g. components). I don't quickly see whether this distinction is made.

robertoostenveld:
for cleaning of ICA components, the channel names remain the same, which affects line 380 (keepchans=yes). In cases where channel names do not stay the same, keepchan=no and the chanpos is recomputed. 

It is an interesting test case to consider a ft_componentanalysis with method=pca, and then ft_rejectcomponent without rejecting anything, but back projecting the components...

robertoostenveld:
I added a section to the test script. But cannot get it to work, since the initial balancing is not present to start with, and I don't know where it should be added. The test script actually reveals an error in the balance bookkeeping. @schoffelen could you have a look at the last section of the test script?

schoffelen:
will do. Domani...

StolkArjen:
elec.label = {'1';'2';'3';'4'};
elec.elecpos = [1 1 1; 2 2 2; 3 3 3; 4 4 4];
elec.chanpos = elec.elecpos;
elec.tra = eye(4);

bipolar.labelold  = {'1', '2', '3', '4'};
bipolar.labelnew  = {'1-2', '2-3', '3-4'};
bipolar.tra       = [
  +1 -1  0  0
   0 +1 -1  0
   0  0 +1 -1
  ];

posweight = abs(bipolar.tra);
posweight = diag(1./sum(posweight,2)) * posweight;

elec.chanpos = posweight * elec.chanpos
produces the expected result:

    1.5000    1.5000    1.5000
    2.5000    2.5000    2.5000
    3.5000    3.5000    3.5000

Wondering whether the same posweight matrix should also be used for (bipolar) derivation of the neural data? line 495 in ft_apply_montage: (data.trial{i} = montage.tra * data.trial{i};)

robertoostenveld:
> Wondering whether the same posweight matrix should also be used for (bipolar) derivation of the neural data?

No. The bipolar channel is the difference between two electrodes, which involves a minus. The bipolar position  is an average of two electrode positions, which involves a plus (and a 1/2).

StolkArjen:
Makes sense. Grazie.

robertoostenveld:
I made the overall handling of ft_apply_montage and the sense.balance field more consistent. It is still not perfect, but should be better. I think my work is complete now.

This also addresses http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1481, http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1990 and http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2634.

I ran some test, not all. @schoffelen, do you care to test more elaborately, or should we just merge and see whether the code holds up to the stress of actual use? If any, there will be only (obvious) bookkeeping errors, no numerical errors.

robertoostenveld:
@StolkArjen, please note that when doing ft_componentanalysis and then ft_rejectcomponent, you will also get weird mixtures of channel positions. I.e. the original channel position will not be retrieved. This is technically correct, but may be undesired. For that purpose you should use the cfg.updatesens=yes/no option, i.e. default=yes, but if you don't want the sense to change, specify no.

schoffelen:
Let me read through the changes, I should probably not devote too much time now to extensive testing.

schoffelen:
Looks good to me. God zegene de greep!

StolkArjen:
Thanks @robertoostenveld, nice work

PR #362, opened 2017-03-08
Partially solved bug (bugzilla 3264). basic functionality did not work for raw …

matsvanes:
…data structures (with parameter 'trial') because of faulthy indexing. Solved this issue by changing faulthy index (i-->1); also added tests for raw data structures in test_ft_math. 
Furthermore, in test_ft_math, I removed two duplicate checks.

where raw data is input, the output of ft_math still contains a dimord field, which thus remains unresolved.

### DIFF
--- a/ft_math.m
+++ b/ft_math.m
@@ -123,9 +123,9 @@ if ~iscell(cfg.parameter)
   cfg.parameter = {cfg.parameter};
 end
 
-if ft_datatype(varargin{1}, 'raw')
+if ft_datatype(varargin{1}, 'raw+comp')
     if length(varargin)>1
-        error('ft_math does not support more than one input argument if the input data is of type "raw"')
+        error('ft_math does not support more than one input argument if the input data is of type "raw" or "comp"')
     end
 end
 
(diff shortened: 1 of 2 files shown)


schoffelen:
OK, good so far!
It looks as if the dimord issue can be easily solved by removing line 447, and treating the dimord in the same way as the fields that are added to the output data structure starting from line 175

schoffelen:
however, the chunk of code starting at line 175 seems a bit outdated. Nowadays, it should work if we write

data = keepfields(varargin{1}, {'label', 'pos', etcetc});

Much more compact

robertoostenveld:
can you outline - perhaps inside the test script or in a comment in ft_math - what the expected behavior should be for raw data when

- there are different numbers of trials
- one dataset has a subset of channels of the other
- one dataset has another grad or elec structure than the other
- one dataset has different (shorter/longer) time axes than the other
- one dataset has partially overlapping time axes with the other
- one dataset has different trialinfo or sampleinfo
- etc.

I am not stating that all these should all be supported in the most elegant fashion, only that the expected behavior should be defined. It is fine with me to give an error, and e.g. only support operations that involve a scalar and not two (or more) datasets.

schoffelen:
@robertoostenveld good points! Yet, I wondered all of a sudden whether any sanity checks are performed on non-raw data with respect to a subset of the issues you outlined, but a call to ft_selectdata in line 152 ensures this.

schoffelen:
We could be inspired by ft_appenddata, where there should be some code that checks the input data structures in order to assess whether a concatenation over trials or channels should be done.

Also, but I don't remember the details, I thought that ft_selectdata has some basic support for raw data structures, or not?

robertoostenveld:
of course it is fine with me if it being dealt with by ft_selectdata. Was not sure about that, just wanted to rase awareness for the less common situations.

schoffelen:
@matsvanes didactisch momentje: what sometimes seems as a small but simple fix, often explodes in Robert and my hands ...

robertoostenveld:
That is because I realise that the suggested code will also be in the hands of a few thousand other people soon…

schoffelen:
Exactly!

schoffelen:
I suggest to build in a check into ft_math, which throws an error (perhaps for the time being) the input data is of type 'raw', and if there's more than one input data argument. In other words, allowing the function only to operate on a single raw data structure.

schoffelen:
We can then later check ft_selectdata (through which the data are passed anyway), and see whether it behaves as expected on multiple raw data structures in the input (probably it doesn't)

matsvanes:
Additionally I implemented some of the requested changes:
- resolved the issue with the dimord field
- updated the old code that JM mentioned (from line 175 onwards)
- throw an error if the input consists of more than one raw/comp structure

schoffelen:
can you change 'chancmb' in line 189 into 'labelcmb', thanks

matsvanes:
done!

PR #322, opened 2017-02-08
FIX+OCTAVE: simplify ft_{pre,post}amble for Octave

nno:
This PR tries to address the issue that surfaced on travis (https://travis-ci.org/nno/fieldtrip/jobs/196967097) where the current implementation of ft_preamble does not work properly in Octave when ft_preamble gets called recursively. The present attempt does so by a temporary ``cd`` into the private directory, and executes commands from there. As far as I can tell Octave is happy (or at least, happier).

Note that this PR also changes behaviour of ft_{pre,post}amble: typically supported commands are ``debug``,``init``,``loadvar``,``provenance``,``randomseed`` and ``trackconfig``. In this PR an error is now thrown if the input argument``cmd`` is referring to an unsupported command (rather than quitting silently). Thus, previous versions of FieldTrip are fine with 
```
    ft_defaults foo
```
whereas with the current PR such a command raises an error.

@robertoostenveld if you agree it is good to raise an error if an unsupported value for the input argument ``cmd`` is used, it is probably a good idea to run the full test suite on the changed code before merging. I tried to find / grep all occurences of ft_{pre,post}amble

PR #363, opened 2017-03-08
improve preamble and postamble for octave and update contrib/spike

robertoostenveld:
See PR #322 from @nno and comments therein. I made some updates to contrib/spike functions that were using an old preamble that used to fail silently, and that now resulted in an explicit error (much better).


nno:
Thanks @robertoostenveld.

PR #365, opened 2017-03-09
fieldtrip2fiff fails if filename is only a file (no path)

dnacombo:
if filename is only a file name, prepending filesep creates a conflict. fieldtrip2fiff tries to write in / and fails if no permission.
Fixed by using fullfile.


robertoostenveld:
thanks

PR #370, opened 2017-03-13
FIX+OCTAVE: fix case where input is empty string

nno:
Addresses http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3269

### DIFF
--- a/compat/matlablt2014b/private/fixname.m
+++ b/compat/matlablt2014b/private/fixname.m
@@ -40,6 +40,9 @@ end
 
 switch version
   case 'default'
+    if isempty(str)
+      str='x';
+    end
     str = lower(str);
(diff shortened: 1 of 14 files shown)


nno:
Note that CoSMoMVPA passes the tests again, but the issue is not resolved - I switched temporarily away from the official FT to use this branch. 

See https://github.com/CoSMoMVPA/CoSMoMVPA/pull/136

nno:
@robertoostenveld since I would like to be able to use the official FT again, it would be great if you could have a look. Thanks.

robertoostenveld:
sorry for the slow response. we have a project workshop and mid-term review this week, so I am very busy.

nno:
No worries, thanks for merging.

PR #371, opened 2017-03-13
adds fontsize to ft_plot_sens

ualsbombe:
Adds fontsize to ft_plot_sens

### DIFF
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -11,6 +11,7 @@ function hs = ft_plot_sens(sens, varargin)
 %   'label'           = show the label, can be 'off', 'label', 'number' (default = 'off')
 %   'chantype'        = string or cell-array with strings, for example 'meg' (default = 'all')
 %   'unit'            = string, convert the sensor array to the specified geometrical units (default = [])
+%   'fontsize'        = number, sets the fontsize of the labels (default = 10)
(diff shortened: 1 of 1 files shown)


robertoostenveld:
this assumes that the previous default was 10. I think that should be fine.

robertoostenveld:
could you figure out what the "default" default would be, and whether the user can set another default that spans multiple figures? I.e. if you make a figure, what is the font size?

ualsbombe:
```
fig = figure;
h = get(gca);
disp(h.FontSize)
```

This returns: 10, which I guess is the default then.

Changing the defaultaxesfontsize
```
set(0, 'defaultaxesfontsize', 30)
fig = figure;
h = get(gca);
disp(h.FontSize)
```

returns 30

So I guess the fontsize could be got from
`get(0, 'defaultextfontsize')`
Be aware that it is default_text_fontsize and not default_axes_fontsize that should be changed

robertoostenveld:
```
get(0, 'defaultaxesfontsize')
```
works on 2016b and 2012b. It returns 10, which is consistent with documentation. I'll change it.

ualsbombe:
I think it is `get(0, 'defaulttextfontsize')` you should use _not_ `get(0, 'defaultaxesfontsize')` since the text function is used to write the text.

PR #372, opened 2017-03-13
improve trialinfo handling in ft_appendxxx functions, related to bug3267

TEMadsen:
I changed an "i" used between for loops to 1, so the first input isn't skipped.  If you want me to make any of the other changes I [asked about](http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3267), feel free to wait on merging this pull request until I commit them. 

Please let me know which are highest priorities.


robertoostenveld:
I would like to combine the other changes with this PR. I'll comment on details on http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3267

TEMadsen:
Those changes are just to ft_appenddata - I haven't gotten to the others yet, and I haven't fully tested these yet.  Just wanted to let you know it's still a work in progress, not yet worth merging.

~Teresa

robertoostenveld:
Hi @TEMadsen, are you still working on this or is it ready for review?

TEMadsen:
I still haven't finished, but had to set it aside temporarily in favor of other urgent priorities.  I'm hoping to get back to it later today or tomorrow.  Sorry for the delay.

robertoostenveld:
no worries if it takes some time. Let me know when you want me to review and merge.

TEMadsen:
Okay, thanks for your patience. 

As I started trying to merge the changes I had made to ft_appenddata into ft_appendfreq, I realized they're dealing with mismatches very differently:  ft_appenddata tends to quietly remove mismatched fields with just a printed message or sometimes a warning, whereas ft_appendfreq throws errors. Is there a justification for this difference, or is it just a result of the functions not being written at the same time or by the same person?  

I was trying to think through which was better and think there might be different cases in which people would prefer to discard the field or throw an error.  What do you think of adding a cfg option to specify which mismatch handling is preferred? Or I could go with the more conservative option of throwing an error, but add documentation clarifying that mismatched fields could be removed before merging if you want it to run through successfully. 

Thoughts?

robertoostenveld:
I expect the inconsistency to be due to different people, different times of implementation. For raw data it is simpler (and the function is older), for freq it needs to be better controlled. 

I would not add an option to influence the silent/warning/error behavior. We do have such an option in "global ft_default" set by ft_defaults: it is ft_default.checkconfig, but that is specific for cfg option parsing and backward compatibility. 

In the append case the users would simply get the error at a later stage when they (or their code) realizes that a data field is missing. So I think that warnings are fine, as long as the output structure is valid.

robertoostenveld:
The whole ft_appendxxx implementation has been fully overhauled in http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3287. According to the test scripts, all works as it should.

PR #374, opened 2017-03-15
enhancement-added support for common filter approach when using refdi…

jimher:
…p with dics-see bug3270

### DIFF
--- a/inverse/beamformer_dics.m
+++ b/inverse/beamformer_dics.m
@@ -526,6 +526,9 @@ switch submethod
     elseif isstruct(refdip) && isfield(refdip, 'pos') % check if only position of refdip is present
       assert(isnumeric(refdip.pos) && numel(refdip.pos)==3);
       lf1 = ft_compute_leadfield(refdip.pos, grad, headmodel, 'reducerank', reducerank, 'normalize', normalize);
+      if isfield(refdip,'mom'); % check for fixed orientation
+        lf1 = lf1.*refdip.mom'; 
+      end; 
       filt1 = pinv(lf1' * invCf * lf1) * lf1' * invCf;       % use PINV/SVD to cover rank deficient leadfield
     else % backwards compatible with previous implementation - only position of refdip is present
       % compute cortio-cortical coherence with a dipole at the reference position
--- a/inverse/beamformer_dics.m
+++ b/inverse/beamformer_dics.m
@@ -527,7 +527,7 @@ switch submethod
       assert(isnumeric(refdip.pos) && numel(refdip.pos)==3);
       lf1 = ft_compute_leadfield(refdip.pos, grad, headmodel, 'reducerank', reducerank, 'normalize', normalize);
       if isfield(refdip,'mom'); % check for fixed orientation
-        lf1 = lf1.*refdip.mom'; 
+        lf1 = lf1.*refdip.mom(:); 
       end; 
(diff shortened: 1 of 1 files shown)


robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3270 for reference.

@schoffelen, I suggest you review it - as I came up with the idea that Jim implemented: but am not sure whether it is a smart idea.

schoffelen:
Looks good to me. But while you are at it, you may want to consider also an orientation constrained dipole, i.e. a refdip.pos + refdip.ori, or perhaps one would expect the fixedori settings to percolate to the refdip

inverse/beamformer_dics.m
```diff
@@ -516,10 +516,23 @@
...
+    if isstruct(refdip) && isfield(refdip, 'filter') % check if precomputed filter is present
+      assert(iscell(refdip.filter) && numel(refdip.filter)==1);
+      filt1 = refdip.filter{1};
+    elseif isstruct(refdip) && isfield(refdip, 'leadfield') % check if precomputed leadfield is present
+      assert(iscell(refdip.leadfield) && numel(refdip.leadfield)==1);
+      lf1 = refdip.leadfield{1};
+      filt1 = pinv(lf1' * invCf * lf1) * lf1' * invCf;       % use PINV/SVD to cover rank deficient leadfield
+    elseif isstruct(refdip) && isfield(refdip, 'pos') % check if only position of refdip is present
+      assert(isnumeric(refdip.pos) && numel(refdip.pos)==3);
+      lf1 = ft_compute_leadfield(refdip.pos, grad, headmodel, 'reducerank', reducerank, 'normalize', normalize);
```
robertoostenveld: after this there should be a check on hasfield(refdip, mom), and if so, the lead field should be multiplied with the mom(which should be 3x1) prior to computing filt1 .

robertoostenveld:
rather than what @schoffelen writes (otherwise a good comment): it should not be "ori" but "mom" for consistency with the scandip.

inverse/beamformer_dics.m
jimher: So, something like this?

if isfield(refdip,'mom'); % check for fixed orientation
        lf1 = lf1.*refdip.mom'; 
 end;

jimher:
I copied from 'disc_refchan' lines 413-416:

 elseif hasmom
        % compute the leadfield for a fixed dipole orientation
        lf = ft_compute_leadfield(dip.pos(i,:), grad, headmodel, 'reducerank', reducerank, 'normalize', normalize) .* dip.mom(i,:)';
      else

inverse/beamformer_dics.m
```diff
@@ -526,6 +526,9 @@
     elseif isstruct(refdip) && isfield(refdip, 'pos') % check if only position of refdip is present
       assert(isnumeric(refdip.pos) && numel(refdip.pos)==3);
       lf1 = ft_compute_leadfield(refdip.pos, grad, headmodel, 'reducerank', reducerank, 'normalize', normalize);
+      if isfield(refdip,'mom'); % check for fixed orientation
+        lf1 = lf1.*refdip.mom';
```
robertoostenveld: if you multiply with refdip.mom(:), then you know for sure that it is a column vector, regardless of the input.

jimher: done

PR #375, opened 2017-03-16
enh: preparing subcortical meshes

smg211:
Added hidden option to prepare_mesh_segmentation that facilitates preparing meshes of oddly shaped volumes (e.g., hippocampus) using vol2surf.

setting cfg.maxsurf = 0 is necessary in case the volume is not continuous
setting cfg.radbound = 2 is necessary to create a surface with more vertices, which facilitates smoothing

See pictures for reference

maxsurf 0 radbound 2 smoothed
maxsurf=0 radbound=2 smoothed
maxsurf 0 radbound 3 smoothed
maxsurf=0 radbound=3 smoothed
maxsurf 0 radbound 2
maxsurf=0 radbound=2
maxsurf 0 radbound 3
maxsurf=0 radbound=3
maxsurf 1
maxsurf=1

### DIFF
--- a/fileio/ft_read_atlas.m
+++ b/fileio/ft_read_atlas.m
@@ -87,7 +87,7 @@ elseif strcmp(x, '.nii') && exist(fullfile(p, [f '.txt']), 'file')
     defaultformat = 'aal';
   end
   fclose(fid);
-elseif strcmp(x, '.mgz') && ~isempty(strfind(f, 'aparc')) && ~isempty(strfind(f, 'aseg'))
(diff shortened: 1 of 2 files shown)


fileio/ft_read_atlas.m
```diff
@@ -87,7 +87,7 @@
     defaultformat = 'aal';
   end
   fclose(fid);
-elseif strcmp(x, '.mgz') && ~isempty(strfind(f, 'aparc')) && ~isempty(strfind(f, 'aseg'))
+elseif strcmp(x, '.mgz') && ~isempty(strfind(f, 'aparc')) || ~isempty(strfind(f, 'aseg'))
```
robertoostenveld: this is also part of another commit.

PR #377, opened 2017-03-17
Improved handling of missing and unspecific stimulus markers

dmatthes1982:
Error messages added for unspecific stimuli
Bug resolved, which occurs, if no stimulus is defined in the marker file


trialfun/ft_trialfun_brainvision_segmented.m
```diff
@@ -97,18 +97,29 @@
...
+    for i=1:length(stim)
+      for j=1:numstim(i)
+        if isempty(value{stim{i}(j)})
+          error('missing a stimulus type definition in the related *.vmrk file');
+        end
+        stimvalue  = sscanf(value{stim{i}(j)}, cfg.trigformat);
+        stimsample = sample(stim{i}(j));
+        stimtime   = (stimsample - begsample(i) + offset(i))/hdr.Fs; % relative to 'Time 0'
+        if isempty(stimvalue)
+          error('the upper case letter of the stimulus value does not match with definition of "cfg.trigformat"');
```
robertoostenveld: thanks for the suggestion. Can you please add the new option to the documentation (help) of the function, otherwise people won't know it exists. Furthermore, I would opt for calling the option "stimformat" rather than "trig format", since the other code and user-feedback also mentions stim/stimulus.

dmatthes1982: Of cource, I can do that. Before I want to notice, that I did not add the option 'cfg.trigformat'. This options has been already in the code. I only added a corresponding error message. But  if nobody disturbs, I will make the change and add the option to the doc. At last, just one formal question: After the commit and the push, do I have to force a new pull request? I'm not so familiar with this procedure, since I'm new on GitHub.

dmatthes1982: Appendix: I just saw cfg.trigformat is already part of the documentation

dmatthes1982: I changed the name and I meanwhile know, that I don't have to force a new push request. ;-)

robertoostenveld: for me the github review process is also still a bit unclear. It took some time to find back my own review request.

robertoostenveld:
thanks

PR #378, opened 2017-03-22
Pull371 fontsize

robertoostenveld:
in PR #371 Lau suggested to use fontsize. As this was not consistent over all plotting functions, I did a clean-up, trying to improve overall font handling. Along the way I detected and fixed a bug in ft_plot_text, when plotting in local axes. I also updated the interpreter (tex->none), cleaned up the help, and made some other changes.


robertoostenveld:
I am not really sure how the "fixmontage" commits ended up in this PR. I'll merge it nevertheless and then check that all is fine.

PR #381, opened 2017-03-23
edit in help: "demean"

sappelhoff:
As opposed to ft_preprocessing, the cfg.demean option in ft_resampledata does not have an associated field "cfg.baselinewindow" to specify a baseline other than the complete trial. The present edit makes this explicit  in the function help.

I am not entirely sure though, whether this is actually true. 

If it IS true however, it would be a nice feature to provide a cfg.baselinewindow for ft_resampledata in the future.


robertoostenveld:
I suggest to solve it differently. You can write in the help
```
% cfg.demean        = 'no' or 'yes', whether to apply baseline correction (default = 'no')
% cfg.baselinewindow = [begin end] in seconds, the default is the complete trial (default = 'all')
```
where 'all' is the default option for cfg.baselinewindow, and then use ft_preproc_baselinecorrect instead of nanmean. That function returns the mean value as 2nd output, which can be added later back to the data (unless the user requested demean=yes, in which case the mean is not added back).

This would be a full improvement. The only difference is that data with nan values is not supported any more (not used much anyway). The best place to add support for that again is much further down in the code, in ft_preproc_polyremoval. I suggest to file that on bugzilla as separate bug.

robertoostenveld:
I don't like the idea of documenting the discrepancies in the implementation, as that suggests that to be the desired state. I rather want them resolved. 

I suggest that you make a new pull request once the baselinewindow option is implemented. For now I will add an explicit error in the code in case the user supplies cfg.baselinewindow.

sappelhoff:
yes, the last weeks have been busy. Hopefully I will get to it soon.

PR #383, opened 2017-03-23
created ft_plot_cloud and added to ft_sourceplot

smg211:
ft_plot_cloud plots functional data as a cloud of points (radius and color depending on funparameter) at sensor positions.  See attached example from 3 sets of hippocampal iEEG depth electrodes.  
ft_plot_cloud example

### DIFF
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -667,6 +667,10 @@ if ~isempty(cfg.title)
 end
 
 %% set color and opacity mapping for this figure
+if hasfun
+  colormap(cfg.funcolormap);
+  cfg.funcolormap = colormap;
+end
 if hasmsk
   cfg.opacitymap = alphamap(cfg.opacitymap);
   alphamap(cfg.opacitymap);
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -30,6 +30,8 @@ function ft_sourceplot(cfg, functional, anatomical)
 %                       'surface',    plots the data on a 3D brain surface
 %                       'glassbrain', plots a max-projection through the brain
 %                       'vertex',     plots the grid points or vertices scaled according to the functional value
+%                       'cloud',      plot the data as point clouds scaled according to the functional value
+%                       
 %
 %   cfg.anaparameter  = string, field in data with the anatomical data (default = 'anatomy' if present in data)
 %   cfg.funparameter  = string, field in data with the functional parameter of interest (default = [])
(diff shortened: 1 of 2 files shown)


ft_sourceplot.m
```diff
@@ -660,11 +666,7 @@ function ft_sourceplot(cfg, functional, anatomical)
   title(cfg.title);
 end
 
-%%% set color and opacity mapping for this figure
-if hasfun
-  colormap(cfg.funcolormap);
```
robertoostenveld: why did you remove the setting of the colormap?

smg211: That was by accident, sorry.  Just added it back

plotting/ft_plot_cloud.m
```diff
@@ -0,0 +1,137 @@
...
+
+% FT_PLOT_CLOUD visualizes iEEG data as a spherical cloud
+%
+% Use as
+%   ft_plot_cloud(pos, 'funparam', dat, ...)
+% where the first argument is the sensor array as returned by FT_ELECTRODEPLACEMENT,
+% FT_READ_SENS or FT_PREPARE_VOL_SENS.
+%
+% Optional input arguments should come in key-value pairs and can include
+%   'funparam'        = Nx1 array where N is the number of sensors, used to
```
robertoostenveld: this should be passed in another way, more like ft_plot_topo and ft_plot_topo3. I'll change that after the merge of this version.

plotting/ft_plot_cloud.m
```diff
@@ -0,0 +1,137 @@
...
+%                       determine color and size of clouds (see below)
+%   'radius'          = scalar, maximum radius of cloud (default = 4)
```
robertoostenveld: what are the units of the radius? The same as the units of pos?

robertoostenveld:
@StolkArjen and @smg211 this requires some more work. Please fetch the latest master, and have  look at test_ft_plot_cloud (in plotting/test). I am not sure whether the unit handling that I added is a smart thing, but anyway: something needs to be done w.r.t. units. Please extend the test script (based on some template EEG electrodes) with useful examples and please review some of the problematic cases that I added there.

Note that I moved the colorbar handling to ft_sourceplot and changed the interface (both for consistency).

robertoostenveld:
@smg211 regarding the units: I put some more thought in it. I think that the code should not try to change the units. The unit varargin option should only be used to know the units of the input data, and to adjust the defaults correspondingly. In the test script I wrote a section with mm/cm/m and commented that it should be the same figure; that is not correct. Consider the following

```
pos_mm = 85*rand(10,3);
pos_cm  = 85*rand(10,3)/10;
pos_m    = 85*rand(10,3)/1000;

ft_plot_unit(pos_mm, 'unit', 'mm');
ft_plot_unit(pos_cm, 'unit', 'cm');
ft_plot_unit(pos_m, 'unit', 'm');
```

Then I would find it desirable that the electrodes show real-world spacing (in mm/cm/m respectively), which means that in the m-case the clouds that default to 4 mm(!) are very small. Like recording surface EEG from an elephant ;-)

PR #384, opened 2017-03-23
improve support for building realtime applications on win64

robertoostenveld:
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3265

The changes in this PR only affect the Makefiles and source code (primarily headers, compiler/platform/platform_includes) and some compatibility functions. 

I have updated the Makefiles for the latest version of mingw-w64 in the 32 and 64 bit version, and for Cygwin, also in the 32 and 64 bit version.


robertoostenveld:
@Horschig I have not recompiled anything (that is: I have recompiled many times for testing, but have not included any fresh executable in the PR). As far as I can tell, the support for windows 64 bit is now as good as it already used to be for windows 32 bit.

PR #385, opened 2017-03-24
Layouttweaks

roemervandermeij:
Two small tweaks:

1st commit:
Allows proper projection of headshape/mri if cfg.projection ~= orthographic or if viewpoint = [], using cfg.rotate and elproj.m as in subfunction sens2lay

2nd commit
Makes sure the headshape/mri outline is reused if possible

I wanted to implement a labellocation key-value pair in ft_plot_lay using text property 'extent', with:
center = nothing changes
north = on top op electrode box
south = below electrode box
west = on left of box (base rotated 90 degrees counterclockwise)
east = on right of box (base rotated 90 degrees clockwise)

But I'd rather resolve the hpos/vpos issue first (http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3252), as it involves ft_plot_lay.

### DIFF
--- a/ft_prepare_layout.m
+++ b/ft_prepare_layout.m
@@ -1499,12 +1499,6 @@ elseif ~isempty(cfg.mri)
   outlbase = ft_prepare_mesh(cfgpm, outlbase);
 end
 
-% check for presence of cfg.viewpoint
-if isempty(cfg.viewpoint)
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks

PR #387, opened 2017-03-27
Added spm12's new cfg_getfile

jens-k:
See comment to #376


schoffelen:
Oh no,  ~1500 lines of additional code, just to get it running on windows.

schoffelen:
Hi Jens, 
I don't get it: if cfg_getfile is called by spm_select, how come this does not crash higher up? spm_select is not in external/spm12

jens-k:
Good point. I had an unneccessary function call in my test script that added spm8 to the path (so external/spm8/spm_select was called which chokes on 'ListDrives' as input). Seems another 100 lines of code for spm_select will be needed to support windows. Totally worth it ;)

A - not very clean - alternative would be to write our own function that handles exactly this one call.

schoffelen:
thanks

PR #388, opened 2017-03-28
ENH - recursive use of cfg.showcallinfo and cfg.feedback in serveral functions

dmatthes1982:
If you call ft_freqanalysis() and ft_singleplotTFR() with cfg.showcallinfo='no' and cfg.feedback='no', contrary to the expectations some output appears on the terminal. This is a result of the missing bypass of these attributes to low-level functions. I fixed these behaivor for the two mentioned functions, but I guess this problem might be a general thing and have to fixed in futher functions.


robertoostenveld:
thanks, I'll merge. If you encounter more of these cases, please fix them as well. 

PS in general the tmp = keepfields(...) syntax is preferred over tmpcfg.something=cfg.something...

dmatthes1982:
thanks for this hint. I let the code in the existing and file-specific 
syntax, but next time I willingly change it to keepfields.

PR #390, opened 2017-04-03
bugfix in external/openmeeg/openmeeg_eeg_leadfield_example.m

hsonntag:
The example script external/openmeeg/openmeeg_eeg_leadfield_example.m produced several errors and stopped.
Some minor problems are fixed and the script runs without errors or warnings in this revision.


robertoostenveld:
thanks!

PR #391, opened 2017-04-04
ENH: plotting now occurs in headcoord space, ensuring up is up. plus,…

StolkArjen:
… a few more optimization/visualization tweaks


robertoostenveld:
I did not run it (too busy with toolkit) but assume it will work. Thanks for also changing the radioboxes to checkboxes.

PR #392, opened 2017-04-05
FIX: multiplied the pos range by 2 to adapt the unit estimation to iE…

StolkArjen:
…EG data


robertoostenveld:
makes sense. The iEEG cloud is usually smaller than the whole head.

PR #393, opened 2017-04-05
ENH: added support for remontaging electrodes

StolkArjen:
in sum,
- ft_preprocessing updates the sens locations according to montage (ensured the output only has channels corresponding to cfg.channel)
- ft_appenddata has a cfg.appendsens input option, which ensures multiple sens (with 'inconsistent' labels, as in the case of iEEG) are concatenated across input arguments
- ft_appendsens received minor updates

### DIFF
--- a/forward/ft_apply_montage.m
+++ b/forward/ft_apply_montage.m
@@ -385,16 +385,16 @@ switch inputtype
       if keepchans
         sens.chanpos = sens.chanpos(sel2,:);
       else
+        if ~isfield(sens, 'chanposold')
+          % add a chanposold only if it is not there yet
+          sens.chanposold  = sens.chanpos;
+          %  also keep the old label, type and unit for reference
+          sens.labelold    = inputlabel;
+          sens.chantypeold = inputchantype;
+          sens.chanunitold = inputchanunit;
+        end
         % compute the channel positions as a weighted sum of the original ones
         sens.chanpos = posweight * sens.chanpos;
-      end
-      if ~isfield(sens, 'chanposold')
-        % add a chanposold only if it is not there yet
-        sens.chanposold  = sens.chanpos;
-        %  also keep the old label, type and unit for reference
-        sens.labelold    = inputlabel;
-        sens.chantypeold = inputchantype;
-        sens.chanunitold = inputchanunit;
       end
     end
     
--- a/forward/ft_apply_montage.m
+++ b/forward/ft_apply_montage.m
@@ -385,16 +385,16 @@ switch inputtype
       if keepchans
         sens.chanpos = sens.chanpos(sel2,:);
       else
-        if ~isfield(sens, 'chanposold')
-          % add a chanposold only if it is not there yet
-          sens.chanposold  = sens.chanpos;
-          %  also keep the old label, type and unit for reference
-          sens.labelold    = inputlabel;
-          sens.chantypeold = inputchantype;
-          sens.chanunitold = inputchanunit;
-        end
         % compute the channel positions as a weighted sum of the original ones
         sens.chanpos = posweight * sens.chanpos;
+      end
+      if ~isfield(sens, 'chanposold')
+        % add a chanposold only if it is not there yet
+        sens.chanposold  = sens.chanpos;
+        %  also keep the old label, type and unit for reference
+        sens.labelold    = inputlabel;
+        sens.chantypeold = inputchantype;
+        sens.chanunitold = inputchanunit;
       end
     end
     
--- a/ft_appenddata.m
+++ b/ft_appenddata.m
@@ -165,12 +165,13 @@ for j=1:Ndata
 end
 
 % check consistency of sensor positions across inputs
-haselec = 1;
-hasgrad = 1;
+haselec = 0;
+hasgrad = 0;
+hasopto = 0;
 for j=1:Ndata
-  haselec = isfield(varargin{j}, 'elec') && haselec;
-  hasgrad = isfield(varargin{j}, 'grad') && hasgrad;
-  hasopto = isfield(varargin{j}, 'opto') && hasopto;
+  haselec = isfield(varargin{j}, 'elec');
+  hasgrad = isfield(varargin{j}, 'grad');
+  hasopto = isfield(varargin{j}, 'opto');
 end
 
 % check whether the data are obtained from the same datafile in case either
@@ -323,7 +324,7 @@ if shuflabel
 end
 
 removesens = 0;
-if haselec || hasgrad || hasopto
+if haselec || hasgrad || hasopto %FIXME all might now theoretically be true due to changes at line 170
   sens = cell(1, Ndata);
   for j=1:Ndata
     if haselec, sens{j} = varargin{j}.elec; end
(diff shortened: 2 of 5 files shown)


StolkArjen:
Still under construction

StolkArjen:
Please double check this. The overarching goal here is to allow automatic updating of sens information that has been appended to the data (i.e. data.elec), in the context of iEEG analysis.

Of course there's the montage way, but in the case the user does the standard way of rereferencing, e.g.

load('SubjectUCI29_data.mat', 'data');
cfg               = [];
cfg.channel = ft_channelselection({'LPG*', 'LTG*'}, data.label);
cfg.reref      = 'yes';
cfg.refchannel = 'all';
reref = ft_preprocessing(cfg, data);
 
until now this would NOT update the sens information (all original electrodes remain in the elec field). However, given that this may be intended behavior (cfg.updatesens = 'yes' by default), now a montage is created on the fly when cfg.reref = 'yes'. 

One of the positive side-effects is that not only the elec field is updated/adjusted to match the actual functional data, but since ft_apply_montage is called, chanposold and labelold are added to the elec structure. This is useful when appending the output with say bipolar re-referenced data, hence the adjustments to ft_appendsens and ft_apply_montage.

forward/ft_apply_montage.m
```diff
@@ -385,17 +385,17 @@
...
-          sens.chanposold  = sens.chanpos;
-          %  also keep the old label, type and unit for reference
-          sens.labelold    = inputlabel;
-          sens.chantypeold = inputchantype;
-          sens.chanunitold = inputchanunit;
-        end
         % compute the channel positions as a weighted sum of the original ones
         sens.chanpos = posweight * sens.chanpos;
       end
+      if ~isfield(sens, 'chanposold')
```
robertoostenveld: why outside the "if"? 
chanposold should only be kept in case keepchan.

ft_appenddata.m
```diff
@@ -336,6 +322,31 @@
...
+    if j>1
+      if ~isequaln(sens{j}, sens{1})
+        removesens = 1;
+        warning('sensor information does not seem to be consistent across the input arguments');
+        break;
+      end
+    end
+  end
+  if strcmp(cfg.appendsens, 'yes')
+    fprintf('concatenating sensor information across all input arguments\n');
```
robertoostenveld: This needs to be redesigned to deal with the combined MEG+EEG case.

robertoostenveld: i.e. the output can have both a grad and an elec if I do appendsens=yes with EEG as input 1 and MEG as input 2.

robertoostenveld:
I understand the use case for iEEG, but am not confident that you considered the other data types well enough. 

please start with a test script that demonstrates what you want to achieve. E.g. call it "test_pull393.m" and use some simple on-the-fly constructed EEG, MEG, iEEG and NIRS data to demonstrate how you want to deal with it.  

Furthermore, the appending should also be supported at later stages or processing. Please demonstrate the desired functionality for ft_appendtimelock and ft_appendfreq as well.

StolkArjen:
Thanks for the prompt feedback. 

1) What would be the desired output for appending an EEG and an MEG dataset? Should the output have both an elec and a grad field, or should it have none of these? I presume the same arguments hold for NIRS.

2) In the case of appending a bipolar-rereferenced iEEG dataset with a monopolar iEEG dataset, I want ft_appendsens to append labelold fields (and chanposold), and if not present for one of the datasets, use the regular label field. That way one keeps information about the original electrode positions (i.e. before rereferencing) and the current channel positions. Would you say this is a preferred way of doing it? And what should happen to the elecpos field? Currently it's has more positions than chanpos because elecpos is not adjusted during rereferencing (only chanpos is). Hence the labelold field also relates to the current elecpos field. For an idea of what an appended structure might end up looking like:

screen shot 2017-04-07 at 10 15 45 pm

StolkArjen:
The proposed behavior, as currently implemented, is (copying over from test_pull393.m):

%% append
% EEG and MEG
cfg = [];
cfg.appendsens = 'yes';
data1 = ft_appenddata(cfg, eeg, meg);
assert(isequal(numel(data1.label),6)) % 6 labels
assert(isequal(numel(data1.elec.label),3)) % 3 elec labels
assert(isequal(numel(data1.grad.label),3)) % 3 grad labels

% EEG and NIRS
cfg = [];
cfg.appendsens = 'yes';
data2 = ft_appenddata(cfg, eeg, nirs);
assert(isequal(numel(data2.label),6)) % 6 labels
assert(isequal(numel(data2.elec.label),3)) % 3 elec labels
assert(isequal(numel(data2.opto.label),3)) % 3 opto labels

% monopolar iEEG and bipolar iEEG
cfg = [];
cfg.appendsens = 'yes';
data3 = ft_appenddata(cfg, ieeg, ieegb);
assert(isequal(numel(data3.label),5)) % 5 ieeg labels
assert(isequal(numel(data3.elec.label),5)) % 5 ieeg labels
assert(isequal(numel(data3.elec.labelold),6)) % 6 ieeg labelolds

%% do not append (discard inconsistent sens information) 
% EEG and MEG
data4 = ft_appenddata([], eeg, meg);
assert(isequal(numel(data4.label),6)) % 6 labels
assert(~isfield(data4, 'elec')) % no elec struc
assert(~isfield(data4, 'grad')) % no grad struc

% EEG and NIRS
data5 = ft_appenddata([], eeg, nirs);
assert(isequal(numel(data5.label),6)) % 6 labels
assert(~isfield(data5, 'elec')) % no elec struc
assert(~isfield(data5, 'opto')) % no opto struc

% monopolar iEEG and bipolar iEEG
data6 = ft_appenddata([], ieeg, ieegb);
assert(isequal(numel(data6.label),5)) % 5 ieeg labels
assert(~isfield(data6, 'elec')) % no elec struc



>> If agreed upon this behavior as desired, we need to implement support for ft_appendsens in concatenation functions other than ft_appenddata. And, a solution may need to be found for appending tra matrices still.

StolkArjen:
Addendum; all old behavior (removal or copying over sens information when consistent) is identical still.

robertoostenveld:
looks good, thanks also for the test script!

StolkArjen:
Thanks for helping out with this complex yet desirable functionality. Running test_pull393.m again, an error occurs when trying to append sens from difference senstypes. This is due a piece of code in ft_appenddata assuming that the different inputs always have the same senstype:

for j=1:Ndata
      if isfield(varargin{j}, 'elec')
        elec{j} = varargin{j}.elec;
      end
      if isfield(varargin{j}, 'grad')
        grad{j} = varargin{j}.grad;
      end
      if isfield(varargin{j}, 'opto')
        opto{j} = varargin{j}.opto;
      end
    end
    if hasgrad, data.grad = ft_appendsens([], grad{:}); end
    if haselec, data.elec = ft_appendsens([], elec{:}); end
    if hasopto, data.opto = ft_appendsens([], opto{:}); end

.. which in the case of varargin{1}.elec and varargin{2}.grad will forward to ft_appendsens an empty grad{1} and a full grad{2}. This causes ft_appendsens to fail when doing checks on the input. It can be solved by either having ft_appendsens dealing with empty inputs, or do use a {end+1} (with say elec ={} above the loop) instead of j in the loop. Any preference?


>>
Another related issue is that the following code

cfg               = [];
cfg.channel       = ft_channelselection({'LPG*', 'LTG*'}, data.label);
cfg.reref      	  = 'yes';
cfg.refchannel    = 'all';
reref_grids = ft_preprocessing(cfg, data);

will produce an output which elec field still has all channels instead of the ones requested for selecting from the original neural data. In case of appending multiple rereferenced data structures, I found a total of 150 channels to be blown up to over 1000 electrodes in the appended data structure (8 copies of itself or so). Thus, ideally we also apply cfg.channel to the data.elec field when rereferencing. Any suggestions for what would the best place to do this, assuming you agree?

StolkArjen:
Regarding the first point, see proposed change in https://github.com/fieldtrip/fieldtrip/pull/411

PR #394, opened 2017-04-05
bugfix in eeg_leadfield4.m for dipoles with negative z-coordinate

hsonntag:
Leadfield computation in eeg_leadfield4.m treats dipoles on the negative z-axis as a special case, which makes totally sense but there was a bug in the indexing of the z-coordinate of electrode positions.
The electrode matrix elc has size Nx3 and hence the z-coordinates are elc(:,3).

### DIFF
--- a/forward/private/eeg_leadfield4.m
+++ b/forward/private/eeg_leadfield4.m
@@ -66,7 +66,7 @@ elc = r4 * elc ./ [dist dist dist];
 
 % rotate everything so that the dipole is along the pos. z-axis
 % only if the dipole is not in the origin or along the positive z-axis
-if R(1)~=0 | R(2)~=0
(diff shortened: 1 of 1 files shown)


hsonntag:
I noticed that my commit message is inaccurate, should I change it to:
FIX: eeg_leadfield4 for dipoles on the negative z-axis
?

robertoostenveld:
Thanks, don't worry about the comment. 

I am not able to quickly test the whole thing, but I assume that you have carefully looked at it and that the spatial topographies in that specific case are consistent with what you would expect, right?

hsonntag:
I tested the leadfield for dipoles on the negative z-axis against respective openmeeg BEM solutions and the results were as expected.

PR #395, opened 2017-04-06
FIX - Handling of the STATUS channel in Biosemi BDF files corrected.

dmatthes1982:
The data of this channel will not longer be „calibrated“.

Biosemi in general puts for the status channel the eeg calibration parameters to the header. (take a look at the Test file at the end of this page: https://www.biosemi.com/faq/file_format.htm / that this is a general approach was also confirmed by the Biosemi developers) However a calibration of the 'Status' channel with the eeg calibration parameters results in nonsense. Therefore i have corrected it.

### DIFF
--- a/fileio/private/read_biosemi_bdf.m
+++ b/fileio/private/read_biosemi_bdf.m
@@ -47,10 +47,10 @@ function dat = read_biosemi_bdf(filename, hdr, begsample, endsample, chanindx)
 %
 % $Id$
 
-if nargin==1
-  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-  % read the header, this code is from EEGLAB's openbdf
-  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+if nargin == 1
+% -------------------------------------------------------------------------
+% read the header, this code is from EEGLAB's openbdf
+% -------------------------------------------------------------------------
   FILENAME = filename;
 
   % defines Seperator for Subdirectories
@@ -194,9 +194,9 @@ if nargin==1
   % close the file
   fclose(EDF.FILE.FID);
   
-  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-  % convert the header to Fieldtrip-style
-  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% -------------------------------------------------------------------------
+% convert the header to Fieldtrip-style
+% -------------------------------------------------------------------------
   if any(EDF.SampleRate~=EDF.SampleRate(1))
     error('channels with different sampling rate not supported');
   end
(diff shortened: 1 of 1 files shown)


fileio/private/read_biosemi_bdf.m
```diff
@@ -254,20 +254,52 @@
...
+  calib = diag(EDF.Cal(chanindx)); % get the calibration coefficients
+  
+  % Change the calibration coefficient of the 'Status' channel to 1
+  %
+  % Biosemi in general puts also for the status channel the eeg 
+  % calibration parameters to the header. A calibration of the 'Status'
+  % channel with the eeg calibration parameters results in nonsense.
+  % Therefore it will be corrected here.
+  chanlabels = cellstr(EDF.Label); 
+  labelindx = find(strncmp(chanlabels, 'Status', 6)); % locate 'Status' label
```
robertoostenveld: this should be a case-insensitive comparison, i.e. use strcmpi rather than strncmp

dmatthes1982:
Hi Robert Oostenveld,

Thanks for your reply. I have attached I file which was recorded by the 
'mBrain train smarting' mobile eeg. In this file are 8 events, which 
have various values of the range from 101 to 104. When I use 
ft_read_event() to read this events I get back for every of this 8 
events the number 48011. Otherwise if I import the data with eeglab I 
get back the correct values.

Therefore I have investigated the problem to find out its root. In doing 
this I saw that the PhysMin, PhysMax, DigMin and DigMax values for the 
'Status' channel in the file header are similar to the values of the eeg 
channels. As a result the 'Status' channel will be 'calibrated'  (with 
factor 0,2233 in my case) in the original version of read_biosemi_bdf an 
later (i forgot in which file) rounded to integer. This behavior leads 
to the wrong value of 48011. EEGLab instead is ignoring the 
Phys/Dig-Values and therefore gets back the right values.

Before I changed read_biosemi_bdf, I asked the developers of mBrain 
train, why they put such nonsense values to the header for the status 
channel. They responded that they doing this in the same way like 
Biosemi. So I asked also Biosemi, if this is a general approach to put 
the eeg Phys/Dig limits to the related header cells of the 'Status' 
channel or if this is a mistake in the example. They said that every 
device/software of their company does it in the same way. Therefore I 
made this changes in read_biosemi_bdf.

You can also choose this example files: 
https://www.biosemi.com/download/BDFtestfiles.zip, to review the Biosemi 
approch.

All the best,

Daniel Matthes

robertoostenveld:
Good to know that it is not a biosemi file, but a BDF file written by another EEG vendor's software. Can you send me an email with a link to the attachment? r.oostenveld - at -donders.ru.nl

robertoostenveld:
Thanks for the file. This code confirms the problem

```

filename = {
  '050327BH_overCZnoAlpha.bdf'
  'Newtest17-2048.bdf'
  'Newtest17-256.bdf'
  'mbrain-train-smarting-mobile-eeg.bdf'
  };

for i=1:numel(filename)
  hdr = ft_read_header(filename{i});
  index = find(strcmpi(hdr.label, 'STATUS'));
  assert(hdr.orig.Cal(index)==1, 'calibration appears incorrect');

  sdat = ft_read_data(filename{i}, 'chanindx', index);
  assert(all((sdat-round(sdat))==0), 'non-integer values in STATUS channel');
end
```

it fails for your file only. 

I consider that a bug in your file, i.e. in the brain train smarting mobile eeg software. For the actual biosemi files (I have three, two from their website and one locally recorded) the calibration is correctly stored in the header, so your statement about Biosemi doing it wrong themselves does not hold.

robertoostenveld:
I have made an alternative fix. I will close this one and merge the other. Please look in the code and test_pull395 script for details.

BogiZmaj:
Hey everyone, I am employed with mBrainTrain. We have just seen this thread. Still have problems with mBT bdf files? I have used the test_pull395.m and with some of our own files, and it seems to be working now. Please let me know if the issue persists and in that case, please share the file that does not work, so that we take a look. Sorry for this late response :)

Cheers,
Bogdan

robertoostenveld:
thanks, much appreciated. I think all is fine for now.

PR #396, opened 2017-04-06
FIX: allow both input from cfg.elec and cfg.channel, e.g. to continue…

StolkArjen:
… working on an earlier project


StolkArjen:
Ready to be pulled.

PR #397, opened 2017-04-07
ENH - fix the calibration of the status channel for mbraintrain BDF files

robertoostenveld:
which is incorrect in the file itself, see issue #395

### DIFF
--- a/fileio/private/read_biosemi_bdf.m
+++ b/fileio/private/read_biosemi_bdf.m
@@ -27,7 +27,7 @@ function dat = read_biosemi_bdf(filename, hdr, begsample, endsample, chanindx)
 %    chanindx        index of channels to read (optional, default is all)
 % This returns a Nchans X Nsamples data matrix
 
-% Copyright (C) 2006, Robert Oostenveld
+% Copyright (C) 2006-2017, Robert Oostenveld
 %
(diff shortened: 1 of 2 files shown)


robertoostenveld:
note that this also fixes the offset, which was not added to the data in the old fieldtrip code. Now it is.

robertoostenveld:
@dmatthes1982 you should have a look at the handling of the status channel in ft_read_event. Your suggested change at which you commented "only the lower 16 bits of the 'STATUS' channel are carrying trigger informations" is dealt with differently in that function. Battery, CMS and other information should remain available to the user. 

Let me know when mbraintrain is using the highest byte differently. In your example it was all zero.

dmatthes1982:
Thanks @robertoostenveld for this commit. I agree with you the Battery, 
CMS and other information should remain, which are used by mbrain train 
in the same way.

Furthermore, I have to apologize. After double-checking the Headers of 
mbrain an Biosemi I noticed an error of mbrain within the PhysMin an 
PhysMax cells of the 'Status' channel. If I had noticed this before, I 
would not have made the pull request. The next time I will be more thorough.

I have rcently contacted mBrain train and they will correct this error 
in the next release.

robertoostenveld:
thanks. Nice to hear that mBrain is listening to their customers for making improvements to their products.

PR #401, opened 2017-04-10
ENH - add cfg.feedback to ft_freqgrandaverage.m

dmatthes1982:


robertoostenveld:
Thanks for the contribution. I cam imagine more of these coming... But I don't think it scales very well. It adds conditional lines of code to each high-level function and the order of the logic can be different every time. 

How about doing it like this

if istrue(cfg.feedback)
  ft_feedback = @fprintf
else
  ft_feedback = @dummy
end

where dummy (another name needed!) would be a function like 

function dummy(varargin)
% do nothing
end function

Could you try that out here? You can insert a local function for the moment, but I think it should be shared between all code.

robertoostenveld:
I realize that my suggestion was rather short, esp. in case you don't know function handles. The idea is to make a function handle (which acts as a normal function, i.e. it is a named pointer to another function) which works like this

```
ft_feedback("whatever")
```
which then results in either fprintf or dummy being called.

dmatthes1982:
Thanks for the replies and the suggestion. I'm familar with funktion 
handles, but thanks for the explanation anyway. However, I have another 
question.  Since this is not the first function using cfg.feedback, is 
there a function with an implementation of cfg.feedback likewise your 
suggestion, which I could use as a template?

robertoostenveld:
cfg.feedback was primarily meant for silencing ft_progress, which prints stuff like "trial 276 of 341287" a million times.

Searching at the top level gives this
```
mac011> grep istrue.*feedback *.m
ft_defacevolume.m:  if istrue(cfg.feedback)
ft_electrodeplacement.m:    [pos, lab] = elec1020_locate(headshape.pos, headshape.tri, nas, ini, lpa, rpa, istrue(cfg.feedback));
ft_redefinetrial.m:fb   = istrue(cfg.feedback);
```

ft_redefinetrial is the closest match in terms of style. I see that forward/ft_headmodel_localspheres also has some related code. In test/test_warp there is a "global fb", which relates to an old way of doing it, but I don't think that is used any more (i.e. private functions would look at the global fb variable).

I know that we have discussed feedback in more general terms in the past, specally relating to throwing warnings if in the code it not 100% sure whether stuff goes as it should. Perhaps a broader plan should be made, that considers the proposed ft_feedback function in relation to ft_progress and ft_warning.

Just to give some ideas: For ft_progress there are different styles. For user feedback you could also imagine different levels. We already have the "warning_once" functionality (now absorbed in the ft_warning function using the timeout). We could consider treating this like https://en.wikipedia.org/wiki/Syslog levels.

robertoostenveld:
> ft_redefinetrial is the closest match in terms of style.

Here I meant that it is the closest to what you are after. 

For ft_progress please search the code if you want to see examples.

dmatthes1982:
I just wanted to implement your suggestions regarding cfg.feedback, while I another solution came to my mind.

Wouldn't it be better, if I create a function like ft_feedback(cfg.feedback, varargin) which generates debug output or not depending on the current cfg.feedback value instead of implementing function handles and building a ft_discardfeedback function.

If not, I would appreciate to get your feedback about my suggested function name ft_discardfeedback an the idea to place this function in the folder fileio/private.

By the way, if you prefer my new suggestion, I would store the ft_feedback function just as fileio/private.

robertoostenveld:
I think a good implementation would be if we would allow 
1) the user to specify which level of feedback he wants
2) the programmer to indicate how important the feedback is

Looking at https://en.wikipedia.org/wiki/Syslog#Severity_level this could for example be implemented with

```
ft_ emergency(...)
ft_alert(...)
ft_critical(...)
ft_error(...)
ft_warning(...)
ft_notice(...)
ft_info(...)  or ft_informational
ft_debug(...)
```

where each function would have the same arguments. The first argument would be the user-specified cfg.feedback flag, subsequent arguments would be similar to those of printf. 

We have ft_warning, although not exactly in this implementation. The relevant ones are the lower ones: notice, informational, debug. I think that most information currently printed is at the informational level. 

If you could start with implementing ft_info(), then we can start with that and extend it.

robertoostenveld:
I propose to make the functions in fieldtrip/utility.

dmatthes1982:
Here is a first draft of the discussed functionality.

dmatthes1982:
Good morning @robertoostenveld, 9 days ago I made the first draft of the discussed functionality.Should I add more changes?

robertoostenveld:
Your suggested change would require a lot of changes to the FT code, and would require cfg.feedback to be known everywhere. 

I was more thinking along the lines of 

% main function 
ft_info('you can ignore this')
ft_notice('but this is getting more important'); 
ft_warning('this is not to be ignored, unless you really insist')
ft_error('sorry, I have to stop now');

and cfg.feedback being 'info', 'notice', 'warning', or 'error'. If unspecified, it would default to something. 

If info, it shows info and all higher levels
If notice, it shows notice and all higher levels
etc.

This requires that cfg.feedback is somehow handled as global variable or so. The present ft_warning "once" handling using the ft_default global might serve as template. This part is still a bit fuzzy to me. It might have to be done in ft_preamble.

For the future programmer it means that he/she can think about the importance of the situation, but ignore the implementation details. 

For the transition/refactoring it would mean that we would replace all printf with ft_notice, and some with ft_info.

robertoostenveld:
@schoffelen could you have a look at this and think along with us on the design of the user-feedback?

robertoostenveld:
I thought about it. How about this

mac011> git commit
[pull401-feedback dad28c3] proposal for user-configurable notification, see https://github.com/fieldtrip/fieldtrip/pull/401
 3 files changed, 100 insertions(+)
 create mode 100644 test/test_pull401.m
 create mode 100644 utilities/ft_info.m
 create mode 100644 utilities/ft_notice.m

robertoostenveld:
this would have to implemented consistently with the current (and long-standing) ft_warning. Its primary reason for existence is that it can show a warning once every xx seconds. 

I also mentioned msgId in the help, but did not add it to the code yet. My idea is that you should also be able to do 

ft_info('off', msgId) and ft_info('on', msgId)

just like MATLAB warning.

robertoostenveld:
I just added demonstration and fix to my branch https://github.com/fieldtrip/fieldtrip/compare/master...robertoostenveld:pull401-feedback, please check it out.

robertoostenveld:
@dmatthes1982 could you have a look at my PR and let me know whether you like it this way?

dmatthes1982:
Hi @robertoostenveld, sorry for my late answer. I was so busy at work and at home in the preceding weeks, that I had no time to consider your suggestion. Your solution seems to be good for me, but I have to be honest, I did not completely understood, how to switch off the ft_info message when required. In which way should I call ft_info in ft_freqgrandaverage at line 185 and 189?

robertoostenveld:
The idea is that the programmer decides on the importance level by calling the associated ft_XXX in the code (with XXX the importance level). 

That would simply be
```
ft_info('not computing average, but keeping individual %s for %d subjects\n', cfg.parameter{k}, Nsubj);
```

And the user gets to choose by specifying cfg.feedback at what level he sees the feedback. In this case (with the "info" level), you could specify 

```
cfg.feedback = 'info' or 'debug' 
```

to show the feedback, or you could specify 

```
cfg.feedback = 'emergency', 'alert', 'critical', 'error', 'warning', 'warning_once(?)', 'notice'
```

to not show the feedback.

To allow the old-style compatibility with cfg.feedback=yes/no, I suggest to map those onto yes=info and no=notice. To me that seems to be the sweet spot where tweaking is desired. 

Note that 'emergency', 'alert' and 'critical' are all higher than error. On case of error, we would just show a MATLAB error and the code would halt. It does not get more severe, so these are not needed. The most silent behavior that the user therefore would get is with cfg.feedback='error', in which case only errors are shown, but warning+notice+info+debug are not shown.

Updating all FT code would involve a search-and-replace (with some care) of fprintf, replacing it with ft_info.

dmatthes1982:
Thank for this explanation in detail. This transformation is very coherent for my point of view. So I will change my own scripts from cfg.feedback='no' to cfg.feedback='error' and if I get some output, which I don't expect, then I will contribute some adaptions.

robertoostenveld:
let's close the discussion here, and move on with #445

PR #402, opened 2017-04-10
implement singletrial analysis in ft_sourceanalysis

mrksh:
concerns the noted bug in the code when cfg.singletrial == 'yes'. The provided code now
- calculates average Cf
- runs the actual source reconstruction with the average Cf
- uses the filter computed thereby as precomputed filter for the single-trial Cf s afterwards

Is this the intended behavior? If I missed something, could you point me to the remaining problems?


schoffelen:
This PR points to a 'unknown repository'. Sorry for not following this up, but it seems so deprecated now, that it's not worthwile to prioritise this.
As a side note, the 'singletrial', 'rawtrial' functionality in ft_sourceanalysis is a poorly maintained corner of the code that has been nominated since already a long time to be re-implemented. Current recommended way to deal with all this, is to call ft_sourceanalysis twice, once with 'keepfilter' to be true, based on the covariance/csd of all data, and then in a second round on a subset of trials (with a single trial in the limit) to compute single trial activity. For frequency domain applications, the 'pcc' method, combined with 'fourierspctrm' in the input gives the single trial fourier coefficients for free. For time domain applications, it's straightforward to compute single trial virtual channel time courses by means of left multiplying the single trial channel data with the selected spatial filter. This requires the user to write some own code.

PR #404, opened 2017-04-11
ENH - use audioinfo/audioread/audiowrite instead of wavread/wavwrite

robertoostenveld:


fileio/ft_read_event.m
```diff
@@ -1186,7 +1186,7 @@
     
   case 'gtec_hdf5'
     % the header mentions trigger channels, but I don't know how they are stored
-    error('this has not been implemented yet due to a lack of a good example file');
+    warning('event reading for hdf5 has not yet been implemented due to a lack of a good example file');
```
robertoostenveld: hmm, this should not have been in this PR. But I don't think it is a problem per see, so I'll just keep it as is.

PR #405, opened 2017-04-13
Bug3280 read electrode positions from res4

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3280

### DIFF
--- a/fileio/ft_read_sens.m
+++ b/fileio/ft_read_sens.m
@@ -70,7 +70,7 @@ end
 
 % get the options
 fileformat     = ft_getopt(varargin, 'fileformat', ft_filetype(filename));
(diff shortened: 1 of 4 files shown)


fileio/ft_read_sens.m
```diff
@@ -168,20 +166,25 @@
     hdr = ft_read_header(filename, 'headerformat', fileformat, 'coordsys', coordsys, 'coilaccuracy', coilaccuracy);
     % sometimes there can also be electrode position information in the header
     if isfield(hdr, 'elec') && isfield(hdr, 'grad')
+      if isempty(senstype)
```
robertoostenveld: default handling is not yet in place for CTF

robertoostenveld: oh, the default is actually applied not only for fif, but for all datasets where ft_read_header is used. So no problem here.

robertoostenveld:
I looked at all changes with Marios and will merge.

PR #406, opened 2017-04-14
Resampledata baselinewindow

sappelhoff:
As discussed here:

https://github.com/fieldtrip/fieldtrip/pull/381

Will not support NaN values anymore. Works well, when compared to the
cfg.baselinewindow option in ft_preprocessing.


ft_resampledata.m
```diff
@@ -193,9 +193,19 @@
...
     end
 
+    % Compute mean based on a window (cfg.baselinewindow) or the whole
+    % trial. This does not support nan values anymore. Fix in
+    % ft_preproc_plyremoval. 
+    % previously --> bsl = nanmean(data.trial{itr},2);
+    if ~strcmp(cfg.baselinewindow, 'all')
+        begsample   = ceil(data.fsample*cfg.baselinewindow(1))+find(data.time{itr}==0);
+        endsample   = floor(data.fsample*cfg.baselinewindow(2))-1+find(data.time{itr}==0);        
+        [~,bsl]     = ft_preproc_baselinecorrect(data.trial{itr}, begsample, endsample);
```
robertoostenveld: the notation ~ for unused output variables is not supported by older MATLAB versions (which we still need to support). This should be changed into "dum".

robertoostenveld:
thanks

PR #407, opened 2017-04-18
Fix import of EGI MFF files with additional PIB channel labels

ThreeFive-O:
Wrong index access of data in the import function for the PIB channel labels for EGI MFF files.
Unfixed version accesses the number of channels for PIB channels twice, whereas the sum of the EEG channels and the PIB channels is desired.


robertoostenveld:
thanks, much appreciated!

PR #410, opened 2017-04-24
ENH: added and tested the ability to use ft_convert_coordsys under SP…

StolkArjen:
…M12, which resorts to using the OldNorm sub toolbox. as a result, explicitly switching to SPM8 is no longer needed. 

Testing was done with both opt = 1 and opt = 2, and it was able to run through under both conditions.


robertoostenveld:
where is the test script?

StolkArjen:
Added the testscript to test_spm12

PR #411, opened 2017-04-24
FIX: support appending inputs from different senstypes (according to …

StolkArjen:
…test_pull393.m)

with this change, test_pull393 runs through as expected


robertoostenveld:
this will have to wait until http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3287 is resolved

robertoostenveld:
Please review the new implementation of ft_appenddata, ft_appendfreq and ft_appendtimelock. They all support keeping grad/elec/opto, and they all work with ICA decomposed data.

At this moment I don't see a reason for further work on this. Please open a bug on bugzilla that clearly outlines the case for revisiting this issue.

PR #413, opened 2017-04-25
Bug3287 reimplemented ft_apendxxx functions

robertoostenveld:
ft_appenddata/timelock/freq now use a shared private/append_common function. This should improve consistency between the three w.r.t. handling trial info, elec, etc.

It was quite of a struggle to get the new implementation working exactly the same as the old one. I made a new test script, ran some of the existing test scripts, and finally ran the full test batch a few times. At the moment it is running for (hopefully) the last time, after which it can be merged.


robertoostenveld:
there were still issues with the test batch, a.o. with topo/topolabel. I will run once more.

robertoostenveld:
it now all looks ok.

robertoostenveld:
we might want to revisit ft_appendsource as well. But not now. I have already wasted 3 days of my life on this.

PR #414, opened 2017-04-25
Fix: Check for signal processing toolbox before loading external drop-ins

ThreeFive-O:
Right now the MATLAB internal signal processing toolbox functions are always overridden by the fieldtrip functions.
From what I understand is, that the drop-ins are a substitute if the corresponding MATLAB toolbox is not available. Hence, the MATLAB toolbox should be favored.


robertoostenveld:
the drop-ins are used to reduce the need for toolboxes. Many sites (like ours) have a network license server with many MATLAB core licenses but much fewer toolbox licenses. For those the existing scheme (for signal+stats+image) serves to reduce toolbox use, so that others that need the toolboxes age larger chances for getting a license. 

What I propose is that you make the following changes
1) that ft_platform_supports is extended such that all three of signal+stats+image are detected. 
2) that ft_defaults is made consistent between the three
3) that ft_defaults is changed such that it detects that it is running at the Donders, and if so, that the compat toolboxes are used even if the Mathworks ones are available. When not at the Donders, the compat toolboxes are only used when the Mathworks one are not available.

ThreeFive-O:
In reply to 3: 
The network license problem might be of interest to other users too. Hence I'm not sure, if the single check for Donders is the right way to go.
What about a switch parameter for ft_defaults? With no argument the drop-ins are loaded. fieldtrip should behave the same; but with an override argument to prefer the MATLAB toolboxes (only if available).
Something like: `ft_defaults('preferMatlabToolboxes')`

robertoostenveld:
agreed. 

Coming up with a good way of bootstrapping the setting-up of ft_default (the global variable) and the path is tricky, since we don't have that yet. Let's consider how it would scale up if there were more (or more detailled) options:

your suggestion would result in
ft_defaults('stringoption', 'anotherstringoption', ...)

you can also image key-value pairs
ft_defaults('option1', 'value1', ...)

or perhaps FT-style
cfg = [];
cfg.toolbox.stats = 'matlab';
cfg.toolbox.images = 'compat';
cfg.toolbox.signal = []; % meaning default
ft_defaults(cfg)

The latter version might also take away part of the mystery of the ft_default global variable. E.g. you could do

cfg.debug = 'saveonerror'
ft_defaults(cfg)

which causes the global ft_default.debug to be assigned the value 'saveonerror'. The global ft_default structure is always "mixed in" with the cfg structure that every FT function gets through the "ft_preamble init" statement, which calls utilities/private/ft_preamble_init.m.

ThreeFive-O:
I've just added an approach which uses the `global ft_default` structure.
- Defaults to using the drop-ins when the struct was never explicitly changed.
- Overrides when the following is done prior calling `ft_defaults`:
```MATLAB
global ft_default
ft_default.useMatlabToolboxes = 'yes';
ft_defaults;
```
Let me know what you think and if the solution fixes 1. and 2. of your comment accordingly.

robertoostenveld:
then I suggest to do it like this

```
global ft_default
ft_default.toolbox.stats = 'matlab' / 'compat'
ft_default.toolbox.images = 'matlab' / 'compat'
ft_default.toolbox.signal = 'matlab' / 'compat'
ft_defaults
```

to make it clear that there are three choices to be made. Also, in cfg options we tend not to use upper case characters. The option of passing the cfg into ft_defaults as alternative to the global variable is independent of this.

ThreeFive-O:
Makes perfect sense to distinguish between all three cases. Great feedback!
I used uppercase for ft_has_toolbox because within the function the different cases are uppercase too. Furthermore it shows a little better that the argument is 'image' and not 'images' like everywhere else.
Shall I make the small change in ft_has_toolbox in the same PR to be consistent throughout?

I also like the idea of bootstrapping ft_default using the fieldtrip convention of using a cfg struct. But I suggest that should be another PR.

robertoostenveld:
yes, please go ahead and let me know when to review and merge. 

upper/lower case is indeed not 100% consistent due to different people contributing. I always use lower case myself. 

I agree about the other PR for "ft_defaults(cfg)".

ThreeFive-O:
Please review and merge if everything is alright.
As with the last commit, the mentioned tasks should have been solved. I tried several variants and it works for me. Unfortunately I don't have any setup available without the MATLAB toolboxes, but the logic should be ok.

robertoostenveld:
looks good to me. Thanks!

PR #415, opened 2017-04-26
ENH: rereferencing with and without montage using ft_preprocessing

StolkArjen:
- ft_preprocessing creates a unit matrix on the fly in case cfg.reref = 'yes' (and in case there's a sens)
- ft_apply_montage keepunused = 'no' still had too many elecpos. the function is pretty oldskool and the most reliable way to ensure that keepunused = 'no' didn't still attach unused stuff was by using a call to ft_selectdata on the input right in the beginning (rather than sifting out fields at the end)
- test_preprocmontage test the different rereferencing options on data with and without a sens

unclear: should ft_preprocessing, when it creates the unit matrix (in case cfg.reref) , also take into account cfg.implicitref if existent? physically, there's no elecpos to assign to it, but there may be a chanpos hypothetically (although no idea where to put it)

### DIFF
--- a/forward/ft_apply_montage.m
+++ b/forward/ft_apply_montage.m
@@ -95,11 +95,11 @@ end
 haschantype = (isfield(input, 'chantype') || isfield(input, 'chantypenew')) && all(isfield(montage, {'chantypeold', 'chantypenew'}));
 haschanunit = (isfield(input, 'chanunit') || isfield(input, 'chanunitnew')) && all(isfield(montage, {'chanunitold', 'chanunitnew'}));
 
-if ~istrue(keepunused)
+if ~istrue(keepunused) && strcmp(ft_datatype(input), 'elec')
   % remove the channels that are not rereferenced from the input
   cfg = [];
   cfg.channel = montage.labelold;
-  input = ft_selectdata(cfg, input);
+  input = ft_selectdata(cfg, input); % FIXME a low-level function (like ft_apply_montage) should never be calling a high-level function
 end
 
 % make sure they always exist to facilitate the remainder of the code
--- a/forward/ft_apply_montage.m
+++ b/forward/ft_apply_montage.m
@@ -99,7 +99,7 @@ if ~istrue(keepunused) && strcmp(ft_datatype(input), 'elec')
   % remove the channels that are not rereferenced from the input
   cfg = [];
   cfg.channel = montage.labelold;
-  input = ft_selectdata(cfg, input); % FIXME a low-level function (like ft_apply_montage) should never be calling a high-level function
+  input = ft_selectdata(cfg, input); % FIXME a low-level function should not be calling a high-level function like ft_selectdata
 end
 
 % make sure they always exist to facilitate the remainder of the code
--- a/forward/ft_apply_montage.m
+++ b/forward/ft_apply_montage.m
@@ -95,6 +95,13 @@ end
 haschantype = (isfield(input, 'chantype') || isfield(input, 'chantypenew')) && all(isfield(montage, {'chantypeold', 'chantypenew'}));
 haschanunit = (isfield(input, 'chanunit') || isfield(input, 'chanunitnew')) && all(isfield(montage, {'chanunitold', 'chanunitnew'}));
 
+if ~istrue(keepunused)
+  % remove the channels that are not rereferenced from the input
+  cfg = [];
+  cfg.channel = montage.labelold;
+  input = ft_selectdata(cfg, input);
+end
+
 % make sure they always exist to facilitate the remainder of the code
 if ~isfield(montage, 'chantypeold')
   montage.chantypeold = repmat({'unknown'}, size(montage.labelold));
@@ -275,7 +282,7 @@ if istrue(keepunused)
   montage.chanunitold = cat(1, montage.chanunitold(:), addchanunit(:));
   montage.chanunitnew = cat(1, montage.chanunitnew(:), addchanunit(:));
 else
-  % add the channels that are not rereferenced to the input of the montage only
+  % add the channels that are not rereferenced to the input of the montage
   montage.tra(:,(n+(1:k))) = zeros(m,k);
   montage.labelold    = cat(1, montage.labelold(:), addlabel(:));
   montage.chantypeold = cat(1, montage.chantypeold(:), addchantype(:));
--- a/ft_preprocessing.m
+++ b/ft_preprocessing.m
@@ -367,7 +367,7 @@ if hasdata
     
   if strcmp(cfg.updatesens, 'yes')
     if isfield(dataout, 'elec') && strcmp(cfg.reref, 'yes') && ~isstruct(cfg.montage) % create EEG montage on-the-fly in case cfg.reref
-      cfg.montage = ft_prepare_montage(keepfields(cfg, 'reref', 'refchannel', 'implicitref'), data);
+      cfg.montage = ft_prepare_montage(keepfields(cfg, {'reref', 'refchannel', 'implicitref'}), dataout);
     end  
     if isstruct(cfg.montage)
       % apply the linear projection also to the sensor description
--- a/ft_preprocessing.m
+++ b/ft_preprocessing.m
@@ -365,13 +365,9 @@ if hasdata
     
   end % for all trials
     
-  if strcmp(cfg.updatesens, 'yes') && (isfield(dataout, 'grad') || isfield(dataout, 'elec') || isfield(dataout, 'opto'))
-    if strcmp(cfg.reref, 'yes') && ~isstruct(cfg.montage)
-      fprintf('creating an identity matrix montage');
-      cfg.montage          = [];
-      cfg.montage.labelold = data.label;
-      cfg.montage.labelnew = data.label;
-      cfg.montage.tra      = eye(numel(data.label)); % do not adjust chanpos
+  if strcmp(cfg.updatesens, 'yes')
+    if isfield(dataout, 'elec') && strcmp(cfg.reref, 'yes') && ~isstruct(cfg.montage) % create EEG montage on-the-fly in case cfg.reref
+      cfg.montage = ft_prepare_montage(keepfields(cfg, 'reref', 'refchannel', 'implicitref'), data);
     end  
     if isstruct(cfg.montage)
       % apply the linear projection also to the sensor description
--- a/ft_preprocessing.m
+++ b/ft_preprocessing.m
@@ -364,25 +364,34 @@ if hasdata
     [dataout.trial{i}, dataout.label, dataout.time{i}, cfg] = preproc(data.trial{i}, data.label,  data.time{i}, cfg, begpadding, endpadding);
     
   end % for all trials
-  
-  if isstruct(cfg.montage) && strcmp(cfg.updatesens, 'yes')
-    % apply the linear projection also to the sensor description
-    if issubfield(cfg.montage, 'type')
-      bname = cfg.montage.type;
-    else
-      bname = 'preproc';
-    end
-    if isfield(dataout, 'grad')
-      fprintf('applying the montage to the grad structure\n');
-      dataout.grad = ft_apply_montage(dataout.grad, cfg.montage, 'feedback', 'none', 'keepunused', 'yes', 'balancename', bname);
-    end
-    if isfield(dataout, 'elec')
-      fprintf('applying the montage to the grad structure\n');
-      dataout.elec = ft_apply_montage(dataout.elec, cfg.montage, 'feedback', 'none', 'keepunused', 'yes', 'balancename', bname);
-    end
-    if isfield(dataout, 'opto')
-      fprintf('applying the montage to the opto structure\n');
-      dataout.opto = ft_apply_montage(dataout.opto, cfg.montage, 'feedback', 'none', 'keepunused', 'yes', 'balancename', bname);
+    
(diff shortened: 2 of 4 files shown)


StolkArjen:
Just to give a feeling of how things pan out: in case the input is MEG data with a grad structure in it and the user  specifies reref = 'yes' (with refchannel 'all' for instance), a unit matrix montage is created on the fly and applied to the grad structure. This is because updatesens = 'yes' by default, reref = 'yes', and' a montage is made in case it doesn't exist. This does not affect the labels and chanpos in the grad structure (because all channels are used in the montage), and it is a obsolete process in this MEG scenario. 

However, in case the user specifies cfg.channel and cfg.reref = 'yes', the rereferencing is done all channels defined in cfg.channel, and automatically a montage is created for the channels in the data (preselected with cfg.channel). Because keepunused = 'no' by default, the channels that weren't recruited for referencing are removed from the sens structure, preventing issues down the line.

So, I'm in two minds in that the depicted MEG scenario is ugly since it calls apply_montage without any good reason (and effect). One is likely never wanting to remove subsets of gradiometers from the data. On the other hand, in the other (iEEG) scenario it is convenient to have a montage being generated on the fly so that unused channels are removed from the elec structure. One could easily prevent the MEG scenario by setting the default of updatesens to 'no', likely saving computational time. What do you think, @robertoostenveld? Note this wasn't issue before when you set the dependency to if isstruct(montage) && updatesens = 'yes'. By having reref = 'yes' create a montage, this dependency got trimmed to if updatesens = 'yes'. Do you see a more elegant solution?

robertoostenveld:
ik ga hier vanavond even naar kijken

StolkArjen:
thanks

2017-05-03 8:56 GMT-07:00 Robert Oostenveld <notifications@github.com>:

robertoostenveld:
test_bug2570.txt:=== test_bug2570 FAILED in 3 seconds
test_pull393.txt:=== test_pull393 FAILED in 2 seconds

forward/ft_apply_montage.m
```diff
@@ -95,6 +95,13 @@
 haschantype = (isfield(input, 'chantype') || isfield(input, 'chantypenew')) && all(isfield(montage, {'chantypeold', 'chantypenew'}));
 haschanunit = (isfield(input, 'chanunit') || isfield(input, 'chanunitnew')) && all(isfield(montage, {'chanunitold', 'chanunitnew'}));
 
+if ~istrue(keepunused)
+  % remove the channels that are not rereferenced from the input
+  cfg = [];
+  cfg.channel = montage.labelold;
+  input = ft_selectdata(cfg, input);
```
robertoostenveld: this will fail on many of the input structures that ft_apply_montage accepts. Not all of those input structures are correctly dealt with in ft_selectdata.

ft_preprocessing.m
```diff
@@ -364,25 +364,34 @@
...
-      fprintf('applying the montage to the grad structure\n');
-      dataout.elec = ft_apply_montage(dataout.elec, cfg.montage, 'feedback', 'none', 'keepunused', 'yes', 'balancename', bname);
-    end
-    if isfield(dataout, 'opto')
-      fprintf('applying the montage to the opto structure\n');
-      dataout.opto = ft_apply_montage(dataout.opto, cfg.montage, 'feedback', 'none', 'keepunused', 'yes', 'balancename', bname);
+    
+  if strcmp(cfg.updatesens, 'yes') && (isfield(dataout, 'grad') || isfield(dataout, 'elec') || isfield(dataout, 'opto'))
+    if strcmp(cfg.reref, 'yes') && ~isstruct(cfg.montage)
+      fprintf('creating an identity matrix montage');
```
robertoostenveld: why would this be the montage that is to be applied. The output cfg will now be incorrect and not describe the operations that are done on the output data. 

Please note that 

cfg = ...
data2 = ft_preprocessing(cfg, data1)
data3 = ft_preprocessing(data2.cfg, data1)
assert(data2==data3)

should hold.

robertoostenveld:
this is not a good implementation. Please redesign and make a better test (not only the stuff you want to work, but all possible cases). The code now shows a lot of warnings which indicate that something is wrong. 

as far as I can tell, the low-level ft_apply_montage is already doing what it should with the following test code:

```
%%

montage1 = [];
montage1.labelnew = {'a', 'b', 'c'};
montage1.labelold = {'1', '2', '3'};
montage1.tra = eye(3);

montage2 = [];
montage2.labelnew = {'1', '2', '3'};
montage2.labelold = {'a', 'b', 'c'};
montage2.tra = eye(3);

% 2 is applied to 1
montage1_2 = ft_apply_montage(montage1, montage2);
assert(isequal(montage1_2.labelold, {'1', '2', '3'}));
assert(isequal(montage1_2.labelnew, {'1', '2', '3'}));

% 1 is applied to 2
montage2_1 = ft_apply_montage(montage2, montage1);
assert(isequal(montage2_1.labelold, {'a', 'b', 'c'}));
assert(isequal(montage2_1.labelnew, {'a', 'b', 'c'}));

%%
montage1 = [];
montage1.labelnew = {'a', 'b', 'c'};
montage1.labelold = {'1', '2', '3'};
montage1.tra = eye(3);

elec1 = [];
elec1.label = {'1', '2', '3', '4'};
elec1.elecpos = randn(4,3);
elec1.chanpos = elec1.elecpos;

elec1_1a = ft_apply_montage(elec1, montage1, 'keepunused', false);
elec1_1b = ft_apply_montage(elec1, montage1, 'keepunused', true);

assert(isequal(elec1_1a.label(:), {'a', 'b', 'c'}'));
assert(isequal(elec1_1b.label(:), {'a', 'b', 'c', '4'}'));

%%
montage1 = [];
montage1.labelnew = {'a', 'b', 'c'};
montage1.labelold = {'1', '2', '3'};
montage1.tra = eye(3);

data1 = [];
data1.label = {'1', '2', '3', '4'};
data1.time{1} = 1:10;
data1.trial{1} = randn(4,10);

data1_1a = ft_apply_montage(data1, montage1, 'keepunused', false);
data1_1b = ft_apply_montage(data1, montage1, 'keepunused', true);

assert(isequal(data1_1a.label(:), {'a', 'b', 'c'}'));
assert(isequal(data1_1b.label(:), {'a', 'b', 'c', '4'}'));
```

I think it would be better to write a private reref2montage function that converts cfg.reref/implicitref/refchannel into a montage. That montage can then be applied, both inside private/preproc like this

if istrue(cfg.reref)
  % convert into montage, don't put in cfg
  ft_apply_montage
else not isempty(cfg.montage)
  ft_apply_montage
else 
 %  keep the data as it is
end

and the same code should be applied in ft_preprocessing on the grad/elec/opto structure. Except that reref=yes only applies to elec, and cfg.montage applies to all three.

StolkArjen:
Starting to lose overview. :) Maybe we can sykpe about this tomorrow (5pm
NL time)?

2017-05-03 14:01 GMT-07:00 Robert Oostenveld <notifications@github.com>:

robertoostenveld:
oh, and note that a low-level function (like ft_apply_montage) should never be calling a high-level function (one that takes a cfg as first input, like ft_selectdata). 

Furthermore, in case you want to call a high-level function from another high-level function, you should use rollback_provenance.

robertoostenveld:
I think I have quite a clear overview. 

Please start (in the test script and in the description of the enhancement) by specifying the current and desired behavior, rather than by describing what (you think) is the  solution.

- Current behavior is that data can be re-referenced with either cfg.reref or with cfg.montage. Conceptually they are identical. However, the details of the output are currently different, specifically in the elec structure. 

- Desired behavior is that the output is the same, regardless whether cfg.reref was used or cfg.montage. The implementation using cfg.montage is correct (I think), which means that only the implementation that involves cfg.reref should change.

robertoostenveld:
Oh, and note that test_tutorial_natmeg_timefrequency also started failing. There are still some test scripts running, so there might be more failures coming.

forward/ft_apply_montage.m
StolkArjen: Should we make it ft_datatype = sens dependent, since it applies only to the elec structure?

StolkArjen:
"I think it would be better to write a private reref2montage function that converts cfg.reref/implicitref/refchannel into a montage."
> what would the implicitref look like montage-wise: a row with zeros appended tot the bottom? and what position to assign it to?

robertoostenveld:
indeed a row with zeros. The electrode (and channel) position would be [nan nan nan], since we don't know it. If you do ft_read_sens, the reference might be included in the polhemus (or whatever) recording of electrode positions. In that case elecpos is not nan.

I suggest to call the function ft_prepare_montage. It can work similar to ft_prepare_layout, i.e. 

```
tmpcfg = [];
tmpcfg.reref = 'yes'; % the function should have this as default, and error when set to no
tmpcfg.implicitref = string or cell-array.
tmpcfg.refchannel = string or cell-array, needs to be matched with ft_channelselect on the data.label+implicitref 

cfg.montage = ft_prepare_montage(tmp cfg, data)
```

robertoostenveld:
oh, and note that you should not worry about elecpos and chanpos. You should only make the montage. That is for ft_apply_montage to solve: it will (should) add nans already.

StolkArjen:
So is the philosophy behind ft_prepare_montage that it is used before calling ft_preprocessing. Or is it called on-the-fly in ft_preprocessing when cfg.reref = 'yes'?

StolkArjen:
Do we update the tra matrix with cfg.refchannel? If so, how to handle more than one refchannel? That is, a single refchannel could be implemented as:

montage.tra      = zeros(numel(data.label));
montage.tra(:,match_str(data.label, cfg.refchannel)) = -1; % fill the refchannel column(s) with -1
montage.tra(eye(numel(data.label))==1) = 1; % fill the diagonal with 1

robertoostenveld:
Backward conpatibility requires that the montage is constructed inside ft_preprocessing. It can also be called directly by power users.

Both is similar to the layout generation.

robertoostenveld:
You *make* the montage.tra with refchannel, not update it. Or you could say that you start with identity and then subtract a fraction from each refchannel column. But of course not the sum of refchannel as you do below.

Note that cfg.refchannel is not a list of indices, but could be 'all' or 'EEG'.  It needs to be matched with data.label.

StolkArjen:
"Or you could say that you start with identity and then subtract a fraction from each refchannel column"

> Not sure what you mean with 'fraction'. Since this montage is used for just the elec (there's a preproc_reref for the signals) and assuming we're not adjusting channel positions (given that we use an identity matrix for refchannel 'all'), should the tra become as follows in the case of chan1 .. chan4 where say cfg.refchannel is chan4?

            chan1 chan2 chan3 chan4
chan1      1         0         0       0
chan2     0         1         0       0
chan3     0         0         1       0

But by the same token, refchannel 'all' (which is updated to include chan1 .. chan4 in ft_preproc) would leave us with no rows in the tra matrix.

Pushing some work-in-progress files.

schoffelen:
spuit 11: does this resolve http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1482 ?

StolkArjen:
Aye cap. If we ever get it implemented, it would. :)

robertoostenveld:
test_pull393 and test_bug2570 both seem fine with my branch. The rest still should be tested.

StolkArjen:
for tracking purposes:
git remote add roboos https://github.com/robertoostenveld/fieldtrip.git
git fetch roboos
git checkout --track roboos/bug1482-reref

StolkArjen:
Looks like you nailed it, @robertoostenveld. Grazie mille. 

A major difference between your and my implementation is due to different expectations about how what should be kept in the output. I tried to apply cfg.channel to the input, essentially restricting what would end up in elecpos and chanposold. This presumably caused more friction than anticipated. As per test_bug1482, you scan scratch lines 126 and 139 or else change 'm into
assert(**~**isequal(reref_eeg_reref.elec.chanpos, reref_eeg_reref.elec.elecpos))
and
assert(isequal(numel(reref_eeg_implref.elec.label),**3**))

A consequence of the current implementation is that after appending multiple sens structures originating from the same dataset, there will be substantial duplicate information in elecpos, labelold, and chanposold. I'll fix this in ft_appendsens itself.

Lastly, how did you manage to update the tra after cfg.reref = 'yes' and cfg.refchannel = 'xxx', without actually adjusting (screwing up) the chanpos? Yet, when manually specifying a cfg.montage with -1 and 1 weights, the chanpos are actually adjusted as desired.

robertoostenveld:
I only made the problem smaller; i.e. changed it into "construct a montage", knowing that ft_apply_montage was recently carefully checked. I will make the changes to the test script, and run the full FT dashboard batch against the branch.

robertoostenveld:
er zijn wel een paar die failen. 

```
test_bug3285.txt:=== test_bug3285 FAILED in 10 seconds
test_tutorial_eeg_preprocessing.txt:=== test_tutorial_eeg_preprocessing FAILED in 10 seconds
test_tutorial_natmeg_preprocessing.txt:=== test_tutorial_natmeg_preprocessing FAILED in 293 seconds
test_tutorial_preprocessing_erp.txt:=== test_tutorial_preprocessing_erp FAILED in 10 seconds
test_tutorial_tmseeg.txt:=== test_tutorial_tmseeg FAILED in 195 seconds
```

met cfg.updatesens='no' als default verwacht ik dat het beter gaat. Dit lijkt me een veiligere optie, ook voor scripts van end-users.

robertoostenveld:
this is superseded by PR #427

PR #417, opened 2017-04-27
FIX - Calculation of endsample

dmatthes1982:
The endsample of one trial is definitely not equal to the next begsample, since the data in the eeg file is segmented and the two data points (endsample and begsample) are with highest probability in many cases not consecutive data points in the raw data.


robertoostenveld:
thanks!

PR #419, opened 2017-04-28
fixed colorbar bug with 2D cloud slice method

smg211:


robertoostenveld:
this causes test_ft_plot_cloud to fail

```
=== test_ft_plot_cloud FAILED in 18 seconds
Error using ft_plot_cloud (line 692)
In an assignment  A(:) = B, the number of elements in A and B must be the same.
Error in test_ft_plot_cloud (line 31)
```

robertoostenveld:
and it causes test_bug3275 to fail.

```
=== test_bug3275 FAILED in 28 seconds
Error using ft_sourceplot (line 1397)
Undefined function or variable 'cloudsurf'.
Error in test_bug3275 (line 81)
```

PR #423, opened 2017-05-02
bugfixes to PR #419

smg211:
Because ft_sourceplot was updated since I started the original depth_plotting branch, I pulled the most recent master, created this new branch, and added my changes there so that there would be no conflicts.  Apologies for the convoluted solution.

The bugs from PR #419 should be resolved now though.

### DIFF
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -327,6 +327,33 @@ elseif ~hasanatomical && cfg.downsample~=1
   tmpcfg.parameter = {cfg.funparameter, cfg.maskparameter, cfg.anaparameter};
   functional = ft_volumedownsample(tmpcfg, functional);
   [cfg, functional] = rollback_provenance(cfg, functional);
+elseif hasanatomical && strcmp(cfg.method, 'cloud') % need to disentangle the volumetric from the surface inputs
+  cloudsurf = {};
+  cloudvol = {};
+  
+  % check that anatomical is a cell array
+  if ~iscell(anatomical)
+    if numel(anatomical) == 1
+      tmp = anatomical;
+      anatomical = cell(1);
+      anatomical{1} = tmp;
+    else
+      error('If anatomical includes multiple meshes and/or volumes as input, they must be organized in a cell array')
+    end
+  end
+  
+  for n = 1:numel(anatomical)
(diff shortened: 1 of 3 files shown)


ft_sourceplot.m
```diff
@@ -327,6 +327,33 @@ function ft_sourceplot(cfg, functional, anatomical)
   tmpcfg.parameter = {cfg.funparameter, cfg.maskparameter, cfg.anaparameter};
   functional = ft_volumedownsample(tmpcfg, functional);
   [cfg, functional] = rollback_provenance(cfg, functional);
+elseif hasanatomical && strcmp(cfg.method, 'cloud') % need to disentangle the volumetric from the surface inputs
+  cloudsurf = {};
+  cloudvol = {};
+  
+  % check that anatomical is a cell array
+  if ~iscell(anatomical)
```
robertoostenveld: the field "anatomical" is a single anatomical MRI or a single structure with mesh information, such as a cortical sheet. It is not (and cannot be) a cell array: that is not how FT works.

plotting/ft_plot_cloud.m
```diff
@@ -94,14 +100,11 @@ function ft_plot_cloud(pos, val, varargin)
 radius             = ft_getopt(varargin, 'radius', 4 * ft_scalingfactor('mm', unit));
 rmin               = ft_getopt(varargin, 'rmin',   1 * ft_scalingfactor('mm', unit));
 scalerad           = ft_getopt(varargin, 'scalerad', 'yes');
+cmap               = ft_getopt(varargin, 'colormap', 'parula');
```
robertoostenveld: why do you think I added the following section 

-if ft_platform_supports('parula')
 -  cmap             = ft_getopt(varargin, 'colormap', 'parula');
 -else
 -  cmap             = ft_getopt(varargin, 'colormap', 'jet');
 -end

removing it ensures that the code (once more) breaks on MATLAB versions prior to 2014.

robertoostenveld:
I suggest you start by making a bug report. Describe what is currently either incorrect, or suboptimal. Describe how the ideal outcome would look like. Following that description (which is purely textual), you can describe the design (architecture) of the changes to the code and to the interface of the functions, and only then you can make the implementation.

PR #427, opened 2017-05-08
Bug 1482 and PR 415 - apply reref options to elec

robertoostenveld:
it now can do the same for cfg.reref options as for cfg.montage. 
the default in ft_preprocessing is cfg.updatesens=no, i.e. the same as before.


robertoostenveld:
most tests ran through, only the lengthy ones are still running. So far no errors, so I'll merge.

PR #432, opened 2017-05-10
enh: chan2source uses posunit if available, avoiding risk of misestimation

StolkArjen:


robertoostenveld:
I am closing this for now, please see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3293#c16.

If it is clear what is solved, it can be reopened.

robertoostenveld:
Ok, I looked into this a bit more. This specific change simply addresses that the geometric unit should be retained in ft_checkdata.

PR #434, opened 2017-05-11
deal with overlapping channels and electrodes in ft_appendsens

StolkArjen:
Trumped down the functionality of appendsens. It's now simply concatenating and removing duplicates, while still checking the consistency between the relevant fields.

Are channels appended and are overlapping channels in multiple input dealt with?
a) yes
b) no
c) I prefer to give fuzzy answers
> A but also C of course

Are electrodes appended and are overlapping electrodes in multiple input dealt with?
a) yes
b) no
c) I prefer to give fuzzy answers
> idem as above


StolkArjen:
Cool, will scrutinize the test script

PR #435, opened 2017-05-13
improvement of ft_appendsens

StolkArjen:
a couple of small fixes:

- removed old code that was somehow still present
- ensured label and labelold have a column orientation
- made computation of the tra matrix work for many more situations (as in test_pull393 and test_pull434)

open tra-related issue:

- ft_appendsens still cannot deal well with the situation in which two datasets have identical or overlapping elecpos but different chanpos, e.g.

elec1.label = {'1'; '2'; '3'};
elec1.elecpos = [1 1 1; 2 2 2; 3 3 3];
elec1.chanpos = [1 1 1; 2 2 2; 3 3 3];
elec1.tra = eye(3);

elec1b = elec1;
elec1b.label = {'1b', '2b', '3b'};
elec1b.chanpos = elec1.chanpos+10;

append11b = ft_appendsens([], elec1, elec1b);
append11b.tra

     1     0     0
     0     1     0
     0     0     1
     0     0     0
     0     0     0
     0     0     0

whereas this should produce two identity matrices on top of each other. Any suggestions per the implementation of the tra matrix computation/pruning are welcome (see FIXME in the code).

### DIFF
--- a/ft_appendsens.m
+++ b/ft_appendsens.m
@@ -99,7 +99,6 @@ haschanori = 0;
 hasoptopos = 0;
 haslabelold = 0;
 haschanposold = 0;
-tramatch = 0;
 for i=1:length(varargin)
   % the following fields should be present in any sens structure
   if isfield(varargin{i}, 'label')
--- a/ft_appendsens.m
+++ b/ft_appendsens.m
@@ -6,9 +6,9 @@ function [sens] = ft_appendsens(cfg, varargin)
 % Use as
 %   combined = ft_appendsens(cfg, sens1, sens2, ...)
 %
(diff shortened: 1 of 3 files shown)


StolkArjen:
It was a pita.. but appendsens now produces (correct) tra matrices for all tests cases in test_pull393.m, test_pull434.m, as well as for the rereferenced iEEG data. For instance,

append11b.tra

     1     0     0
     0     1     0
     0     0     1
     1     0     0
     0     1     0
     0     0     1

Assuming this is the desired behavior for a situation with multiple chanpos based on the same set of elecpos, appendsens is ready to be pulled.

PR #436, opened 2017-05-15
implement ft2audio, i.e. application that streams data from FT buffer to audio

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3281


robertoostenveld:
Since there are some changes to the buffer source code (interface.[ch]), I tested recompilation of all realtime executables. That seemed to work fine.

PR #437, opened 2017-05-16
Bug2706 - merge the crosshair plotting

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2706


test/test_pull433.m
```diff
@@ -1,10 +1,11 @@
-function test_ft_electroderealign_headshape
```
robertoostenveld: hmm, this commit should not have been here in this PR. But I think that the commit is fine in itself, so I'll leave it in.

PR #438, opened 2017-05-16
Conversion of EGI data to uV

thomaskustermann:
As per the EGI documentation on file formats ([rawformat.pdf](https://github.com/fieldtrip/fieldtrip/files/1004869/rawformat.pdf)), Net Station simple binary files' header contains information about whether the data is in microvolts or A/D units.

> The header of a simple binary file has two fields that need to be consulted to determine
whether the EEG data stored in the file are in microvolts or A/D units. These fields are the ‘bits’
field at offset 26 and the ‘range’ field at offset 28. If the values in the ‘bits’ and ‘range’ fields are
both 0, then the file’s data is already in the form of microvolts. Users who save to simple
binary format should note that this is the only method to determine if a file’s data are in A/D
units or microvolts.

These values are currently not evaluated to determine the unit of data or to convert data to uV as far as I can tell. EGI proposes using the following formula for unit conversion:
>Converting the A/D units to microvolts can be done programmatically by applying the following formula:
microvolt value = (range / 2^bits) x A/D value,
where ‘bits’ and ‘range’ are read from the header of the file.

This approximates the value that would be obtained using gains and zeros.
In addition, even when data are stored in units of A/D, fieldtrip will return 'uV' as the unit-type for all channels from a later call to ft_chanunit.

In some cases it is possible to avoid this unit issue by converting the data to microvolt using EGI's NetStation program before reading the data in fieldtrip. Problems arise when recording a longer session where a conversion to microvolt (resulting in float instead of int) would raise the file size above 2GB. At least on older EGI systems (i.e. NetStation 4.3) exported file size is simply capped at 2GB and becomes unusable, making unit conversion to uV with single- or double-precision impossible in NetStation.

In any case, I suspect many people use the EGI-provided simple-to-use NetStation File Exporter Apple which will only export to A/D and is afaik the only way to obtain the impedance values of a recording.

The lines of code I added implement the above formula to automatically convert data to uV when fields for 'bits' and 'range' both are not 0.
A more accurate calculation would be possible using the .GAIN and . ZERO file provided by exporting EGI-data using the NetStation File Exporter Applet, but is not implemented (yet).

egi_format_comparison
This is a graph showing an examplary .2 sec snippet of one channel using different methods of export in NetStation, unit conversion with the above formula and using .GAIN and .ZERO files.
Upper row depicts uV-converted data (RAW is from the EGI File Exporter), lower row depicts A/D values (as read by fieldtrip).
Red and blue lines are numerically identical, I added a small intercept for visibility.


robertoostenveld:
Hi Thomas,

Thanks for the suggestion. Let me CC Joe (by email); he contributed the EGI reading functions and who has the best understanding of them. @Joe, do you have a github account? If not, would you mind making one?

I realize that I do not have any EGI sbin test file. @Thomas: could you send me one (or a few) by private email? 

As I understand it, the suggested code change changes the values from raw AD values (i.e. integers?) to uV. Overall I think that makes sense. But that would be bad for people that are in the middle of the analysis, especially since they cannot see what the units are.  For most EEG formats FieldTrip defaults to uV. A few years back (which is after most of the low-level reading functions were implemented) we started with more explicit unit handling, among others implemented by extending the header with hdr.chantype and hdr.chanunit. That defaults to repeat({‘unknown’}, size(hdr.label)), but can be used to specify the units when they can be determined from the file format.

The long-term goal is to have units explicitly represented everywhere in the analysis, and to have units match. E.g. a source reconstruction of uV potentials on a mm-sized headmodel should  result in correct nAm dipole moments.

I suggest that we first add the units to the header, i.e. ensure that hdr.chanunit is not unknown. The ft_read_data function already has the ‘chanunit’ option which allows more low-level users to specify whether they want data in V or uV (or T versus fT, or T/m versus fT/cm, etc). The conversion happens on the fly on line 1417 in ft_read_data. Note that itthe specific conversion would not work for AD to uV, but would work for V or uV. Please review that code. 

I also think that the conversion of the units is better done in ft_read_data than in read_sbin_data, because ft_read_data has a more explicit description of unit details (which comes from ft_read_header).

So in short I suggest (but Joe might overrule me on this): please change the PR to add explicit hdr.chantype and hdr.chanunit in ft_read_header, please move the unit conversion from read_egi_sbin to ft_read_data (consistent with the units as specified in ft_read_header), please document in the help of read_egi_sbin what its output units are.

best
Robert

robertoostenveld:
Thomas wrote by email:

-------
thanks for your swift reply on github. I'm sending you some sample EGI data, that is:
a) Output from using the EGI File Exporter Applet (integers, A/D values)
b) Output from NetStation (integers, A/D values)
c) Output from NetStation (float, uV values)

In case you are interested in converting to the 'exact' values using the .GAIN and .ZERO file you could use the following code:
```
fg = fopen(gainfile);
        gains_val = textscan(fg,'%f %*s %f ','HeaderLines', 3 );
        fclose(fg);
        fz = fopen(zerofile);
        zeros_val = textscan(fz,'%f %*s %f ','HeaderLines', 3 );
        fclose(fg);
        for i = 1:numel(chanindx(ismember(chanindx,zeros_val{1})))
            trialData(i,:) = (trialData(i,:)-zeros_val{2}(chanindx(i)))*400/gains_val{2}(chanindx(i)); %Scaling Factor (acc. to manual): uV = (A/D Unit - channel zero) * (calibration signal amplitude) / channel gain
        end;
```

Since I haven't heard anything to the contrary I will try and implement your suggestions and then get back to you.

-------
Thanks, I received the data. 

On our local system I unpacked it to /home/common/matlab/fieldtrip/data/test/pull438 with inside the RAW_egi_exporter, RAW_fl_uV and RAW_int_ad directories. If you make a test script (look in fieldtrip/test for examples), please consider that I will be adding a "cd /home/common/matlab/fieldtrip/data/test/pull438" to that script and be running it from there.

jdien07:
I just happened across this thread.  I never got the message Robert mentions cc'ing me back in May.  For that matter, I never got a message about FieldTrip shifting to github.  Is it possible e-mail messages are being sent to the wrong e-mail address?  Anyway, I do now have a github account, which I just used to submit a BrainVision I/O modification.  I don't have any comments on this particular issue but would be happy to test out the proposed changes.  Is there anything I can do to help get this wrapped up?

thomaskustermann:
The hold up is all on me, but I should be able to get going on it again in
the next few weeks as I finally submit my PhD thesis.

schoffelen:
Hi @robertoostenveld and @jdien07 I am reviewing some long-standing issues/PRs. This one has been open for the past years. If I read the comments, I have the impression that a lot of thought has already gone into this, but it has never been finalized. Given that this has been off the radar for so long kind of suggests that there's no acute need to really follow this up, but on the other hand it could be a waste to just close this, if with some relatively small changes the code can be improved. What do you think?

PR #439, opened 2017-05-16
bugfix: verify label is a column vector

TEMadsen:
In case of one unique channel name per file, label became a row vector instead and threw an error when concatenation was attempted at line 681 (now 682).

This was such a small bug with such an easy fix, I didn't bother to make a bug report, but let me know if you'd prefer I do.  Also, let me know if it would be easier in the future for me to just suggest tiny changes like this, rather than creating a pull request.

### DIFF
--- a/utilities/ft_selectdata.m
+++ b/utilities/ft_selectdata.m
@@ -671,6 +671,7 @@ for k = 1:ndata
   end
   label      = union(label, selchannel);
 end
+label = label(:);   % ensure column array
 
 % this call to match_str ensures that that labels are always in the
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Hi Theresa,

A pull request is really easy to review and merge for me, and takes me less time than understanding and implementing the suggestion myself (although this oen was really simple). So please do send pull requests. If you wish, you can just keep on using the same branch for that.

It is not a problem that there is no corresponding report on bugzilla. I think that is especially useful when there is some discussion prior to coming up with a (proposed) solution. But discussing it here is also fine.

thanks
Robert

PR #441, opened 2017-05-19
Bug3295 - fix appenddata for data with non-identical but partially overlapping channels 

robertoostenveld:
This was reported by @thht 
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3295


robertoostenveld:
@thht and @StolkArjen do you want to review the suggested change that I made to the code?

thht:
on it.

thht:
my own tests went fine. and i did not see any obvious problems in the code. but of course, i do not know all the ins and outs of the fieldtrip code.

the bug i reported is fixed with this commit. so, from my point of view, it could be merged.

thanks for fixing, @robertoostenveld!

PR #444, opened 2017-05-29
Fix bug in UNIX with symlinked CTF dataset files

dtxe:
Modified CTF I/O functions. When reading CTF datasets, append file separator to the path provided to the MATLAB dir() command to indicate that symlinks should be dereferenced.

The dir() command is used to retrieve file size. Without dereferencing symlinks, the size of the link itself, as opposed to the actual meg4 file is returned, causing errors when loading data.


robertoostenveld:
I guess the suggested change will work fine on linux and OSX. It is hard to oversee whether it might break something for windows. Did you test it for windows?

dtxe:
That's a good point. I just tried the dir() command in windows to retrieve the size of a file with and without a trailing slash, and it seems to work the same way.

http://imgur.com/a/tQnhn

dtxe:
I added if-statements so that this code works on both non-symlinked and symlinked ds files across all platforms.

robertoostenveld:
thanks

PR #445, opened 2017-05-30
refined strategy for providing user feedback

robertoostenveld:
This follows from a discussion with @dmatthes1982 in #401 to refine the feedback that the user gets, and the control over that feedback.


robertoostenveld:
@schoffelen, would you care to review this? It should not affect any existing code yet, but I only want to merge this if it you consider it a good plan.

StolkArjen:
Are you in the US? Otherwise bed time! :P

robertoostenveld:
Blame Canada!

StolkArjen:
West coast? If so, coffee in Berkeley? (it's only a two-day drive)

robertoostenveld:
I am in Vancouver for the HCP workshop. Quite a drive!

StolkArjen:
The weather's worth it. ;)

robertoostenveld:
I have tested it on the dashboard, all seems fine. So I'll merge. 

@dmatthes1982 please check out the new ft_warning, ft_info, ft_notice functions (in utilities). There are still changes needed in many FT functions: replacing fprintf with ft_notice/ft_info/ft_debug. That  then allows you to turn them off one-by-one.

dmatthes1982:
@robertoostenveld thank you very munch for this contribution. In general this works very good for me. I changed some warning('off', all) to ft_warning off in my own scripts and the results meet my expectations.

Furthermore I replaced two fprintf lines in ft_freqgrandaverage with ft_info, but in this case the results doesn't meet my expectations completely. Now I got the following output

```
computing average powspctrm over 7 subjects
 In DP_testAlphaMu_preproc at line 121
 In ft_freqgrandaverage at line 185
```
instead of the message only. On my opinion ft_info and ft_notice should not add the lines of appearance to the output.

At last I have one remaining question, how do you distinguish the use of ft_info and ft_notice. In which case to you use what function?

robertoostenveld:
thanks for testing. The line number details are the backtrace, i.e. the "stack" details from which the print command is made. 

I suppose that the backtrace should be on by default for error/warning, but off for notice/info/debug. I will change that. 

The decision to use notice/info/debug should be done by the programmer that understands the specific code. He/she should weigh the importance of the information presented.

Since we have three levels, I propose to switch (search-replace) the current print commands all to the middle level (info) as a default. From that point on we can start thinking, move some up (notice) and some down (debug) in importance.

Although previously discussed, the idea that is not yet implemented is to relate this to the cfg.verbose option. That is something that can be done in a ft_preamble_init. E.g. if the user specifies

cfg.verbose = 'info'

then that should cause

ft_debug off
ft_info on
ft_notice on
ft_warning on

and e.g. if cfg.verbose=error, all 4 would be off, not even showing warnings.

robertoostenveld:
dhcp-206-12-55-226> git commit -a
[master d421a6c] ENH - set notification backtrace off by default for debug/info/notice, on for warning/error
 1 file changed, 7 insertions(+), 1 deletion(-)

robertoostenveld:
I just looked at ft_preamble_init. It is not matched with a postamble, i.e. reverting the warning/notice/info state at the end of the high-level FT function is not possible.

But there are ft_preamble_debug and ft_postamble_debug, which act as a pair. So these can be used to set the state for the limited time of the function, and restore at the end.

robertoostenveld:
[master 06d17ec] ENH - allow cfg.verbose to be used to change the warning/notify/info/debug level for a single high-level FT function. See https://github.com/fieldtrip/fieldtrip/pull/445#issuecomment-310660686
 4 files changed, 112 insertions(+), 64 deletions(-)

robertoostenveld:
@dmatthes1982 you can try it with ft_freqsimulation. It prints one info and two debug messages.

robertoostenveld:
One issue that I now realize is that the behavior of repeated printing of warnings is now different than before. The default is now "ft_warning on", which (conceptually) used to be "ft_warning once".

Should we change the default? Where in the code should that be set?

dmatthes1982:
@robertoostenveld I have tested your recent changes in another function, where I changed fprintf to ft_info (the related pull request follows later) and it works like expected.

PR #447, opened 2017-05-31
Bug 3267 - ft_redefinetrial incorrectly claims trialinfo cannot be combined

StolkArjen:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3267

### DIFF
--- a/ft_appenddata.m
+++ b/ft_appenddata.m
@@ -172,6 +172,12 @@ switch cfg.appenddim
       data.label;       % keep it as determined by append_common
     end
     
+    % check whether the trials are consecutive segments
+    if isfield(data, 'sampleinfo') && any(diff(data.sampleinfo(:,1))<0)
+      warning('removing sampleinfo because the appended trials are not consecutive segments')
+      data = rmfield(data, 'sampleinfo');
+    end
+    
   otherwise
     error('unsupported cfg.appenddim');
 end % switch
--- a/test/test_bug3267_thesequel.m
+++ b/test/test_bug3267_thesequel.m
@@ -1,0 +1,67 @@
+function test_bug3267_thesequel
+
+% MEM 1500mb
+% WALLTIME 00:10:00
+
+
+%% situation 1: without sampleinfos (works fine)
+data1a.label = {'1';'2';'3'};
+for trl = 1:5
+  data1a.trial{1,trl} = randn(3,3);
+  data1a.time{1,trl} = [1 2 3];
+end
+data1a.trialinfo(:,1) = 1:5;
+
+data1b = data1a;
+data1b.trialinfo(:,1) = 6:10;
+
+% append
+data1 = ft_appenddata([], data1a, data1b);
+
+% this works fine
+cfg             = [];
+cfg.length      = 1;
+cfg.overlap     = 0;
+redef1 = ft_redefinetrial(cfg, data1);
+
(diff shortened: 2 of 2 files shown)


robertoostenveld:
this commit causes test_bug3267 to fail on line 48 (since sample info is missing).

Reading the report on bugzilla, it seems to me that it is a feature rather than a bug. In the following, there is no reason for data12 not having sampleinfo.

```
data1 = [];
data1.label = {'1'};
data1.time{1} = 1:10;
data1.trial{1} = ones(1,10)*1;
data1.sampleinfo = [1 10];

data2 = [];
data2.label = {'1'};
data2.time{1} = 1:10;
data2.trial{1} = ones(1,10)*2;
data2.sampleinfo = [1 10];

cfg = [];
data12 = ft_appenddata(cfg, data1, data2);
```

The situation that ft_redefinetrial has to deal with overlapping data needs to be dealt with there. The ft_fetch_data function has the 'allowoverlap' option for that. And with master I get

```
cfg = [];
cfg.trl = [
  1 2 0
  3 4 0
  5 6 0
  7 8 0
  9 10 0
  ];
data_cut = ft_redefinetrial(cfg, data12);

cfg = [];
cfg.trl = [
  1 2 0
  3 4 0
  5 6 0
  7 8 0
  9 10 0
  ];
data_cut = ft_redefinetrial(cfg, data12);

cfg = [];
cfg.trl = [
  1 2 0
  3 4 0
  5 6 0
  7 8 0
  9 10 0
  ];
data_cut = ft_redefinetrial(cfg, data12);

Error using ft_fetch_data (line 140)
some of the requested samples occur twice in the data and have conflicting values

Error in ft_redefinetrial (line 248)
    data.trial{iTrl} = ft_fetch_data(dataold, 'header', hdr, 'begsample', begsample, 'endsample',
    endsample, 'chanindx', 1:hdr.nChans, 'skipcheckdata', 1);
```

just as I had hoped.

Can you give an example with two datasets and each one trial that demonstrates the problem?

StolkArjen:
Can you look at situation 2 or 3 in test_bug3267_thesequel? Its a simplification of a real data scenario.

robertoostenveld:
in situation 3 the sampleinfo appears correct in redef3. The trialinfo is not correct. But that is not due to the appending. The appending-with-overlap results in a complex data structure to redefine, but the outcome of that should simply be 10 times 3 slices of one sample each. The fix should therefore be sought in ft_redefinetrial.

StolkArjen:
I agree ft_appenddata is not the problem, since the sampleinfo and trialinfo it puts out are simply correct. It's just that inconsistent sample ordering in the sampleinfo after appending confuses ft_redefinetrial. Accordingly, ft_redefinetrial (at line 253) would need to use a variable other than the sampleinfo to figure out to what original trial a segment belonged to. But what?

StolkArjen:
screen shot 2017-05-31 at 9 41 59 am

just to give a feeling for why ft_redefinetrial, when retrieving the original trialID on the basis of the sampleinfo, may get confused

StolkArjen:
Closing this bugreport since the change in ft_appenddata is not where the fix should be implemented. For now, it'd be advisory to have ft_redefinetrial refuse to update the trialinfo field when the sampleinfo segments aren't properly aligned (which causes incorrect trial info).

PR #448, opened 2017-06-01
fix: added check so ft_redefinetrial does not tacitly create incorrec…

StolkArjen:
…t trialinfo fields due to nonconsecutive sampleinfo segments

related to http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3267

Ideally, ft_redefinetrial has a way to correctly retrieve iTrlorig at line 253

### DIFF
--- a/ft_redefinetrial.m
+++ b/ft_redefinetrial.m
@@ -251,15 +251,16 @@ elseif ~isempty(cfg.trl)
     % ensure correct handling of trialinfo.
     % original trial
     iTrlorig  =  find(begsample <= dataold.sampleinfo(:,2) & endsample >= dataold.sampleinfo(:,1)); % Determines which old trials are present in new trials
-
+    
     if size(cfg.trl,2)>3 %In case user specified a trialinfo
       data.trialinfo(iTrl,:) = cfg.trl(iTrl,4:end);
       if isfield(dataold,'trialinfo')
         ft_warning('Original data has trialinfo, using user specified trialinfo instead');
       end;
     elseif isfield(dataold,'trialinfo') % If old data has trialinfo
-      if numel(iTrlorig) == 1 || ...  % only 1 old trial to copy trialinfo from, or
-          size(unique(dataold.trialinfo(iTrlorig,:),'rows'),1) % all old trialinfo rows are identical
+      if (numel(iTrlorig) == 1 ...  % only 1 old trial to copy trialinfo from, or
+          || size(unique(dataold.trialinfo(iTrlorig,:),'rows'),1)) ... % all old trialinfo rows are identical
+          && ~any(diff(dataold.sampleinfo(:,1))<=0) % and the trials are consecutive segments
         data.trialinfo(iTrl,:) = dataold.trialinfo(iTrlorig(1),:);
       else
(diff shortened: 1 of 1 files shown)


robertoostenveld:
it would have been nice to receive a test script together with this. It is not really clear what the original problem was.

StolkArjen:
This bug was a follow-up to a few other (closed) attempts, and the test script got lost in the process. I reran it with the new version of ft_redefinetrial which now does catch the problem, and throws an (informed) error. For completion, the original test script. We cannot really add it in this form, because ft_redefinetrial is now meant to throw an error in situation 2 and 3.

function test_bug3267_thesequel

% MEM 1500mb
% WALLTIME 00:10:00

%% situation 1: without sampleinfos (works fine)
data1a.label = {'1';'2';'3'};
for trl = 1:5
  data1a.trial{1,trl} = randn(3,3);
  data1a.time{1,trl} = [1 2 3];
end
data1a.trialinfo(:,1) = 1:5;

data1b = data1a;
data1b.trialinfo(:,1) = 6:10;

% append
data1 = ft_appenddata([], data1a, data1b);
% this works fine

cfg             = [];
cfg.length      = 1;
cfg.overlap     = 0;
redef1 = ft_redefinetrial(cfg, data1);

% this should be 1 1 1 2 2 2 ... 10 10 10
assert(all(abs(diff(redef1.trialinfo))<2)); % PASSES

%% situation 2: with identical sampleinfos (produces incorrect trialinfo)
data2a = data1a;
data2a.sampleinfo = [1 3; 4 6; 8 10; 11 13; 16 18];
data2b = data1b;
data2b.sampleinfo = data2a.sampleinfo;

% append
data2 = ft_appenddata([], data2a, data2b);

% tacitly fails because of intertwined sampleinfos (at line 263)
cfg             = [];
cfg.length      = 1;
cfg.overlap     = 0;
redef2 = ft_redefinetrial(cfg, data2);

% this should be 1 1 1 2 2 2 ... 10 10 10
assert(all(abs(diff(redef2.trialinfo))<2)); % FAILS


%% situation 3: with intertwined and partly overlapping sampleinfos (produces incorrect trialinfo, as in situation 2)
data3a = data1a;
data3a.sampleinfo = [1 3; 4 6; 8 10; 11 13; 16 18];
data3b = data1b;
data3b.sampleinfo = [2 4; 5 7; 9 11; 12 14; 15 17];

% append
data3 = ft_appenddata([], data3a, data3b);

% tacitly fails because of intertwined sampleinfos (at line 263)
cfg             = [];
cfg.length      = 1;
cfg.overlap     = 0;
redef3 = ft_redefinetrial(cfg, data3);

% this should be 1 1 1 2 2 2 ... 10 10 10
assert(all(abs(diff(redef3.trialinfo))<2)); % FAILS

robertoostenveld:
I added it to the existing script like this 

```
try
  detectederror = true;
  % the next tacitly fails because of intertwined sampleinfos (at line 263)
  cfg = [];
  cfg.length = 1;
  cfg.overlap = 0;
  redef2 = ft_redefinetrial(cfg, data2);
  detectederror = false;
end
assert(detectederror, 'the expected error was not detected');
```

StolkArjen:
Didn't know about the detecterror option, thanks

PR #449, opened 2017-06-01
ENH - more output for paired-ttest added

dmatthes1982:
Output of mean of difference, standard deviation of difference, confidence interval, degree of freedom and Cohen's d to output structure


robertoostenveld:
thanks. See also http://www.fieldtriptoolbox.org/example/effectsize, where we plan to document more w.r.t. effect size estimates (also for cluster stats).

PR #451, opened 2017-06-03
built in warning for when cfg.padding is shorter than the trial length

StolkArjen:


robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3301

PR #452, opened 2017-06-07
updated figure axes for a better fit on linux platforms

StolkArjen:
minor tweak, no need to double check really


robertoostenveld:
lets check whether the double-check can be skipped ... ;-)

robertoostenveld:
I triple checked, looks good!

StolkArjen:
Thanks! :)

PR #453, opened 2017-06-08
improve consistency of coordsys

robertoostenveld:
moved assignment of labels to axes on basis of coordsys to stand-alone function, see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3304


robertoostenveld:
Sorry, coordsys2label got stuck between two branches. It is there now.

schoffelen:
Would you want me to approve and merge?

robertoostenveld:
I propose that I first run the test batch on this. 

regarding the 4th output: I had that in my initial implementation, then realized that I did not have a sufficient handle on it yet. I was in the original code the tokenize on the coordsys; where in the code is a coordsys assigned that has a ‘_’ in it? I could only find ft_headcoordinates, which has als_ctf and ras_itab. This suggests that there are other ctf and itab coordinate systems as well. 

Should we not get rid of the als_ctf and ras_itab for consistency with the other systems, in which the first (and only) token represents both orientation and origin?

Furthermore, if we don’t use the 4th output, we should also not have it. The only use for it that I can see is in ft_determine_coordsys, where we could add the informative print statement

fprintf(‘The origin of the coordsys is %s\n’, labelo)

robertoostenveld:
I started the dashboard test batch

robertoostenveld:
the batch completed without errors. So it could be merged. 

I suppose to wait for @StolkArjen to review and possibly @smg211 to comment.

specific action points to consider: 
- should we remove the als_ctf and ras_itab? 
- anything else that needs to be fixed?
- any other code that is affected?

StolkArjen:
Assuming my understanding of this PR is to centralize the coordsys handling is correct, I much welcome it. For a next and related step, would it be possible to allow ft_convert_coordsys to take input like RAS and LAS? The current implementation:

 switch target
    case {'spm' 'mni' 'tal'}
      switch obj.coordsys
        case {'ctf' 'bti' '4d'}
          fprintf('Converting the coordinate system from %s to %s\n', obj.coordsys, target);
          if hastemplate
            obj = align_ctf2spm(obj, opt, template);
          else
            obj = align_ctf2spm(obj, opt);
          end
        case {'itab' 'neuromag'}
          fprintf('Converting the coordinate system from %s to %s\n', obj.coordsys, target);
          if hastemplate
            obj = align_ctf2spm(obj, opt, template);
          else
            obj = align_itab2spm(obj, opt);
          end
        otherwise

And finally, does this imply we will reserve labels like RAS and  LAS for native/unscaled space, and labels that are atlas-interpretible such as Tal and MNI for scans that are in normalized/standardized space?

StolkArjen:
Thus,

should we remove the als_ctf and ras_itab?
> my last note on scaled vs unscaled convention

anything else that needs to be fixed?
any other code that is affected?
> ft_convert_coordsys seemingly, as above
> ft_volumerealign seemingly, at line 259 and 304
> ft_electroderealign maybe, at line 190
> ft_volumereslice, at line 102
> ft_read_headshape uses head/ctf and dewar for a coordsys
> ft_read_mri attaches coordsys throughout

robertoostenveld:
> And finally, does this imply we will reserve labels like RAS and LAS for native/unscaled space, and labels that are atlas-interpretible such as Tal and MNI for scans that are in normalized/standardized space?

On https://mipav.cit.nih.gov/documentation/presentations/talairach.pdf is a nice set of slides that explains the scaling involved in Tal.

Although RAS does not specify the origin, I believe it is better than tal.

> More generally, for clarity we should use 'coordinate space' in the context of mni or tal (coordinate spaces that are interpretable atlas-wise), and 'coordinate system' for say ctf, ras, etc

Hmm, that is a subtle one. I don't expect that to work in the long run, i.e. people will not appreciate the difference. I think it is better to explicitly document that mni and tal are scaled, and that (most) others are not.

StolkArjen:
> On https://mipav.cit.nih.gov/documentation/presentations/talairach.pdf is a nice set of slides that explains the scaling involved in Tal.
>Although RAS does not specify the origin, I believe it is better than tal.
Indeed, the 'piecewise linear deformation/scaling' step is what isn't included.

ft_volumerealign and ft_convert_coordsys would need to support 'ras' too then (to be specified as cfg.coordsys)?

robertoostenveld:
ras is a superset of other (more explicit) coordinate systems , but with an unspecified origin. In ft_volumerealign I would not know how to specify the direction of the axes without specification of the origin.

StolkArjen:
Agreed, but it's not Talairach coordinate space either, since the coordinates are not scaled according to the actual Talairach atlas. So .. it's in Talairach-alike native space, which we maybe should dub 'RAS-AC' (which would be the same for MNI native space). But then again, the origin is not really crucial for aligning different scans, which mainly requires identical orientation.. (we may have a catch-22 here)

robertoostenveld:
see attached figure, which is from an (quite old) book on brain warping, of which I happen to have a paper copy.

it refers to the 'acpc' coordinate system. 

screen shot 2017-06-11 at 12 02 54

robertoostenveld:
I want to avoid 'RAS' in fieldtrip, since does not specify the origin, and is used in different software with a different specification of the origin.

The 'ACPC' coordinate system that is specified on https://books.google.nl/books?id=8WdlWJepgWMC&pg=PA347&dq=native+coordinate+system+anterior+commissure&hl=en&sa=X&ved=0ahUKEwirk9TPxbXUAhXQJVAKHatcBI8Q6AEIKDAA#v=onepage&q=native%20coordinate%20system%20anterior%20commissure&f=false on page 347 (in a chapter by Avi) seems to fulfill the requirements (native scale, RAS, AC as origin). 

See also http://www.brainvoyager.com/bvqx/doc/UsersGuide/BrainNormalization/AutomaticACPCAndTalairachTransformation.html

robertoostenveld:
@schoffelen @StolkArjen: how about using the following as conventional in FT for the RAS coordinate systems that have the anterior commissure as their origin.

spm==mni: the geometry is scaled according to MNI152 template
tal: the geometry is scaled according to TT88 atlas 
acpc: the geometry is non-scaled, i.e. native subject size 

This forms a clear and unambiguous principle that we can use to check (and where needed) update the corresponding fieldtrip functions. 

I don't have the overview yet as to what needs to be updated (in the code and on the wiki).

StolkArjen:
Great catch, I like it. It indeed seems brainvoyager uses it the acpc label for intermediate output in the process of normalizing a scan to tal space. One of the implications is that we rename align_ctf2spm into align_ctf2acpc?

robertoostenveld:
I also considered that for align_ctf2spm but am not 100% sure with all the SPM code in there that it is only doing a rigid body transform. Actually I had an initial go at making changes, but then realized that I am not 100% sure about all of them.

You made the nice list of affected software. I suggest we review that list and make (+share) a plan for each, prior to implementing the code changes. We should also prioritize them for the submission.

schoffelen:
Spuit 11: as far as I can see, the way SPM is called for align_abc2spm is only using rigid body transformation, so in that respect we are safe to equate the 'spm', to 'acpc', I think.

StolkArjen:
From that initial list, I'd say the following may not some adjustment:

ft_convert_coordsys, now that align_xxx is updated
ft_volumerealign, at lines 259 and 304
ft_volumereslice, at line 102
.. and of course in coordsys2label

There may be more. I could testrun the iEEG pipeline with the new coordsys once this has been pulled.

robertoostenveld:
I added this FAQ http://www.fieldtriptoolbox.org/faq/acpc, could you please review?

StolkArjen:
Same here, looks good as far as I can tell from the code changes. Overall, the introduction of the acpc system is a highly welcome conceptual advance!

robertoostenveld:
another principle to follow from now on: fsaverage==mni305. We already identified earlier that mni152==mni==spm.

based on FS documentation, I implemented (a very simple) align_fsaverage2mni

robertoostenveld:
I added "MNI305 or FSAVERAGE is used if the geometry is spatially warped to the MNI305 template brain" to http://www.fieldtriptoolbox.org/faq/acpc.

We are not using mni152, nor mni305 or fsaverage in FieldTrip so far.

robertoostenveld:
"we should not use it" -> FT code should not assign the fsaverage to any FT output object, since at this moment there is not a single function that spatially transforms input data to the fsaverage coordinates.

In case some FT input data object has the fsaverage coordsys, that is fine.

StolkArjen:
Actually, there's one location. Ft electroderealign attaches 'fsaverage' to elec.coordsys after warping (cfg.warp = 'fsaverage') the electrodes to the fsaverage brain. But those electrodes are likely only used for plotting data (ft plot sens, though also ft sourceplot maybe - still have to try out).

robertoostenveld:
I just did a test of this branch; the following scripts are now failing

```
test_bug1297.txt:=== test_bug1297 FAILED in 4 seconds
test_bug1665.txt:=== test_bug1665 FAILED in 0 seconds
test_bug1828.txt:=== test_bug1828 FAILED in 1 seconds
test_bug1832.txt:=== test_bug1832 FAILED in 1 seconds
test_bug1902.txt:=== test_bug1902 FAILED in 1 seconds
test_bug1954.txt:=== test_bug1954 FAILED in 1 seconds
test_bug1956.txt:=== test_bug1956 FAILED in 1 seconds
test_bug2621.txt:=== test_bug2621 FAILED in 6 seconds
test_bug2928.txt:=== test_bug2928 FAILED in 18 seconds
test_bug62.txt:=== test_bug62 FAILED in 5 seconds
test_ft_volumesegment.txt:=== test_ft_volumesegment FAILED in 2 seconds
test_spm12.txt:=== test_spm12 FAILED in 3 seconds
```

robertoostenveld:
I fixed a missing bracket. I now have 

```
test_tutorial_MNE.txt:=== test_tutorial_MNE FAILED in 8 seconds
```

note that there are other tests still running.

robertoostenveld:
roboos@mentat001> ll *.txt | wc -l
591

test_ft_componentanalysis.txt:=== test_ft_componentanalysis FAILED in 71 seconds
test_ft_freqanalysis.txt:=== test_ft_freqanalysis FAILED in 581 seconds

these always fail due to small numerical inaccuracies.

That means that all others are working.

PR #456, opened 2017-06-13
bugfix - allow for NaNs in trialinfo/sampleinfo/cumsumcnt/cumtapcnt w…

roemervandermeij:
…hen appending datasets

NaNs can be useful in the trialinfo field, e.g. when a numeric trial type code is not defined for a subset of trials.

For sampleinfo/cumsumcnt/cumtapcnt, the other affect fields when appending using the new append_common.m, I see no immediate reason for having a meaningful NaN. But, in case there is a NaN in identical locations of the input datasets, then why not keep them and their field? Otherwise, an exception is also fine of course.

### DIFF
--- a/private/append_common.m
+++ b/private/append_common.m
@@ -104,7 +104,7 @@ switch cfg.appenddim
     for i=1:numel(fn)
       keepfield = isfield(varargin{1}, fn{i});
       for j=1:numel(varargin)
-        if ~isfield(varargin{j}, fn{i}) || ~isequal(varargin{j}.(fn{i}), varargin{1}.(fn{i}))
+        if ~isfield(varargin{j}, fn{i}) || ~isequalwithequalnans(varargin{j}.(fn{i}), varargin{1}.(fn{i}))
           keepfield = false;
           break
         end


roemervandermeij:
Ha, fine fine, I'll go ahead with the boring.. tests ;), even for such a ridiculously small change

roemervandermeij:
Okay, come on, this is a bit ridiculous, it already takes more than 1 hour and the output already is over 145 pages long. I'm not gonna skim through that to find any unintended effects of 1 isequal vs isequalwithequalnans change (also considering that running it without the change produced plenty of errors). I understand the need for testing, but the current call result is obviously not practical. I assume it's because there's no test_ft_appenddata?

robertoostenveld:
Note that all changes to ft_appenddata w.r.t. trial handling need to be implemented for the other append functions as well. Trial information is not only for raw dat structures. And there are test functions for those:

mac011> cd test/
mac011> ls *append*
test_ft_appendfreq.m		
test_ft_appendtimelock.m
test_old_appenddata_2datasets.m

and test_bug946 is relevant as well. 

You don't have to run the whole batch. Even without the test data present (which ft_test detects), it takes too long on a single computer. But ideally I would like to see 

>> ft_test report fieldtripversion 0571206

to return some test results that don't fail.

schoffelen:
Sorry if I started this, but I mean to ask you to do something like ft_test run ft_appendtimelock (or so), which should just test a single function

roemervandermeij:
Sorry for the quick reply and not being to delve into the testing framework just now, obviously it evolved a lot.

The change is only to append_common, where an isequal is replaced with isequalwithequalnans when checking trialinfo/sampleinfo/cumsumcnt/cumtapcnt (in the case of trialinfo, reimplementing lost functionality). The rationale is simple, if in these fields there's a NaN in one dataset in the exact same spot as another for these fields, no reason to throw them all out.

test_ft_appendfreq runs through fine
test_ft_appendtimelock runs through fine until a hard coded dccn path
test_old_appenddata_2datasets starts with a hard coded dccn path

schoffelen:
Geeft helemaal niet erg 😄 . Ik durf bij deze commit het risico wel te nemen (ik was even vergeten dat sommige testfuncties natuurlijk alleen maar inhouse werken). I ran the other test function locally (without the ft_test, because my local repo wasn't up to date), they ran through fine.

PR #458, opened 2017-06-14
BUGFIX - important for ft_preproc_polyremoval

dmatthes1982:
typ hast to be set in line 62, since it is used in the last line in ft_preproc_polyremoval


robertoostenveld:
thanks. I also realized this bug and just committed a fix independent of this PR.

PR #459, opened 2017-06-14
use gammaln instead of lgamma

gllmflndn:
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3312


robertoostenveld:
that was fast! Do you want to include the other changes (e.g. ON/OFF) in here as well?

gllmflndn:
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3310

robertoostenveld:
thanks. Anything else that you want in here? If not, I'll merge.

gllmflndn:
I think that should be it. Thanks!

robertoostenveld:
just create another PR if there is more.

PR #461, opened 2017-06-15
A little FIX in ft_preprocessing

dmatthes1982:
Throw padding related warning only if padding is selected  (cfg.padding > 0)


robertoostenveld:
good point! thanks.

PR #462, opened 2017-06-16
second FIX in ft_preprocessing - similar problem

dmatthes1982:
throw padding related warnings only if padding is selected 
no further cases found


StolkArjen:
Thanks for fixing this, @dmatthes1982. I introduced those warnings, but didn't test it under the condition that no padding was specified.

PR #465, opened 2017-06-20
Bug3313 bias

schoffelen:
Hmmm. I have built in support for SPM12, which for the time being required me to adjust align_itab2spm a bit (to give it support for spm12). I would expect this to transparently change once the new coordsys business comes online (which at the moment is in one of Robert's other branches). The reason I needed it for now, is that the example MR-image, associated with bug3313 is in 'neuromag' (='itab') convention.


schoffelen:
There is however something conceptually funny with the spm12 version.

Long story short, spm_preproc8 does bias estimation/correction, segmentation and spatial normalisation in a single function. Subsequently (this is how it is in ft_volumesegment/ft_volumenormalise, and now also in ft_volumebiascorrect) spm_preproc_write8 is called, which, depending on the specification some input flags, writes out some volumes (this can be warped/unwarped, the anatomical info, the segmentation volumes, the biascorrected image, and also the bias fields). Now, if the main purpose of ft_volumebiascorrect would be to biascorrect the image in order to achieve a better segmentation, in case of spm12 this function does not necessarily have to exist as a separate function to begin with.

jens-k:
This seems to break the segmentation for me on windows (not on linux). We had a similar problem before, do you remember?

**Test script:**

mri = ft_read_mri(abpath('Subject01.mri'));

cfg     = [];
cfg.dim = mri.dim;
mri     = ft_volumereslice(cfg,mri);

cfg           = [];
cfg.output    = {'gray','white','csf','skull','scalp'};
segmentedmri  = ft_volumesegment(cfg, mri);

**Error:**

Undefined function or variable 'cfg_branch'.

Error in cfg_getfile>reg_filter (line 1185)
    dprms  = {cfg_branch};

Error in cfg_getfile (line 177)
            t = reg_filter(varargin{2:end});

Error in spm_select (line 115)
            cfg_getfile('regfilter', 'mesh',  {'\.gii$'});

Error in spm_select (line 110)
            spm_select('regfilter');

Error in spm_select (line 97)
    spm_select('init');

Error in spm_platform>init_platform (line 233)
    driveLett = spm_select('ListDrives');

Error in spm_platform (line 62)
if isempty(PLATFORM), PLATFORM = init_platform; end

Error in spm_smoothto8bit>smoothto8bit (line 46)
VO.dt     = [spm_type('uint8') spm_platform('bigend')];

Error in spm_smoothto8bit (line 16)
    VO = smoothto8bit(V,fwhm);

Error in spm_normalise (line 157)
VF1 = spm_smoothto8bit(VF,flags.smosrc);

Error in align_ctf2spm (line 180)
  params           = spm_normalise(V2,V1,[],[],[],flags);

Error in ft_convert_coordsys (line 97)
            obj = align_ctf2spm(obj, opt);

Error in ft_volumesegment (line 307)
    mri = ft_convert_coordsys(mri, 'spm');

jens-k:
Yeah, here it was:

https://github.com/fieldtrip/fieldtrip/commit/b79ad1505df122912c4c0e71729c8040226f2fc1#commitcomment-21501648

schoffelen:
sorry Jens, my bad. I updated ft_volumesegment to use spm12 by default. Could you check whether specifying cfg.spmversion='spm8' works?

Or did you have it working on windows before with spm12?

schoffelen:
Hmmm, now it seems to need cfg_branch. Shall I just try and copy it over to external spm12, and see whether it runs through?

jens-k:
Yes, I just tried that and it works.

schoffelen:
but that's then the @cfg_branch subdirectory?

schoffelen:
Oh, and note that I have added functionality for 'mars'
cfg.spmversion='spm12';
cfg.spmmethod='mars';

schoffelen:
does a postprocessing on the tpm's to get smoother results
don't know whether it's useful, but looks like state-of-the-art

jens-k:
Hey, sounds cool. Does that automatically perform the bias correction?

schoffelen:
well, that's what is in the spm_preproc8

Yet, optimal correction parameters are now at least configureable

jens-k:
I just moved the cfg_branch into /matlabbatch/ but to be consistent with the spm folder structure maybe you wanna move it into /matlabbatch/@cfg_branch/cfg_branch.

All the other files in @cfg_branch seem not to be needed.

jens-k:
Oh, now that I have you here: When creating a mesh from that segmentation, some of the prior changes changed the order of tissue types in mesh.tissuelabel. That breaks the FEM headmodel tutorial (now the wrong conductivities are assigned to the tissues).

http://www.fieldtriptoolbox.org/tutorial/headmodel_eeg_fem#mesh

The order of tissues used to be the way it was specified, now its alphabetical.

jens-k:
I think we should either change the order or the tutorial. This here would be robust against any order changes in the mesh (I'm sure there is a more elegant way):

conductivities{1}				= {'gray'; 'white'; 'csf'; 'skull'; 'scalp'};
conductivities{2}				= {0.33; 0.14; 1.79; 0.01; 0.43};
for t = 1:numel(mesh.tissuelabel)
	idx(t) = find(cellfun(@(x) strcmp(mesh.tissuelabel{t}, x), conductivities{1}));
end

% Create the actual head model
cfg                         = [];
cfg.method           = 'simbio';
cfg.conductivity    = [conductivities{2}{idx}];   % order follows mesh.tissuelabel
vol                         = ft_prepare_headmodel(cfg, mesh);

jens-k:
Concerning the bias correction, should we also add the following to the headmodel tutorials?

mri = ft_volumebiascorrect([], mri);

I seem to not be able to edit the tutorials though. Do I have to be at Donders to do that?

jens-k:
ft_volumebiascorrect currently does not work with spmversion = 'spm12' since it requires spm_bias_estimate, which only exists in external/spm8/ and does not exist in spm12 (also not in the full spm12). I guess it would be desirable to consistently use spm12 throughout the processing pipeline, and if its only to prevent path problems.

Once that's settled, should we  add the following to the headmodel tutorials?

cfg							= [];
cfg.spmversion				= 'spm12';
mri   						= ft_volumeunbias(cfg, mri);

I seem to not be able to edit the tutorials though. Do I have to be at Donders to do that?

schoffelen:
OK, back at my desk again:

I think you may want to ask Robert to grant you edit access. I had problems with editing after we moved over to the new wiki layout too..

Biascorrection is automatically done in ft_volumesegment (at least when using spm12 as version (and probably also spm8), so no bias correction is needed.

I guess that volumebiascorrect can be used to probe the effect of different parameter setttings, although the effect of the biascorrection parameters is typically evaluated on the segmentation output, so as such there's no reason for this function to exist.

Which I only realized after I implemented it, so that's why it still exists...

jens-k:
There was one more file for ft_volumesegment to work on windows (apparently only needed when using spm12 but the old segmentation). Here is a pull request adding both files:

https://github.com/fieldtrip/fieldtrip/pull/469

The above problem with ft_volumebiascorrect is however still on.

schoffelen:
hmm, I am curious why ft_volumebiascorrect doesn't work, or are you perchance calling ft_volumeunbias, which should be removed?

jens-k:
No, unfortunately not. this script yields the same error on windows and linux:

mri = ft_read_mri('Subject01.mri');
cfg							= [];
cfg.spmversion				= 'spm12';
mri_un						= ft_volumebiascorrect(cfg, mri);

--

Error using chol
Matrix must be positive definite.

Error in spm_preproc8>log_likelihoods (line 945)
    C      = chol(vr(:,:,k));

Error in spm_preproc8>latent (line 985)
Q   = log_likelihoods(f,bf,mg,mn,vr);

Error in spm_preproc8 (line 412)
                    [q,dll] = latent(buf(z).f,buf(z).bf,mg,mn,vr,B,lkp,wp);

Error in ft_volumebiascorrect (line 114)
        p = spm_preproc8(opts);

jens-k:
The errors don't occur if you resluce the MRI before (so maybe it's my mistake?):

cfg     = [];
cfg.dim = mri.dim;
mri     = ft_volumereslice(cfg,mri);

However, under windows, the warning appears (just to let you know):
Warning: Using 'state' to set RAND's internal state causes RAND, RANDI, and RANDN to use legacy random number
generators.  This syntax will be removed in a future release.  See Updating Your Random Number Generator Syntax to use
RNG to replace the old syntax.

schoffelen:
Strange, as far as I can see, the same 'preprocessing' is applied to the mri image before it is passed to spm_preproc8 as in ft_volumesegment (when using cfg.spmversion='spm12', and cfg.spmmethod='new'). This does not fail

schoffelen:
no, correction. In ft_volumesegment a magical thing is done on line 313, which could have to do with the discrepancy between ft_volumesegment and ft_volumebiascorrect

schoffelen:
looks like it was the scaling step that caused the problem. Running a test now.

jens-k:
The linux error is gone for me after your last commits.

PR #466, opened 2017-06-22
comment printf() calls in bemcp

gllmflndn:
Cosmetic changes to remove unnecessary printf statements.


robertoostenveld:
the removal of printf only has effect when running matlab on the command line, right?

gllmflndn:
Display happens all the time, whether the MATLAB desktop is used or not.

PR #467, opened 2017-06-26
bugfix for bug3307 

kdijkstra:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3307

Changed biosemi section in read_trigger to avoid call to bitcmp (arguments changed after matlab2012). It now matches the biosemi implementation for bdf files (in ft_read_events) more closely, with the only difference that apparently the status channel in biosemi .gdf files (at least in our research group) still require correction for having been shifted by one byte.

(see also: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2409)

### DIFF
--- a/fileio/private/read_trigger.m
+++ b/fileio/private/read_trigger.m
@@ -89,6 +89,12 @@ if denoise
 end
 
 if fixbiosemi
+  if ft_platform_supports('int32_logical_operations')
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks!

PR #468, opened 2017-06-26
Control of status output for some functions improved

dmatthes1982:
- Forwarding of cfg.showcallinfo by using keepfields in two functions added.
- Replacement of fprintf/warning/error with ft_info/ft_warning/ft_error in another six functions.


robertoostenveld:
I ran the test scripts versus this PR. So far 409 scripts are ready, one gives an error. 

test_bug1168.txt:=== test_bug1168 FAILED in 17 seconds
 
This is due to an assertion about the content of a specific error message. So no content-wise error, but something where a subtle change seems to have happened. 

I will merge and then fix that test script.

PR #470, opened 2017-06-27
Another spm file for ft_volumesegment

jens-k:
This one seems to be needed only when using the new mars segmentation.


schoffelen:
Thanks, Jens!

PR #471, opened 2017-06-27
Propose fix for bug 3320

jens-k:
http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3320

This solution worked for me. Please double-check, though, there are probably better ways to deal with superfluous fields..

### DIFF
--- a/ft_volumesegment.m
+++ b/ft_volumesegment.m
@@ -753,8 +753,9 @@ else
   error('unknown output %s requested\n', cfg.output{:});
 end
 
-% remove unnecessary fields
-segmented = removefields(segmented, remove);
+% remove unnecessary fields (use previously collected 'remove' and further
+% make sure only desired tissue types are present)
+segmented = removefields(segmented, unique([remove setdiff({'white' 'gray' 'csf' 'brain' 'skull' 'scalp' 'skullstrip' 'softtissue' 'bone' 'air'}, outp)]));
 
 % do the general cleanup and bookkeeping at the end of the function
 ft_postamble debug


schoffelen:
Hmmm, I would prefer it to be slightly differently solved higher up in the code (rather than once more enumerating all possible output fields). I suspect that something similar can be achieved by adjusting lines 686,690 etc. (probably adding 'bone' 'air' 'softtissue' there would do the trick. Can you give this a shot?

jens-k:
I'm not sure if I follow the strategy in those lines. If we keep it the way it is, we first put all possible tissue types into 'remove' and then repeatedly intersect those strings again with all possible tissue types minus the one tissue that we just created.

Why don't we feed all tissue types into 'remove' and then explicitly remove particularly that one tissue from it that we just created (its effectively the same but spares redundancy and is probably more intuitive to read).

remove = {'anatomy' 'csf' 'gray' 'white' 'bone' 'softtissue' 'air' 'skull' 'skullstrip' 'brain'}; % added last three

create csf mask here
remove = remove(~strcmp(remove, 'csf'))

create skull here
remove = remove(~strcmp(remove, 'skull'))

schoffelen:
yeah, that sounds good, i think.
Perhaps, rather than saying remove=remove(~strcmp(remove,'fieldname'));
one could do remove(strcmp(remove,'fieldname')))=[];

schoffelen:
Looks good. I am willing to take the risk...
Thanks!

schoffelen:
Feel free to close bug 3320 if it works for you now

jens-k:
I tested a couple of output parameter combinations (binary masks and tpms) and it worked fine for me.

PR #472, opened 2017-06-28
Bug2997 alpha

schoffelen:


schoffelen:
OK, this is my first attempt at addressing http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2997

Basically, I now implemented a rg_rgba2rgb function, that essentially allows for the conversion of a colored pixel with a certain background color, with on top of it a opacity-masked other color, into a single rgb value.

I built it into ft_plot_mesh, and added the colorlimit and opacitymasking handling as options to this function.
For now I named the option 'maskstyle', where the default value is 'opacity', and the new functionality is 'saturation'. I am not so happy with the naming, because it suggests a strategy that is similar to how it's done in ft_plot_topo and ft_plot_matrix. So any suggestion of a better name is welcome.

The idea would now be to also build this into ft_plot_slice and ft_plot_ortho and ft_plot_montage, and think about consistency with respect to ft_plot_topo and ft_plot_matrix.

By the way, working in ft_plot_mesh, I encountered some inconsistent variable handling which we may consider to iron out while I am at it. Particularly, the masking parameter is dealt with as 'facealpha', while it is inputted as a Number-of-vertices vector. It would probably be better to call this vertexalpha instead.

schoffelen:
@KristijanArmeni can you evaluate whether the current functionality in this branch works as expected?

i.e. 

cfg=[];
cfg.method='surface';
cfg.maskstyle='saturation';
cfg.funparameter='something'
cfg.maskparameter='somethingelse', but can be 'something'
ft_sourceplot(cfg, source);

schoffelen:
@robertoostenveld could you approve of the general strategy?

plotting/private/bg_rgba2rgb.m
```diff
@@ -0,0 +1,116 @@
+function [rgb] = bg_rgba2rgb(rgba, bg, varargin)
```
robertoostenveld: I suggest to make the order of the input arguments consistent with the name of the function.

plotting/private/bg_rgba2rgb.m
```diff
@@ -0,0 +1,116 @@
...
+
+if size(bg,1) ~= size(rgba,1)
+    error('the background and data should have the same number of points in the rows');
```
robertoostenveld: points is (or will be) ambiguous. In the future it can be either vertices or voxels. I suggest something like

"the background and color data should have the same number of rows"

plotting/private/bg_rgba2rgb.m
```diff
@@ -0,0 +1,116 @@
...
+end
+
+if size(rgba,2)==1,
```
robertoostenveld: please no "," at the end

robertoostenveld:
I general I think it is a fine solution. Getting the colormap and translating the values into colors requires careful attention.

KristijanArmeni:
@schoffelen I ran ft_sourceplot on your branch and works as it should (compared to when checking out on master)

schoffelen:
what about 'colormix'?

robertoostenveld:
I like that better. Saturation is a feature of a single color, doing a “colormix” with white results in saturation.

plotting/private/bg_rgba2rgb.m
```diff
@@ -10,64 +10,89 @@
...
+    error('invalid specification of colormap, should be nx3');
+  end
+  clim  = varargin{2}; if isempty(clim), clim(1) = min(rgba(:)); clim(2) = max(rgba(:)); end
+  alpha = varargin{3};
+  amap  = varargin{4};
+  alim  = varargin{5}; if isempty(alim), alim(1) = min(alpha(:)); alim(2) = max(alpha(:)); end
+        
+  % deal with the color data
+  dat     = rgba;
+  siz     = size(dat);
```
KristijanArmeni: I think l. 56 needs to be `siz = size(dat, 1),` otherwise the reshape step in l. 91 outputs a rgb with a singleton dim 'npos x 1 x 3' which makes it crash in ft_plot_mesh.

schoffelen:
Yeah, you're right.
I had fixed this alreadly locally, now pushed it

schoffelen:
OK, this branch is now at a stage that the 'colormix' maskstyle is implemented in the intermediate/low-level plotting functions.

it still needs to be broadcast to the 'ortho' and 'montage' methods for ft_sourceplot, but this would require a bit more work, which I am not willing to spend at the moment. 

Preliminary testing seems to work fine (i.e. no error when running test_tutorial_plotting)

Still to do: document this option, e.g. in a FAQ

PR #473, opened 2017-06-29
Bug3318 - implement support for the 16-channel OpenBCI cyton board

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3318


robertoostenveld:
I still need to recompile on some of the other platforms

robertoostenveld:
I have recompiled on the current platforms.

PR #474, opened 2017-07-04
code cleanup - use ft_warning instead of warning

robertoostenveld:
and some other cleanups, e.g. capitalization of fieldtrip


robertoostenveld:
@schoffelen I have been running the full dashboard test. I receive some weird errors and some PBS jobs failing. However, these are not consistent and I don't think they relate to the m-code. I think this is due to the cluster having difficulties with fileserver access. 

I'll merge it anyway. Let's keep an eye open for weird errors.

besa2fieldtrip.m
```diff
@@ -173,9 +173,9 @@
     % convert into a TIMELOCKANALYSIS compatible data structure
     data = [];
     data.label = [];
-    if isfield(tmp, 'ChannelLabels'),
+    if isfield(tmp, 'ChannelLabels')
         data.label = fixlabels(tmp.ChannelLabels);
-    end;
+    en;
```
mmagnuski: should be `end` :)

schoffelen: well spotted! Apologies and I fixed it.

PR #475, opened 2017-07-09
Blackrock

awutz:
This a pull request to edit ft_read_data and ft_read_event for blackrock data.


schoffelen:
Hi Andi,

You shouldn't necessarily be doing a pull request at this time. Typically, the pull request is done once the developer(s) of the branch in question agree that the code is sufficiently stable to be merged into the main repository. For now, I think we first should agree on the current implementation, and probably improve it, before it can actually be pulled. The fact that you were able to initiate the pull request, however, indicates that you managed to pull my branch into your local repository, which  is a good thing. This would mean that you now also have a 'blackrock' branch on your repository, which you should be able to freely edit. If you have improvements to commit which you want to have checked by me, you need to do a pull request into my branch (i.e. the blackrock branch of schoffelen/fieldtrip.git), so that I can get the same code changes, and test them. Eventually, once we agree, you could pull request your branch directly into fieldtrip/fieldtrip, or go through my branch.
I can imagine that it sounds complicated, but I think that this would be the cleanest way to go.

I suggest to keep this pull request open, and use the conversation section to discuss further

awutz:
Hi Jan-Mathijs,

sorry for the delay here. I was on vacation last week.
I still have not managed to find your blackrock branch in my local resp. but this will very likely be rooted in my very beginner knowledge of how git works. I will now attach comments with my suggested changes to the files ft_read_data and ft_read_event here.

fileio/ft_read_data.m
```diff
@@ -1372,6 +1372,21 @@
...
+    % ensure that the filename contains a full path specification,
+    % otherwise the low-level function fails
+    [p,f,e] = fileparts(filename);
+    if ~isempty(p)
+      % this is OK
+    elseif isempty(p)
+      filename = which(filename);
+    end
+    orig = openNSx(filename, 'duration', [begsample endsample], 'channels', chanindx);
+    dat  = double(orig.Data);
```
awutz: ft_read_data works fine with the current code if one continuous piece of recording is read in. However, this code will fail, if one stopped the recoding during the session and then started again. In this case,  orig.Data will be a {1xN cell} with N being the number of recording segments between start and end of recording. Any ideas how to deal with this?

fileio/ft_read_event.m
```diff
@@ -2022,6 +2022,47 @@
...
+      % this is OK
+    elseif isempty(p)
+      filename = which(filename);
+		end
+		
+    % 'noread' prevents reading of the spike waveforms 
+    % 'nosave' prevents the automatic conversion of
+    % the .nev file as a .mat file
+    orig = openNEV(filename, 'noread', 'nosave')
+
```
awutz: This code should work. I would delete lines 2044 to 2047 and 2054 to 2057. 
In line 2062, it should say:
event(k).value     = eventCodes(k);

This will be the event codes coming from the serial-digital-IO. There are no codes associated with "Comments", "Spikes", etc. These are coded in separate fields of the orig.Data structure. The field orig.Data.SerialDigitalIO.InsertionReason does NOT dissociate between those codes. Unfortunately, I cannot contribute any clever way to implement more here because I don't have data that contains "comments" or "spikes". 

One crucial point: The .nev files (read in here) will always contain the event codes with 30K sampling but the actual data (.nsx files) can come with different sampling rates. I take care of this in a custom-written 'blackrock_to_ft_trialfun' which can be called like this 

filename1 = 'session1.nev'; % event file
filename2 = 'session2.ns2'; % different file formats (.ns2, .ns6, etc.) have different sampling freqs

cfg = [];
cfg.dataset = filename2;
cfg.eventfile = filename;
cfg.trialdef.eventtype='trigger';
cfg.trialdef.eventvalue=1;
cfg.trialdef.prestim        = 1;
cfg.trialdef.poststim       = 1;
cfg.trialfun                = 'blackrock_to_ft_trialfun';
cfg                         = ft_definetrial(cfg); 

%% 
The trial function looks like this:

function trl=blackrock_to_ft_trialfun(cfg)
hdr=ft_read_header(cfg.dataset, 'headerformat','blackrock_nsx');
event = ft_read_event(cfg.eventfile, 'eventformat', 'blackrock_nev');
trl=[];

 for i=1:length(event)
  if strcmp(event(i).type, 'trigger')
    % it is a trigger, see whether it has the right value
    if ismember(event(i).value, cfg.trialdef.eventvalue)
      % add this to the trl definition
      begsample     = event(i).sample*hdr.Fs - cfg.trialdef.prestim*hdr.Fs;
      endsample     = event(i).sample*hdr.Fs + cfg.trialdef.poststim*hdr.Fs - 1;
      offset        = -cfg.trialdef.prestim*hdr.Fs;  
      trigger       = event(i).value; % remember the trigger (=condition) for each trial
      trl(end+1, :) = [round([begsample endsample offset])  trigger]; 
    end
  end
end

fileio/ft_read_data.m
schoffelen: we could replace line 1387/1388 with a conditional statement. This would require that it is known a priori, whether the datafile consists of a single continuous piece of data, or of multiple small pieces of (discontinuous) stuff. I guess that begsample/endsample handling would need to be carefully implemented and checked in that case as well. How easy would it be for you to create a small example dataset?

fileio/ft_read_event.m
schoffelen: Am I correct to understand, that even for a given recording, there can be a set of nsx files with different sampling rates? This would indeed make things a bit complicated

fileio/ft_read_data.m
schoffelen: No worries, I think I have found an example of this.

awutz: In fact, the begsample/endsample argument won't work in case of a paused file. The low-level function does not permit these arguments in that case. You can only read in the whole file then. I know that Blackrock is working on that issue too. I shot an email to the technical support there and invited him to join us here.

fileio/ft_read_event.m
awutz: Yes, this is correct. Several nsx files with different sampling rates can be created for any given recording.

fileio/ft_read_data.m
schoffelen: I thought that I had access to a paused file, but it turns out I don't. It would be great if the support staff could chime in here.

awutz: The latest version of the NPMK toolbox should allow begsample / endsample to work properly. I will test this as soon as I find time. 

(https://github.com/BlackrockMicrosystems/NPMK/tree/master/NPMK)

PR #477, opened 2017-07-21
ft_databrowser timestamp

ludo-neuro:
Hi,
Here is a very simple fix of a bug preventing the proper display of custom data cursor text in ft_databrowser (see attached picture). Now I can get accurate timing of interesting events super easily! =)
Best,
Ludovic

ft_databrowsertimestampfix


schoffelen:
Thanks Ludovic, I will pull this, but could you please document here which Matlab version (+operating system) gave you problems? I just quickly tried to reproduce your original problem in Matlab2016b on a linux system, and could not reproduce. The case-sensitivity of the low-level object properties may be case-sensitive in certain setups only. 
At least changing the 'position' to 'Position' in my case did not lead to an error, so I think it's safe to pull this one.

ludo-neuro:
Sure: I originally encountered this issue using MATLAB 2015a on CentOS 
(release 6.9), and I just got this same issue using MATLAB 2016a on 
Windows 10 (version 10.0.15063). Good to know case-sensitivity seems not 
to be a problem anymore on further MATLAB versions. Thank you and have a 
great day!

PR #480, opened 2017-07-28
Bugfixes and enhancements to ft_volumelookup

smg211:
Previously, when going from an ROI to a label with ft_volumelookup, the labels.count field in the output did not represent an accurate tally of how many times each label was found around the ROI. The problem appeared to have 2 causes, both of which I outlined in comments in the code, but I will also describe here:

1. atlas_lookup was eliminating repeated labels found within a search area by using the setdiff and unique commands on the labels. Since only one instance of each label that was found was then returned to ft_volumelookup, the count would be lost. Thus, I changed the way atlaslookup ignores zeros that it finds in the search area, to avoid using the setdiff and unique functions.

2. Even when atlas_lookup passes a list of labels to ft_volumelookup, with repeats for labels that were found multiple times in the search area, ft_volumelookup would generate the same count for each label. Thus, I changed it to actually sum up the number of times each label appeared.

I also added in a new method to atlas_lookup where voxels surrounding the voi are searched in a spherical method, rather than in a cube, because searching in a cube would give equal weight to voxels (and their corresponding labels) that may be an unequal distance from the voi.


StolkArjen:
It looks like this branch may have overlap with the branch in PR479, in which you improved the write_bioimage_grid implementation, which could produce a conflict when both get pulled in. And for some reason, git thinks ft_volumelookup and atlas_lookup might also be at risk of conflict?

robertoostenveld:
@smg211 or @StolkArjen can you please resolve the conflicts? If not, I will reject this PR and close it.

smg211:
Conflicts resolved.

robertoostenveld:
thanks

PR #481, opened 2017-08-01
Fix IBTB toolbox check

ThreeFive-O:
Website http://ibtb.org is down and the toolbox no longer available there.
Workaround is to use repository of https://github.com/selimonat/InformationBreakdownToolbox to get toolbox.

Dependency logic was adapted to check for `binr.m` and `information.m` as used by `ft_connectivity_mutualinformation.m`. So far the only field trip function that checks for IBTB toolbox.

### DIFF
--- a/fileio/private/ft_hastoolbox.m
+++ b/fileio/private/ft_hastoolbox.m
@@ -131,7 +131,7 @@ url = {
   'MYSQL'         'see http://www.mathworks.com/matlabcentral/fileexchange/8663-mysql-database-connector'
   'ISO2MESH'      'see http://iso2mesh.sourceforge.net/cgi-bin/index.cgi?Home or contact Qianqian Fang'
   'DATAHASH'      'see http://www.mathworks.com/matlabcentral/fileexchange/31272'
-  'IBTB'          'see http://www.ibtb.org'
+  'IBTB'          'see https://github.com/selimonat/InformationBreakdownToolbox'
   'ICASSO'        'see http://www.cis.hut.fi/projects/ica/icasso'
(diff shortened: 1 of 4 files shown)


schoffelen:
Thanks for catching this. Rather than using a workaround, we would like to point to the location where the software is maintained. I have locally a version 1.0.5 of the code, and the repository you refer to is 1.0.3, so it does not seem the most up-to-date one. If the code is not actively hosted anywhere anymore, I think we could consider shipping (the most recent) a version with fieldtrip

ThreeFive-O:
Sounds good to me. Would you do a PR for the 1.0.5 code base?
If the code will be shipped with fieldtrip, then this PR can be considered obsolete.
Probably makes more sense to open up another PR to kick out the unnecessary checks, once the IBTB code is within the fieldtrip repository.

schoffelen:
Although the referenced code on Github seems to be a derivation of the original code, it is good enough for now. I contacted Cesare to get a fresh copy of versoin 1.0.5, but I haven't heard back from him.

asadmalik85:
Hi, this github link no longer works - not sure where to find the most recent version.

schoffelen:
Neither do I

schoffelen:
It's unfortunate that the original toolbox contributors are not maintaining and hosting their code anymore. As per discussed above in this issue, I have now added my own local copy of ibtb (version 1.0.5, with possible minor tweaks (undocumented)) to fieldtrip/external/ibtb.

It is now present in the master branch. If all goes well with the test scripts that we run each night, it should become available in the release version soon.

PR #484, opened 2017-08-03
ENH - increased transparency of selected artifacts

roemervandermeij:
ENH - increased transparency of selected artifacts to improve visibility when comparing artifacts from different sources.

### DIFF
--- a/ft_databrowser.m
+++ b/ft_databrowser.m
@@ -1813,7 +1813,7 @@ for j = ordervec
   
   for k=1:numel(artbeg)
     xpos = [tim(artbeg(k)) tim(artend(k))] + ([-.5 +.5]./opt.fsample);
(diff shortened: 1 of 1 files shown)


schoffelen:
Brilliant, thanks!

roemervandermeij:
I pondered long and hard on how to solve this. Many a night of sleep lost, waking with the same question over and over again. Haunting me. Nightmares, oh, so many nightmares. It makes one's mind wander. If we cannot solve this, how can we even imagine figuring out the slightly bigger mysteries in life? What is the meaning of Stonehenge? If life exist beyond our own sun, why have we not seen it? Why do we need semi-colons? In those moments, I feel truly lost in our ineptitude. Yet, unlikely as it sounds, I found a solution to that which could not leave my mind. On a faithful day this summer, it finally came to me. I was taking a bath, of course, when it hit me. Shouting out something that sounded Greek, I think, I ran to my laptop, opened matlab, and made that fateful change. It is what you see before you today. So elegant, so impactful. It indeed was, as you say, a true stroke of Brilliance.

schoffelen:
😂

StolkArjen:
Are you sure it happened while taking a bath? For me, it's usually the shower. But then again, never had an idea this brilliant.

PR #486, opened 2017-08-09
ENH: peakweighted method for faster localization, adjustable magnet r…

StolkArjen:
…adius option

Tested it several times. It works reliably and electrode selection/extraction has never been this fast (!).


StolkArjen:
Note that the new 'peakweighted' option (now default) is a hybrid between the local maximum and the center-of-mass. That is, the cursor is first moved to the peak voxel, and then to the center of mass, all in a single step (the peak is not necessarily always in the middle of an electrode, but going there helps getting it closer at a first step). As a result, it's hardly no longer needed to adjust initial guesses (one click, one electrode).

PR #492, opened 2017-08-15
ft_multiplotER can now handle cfg.trials

juliankeil:
As described on bugzilla in the comments to bug 3064, I hacked the ft_multiplotER-function (see attachment) to accept the cfg.trial-selection.

Specifically, I included "trials" field in the keepselection (line 317 of the hacked function), I kept the old trial selection (line 350), temporarily switched the trial selection off (line 351) and transferred the old trial selection to the updated "varargin" (line 368).


robertoostenveld:
looks good to me. Should the same be done in ft_multiplotTFR?

juliankeil:
ft_multiplotTFR does seem to work with single trial data, as far as I can tell.

PR #494, opened 2017-08-20
enh: improved output handling of ft_regressconfound

StolkArjen:
Following up the discussion at PR 359 (https://github.com/fieldtrip/fieldtrip/pull/359), ft_regressconfound now deals with the output differently in order to avoid having incompatible fields in the output data structure. Previously, specifying cfg.beta = 'yes' (the default) would break at using cfg.trials in ft_selectdata during a follow-up step as the beta field does not contain trials. Now, and similarly to ft_mvaranalysis, the beta and data (residuals) fields are mutually exclusive, and cfg.output can be specified to have output other than the default residuals:

%   cfg.output      = 'residual' (default), 'beta', or 'model'.
%                     If 'residual' is specified, the output is a data
%                     structure containing the residuals after regressing
%                     out the in cfg.reject listed confounds. If 'beta' or 'model'
%                     is specified, the output is a data structure containing
%                     the regression weights or the model, respectively.

### DIFF
--- a/ft_regressconfound.m
+++ b/ft_regressconfound.m
@@ -22,19 +22,15 @@ function [data] = ft_regressconfound(cfg, datain)
 %                     are to be rejected (default = 'all')
 %   cfg.normalize   = string, 'yes' or 'no', normalization to
 %                     make the confounds orthogonal (default = 'yes')
-%   cfg.statistics  = string, 'yes' or 'no', whether to add the statistics
-%                     on the regression weights to the output (default = 'no')
-%   cfg.beta        = string, 'yes' or 'no', whether to add the beta
-%                     values as a field to the output (default = 'yes')
-%   cfg.model       = string, 'yes' or 'no', whether to add the model to
-%                     the output (default = 'no')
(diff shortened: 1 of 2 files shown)


ft_regressconfound.m
```diff
@@ -88,15 +93,14 @@
 datain = ft_checkdata(datain, 'datatype', {'timelock', 'freq', 'source'}, 'feedback', 'yes');
 
 % ensure that the required options are present
-cfg = ft_checkconfig(cfg, 'required', {'confound'}, 'renamed', {'Ftest','ftest'});
+cfg = ft_checkconfig(cfg, 'required', {'confound'}, 'renamed', {'Ftest','ftest'}, 'deprecated', {'beta','model'});
```
robertoostenveld: since cfg.beta and cfg.model are not supported any more, they should be flagged as forbidden rather than deprecated. 

They would be deprecated if you would have added code for backward compatibility, mapping the old options onto the new one.

StolkArjen:
Grazie mille

PR #499, opened 2017-08-23
added a defaultId function to private and replaced ft_warning 

robertoostenveld:
by warning(defaultId, ...). Idem for ft_error. See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3335#c4


robertoostenveld:
the defaultId is a good function, but replacing all ft_warnings by a normal warning removes the warning-once functionality. Better only replace the ft_error by error, since ft_error does not do anything special. And ft_warning should only be replaced by warning in the qsub and the external functions.

PR #501, opened 2017-08-24
get rid of ft_error in qsub, peer and engine

robertoostenveld:
http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3336


robertoostenveld:
this is the first pass at it, only getting rid of ft_error in engine and peer, and adding the defaultId function. 

For qsub I already removed ft_error and ft_warning in relation to bug 3335.

PR #504, opened 2017-08-25
make realign functions more consistent

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1830

This is the first set of changes, there are more to come (changing the 'interactive' and 'fiducial' cfg.method handling), but those would make the code incompatible and therefore I decided to push this now already. @simonhomoelle needs it for his tutorial.


robertoostenveld:
@schoffelen this one does not need a review yet.

robertoostenveld:
@simonhomoelle the ft_meshrealign is now available for the tutorial

PR #506, opened 2017-08-28
bugfix: dissimilar structure error when loading dicom series

StolkArjen:
This fixes an (annoying) bug that prevented the reading of several dicom series. However, it's a fix in code from an external toolbox (freesurfer), and I don't see any way of fixing it somewhere else. What do you think, RO and/or JM?

The problem: when a dicom file in a series doesn't have exactly the same fields as any other file in that series, matlab will throw a 'Subscripted assignment between dissimilar structures' error. See also report here: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1772 
The consequence is that this dicom series is lost for use with fieldtrip.

The fix: a subfunction in load_dicom_series and load_dicom_fl (checkstructsim) that removes fields from files not present in the first file of the series.  Ideally, this fix was implemented internally, as it may get lost when the external (freesurfer) code is updated. But I don't see how.

### DIFF
--- a/external/freesurfer/load_dicom_fl.m
+++ b/external/freesurfer/load_dicom_fl.m
@@ -15,7 +15,9 @@ function [vol, M, dcminfo, mr_parms] = load_dicom_fl(flist)
 %
 % Does not handle multiple frames correctly yet.
 %
-
+% Bugfix: added feature that ensures similar structures across a dicom 
+% series by removing fields not present in the first file of the series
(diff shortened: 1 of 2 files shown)


robertoostenveld:
I agree that the easiest and possibly only solution is to change it in fieldtrip/external. I think that the freesurfer code is not 100% original anyway.

StolkArjen:
Ok, thanks

PR #513, opened 2017-09-04
fix: fixed missing querymethod in mask2label functionality

StolkArjen:
test_ft_volumelookup passed


robertoostenveld:
thanks for the quick fix

PR #514, opened 2017-09-05
Messy pull request (given the confusing commit messages)

schoffelen:
that makes ft_preproc_polyremoval correctly nan-transparent.
I also updated test_bug2372 to do correct assertions for the cases that depend on ft_preproc_polyremoval.

Please ignore the commit messages that are dated september 4, they somehow ended up in my master branch, but I thought to have pushed them only to another branch. I reverted these commits, and the current PR therefore seems clean in terms of files affected


schoffelen:
Ah, it seems the date information is lost for the commit messages. Ignore all of them, apart from the 4th one.

robertoostenveld:
I pulled the changes to my computer, did a diff, and also here it all seems clean. So I'll merge the PR.

schoffelen:
merci bien

PR #517, opened 2017-09-08
simbio EEG forward problem

mcpiastra:
the function ft_prepare_vol_sens.m will now check if the transfer matrix (TM) exists or not. if TM does not exist, ft_prepare_vol_sens.m will compute TM by calling sb_transfer.m; if TM exists, ft_prepare_vol_sens.m will copy TM to the headmodel in output, making sure that the electrode ordering is preserved. Other minor corrections: checking the electrode positions in sb_find_elec.m and allow for general meshes in ft_inside_vol.m


schoffelen:
Thanks for the reminder Simon, and MC for the fix!

PR #518, opened 2017-09-11
FIX, moved onesvec assignment out of the if control block otherwise t…

KristijanArmeni:
…he statement cannot be reached in cases when cfg.demean is set to false and results in error. @schoffelen


schoffelen:
thanks.

PR #522, opened 2017-09-15
clean up the plotting functions and make data handling more consistent

robertoostenveld:
This should make them more robust, faster, and deal with issues such as selection of data using cfg.trials. 

See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3345

Since this involves quite a some changes (which I have already implemented in a private branch at the moment), I will iteratively apply these changes to the PR here and try test them sequentially.

### DIFF
--- a/contrib/spike/ft_spike_jpsth.m
+++ b/contrib/spike/ft_spike_jpsth.m
@@ -16,7 +16,7 @@ function [stat] = ft_spike_jpsth(cfg, psth)
 %   [jpsth] = ft_spike_jpsth(cfg,psth)
 %
 % The input PSTH should be organised as the input from FT_SPIKE_PSTH,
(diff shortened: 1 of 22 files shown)


robertoostenveld:
the first 4 commits up to 4681c71 relate to actual desired functionality. After that I started running the dashboard and fixing regression errors introduced by the change that avg and trial are now mutually exclusive.

robertoostenveld:
from commit c80b872 onward it is again specific to the desired functionality changes.

robertoostenveld:
Running the test batch one more time, I found some more regression errors that I fixed in the previous 2 commits. The batch almost completely runs through, except for test_tutorial_clusterpermutationtimelock which has an error due to data.individual not being handled by ft_topoplotER yet. That can wait for a bit.  

I will merge this set of changes into the master, so that I can start with a clean PR to follow up with the changes in the actual plotting functions.

PR #523, opened 2017-09-17
enh: added support for 4x6 grids

StolkArjen:


StolkArjen:
This pertains to a non-default functionality in electrode grid snapping, which is in the test phase (hence, undocumented).

PR #526, opened 2017-09-18
FIX+OCTAVE: use ft_warning with parenthesis, because without them Octave throws a syntax error

nno:
Recent tests using Travs of FieldTrip results in failures (e.g. https://travis-ci.org/CoSMoMVPA/CoSMoMVPA/jobs/276506018)

This is because expressions such as 
```
ft_warning backtrace on
```
give an error in Octave when used inside a function. Using the functional form, as in 
```
ft_warning('backtrace','on');
```
works fine.

Commit e1519009c97585bfcb2c923590e7bcd6ff826c96 introduced the use ft_warning without parenthesis. 

This PR partially reverts that commit.

### DIFF
--- a/fileio/private/ft_hastoolbox.m
+++ b/fileio/private/ft_hastoolbox.m
@@ -495,9 +495,9 @@ if isdeployed
   status = 1;
 elseif exist(toolbox, 'dir')
   if ~silent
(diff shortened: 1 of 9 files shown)


robertoostenveld:
thanks for catching this.

PR #527, opened 2017-09-20
Firws

matsvanes:
introduced filter direction "onepass-reversal-zerophase", a combination of "onepass-reverse" and "onepass-zerophase".


schoffelen:
thanks.

Issue #509, opened 2017-08-30
Implement SSVEP tutorial and simulation function


robertoostenveld:
As discussed with Stefan and Malina, we should extend and upgrade the seven example to a tutorial. As part of that, we should implement an ft_steadystatesimulation function.

robertoostenveld:
how about this as a function interface for the end-user to simulate data?

```
% Use as
%   data = ft_steadystatesimulation(cfg)
% where cfg is a configuration structure that should contain
%   cfg.fsample   = scalar, sampling frequency in Hz (default = 512)
%   cfg.duration  = scalar, trial length in seconds (default = 4.56)
%   cfg.baseline  = scalar, baseline length in seconds (default = 0)
%   cfg.ntrials   = integer N, number of trials (default = 320)
%
% Furthermore, the behaviour of the stimuli  is specified with
%   cfg.stimulus1.mode = 'periodic' or 'transient' (default = 'periodic')
%   cfg.stimulus2.mode = 'periodic' or 'transient' (default = 'transient')
% 
% If the stimulus is periodic, the following options apply
%   cfg.stimulus1.onset       = in seconds, first stimulus relative to the start of the trial (default = 0)
%   cfg.stimulus1.onsetjitter = in seconds, max jitter relative to the start of the trial (default = 0)
%   cfg.stimulus1.isi         = in seconds i.e. for 10Hz you would specify 0.1 seconds (default = 0.1176)
%   cfg.stimulus1.isijitter   = in seconds, max jitter relative to the previous stimulus (default = 0)
%   cfg.condition             = 1*Ntrial vector with stimulus codes for each trial  (default = repmat([1 2 3 4], 1, 80))
%   cfg.gain                  = 1*Ntrial vector with gain for each for each trial (default = repmat([1 2 3 4], 1, 80))
%
% If the stimulus is transient, the following options apply
%    cfg.stimulus2.number      = scalar M, how many stimili are to be presented per trial
%    cfg.stimulus2.onset       = in seconds, first stimulus relative to the start of the trial (default = 1.0)
%    cfg.stimulus2.onsetjitter = in seconds, max jitter relative to the start of the trial (default = 0.2)
%    cfg.stimulus1.isi         = in seconds (default = 0.9)
%    cfg.stimulus2.isijitter   = in seconds, max jitter relative to the previous stimulus ((default = 0.2)
%    cfg.condition             = 1xM vector with stimulus codes (default = [1 1 2 2])
[...]

PR #528, opened 2017-09-20
implement ft_steadystatesimulation 

robertoostenveld:
to complement the new SSVEP tutorial


robertoostenveld:
see also issue #509

robertoostenveld:
Along the way I am also trying to improve support for the MATLAB table object; it would be really handy if the trialinfo (and trl in the trial function) could be represented as table rather than as array.

robertoostenveld:
Please note that there are more changes to be expected. But this merge allows @StefanWiens to already have a look at it in the fieldtrip release. I will follow this further up in issue #509

PR #529, opened 2017-09-24
enh: copy over unit field from input

StolkArjen:
Units are hard to be determined in iEEG. In case a unit field is already present, it's better to copy it over than risk any misdetermination by ft_datatype_sens.


robertoostenveld:
should we also do this for "type" (which is e.g. CTF151, EEG, etc),  "chantype" (cell array), and for ""chanunit" (also cell-array)?

StolkArjen:
Sounds right. Will add those.

PR #531, opened 2017-09-25
enh: added option to keep duplicates (default = no)

StolkArjen:
This required a change to private/append_common.m, in which cfg is also forwarded to ft_appendsens.

### DIFF
--- a/ft_appendsens.m
+++ b/ft_appendsens.m
@@ -265,7 +265,7 @@ if hascoilori
 end
 
 % keep the following fields only when identical across inputs
-if haslabelold && all(strcmp(labelold{1}, labelold)) % labeloldmatch
+if haslabelold && all(isequal(labelold{1}, labelold{:})) % labeloldmatch
   sens.labelold = labelold{1};
 end
 if haschanposold && all(isequal(chanposold{1}, chanposold{:})) % chanposoldmatch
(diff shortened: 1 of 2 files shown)


StolkArjen:
This PR doesn't change the default behavior of ft_appendsens, except that it fixes a bug. In that respect, it's a safe pull.

However, there's one detail that needs your attention. In order for cfg.keepduplicates to reach ft_appendsens, a change in append_common was needed. That is, from ..
if hasgrad, data.grad = ft_appendsens([], grad{~cellfun(@isempty, grad)}); end
.. to ..
if hasgrad, data.grad = ft_appendsens(cfg, grad{~cellfun(@isempty, grad)}); end
.. adding the cfg as input.  

As far as I can tell, the cfg information trickling down to ft_appensens is not represented in its output. In fact, I just noticed ft_appendsens has no postamble. It does have ft_preamble trackconfig though. If you think this does not add the cfg information anywhere, then please pull in this PR. If you think the cfg information is stored but that that's meant to be so, then also pull this in this PR. ;)

StolkArjen:
The bug fixed by this PR (at line 263: all(isequal(labelold{1}, labelold{:}))) is a blocker. It'd be great for that reason if this PR could be expedited.

StolkArjen:
superseded by https://github.com/fieldtrip/fieldtrip/pull/637

PR #533, opened 2017-09-26
enh - added option to NaN edge artifacts for hilbert approach (defaul…

roemervandermeij:
…t is no)

I've added a flag to replace edge artifacts by NaNs (when using an FIR type filter). This is fully equivalent to leaving data points NaN in specest_mtmconvol when the wavelet is not fully immersed in the data. (Edge artifacts are caused after all, when the FIR filter is not fully immersed in the data).

The default is currently 'no', as it is used by others, e.g. external toolboxes (SPM?).

(also updated the documentation to reflect the changes to filtering a while ago)

### DIFF
--- a/specest/ft_specest_hilbert.m
+++ b/specest/ft_specest_hilbert.m
@@ -67,8 +67,8 @@ if isempty(fbopt)
 end
 
 % edge artifacts from filters are only exactly defined for FIR filters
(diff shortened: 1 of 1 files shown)


robertoostenveld:
I have no idea why this PR was open for such a long time, seems a clean contribution to me. Thanks.

PR #543, opened 2017-10-05
FIX - Completed the bivariate_common imple,entation, also for labelcm…

robertoostenveld:
…b. Fixed the plotting of markers and highlights.


robertoostenveld:
this is related to http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3359

The implementation of bivariate plotting requires some additional testing by @schoffelen

PR #550, opened 2017-10-18
enhancement: changed 'sum' to 'nansum' in ft_artifact_zvalue

johanna-zumer:
This is a useful change to me, as some of my data already had other artifacts replaced with 'nan' and I wanted to run additional artifact selection.  

Note: the 'numsmp' is not altered: perhaps it should be?  If so, I can change that and then resubmit the change.

[Finally my first pull request to fieldtrip!  Sorry I didn't used a branch, but wanted to keep it simple at first.]


schoffelen:
Brilliant, thanks! Don't let it happen again.

robertoostenveld:
👍

PR #551, opened 2017-10-20
FIX & ENH - ft_redefinetrial

dmatthes1982:
FIX - addition of rollback_provenance  during the last change of this file (in line 313) led to corrupted cfg.previous in the output data, since the variable dataold in line 358 does not exist, when cfg.trl was not defined in the initial function call →  the existence of dataold will be additionally checked in line 356

ENH – trial markers from data.trialinfo will be added to the new trial definition in line 304 → so trialinfo can be removed from input data before the recall of ft_redefinetrial and this avoids a warning message


schoffelen:
Sorry for picking up on this only now. Can I ask you to initialize the newtrl in the presence of a trialinfo field to zeros(0,size(data.trialinfo,2)+3), and accordingly fill the newtrl with the full trialinfo matrix?

Reason: data.trialinfo can be multi-column

schoffelen:
Thanks!

robertoostenveld:
this commit causes a regression error in the test script test_bug3267.m

The test code there uses ft_appenddata to append two exact copies. Cutting trials into shorter pieces now results in the first  half of the data being used twice.

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3379

PR #554, opened 2017-10-26
Bml read

abush84:
Dear Fieldtrip developers, 

I recently started a postdoc at the Brain Modulation Laboratory of the University of Pittsburgh, in which I’ll be analyzing ECoG and MER data. We decided to base our pipeline on Fieldtrip as it appears to be a well-designed and active project and has most of the analysis we are interested in already implemented. Hence, I’ve been extending ft_read_header/data and ft_filetype to be able to load our files, namely, Blackrock, NeuroOmega and Audio files. I also added a new ‘chantype’ option to ft_prerpocessing and ft_read_header, which allows to select the channel before loading the header. Finally, I extended ft_appenddata to be able to append in the ‘time’ dimension.

There are some things in the code I hope to improve in the near future, but didn’t want to delay any more the pull request. Hope you find these modifications useful. 

Best regards,
Alan Bush


schoffelen:
Thanks for your efforts, Alan.

Would it be possible to split the various commits into more logical chunks?

That is, I am not sure whether we should include the recompiled files (perhaps @robertoostenveld can comment on this), but if they are pulled, I think this should go into a separate PR.

Also I am not sure about the changes to ft_appenddata. I did not read in detail, but I came across some code that makes me think that you have a different idea about what 'isequalfreq' means than what is intended. Either way, I think that this should be separated from the fileio contributions in this PR.

Regarding the fileio contributions, it seems there are several file formats at stake here. I have a developers branch on schoffelen/fieldtrip, called blackrock, which is a bit ahead of the stuff in the current fieldtrip main branch. I think it would be good to coordinate your changes with the improvements (at least I hope they are improvements), which are already in that branch.

The blackrock stuff relates to http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2964
I am not actively pursuing it at the moment, but I wouldn't mind coordinating things a bit, to get things coordinated by the different people that are interested at pushing this.

schoffelen:
Oh, and as a side note, we try and be backward compatible for matlab version up to 5 years old, so indeed I am afraid that using the "string" data type as of yet is forbidden ...

abush84:
Hi Jan,

I had to recompile the MEX files to be able to go through the ECoG tutorial
(http://www.fieldtriptoolbox.org/tutorial/human_ecog) on macOS 10.12.6.
Others in the lab had the same issue. I just submitted a separate PR with
only the replaced MEX files.

Regarding ft_appenddata, 'isequalfreq' is a variable I defined in order to
assert that the data being merged has the same sampling frequency. If
you've used this variable name with another meaning elsewhere, I can change
it to something else, but the variable wasn't defined in ft_appenddata.

I'll split the fileio contributions into different PRs, for NeuroOmega,
audio files and BlackRock.

Alan

schoffelen:
Great, thanks! Would you mind PR'ing the blackrock stuff into my blackrock branch, rather than right away into the fieldtrip main tree. I think it would be good to work on it a bit more, perhaps with Andi's input.

schoffelen:
thanks for the clarification w.r.t. ft_appenddata. I only read superficially, so I might have been a bit rash in my judgement.

I was wondering about the use case for this functionality. I think that in general it would be dangerous to allow 'generic' users to have this functionality, because it would allow them to glue stuff together which should not go together. I guess, that your implicit assumption here is that the user (which is often somebody who does not deeply understand what (s)he is doing) has 2 data structures, where the last sample of the first set is temporally adjacent with the first sample of the second set. This in general is not the case, and even if it were the case, it's very easy to violate the continuity across samples (e.g. trying to glue stuff together after separate filtering of the individual data segments).

schoffelen:
I suspect that you have a specific user scenario, where the acquisition software has split the raw data in separate data files, while guaranteeing their continuity. The MEG-community faces a similar challenge, and so far we have solved it in the low-level reading routines.
I don't know how the metadata is represented in this respect in the blackrock folders, but I could imagine that something similar could be achieved here. I think this would be a safer solution than opening up the possibility for many people to do strange things (and later blame the software for their b...s... results).

abush84:
Ok, so I did 4 PRs:

bml-mexmaci64: recompiled the mex files for spm8, spm12 and bci2000 for
macOS 10.12.6
bml-neuroomega: support for NeuroOmega files
bml-audiowav: support for audio waveform (.wav) files
bml-blackrock: support for blackrock/trellis ns2/ns5 files. I cloned this
from schoffelen/fieldtrip:blackrock, added my code and did PR to
schoffelen/fieldtrip:blackrock and, by mistake, also to fieldtrip:master.

Tested the code in Matlab 2015aSP1, I don't have any older version. I'm
also relatively new to matlab (coming from python and R) so I don't really
know what functions are from which version.

Regarding ft_appenddata, I understand your point, so I haven't done a PR
for that. The motivation was mainly for neuroomega files that come chunked
in contiguous ~100Mb files, and need to be concatenated. I thought it was
the obvious place to implement this functionality, but its true that its
not 'fool proof'.

Alan

PR #556, opened 2017-10-26
Bml neuroomega

abush84:
Support for NeuroOmega .mat files. Tested on Matlab 2015a

### DIFF
--- a/fileio/ft_filetype.m
+++ b/fileio/ft_filetype.m
@@ -1163,6 +1163,10 @@ elseif filetype_check_extension(filename, '.mat') && filetype_check_header(filen
   type = 'ced_spike6mat';
   manufacturer = 'Cambridge Electronic Design Limited';
   content = 'electrophysiological data';
+elseif filetype_check_extension(filename, '.mat') && filetype_check_header(filename, 'MATLAB') && filetype_check_neuroomega_mat(filename)
+  type = 'neuroomega_mat';
+  manufacturer = 'Alpha Omega';
+  content = 'electrophysiological data';
 elseif filetype_check_extension(filename, '.mat') && filetype_check_header(filename, 'MATLAB')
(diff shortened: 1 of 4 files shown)


schoffelen:
Looks like a clean contribution to me. I don't have my own data to test this with, so let's just pull this and see whether our battery of test functions handles this well.

As a side node (regarding startsWith and endsWith backward compatibility: we have a setup that allows for drop-in replacement for newer functionality in the folders e.g. fieldtrip/compat/matlablt2016b. I recently made a startsWith, to allow this function to be called from older matlab versions. We may consider adding some other functions as well.

robertoostenveld:
The way of handling of channels with different sampling rates seems to be different than for other data formats (mainly EDF). That might be problematic, since the API (the way the code is to be called) would then be different for EDF files and for Omega files. 

I think that we should make a test script that shows how it works for both EDF and Omega. Could you please share an example (small) dataset for testing, or preferably two: one with homogenous sampling and one with variable sampling frequencies?

abush84:
```
cfg=[];
cfg.dataset='LT1D9.975F0001.mat';
dat = ft_preprocessing(cfg);
```
Prints out a table with available channels, chantypes, Fs and then throughs an error. 

```
cfg=[];
cfg.dataset='LT1D9.975F0001.mat';
cfg.chantype={'macro','micro'};
dat1 = ft_preprocessing(cfg);
```
Loads the specified channel types. 

Here is a small example dataset.  All datasets have channels with different sampling frequencies.  

[LT1D9.975F0001.mat.zip](https://github.com/fieldtrip/fieldtrip/files/1422329/LT1D9.975F0001.mat.zip)

robertoostenveld:
macro and micro are currently not valid channel types. See the implementation in ft_chantype.

PR #557, opened 2017-10-26
Bml audiowav

abush84:
Added support for audio waveform (.wav) files


schoffelen:
Hmmm, this now has some conflicts which need to be looked into. I'll check soon.

By the way,  I noticed the use of uniquetol, which seems to exist only from matlab2015a or so. I don't know whether you used it elsewhere as well, but this might lead to compatibility issues.

robertoostenveld:
The code already supports audio files in *.wav format: please search the ft_read_xxx functions for "riff_wave". I guess the documentation/help should be improved. You can send a PR for  the documentation.

PR #571, opened 2017-11-06
Added more robust method for calculating dimensions of iEEG grids

smg211:
To enhance the recently added cfg.pairmethod = 'label' functionality of ft_electroderealign (PR #505 ), I created a more robust function for calculating the dimensions of sets of iEEG electrodes (grids, strips, depths) that takes into account electrode labels and positions. I have tested the function on over 50 different sets of iEEG electrodes, including depths, strips, and grids (both intact grids, and grids that have had several electrodes cut out). This function can also detect the dimensions of L-shaped grids, which are rarely used and still cannot be handled by warp_dykstra2012, but at the least, feedback is provided to the user for these cases.

Because determining grid dimensions is also a necessary part of write_bioimage_mgrid, I created a private function, determine_griddim, that is called by both warp_dykstra2012 and write_bioimage_mgrid, so that any updates to the method of determining grid dimensions can be centralized to this one private function.

### DIFF
--- a/fileio/private/write_bioimage_mgrid.m
+++ b/fileio/private/write_bioimage_mgrid.m
@@ -79,102 +79,10 @@ for g = 1:ngrids % grid loop
   fprintf(fid, '#Dimensions\n');
   
   % determine grid dimensions
-  if isequal(numel(Grid2Elec{g}), 256)
-    GridDim(1) = 16; GridDim(2) = 16;
-  elseif isequal(numel(Grid2Elec{g}), 64)
-    GridDim(1) = 8; GridDim(2) = 8;
-  elseif isequal(numel(Grid2Elec{g}), 48)
-    e6 = elec.elecpos(match_str(elec.label, [GridDescript{g} num2str(6)]),:);
-    e7 = elec.elecpos(match_str(elec.label, [GridDescript{g} num2str(7)]),:);
-    e8 = elec.elecpos(match_str(elec.label, [GridDescript{g} num2str(8)]),:);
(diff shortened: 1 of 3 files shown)


robertoostenveld:
@StolkArjen can you review and let me know when this is ready to merge?

StolkArjen:
Sure, will do

StolkArjen:
This is a significant improvement over the previous implementation for determining electrode grid dimensions. Please pull it in as is, and I will copy over bits to the Hermes 2010 snapping technique that is currently also under development. 

One question, though: for consistency with the other function names in the private folder, should we maybe rename 'determine_griddim' into something like 'get_griddim', or 'find_griddim'?

schoffelen:
There's a typo in the help of determine_griddim: line 4 

I think that the functionname is fine, but I'd consider making this a subfunction, unless the code needs to be called from somewhere else as well. Is that the case?

StolkArjen:
I'll fix the type in a next PR. determine_griddim will be called by 3 different functions in the end.

StolkArjen:
Assuming the name is fine then, can you pull this in so I can implement this functionality in another project?

robertoostenveld:
the typo in the help is not a show stopper. It would be good to get more documentation on the wiki.

StolkArjen:
https://github.com/fieldtrip/fieldtrip/pull/591 builds on the functionalities created by Sandon. Per the overall documentation, we're planning on creating a FAQ on the wiki that explains how to import Bioimage Suite mgrid files and that also addresses the functionalities contributed by this commit.

PR #574, opened 2017-11-09
Add REST in FieldTrip

youlong1230:
Add Reference Electrode Standardization Technique (REST) in FieldTrip functions (ft_preprocessing.m, ft_preproc_rereference.m, preproc.m).

REST is a method to standardize a reference of scalp EEG recordings to a point at infinity. More details about REST can be seen at http://www.neuro.uestc.edu.cn/rest/


robertoostenveld:
thanks for this PR. Could you send an example script that shows how an end-user would do this? E.g. starting from this EEG tutorial http://www.fieldtriptoolbox.org/tutorial/continuous

youlong1230:
Ok, I will send an example later.

youlong1230:
Example script:

clear all; clc;
ft_defaults;

load F:\GitHub_work\test_data\Sub02_leadfield.mat % load leadfield

cfg = [];
cfg.dataset     = 'F:\GitHub_work\test_data\SubjectEEG\subj2.vhdr';
cfg.reref       = 'yes';
cfg.refmethod     = 'rest';     %  if select 'rest','leadfield' is required.
cfg.leadfield = lf;            
%              The leadfield can be a matrix (sources X channels)
%              which is calculated by using the forward theory, based on
%              the electrode montage, head model and equivalent source
%              model. It can also be the output of ft_prepare_leadfield.m
%              (e.g. lf.leadfield) based on real head modal using FieldTrip.
% refchann = [];
% for i = 1:60
%     refchann{i} = num2str(i);
% end
cfg.refchannel     = {'all'}; % vector with indices of the selected channels 
                              %  (re-referenced channels), or 'all'.
data_eeg        = ft_preprocessing(cfg);

youlong1230:
@robertoostenveld 

Here is the sample example script for REST:

clear all; clc;
ft_defaults;

load F:\GitHub_work\test_data\Sub02_leadfield.mat % load leadfield

cfg = [];
cfg.dataset = 'F:\GitHub_work\test_data\SubjectEEG\subj2.vhdr';
cfg.reref = 'yes';
cfg.refmethod = 'rest'; % if select 'rest','leadfield' is required.
cfg.leadfield = lf;
% The leadfield can be a matrix (sources X channels)
% which is calculated by using the forward theory, based on
% the electrode montage, head model and equivalent source
% model. It can also be the output of ft_prepare_leadfield.m
% (e.g. lf.leadfield) based on real head modal using FieldTrip.
% refchann = [];
% for i = 1:60
% refchann{i} = num2str(i);
% end
cfg.refchannel = {'all'}; % vector with indices of the selected channels
% (re-referenced channels), or 'all'.
data_eeg = ft_preprocessing(cfg);

schoffelen:
@youlong1230 is there any interest in finishing this? There are now 2 conflicting files, which at least need to be fixed on the end of the PR. Also, I think it would be good for robustness if at least some basic checks are performed on the leadfield, with respect to included channels etc

youlong1230:
@schoffelen The two conflicts were resolved. In the subfunction of ft_preproc_rereference, i.e. 'rest_refer',  the demension of leadfield will be checked to make sure the no. of channels of leadfield matrix and  re-referencing EEG data are equal.

schoffelen:
there is no guarantee that the channel order in the leadfields is the same as in the data.
Fieldtrip computed leadfields have a .label field these days, so the check can be easily performed.

Also, it would be good to have a test function that demonstrates the functionality (and ideally checks for correct algorithmic implementation). This test function should go into fieldtrip/test/.

youlong1230:
@schoffelen That's great. And I think it can update code to check the channel order according to the .label field in the fieldtrip computed leadfield. I will update the funtion ft_preproc_rereference as soon. Thanks!

youlong1230:
@schoffelen I updated the code according to your comments. Thanks!

schoffelen:
Thanks!. We're getting there.

Could you please update the contributed code such that the number of spaces per tab is 2, rather than the default 4? This can be specified in the matlab preferences, and if you then do 'select all', followed by 'smart indent', you're already there.

Also, we cannot rely on externally hosted data for the test functions. Are your datafiles big? If so, consider pruning them to only a few MB (e.g. reducing the length of the EEG data, and the number of source locations in the leadfield), and send the data to us. 

Also, in the test script, is there any way in which the effect of the re-referencing can be evaluated?

youlong1230:
it's done.

schoffelen:
Are you sure the indentation worked out well. Now it looks even messier than before.

youlong1230:
I am sure the rest code is correct, and the test figure just shows the difference between two reference methods, 'median' reference and rest reference. The original reference of the example data is avg. In addition, according to my experience, sometimes, the wave of rest may look like the wave of avg, but they are different reference methods and there is only a little difference between waves of rest and avg, visually. 

For more about rest, see http://www.neuro.uestc.edu.cn/rest/
or the reference: Yao D (2001) A method to standardize a reference of scalp EEG recordings to a point at infinity. Physiol Meas 22:693?11.

youlong1230:
The smoother blue line is rest.  In addition, the leadfield is calculated by fieldtrip using standard BEM (dipoli method). Thanks!

schoffelen:
Once again. Please fix the indentation of the tabs. When I compare the old and new versions (submitted by you) yours have 4 spaces per tab, our code guidelines require 2 spaces per tab.
Please refer to the 'use consistent spacing' section on http://www.fieldtriptoolbox.org/development/guideline/code/

schoffelen:
also, in the test function, please remove the call to your local F drive, because we cannot access it. And also, include a snippet of code to read in or create the leadfield

youlong1230:
I am really sorry for misunderstanding your comments, it's my fault. Now I updated and double checked the codes. Many thanks!

schoffelen:
Thanks. That looks much better indeed! I have 2 final comments (sorry to keep bothering you about this, but we aim at the best possible quality of code in terms of consistency and maintainability).

1) why are there now both a 'set' and an 'fdt' file? Are both needed? Would it be possible to convert the data into a fieldtrip data structure, and save it as a mat file on your end? Ideally, this file should be named with an understandable name ('sub-01' is too generic). I suggest to save it as test_pr574.mat, and that it contains both the 'data' and the 'lf' variables. 

2) right now, the ft_preproc_rereference is doing the struct-to-matrix conversion for the leadfield, if it is detected. I suggest to move this conversion into preproc, where the order of the channels is checked anyway. In this way, the labour is cleanly divided between bookkeeping (channel selection, re-formatting the data), and actual computation (rereferencing).

youlong1230:
Get it!
The data is EEGLAB format, and 'set' and 'fdt' files are both needed. Anyway, I will saved it as a mat file loaded by fieldtrip later.
I will update codes and example files as soon, according to your sugggestions.

PR #578, opened 2017-11-13
neuroomega events

abush84:
Added support for neuroomega events in ft_read_event.m
Also solved a minor issue in ft_read_data for neuroomega_mat and placed neuroomega_mat sections of ft_read_data/header in alphabetic order.


schoffelen:
Thanks! Would you by any chance have a small example data file, that we can add to our collection of test data?

PR #580, opened 2017-11-14
fix: removed pre/postambles

StolkArjen:
as discussed in http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1110


robertoostenveld:
👍

PR #588, opened 2017-11-22
add accurate coil defs for NM T3 mags

### LINKED ISSUE #586: missing accurate coil defs for Elekta T3-magnetometers
The `mne` matlab toolbox bundled with FT is missing the 'accurate' integration points for Elekta type T3 magnetometers. Compare this

https://github.com/fieldtrip/fieldtrip/blob/8e09fc3ed4ca3bb8d4b8216a10b091ef7084b236/external/mne/coil_def.dat#L148

to  [`mne-python`](https://github.com/mne-tools/mne-python/blob/44bac44efbc85f234c1cc642fbb6b6c9524fc101/mne/data/coil_def.dat#L157) and [`brainstorm3`](https://github.com/brainstorm-tools/brainstorm3/blob/ae010b5d8dc820b473bdbaa291a67e461ad4a355/toolbox/io/private/coil_def.dat#L148) (FYI: @ftadel)

What started me off down this route was coming across the function [`fix_mag_coil_types`](https://github.com/mne-tools/mne-python/blob/44bac44efbc85f234c1cc642fbb6b6c9524fc101/mne/channels/channels.py#L1196) (see docstring for more context). We've had a Triux since 2010, and several DACQ updates since: _our magnetometers are still defined as type T1 (3022) instead of T3 (3024)_. I believe this is the case at _all_ Elekta sites, unless the local admin has changed `collector.pos` to reflect reality. I've been in touch with Elekta, and they've started an internal process to change the header definitions for the next DACQ release. They also ran some tests, and found the potential errors in forward modeling to be insignificant at realistic source-to-sensor distances. This is consistent with both  @agramfort @larsoner et al.'s comments in their code, and @sarangnemo 's beamformer testing (CC: @britta-wstnr). FYI Elekta software does a hack by checking the magnetometer calibration values, and 'fixes' T1 to T3 if the value is above a (heuristic) threshold.

In conclusion, fixing `coil_def.dat` in FT will bring the three packages mentioned above to parity. Dealing with data files with potentially erroneous coil type defs is another matter, but nothing to get too worked up about, it seems.

cjayb:
These are copied from [`mne-python`](https://github.com/mne-tools/mne-python/blob/master/mne/data/coil_def.dat)

Closes #586


robertoostenveld:
As I just mentioned in #586 the ones from BrainStorm are more elaborate. If you agree, I'll merge #587 instead.

PR #594, opened 2017-11-24
Bug3354

schoffelen:
This PR aims to keep the sampleinfo consistent if ft_selectdata with latency selection is invoked.
In addition, it allows for sampleinfo to be kept in multi-trial timelock structures (v2017), which requires a ad hoc'ish  solution ft_datatype_timelock.


robertoostenveld:
looks good

PR #597, opened 2017-11-28
ENH - changed some lines to prevent crash, let Roemer think about whe…

schoffelen:
…ther this is now resulting in desired behavior


schoffelen:
This pertains to bug 3279. I understand the rationale in the code, but I don't know what the correct behavior in this case would be. Please review and fix if needed.

schoffelen:
This still should be followed up.

schoffelen:
Ach so. Wohin?

roemervandermeij:
For the moment NL, afterwards, TBD!

schoffelen:
Well, welcome back, and happy belated b-day!

PR #600, opened 2017-12-04
FIX - Bug 3378 - all true mask in ft_singleplotTFR

demaddis:
This aims to fix conversion of alpha/saturation masking values to all ones when using a logical mask in ft_singleplotTFR.

### DIFF
--- a/ft_multiplotTFR.m
+++ b/ft_multiplotTFR.m
@@ -14,11 +14,14 @@ function [cfg] = ft_multiplotTFR(cfg, data)
 % The configuration can have the following parameters:
 %   cfg.parameter        = field to be represented as color (default depends on data.dimord)
 %                          'powspctrm' or 'cohspctrm'
-%   cfg.maskparameter    = field in the data to be used for opacity masking of data
-%   cfg.maskstyle        = style used to masking, 'opacity', 'saturation', 'outline' or 'colormix' (default = 'opacity')
-%                          use 'saturation' or 'outline' when saving to vector-format (like *.eps) to avoid all
-%                          sorts of image-problems (currently only possible with a white backgroud)
-%   cfg.maskalpha        = alpha value between 0 (transparant) and 1 (opaque) used for masking areas dictated by cfg.maskparameter (default = 1)
+%   cfg.maskparameter    = field in the data to be used for masking of data, can be logical (e.g. significant data points) or numerical (e.g. t-values).
+%                        (not possible for mean over multiple channels, or when input contains multiple subjects
+%                        or trials)
+%   cfg.maskstyle        = style used to masking, 'opacity', 'saturation', or 'outline' (default = 'opacity')
+%                        'outline' can only be used with a logical cfg.maskparameter
+%                        use 'saturation' or 'outline' when saving to vector-format (like *.eps) to avoid all sorts of image-problems
+%   cfg.maskalpha        = alpha value between 0 (transparent) and 1 (opaque) used for masking areas dictated by cfg.maskparameter (default = 1)
+%                        (will be ignored in case of numeric cfg.maskparameter or if cfg.maskstyle = 'outline')   
 %   cfg.masknans         = 'yes' or 'no' (default = 'yes')
 %   cfg.xlim             = 'maxmin' or [xmin xmax] (default = 'maxmin')
 %   cfg.ylim             = 'maxmin' or [ymin ymax] (default = 'maxmin')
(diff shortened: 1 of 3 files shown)


schoffelen:
Hi Mathis,

Thanks for making the PR. Before merging I was just thinking over the suggested change, which in itself is straightforward and simple, and it addresses the issue you ran into, I started thinking about the overall philosophy of the code, and the expected functionality in general. This is how it often goes, somebody asks for a simple change to fix a concrete problem, and I (or usually it's Robert) start nagging about more philosophical/generic things...

Also, before I forget, once we settle on a good solution, please also add the change to ft_multiplotTFR, if applicable, because it likely supports similar functionality.

Anyway, the generic issue here is the combination of the cfg.maskparameter with cfg.maskalpha.
Very broadly, there's a 2x2 'case' here, where the maskparameter is either a matrix with logicals, or with numeric data, and maskalpha can be either 1, or something else.

Your proposed fix addresses one of the cases, i.e.:

* maskalpha~=1, and maskparameter is logical

What happens (or what is the expected behavior if:
* maskalpha == 1, and maskparameter is logical (I think all is fine here, this is the most common case I think)
* maskalpha ~= 1, and maskparameter is numeric -> I don't know what to expect here?
* maskalpha == 1, and maskparameter is numeric -> I don't know what to expect here?

I would expect that a numeric maskparameter should override the maskalpha. In the current code, this coincidentally seems to be happening if maskalpha==1, but I don't know what would happen with a numeric mask. Also, I am not sure what happens if the mask is numeric, but not scaled between 0 and 1.

In order to exhaustively address this, we typically create a little snippet of test code (prior to start fixing the functions), which investigates (and possibly demonstrates) the issue. This test-code will be a part of the PR, and will be included in our test-function battery, that we deploy on a regular basis, in order check whether our code changes don't break things elsewhere in the code base. We have the tradition to name these functions along the lines of test_bug3378, so that the code refers to something we can easily look up on bugzilla. 

In this particular case, I can reproduce your issue with something like this:

++++++++++++++++++

function test_bug3378

% WALLTIME 00:01:00
% MEM 1500mb

data = [];
data.powspctrm = randn(1,10,20);
data.time = 1:20;
data.freq = 1:10;
data.label = {'chan01'};
data.dimord = 'chan_freq_time';
data.mask = data.powspctrm>0;

cfg = [];
cfg.maskparameter = 'mask';
cfg.maskalpha = 0.5;
figure;ft_singleplotTFR(cfg, data); % your original issue, point taken

% the other cases
cfg = [];
cfg.maskparameter = 'powspctrm';
cfg.maskalpha = 0.5;
figure;ft_singleplotTFR(cfg, data); % currently throws a(n rather uninformative) error (that is: without your suggested fix)

cfg = [];
cfg.maskparameter = 'powspctrm';
cfg.maskalpha = 1;
figure;ft_singleplotTFR(cfg, data); % this works as expected

cfg = [];
cfg.maskparameter = 'mask';
cfg.maskalpha = 1;
figure;ft_singleplotTFR(cfg, data); % this works as expected
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

Ideally, we would also address the other issue in the PR. 

Would you be willing to give it a shot? (i.e. add the test function to the PR, and propose a robust and expected behaviour of the combined cfg.maskalpha/cfg.maskparameter options?

schoffelen:
One possibility would be to either use the test-statistic (e.g. T-value) where more extreme T-values are less transparent, or to use a (1-pvalue), where data points with lower p-values are less transparent. This would make the functionality consistent with what we often do with visualizations of source reconstructed data (where some functional map is plotted on top of an anatomical image).

schoffelen:
Admittedly, a numerical mask for the singleplotTFR/multiplotTFR in combination with a cfg.maskalpha~=1 does not really make sense in my opinion, so my suggestion would be to check for this combination of options explicitly in the code, and then either throw an informative error, or disregard the cfg.maskalpha (and optionally print some statement to the commandline).

demaddis:
Please check the suggested fixes. I made some corrections to the help (to clarify expected combinations of mask properties in the cfg) and implemented a robust handling of these combinations: 
- Outline masking should only be used with logical masks, nondefault cfg.maskalpha should be silently ignored
- Numeric masking should ignore cfg.maskalpha and throw a warning if it was set to a nondefault value
- Logical masking with default cfg.maskalpha = 1 should result in an unmasked plot

schoffelen:
Thanks Mathis, and sorry for the delay in getting back to you.
Your suggested changes look fine, and thanks for the test function.
Could you also check whether there are similar sections in ft_multiplotTFR and ft_topoplotTFR (or perhaps private/topoplot_common.m) which would deserve the same changes as the ones you suggest for ft_singleplotTFR? Both for the documentation section, and the mask-handling.

schoffelen:
Thanks, much appreciated.

PR #603, opened 2017-12-06
accelerate ft_artifact_threshold 

dmatthes1982:
function _preproc_ should only be called during threshold artifact estimation, if necessary (if bandpass filtering is selected). If many trials are processed and bandpass filtering is NOT selected, this little change reduces processing time significantly.


ft_artifact_threshold.m
```diff
@@ -161,7 +161,9 @@
   else
     dat = ft_read_data(cfg.datafile, 'header', hdr, 'begsample', cfg.trl(trlop,1), 'endsample', cfg.trl(trlop,2), 'chanindx', channelindx, 'checkboundary', strcmp(cfg.continuous, 'no'), 'dataformat', cfg.dataformat);
   end
-  dat = preproc(dat, channel, offset2time(cfg.trl(trlop,3), hdr.Fs, size(dat,2)), artfctdef);
+  if strcmp(artfctdef.bpfilter, 'yes')
```
robertoostenveld: how do you know that only artfctdef.bpfilter might be set yes/no?  It might also be that some other filter is being used (e.g. lowpass).

robertoostenveld: it should be fine If you would check that artfctdef is otherwise or that all fields are 'no'

dmatthes1982: Since the reference document of ft_artifact_threshold only includes these specifications

% The following configuration options can be specified
%   cfg.artfctdef.threshold.channel   = cell-array with channel labels
%   cfg.artfctdef.threshold.bpfilter  = 'no' or 'yes'
%   cfg.artfctdef.threshold.bpfreq    = [0.3 30]
%   cfg.artfctdef.threshold.bpfiltord = 4

and the code from line 112 to 115 only considers the bpfilter, I reasoned that the application of a bandpass filter is the only selectable option. But of course I can expand the request.

dmatthes1982:
I commited a more comprehensive solution including also a addition in the reference documentation.

robertoostenveld:
I think that demean, detrend, polyremoval and possibly rectify would also be valid options. I think that a better test would be to see whether the preproc specific options are set to the default. If so, the user did not specify any preprocessing.

You could add all options from private/preproc to the private/ignorefieds function, and then

preproccfg = keepfields(artfctdef, ignorefields('preproc'))

If subsequently preproccfg only contains the three bpfilter/bpfreq/bpfiltord options *and* if bpfilter=no, then you can skip. In all other cases you don't know whether you can skip (since you don't know all possible options and values). 

Just taking a step back: would it not be better to try to implement the skipping at the start of private/preproc instead?

dmatthes1982:
Based on your suggestion, I run though all preprocessing options and got the following idea:

I could do of course this operation just without the previous keepfields step

status = struct2cell(artfctdef);
if any(strcmp(status, 'yes'))
...
end

If no preprocessing step is selected no field will be by default 'yes', therefore the preproc function will be correctly skipped. It would be also possible to place this request in private/preproc around line 160. Would this be a adequate solution? And if yes, which one do you prefer?

robertoostenveld:
thanks for following this up. This is a generic solution that would not only benefit the code in ft_artifac_threshold, but also elsewhere.

The check should only be for the string-valued cfg options. It should not only deal with 'yes', but also include the possible values for cfg.hilbert (which is 'no' or something specific, but 'yes' is not one of the options).

To prevent commit conflicts, please pull the latest changes from the master branch prior to making further changes, since I have been doing some updates yesterday evening in private/preproc. And I have been making some changes that involve private/preproc in ft_artifact_threshold, ft_artifact_clip and in ft_heartrate and ft_electrodermalactivity.

PR #604, opened 2017-12-06
accelerate private function defaultID

dmatthes1982:
Calling ft_version in this function returns beside the desired path information also the fieldtrip version. This information is not used in this function, but effort to estimate it is 100 times higher compared to the path information. This small change in default ID accelerate the data processing significantly in functions where notification output is periodically generated.


robertoostenveld:
thanks, much appreciated

PR #605, opened 2017-12-09
enh: output tra field, warp_fsaverage, bugfixes, doc

StolkArjen:
- created separate function for fsaverage warping functionality (warp_fsaverage.m)
- add a tra matrix to the output of ft_electroderealign. this got lost before
- only output chanunits for present channels in ft_electroderealign. before, this field was simply copied over, containing chanunits of all original channels
- minor doc updates


StolkArjen:
Please check/accept this PR. I just resolved a conflict with 'master' that was due to a recent change to ft_electroderealign (isdir into isfolder) - grmbll.

PR #608, opened 2017-12-13
Bug3375

simonhomoelle:
Added new functionality to ft_realtime_headlocaliser:
-Visualisation of head and dewar as meshes now supported (online/offline)
-supports offline visualisation with Elekta/Neuromag with  maxfilter preprocessed data.
-Restructured the visualisation options


StolkArjen:
Sounds nice. I'm curious to see a screenshot of the dewar and head mesh!

Did you guys solve the realtime localization for neuromag too? That was a pesky issue we weren't able to solve for a long time.

robertoostenveld:
@StolkArjen: I will send you a movie that shows the new options in the interface. 

We decided that the realtime HPI localization is to be done by Elekta and that the position of the head relative to the dewar (as1x6  quaternion vector) is added to the channel data. That is consistent with how it is done offline using Maxfilter.

robertoostenveld:
the dewar should also be added to the fieldtrip repository. I propose to add them as
```
fieldtrip/template/dewar/ctf275.mat
fieldtrip/template/dewar/triux.mat
```

robertoostenveld:
the dewar files should also be documented on http://www.fieldtriptoolbox.org/template/dewar

StolkArjen:
@robertoostenveld: is the Neuromag HPI information accessible in real-time (for visualization), nowadays?

robertoostenveld:
the documentation mentions that repositioning to a previous recording is only possible for CTF. I presume that is not the case any more and that it also works for Elekta?

robertoostenveld:
I am currently reviewing the code in ft_realtime_headlocalizer.m and will push my changes

StolkArjen:
"the documentation mentions that repositioning to a previous recording is only possible for CTF. I presume that is not the case any more and that it also works for Elekta?"

In order to reposition to a previous recording, you need the current position in real-time.

robertoostenveld:
regarding difference between CTF and Elekta: I think that we should write that it works for both systems, conditional on the head position being present in the real-time data stream. We know that at the moment that condition is not yet fulfilled for Elekta, but once it is, the code (probably) works without changes. We should not have to revisit the documentation. Also getting the new Acquisition software in all Elekta labs is probably going to take years, so even if it works in Helsinki, it might not work for other labs. Hence the "conditional".

PR #611, opened 2017-12-15
[FIX] ft_plot_vector correctly draw axes when axes do not cross zero

thht:
ft_plot_vector did not behave correctly when the axes did not cross zero.
you can easily see the behavior if you plot a spectrum with ft_multiplotER and set cfg.ylim = 'minmax';
this PR fixes it for me. i have also tested the new code with a bunch of other data and it works for me. however, i have no idea whether i missed some special case or if the old behavior was, in fact, the desired one.


robertoostenveld:
I have seen weird axes in figures, but never bothered debugging it. I your judgement on this being a good fix. Thanks.

Issue #607, opened 2017-12-12
Field .sample is represented in seconds


jens-k:
This concerns Blackrock data.

https://github.com/fieldtrip/fieldtrip/blob/4fbb2b6813690f82c603b8e1b86d63af81ddfc41/fileio/ft_read_event.m#L2088

Not sure if this is intended behavior. This way event.sample is given in seconds:
https://github.com/fieldtrip/fieldtrip/blob/4fbb2b6813690f82c603b8e1b86d63af81ddfc41/fileio/ft_read_event.m#L2098

jens-k:
Mmh, the problem is that the data in the corresponding .ns* file can be sampled at different frequencies, e.g. when concurrently collecting lfp and spike data. The triggers in the .nev file, however, seem to be sampled at 30k. But I'm recording with 'Trellis Neuro' and this might differ between programs or settings?

Maybe the field shouldn't be called .sample then?

schoffelen:
Hi Jens,
There's a 'blackrock' bug in bugzilla, which already identified the different sampling rates. I have CC'ed you to it. Currently, activity on that thread is pretty low indeed, but if we gain sufficient critical mass it may move forward again.

jens-k:
Oh, sorry, I should have checked. I'll try to get my hands on some blackrock hardware soon and will try to work on it then.

PR #612, opened 2017-12-15
Fix error when reading in blackrock header

jens-k:
..if data contains only one channel


jens-k:
(this is unrelated to issue #607)

schoffelen:
No worries Jens. With respect to this PR, I am not so familiar with the table-object, and in which format the openNsX outputs the header data. I wonder why this has gone unnoticed before. Are you sure that your intended fix is robust? (e.g. the old syntax might have worked for a different openNsX version/ matlab version etc.)

jens-k:
Good point. The error occurred only with single-channel data (because then table(...) didn't receive a cell array but a character array as input, which it didn't like). Maybe no one tried single-channel data before?

I now tested the code on Matlab 2013b, 2014b, 2015b and 2016b on Linux and 2015b on Windows 7. In all of those versions the error occurred with one-channel data, in none of them with multi-channel data. The PR works for me on all of those versions.

I tested it using this code:
hdr = ft_read_header('singlechanneldata.ns2')

...and these data (20 KB):
https://www.dropbox.com/s/hjq3oginoiyp4qy/singlechanneldata.ns2?dl=0

schoffelen:
Thanks for testing! As sure as we are sure that it doesn't break on multichannel data, I believe you on your blue eyes.

PR #613, opened 2017-12-16
ENH: added support for marking electrodes on headshapes, using prespecified elec labels

richard9j:
Hi Robert & Simon,

This is functional, and can be pulled in (after review). Electrodes can be prespecified using cfg.channel (just like is done for iEEG cases). Clicking the object will place a black marker (the crosshair). Clicking a label in the list of labels will assign the label to the crosshair, and leave behind a red marker (double clicking the label removes the red marker).

cfg = [];
elec = ft_electrodeplacement(cfg, headshape)

A couple of notes:
- ft_select_point3d.m gets stuck in a loop waiting for a buttonpress to continue. As a workaround, we copied relevant bits of ft_select_point3d.m to the cb_getposition subfunction and the entire of select3d.m (from plotting/private) to the bottom of ft_electrodeplacment.m. Ideally, ft_select_point3d is fixed such that ft_electrodeplacement looks less ugly.
- another problem we foresee is lighting of the 3d object: this needs to be made adjustable otherwise sites in the shade won't be visible?

Best regards,
Richard & Arjen


StolkArjen:
@simonhomoelle: could you maybe look into this PR and see whether you think it shows intended behavior? the code refinement can be done at a later stage

StolkArjen:
This PR is currently blocking ongoing development of ft_electrodeplacement. 

In brief, the placement and label assignment of electrodes on a surface mesh (headshape) is fully operational. The code implementation is a tad cumbersome given that whole sections of ft_select_point3d were copied over into a subfunction in ft_electrodeplacement.

StolkArjen:
Thanks for resolving that merge conflict. Can you pull this in? We have newer functionalities related to the updating of the camlight as a function of viewing positions that we want to add in a new PR.

PR #620, opened 2017-12-20
Bug 3325: enhancement to representation of electrodes/sensors

smg211:
Added functionality to ft_plot_cloud to allow sensor positions to be represented as spherical meshes or 2-dimensional points. This is useful because users can now represent iEEG electrodes as spheres or points (most common in publications) with color/size scaled to some continuous functional parameter. Previously, this would have required repetitive calls to ft_plot_sens. The changes to ft_plot_cloud are also compatible with ft_sourceplot.

In the future, it may be more intuitive to rename ft_plot_cloud to ft_plot_ieeg, call this in ft_sourceplot using cfg.method = 'elec' or 'ieeg', and/or change 'cloudtype' in ft_plot_cloud (and cfg.cloudtype in ft_sourceplot) to 'elecshape' (and cfg.elecshape), similar to ft_plot_sens. However, because of the current status of Stolk et al., (under review and available on BioRxiv), we definitely don't want to make any changes that would break the functionality described in the manuscript, so I'm not sure if/when this should be done.

If the proposed structure of this commit is not ideal, we could also take the code for plotting sensors with color/size scaled (to some continuous functional parameter) that exists in ft_plot_cloud and copy it into ft_plot_sens, but I do not think ft_plot_sens is supposed to handle functional data as input.
screen shot 2017-12-20 at 11 08 30 am

### DIFF
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -30,7 +30,7 @@ function ft_sourceplot(cfg, functional, anatomical)
 %                       'surface',    plots the data on a 3D brain surface
 %                       'glassbrain', plots a max-projection through the brain
 %                       'vertex',     plots the grid points or vertices scaled according to the functional value
-%                       'cloud',      plot the data as point clouds scaled according to the functional value
+%                       'cloud',      plot the data as clouds, spheres, or points scaled according to the functional value
 %
 %
 %   cfg.anaparameter  = string, field in data with the anatomical data (default = 'anatomy' if present in data)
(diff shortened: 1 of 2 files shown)


StolkArjen:
@robertoostenveld, given that we can still accommodate changes to the manuscript, what do you think of renaming ft_plot_cloud into ft_plot_ieeg, and cfg.method = 'cloud' into 'elec'?

robertoostenveld:
I don't like this. Although it is used for iEEG data, the visualization itself is a cloud. The same visualization can also be used for other scalar data that is defined in 3D coordinates. It is also of interest to MEG virtual channel data.

StolkArjen:
Ok. Will keep it as is.

Issue #623, opened 2017-12-26
Error running ft_qualitycheck


suxpert:
Running the following code will lead to an error
``` matlab
cfg = [];
cfg.dataset = 'task1.vhdr';
[info, timelock, freq, summary] = ft_qualitycheck(cfg);
```
After several minutes, matlab complained
> Undefined function or variable 'headpos'

Then I looked into the source code, and found that this error is due to
> Reference to non-exist field 'dataset'

at line 421 of [ft_qualitycheck](https://github.com/fieldtrip/fieldtrip/blob/master/ft_qualitycheck.m#L421)

I manually add this field to `timelock.cfg`, the code can run without error, but with a strange report:

![task1](https://user-images.githubusercontent.com/1191414/34345365-53767e7e-ea29-11e7-80c4-5fc9ed643e39.png)

Did I mess up something?

suxpert:
BTW, I was using the latest version from github (2017.12.26 UTC)
> $ git pull
Already up-to-date.

StolkArjen:
Thanks for the bug report. The 'strange' output is a summary of the data and is intended to look like this. For MEG data, it would show more information such as the head position dynamics.

The bug you ran into is not intended. I think it is fixed now, as soon as the above PR (626) gets pulled in. In essence, I changed line 412 into:
[iseeg, ismeg, isctf, fltp] = filetyper(info.hdr.orig.FileName);

which worked for the few different datasets I tested.

robertoostenveld:
See fieldtrip/test/test_issue623.m for an example that uses a BrainVision EEG dataset.

PR #626, opened 2018-01-02
fix: ft_qualitycheck no longer was able to access the filename in a s…

StolkArjen:
…ubfunction

the issue was described in https://github.com/fieldtrip/fieldtrip/issues/623

### DIFF
--- a/ft_qualitycheck.m
+++ b/ft_qualitycheck.m
@@ -418,7 +418,7 @@ end
 
 % determine whether it is EEG or MEG
 try
-  [iseeg, ismeg, isctf, fltp] = filetyper(timelock.cfg.dataset);
+  [iseeg, ismeg, isctf, fltp] = filetyper(info.hdr.orig.FileName);
 catch % in case the input is a matfile (and the dataset field does not exist): ugly workaround
   [iseeg, ismeg, isctf, fltp] = filetyper(headpos.cfg.previous.dataset);
 end
--- a/ft_qualitycheck.m
+++ b/ft_qualitycheck.m
@@ -418,7 +418,7 @@ elseif nargin == 5
 end
 
 % determine whether it is EEG or MEG
-if isfield(info, 'filename') % supported as of January 2017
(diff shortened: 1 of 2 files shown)


schoffelen:
Out of curiosity, why does this change fix an issue? If I read the edit to the file, the only change occurred within a try-part of a try-catch statement, which would silently go to the catch part, if an error is encountered. Also, I am not sure whether the low-level field ...hdr.orig.FileName exists for every type of file in the input.

StolkArjen:
The catch part only works for meg given that the headpos variable does not exist / is empty for eeg datatypes. Hmm, the hdr.orig.filename existed for test data, but if you think it’s not a given, we’ll need to come up with another way of retrieving the cfg.dataset input. Any suggestions? This used to be in timelock.cfg.dataset but apparently no longer.

schoffelen:
I see. Well,  I haven't looked into this, but you changed timelock.cfg.dataset into info.hdr.orig.FileName, and I am not so sure whether this is more robust in general. 
It feels to me that in order to make these lines fully robust, we would need the equivalent of a double try/catch. In general, an error would occur, if a particular field does not exist in the input argument to filetyper, right? Why not do something like this:

if isfield(timelock.cfg,'dataset'),
filename = timelock.cfg.dataset;
elseif isfield(info.hdr.orig, 'FileName')
filename = info. blabla
elseif some other conditional statement
elseif some even other conditional statement
else
error('could not determine the filename');
end
[ ] =filetyper(filename);


Moet jij trouwens niet gaan slapen, of ben je in Europa?

StolkArjen:
Gracias for this suggestion. Also added another field to info (info.filename) to make the extraction of filename information futureproof.

StolkArjen:
En ja, 't was een latertje. ;)

schoffelen:
wil je de comment in de code nog veranderen? (van january 2017 naar january 2018), of zal ik maar mergen?

StolkArjen:
Oh hahah, hang on

schoffelen:
check!

PR #627, opened 2018-01-04
ENH: cfg.refmethod = 'bipolar'

StolkArjen:
Following a great suggestion by an anonymous referee, this PR adds the possibility to specify cfg.refmethod = 'bipolar'. As a result, a cfg.montage is created on the fly in preproc.m to create bipolar combinations between consecutive channels. This addition avoids users having to design eyesoring weight matrices (from a manuscript point of view) in order to simply combine consecutive channels.

### DIFF
--- a/ft_preprocessing.m
+++ b/ft_preprocessing.m
@@ -104,7 +104,7 @@ function [data] = ft_preprocessing(cfg, data)
 % Preprocessing options that you should only use for EEG data are
 %   cfg.reref         = 'no' or 'yes' (default = 'no')
 %   cfg.refchannel    = cell-array with new EEG reference channel(s), this can be 'all' for a common average reference
-%   cfg.refmethod     = 'avg' or 'median' (default = 'avg')
+%   cfg.refmethod     = 'avg', 'median', or 'bipolar' for bipolar derivation of consecutive channels (default = 'avg')
 %   cfg.implicitref   = 'label' or empty, add the implicit EEG reference as zeros (default = [])
 %   cfg.montage       = 'no' or a montage structure, see FT_APPLY_MONTAGE (default = 'no')
 %
@@ -672,7 +672,7 @@ if strcmp(cfg.updatesens, 'yes')
       dataout.grad = ft_apply_montage(dataout.grad, montage, 'feedback', 'none', 'keepunused', 'no', 'balancename', bname);
     end
     if isfield(dataout, 'elec')
-      ft_info('applying the montage to the grad structure\n');
+      ft_info('applying the montage to the elec structure\n');
       dataout.elec = ft_apply_montage(dataout.elec, montage, 'feedback', 'none', 'keepunused', 'no', 'balancename', bname);
     end
     if isfield(dataout, 'opto')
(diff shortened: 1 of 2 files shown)


StolkArjen:
In addition, new channel labels are automatically created such that the bipolar combination of RAM1 and RAM2 becomes RAM1-RAM2.

schoffelen:
Is there any good reason to assume that the consecutive ordering of the channels is in general meaningful enough to generate meaningful bipolars?

StolkArjen:
It's straightforward for depths, where the electrodes are ordered consecutively. For instance (example taken from the manuscript):

depths = {'RAM*', 'RHH*', 'RTH*', 'ROC*', 'LAM*', 'LHH*', 'LTH*'};
for d = 1:numel(depths)
  cfg                       = [];
  cfg.channel       = ft_channelselection(depths{d}, data.label);
  cfg.reref             = 'yes';
  cfg.refchannel  = 'all';
  cfg.refmethod  = 'bipolar';
  cfg.updatesens = 'yes';
  reref_depths{d} = ft_preprocessing(cfg, data);
end

It's true that this does not hold for situations where there is no consecutive ordering. For those situations cfg.montage may provided an answer. This current implementation is just to make bipolar re-referencing of consecutively ordered channels more accessible.

StolkArjen:
Having some afterthoughts: do you think we should define pairs on the basis of the distance between electrodes? We have recently added functionality (determine_griddim.m) to tell the dimensions of a grid/strip/depth. The functionality is robust but requires sensible labels. Alternatively, electrode pairs can be defined on the basis position/distance alone (e.g. anything within a certain radius), but this is likely less robust.

StolkArjen:
Looking at the code and thinking this out a bit more, I see a few requirements and obstacles for automating this:

- we could extend ft_prepare_montage to support bipolar rereferencing. the function would then need to make an early appearance in ft_preprocessing to create a montage that is fed down to preproc.m
- the bipolar reref functionality would either 1) use elecpos and labels for best performance on iEEG grids, or when not available 2) do consecutive channel rereferencing as the current PR already allows
- most of the ingredients for 1 are available (determine_griddim.m, create_elecpairs subfunc in warp_dykstra2012)
- one obstacle is that it should do all this for iEEG data and not scalp-EEG data, where the rules for determining adjacent electrodes in grids do not apply. it thus is iEEG specific, meaning it requires a safety check to determine it's iEEG data (much harder to implement than it sounds)

Overall, perhaps this is better solved in the context of a FAQ/example page one day as it requires quite an investment for a possibly little-used algorithm. I do think it would be useful for now to have an easy to use consecutive rereferencing option, like the one implemented in this PR. I feel it would be overkill to either create a montage for that or to loop around each and every next electrode pairing. But perhaps to make clear it's not necessarily 'bipolar', we could rename it into 'consecutive' (cfg.refmethod)?

robertoostenveld:
I think that the initial implementation is valuable as it is. The scope is limited and the functionality can be explained in one sentence (the option cfg.refmethod='bipolar' causes the data to be referenced using a bipolar scheme, taking the difference between subsequent channels).

More sophisticated referencing schemes are better explained in full documentation, i.e. on the wiki, and left for the researcher to construct him/herself as cfg.montage.

PR #630, opened 2018-01-07
further improvements for dealing with iMotions data

robertoostenveld:
see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3394
and http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3385


robertoostenveld:
this also improves the support for MATLAB tables as cfg.trl or as data.trialinfo

PR #635, opened 2018-01-12
implement improved bipolar reference

robertoostenveld:
This addresses PR #627


StolkArjen:
Thanks. Could you maybe change the 'sequential' into 'bipolar' in the documentation of ft_preprocessing as well? Currently, reads as:

cfg.refmethod     = 'avg', 'median', or '**sequential**' for bipolar derivation of sequential channels (default = 'avg')

StolkArjen:
See https://github.com/fieldtrip/fieldtrip/pull/636, which addresses this documentation change as well as another fix related to the present PR

PR #638, opened 2018-01-17
fix the ctf realtime interface

robertoostenveld:
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3185


StolkArjen:
This sounds like a pesky issue. Keep up the good work!

robertoostenveld:
I tested compilation and execution on the old acquisition computer (odin): all is fine.

I tested compilation on the new acquisition computer: all is fine. Execution could not be tested since the electronics would not start when connected to the new machine. I assume for now that it will work, and merge this with the release.

PR #641, opened 2018-01-24
ENH - changed logical order of condition checks, to avoid license leak

schoffelen:


schoffelen:
this hopefully addresses the 'license leak' of not-used toolboxes when deploying multiple jobs.

PR #642, opened 2018-01-24
enh: added xml_findstr mex-file compatible with newer matlab versions on mac

StolkArjen:
tested on matlab v2017b, mac OS Sierra

source of the mex-file: https://github.com/spm/spm12/blob/master/%40xmltree/private/xml_findstr.mexmaci64

bug reported: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3402


robertoostenveld:
please test on MATLAB 2012b  (5 years old)  and onwards and onwards, otherwise I will not accept the PR. See http://www.fieldtriptoolbox.org/development/guidelines/code#ensure_that_it_runs_on_older_matlab_versions

StolkArjen:
don't have installs of or access to versions other than 2017b on mac, so that won't work
but will double check with Guillaume to see whether he has done it. otherwise closing this bug

StolkArjen:
The SPM12 MEX-files, including the gifti and xmltree related files we keep in a separate folder in external, are suggested to be compatible with matlab versions going back to 2007a. See doc here: https://en.wikibooks.org/wiki/SPM/MATLAB

robertoostenveld:
I won't merge this pull request, as this code is maintained externally by @gllmflndn.  I did update to the latest available version from Guillaume, see #762 642

PR #645, opened 2018-01-25
Bug3397

schoffelen:
This is an attempt to generically do a check on the first input argument into an FT-function, to avoid that the user can silently get away with inputting a data objects as first argument into functions like ft_appenddata, ft_timelockstatistics etc, which then happily provide output with the first data structure omitted from the computation

### DIFF
--- a/utilities/private/ft_preamble_init.m
+++ b/utilities/private/ft_preamble_init.m
@@ -48,7 +48,7 @@ end % if nargin
 
 % check if there are fieldnames in the cfg that suggest as if the user
 % erroneously inputted a data argument
-checkdatafields = isfield(cfg, {'label' 'dimord' 'time' 'trialinfo' 'avg'});
+checkdatafields = isfield(cfg, {'cfg' 'label' 'dimord' 'time' 'trialinfo' 'avg'});
 if any(checkdatafields)
   stack = dbstack('-completenames');
   stack = stack(3);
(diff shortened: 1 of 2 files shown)


utilities/private/ft_preamble_init.m
```diff
@@ -46,6 +46,20 @@
...
 end % if nargin
 
+% check if there are fieldnames in the cfg that suggest as if the user
+% erroneously inputted a data argument
+checkdatafields = isfield(cfg, {'label' 'dimord' 'time' 'trialinfo' 'avg'});
+if any(checkdatafields)
+  stack = dbstack('-completenames');
+  stack = stack(3);
+  help(stack.name);
+  % throw the error as if it happened in the original function
```
robertoostenveld: this is interesting, this is also how we should do it in ft_notification when called by ft_error.

utilities/private/ft_preamble_init.m
```diff
@@ -46,6 +46,20 @@
   error(msg);
 end % if nargin
 
+% check if there are fieldnames in the cfg that suggest as if the user
+% erroneously inputted a data argument
+checkdatafields = isfield(cfg, {'label' 'dimord' 'time' 'trialinfo' 'avg'});
```
robertoostenveld: I propose to also check on "cfg". We never use cfg.cfg, but we do use data.cfg very frequently.

schoffelen:
Good one!

schoffelen:
Done! Shall I merge myself?

robertoostenveld:
let me merge while I am here...

PR #652, opened 2018-01-31
FIX regression error detected in test_bug3267 due to timelock structures now having a sampleinfo field.

robertoostenveld:
For this I had to make the sampleinfo handling consistent between ft_dataype_raw, comp and timelock and ensure that it is consistently called from within ft_checkdata. 

In ft_appenddata and ft_appendtimelock I am now explicitly passing hassampleinfo=no  to ft_checkdata (which passes it to the ft_datatype_xxx functions). 

So appended timelock structures will not have sampleinfo.

### DIFF
--- a/test/test_bug3267.m
+++ b/test/test_bug3267.m
@@ -71,9 +71,9 @@ timelockAppend = ft_appendtimelock(cfg, timelockFIC, timelockFC);
 
 %% compare the two versions
 
-% these should not have sampleinfo as per specification in FT_DATATYPE_TIMELOCK
-assert(~isfield(timelockAppend, 'sampleinfo'));
-assert(~isfield(timelockAll,    'sampleinfo'));
+
+assert(~isfield(timelockAppend, 'sampleinfo')); % this should not have it
+assert( isfield(timelockAll,    'sampleinfo')); % and this one should
 
 % the order should be different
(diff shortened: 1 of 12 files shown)


test/test_bug3267.m
```diff
@@ -71,9 +71,9 @@
 
 %% compare the two versions
 
-% these should not have sampleinfo as per specification in FT_DATATYPE_TIMELOCK
```
robertoostenveld: this is the relevant change in the test script

utilities/ft_datatype_timelock.m
```diff
@@ -99,9 +129,16 @@
...
-    timelock = ft_datatype_timelock(timelock, 'version', '2011v2');
-    if exist('sampleinfo', 'var'), timelock.sampleinfo = sampleinfo; end
+    % ensure that the sensor structures are up to date
+    if isfield(timelock, 'grad')
+      timelock.grad = ft_datatype_sens(timelock.grad);
+    end
+    if isfield(timelock, 'elec')
+      timelock.elec = ft_datatype_sens(timelock.elec);
+    end
+    if isfield(timelock, 'opto')
```
robertoostenveld: I also made the checking of grad/elec/opto more consistent

PR #653, opened 2018-01-31
Fix error when giving single TPMs as string

jens-k:
If cfg.tpm is a single string, char(cfg.tpm(:)) results in a long column of single characters, which SPM interprets as one TPM each.

char(cfg.tpm) behaves as intended for cell-arrays of strings but works also for a single string.

Tested with single strings and cell arrays of strings on Matlab2015b/Windows 7 and Matlab2016b/Linux.

### DIFF
--- a/ft_volumesegment.m
+++ b/ft_volumesegment.m
@@ -426,7 +426,7 @@ if dotpm
         
       elseif strcmp(cfg.spmmethod, 'new') || strcmp(cfg.spmmethod, 'mars')
         cfg.tpm = ft_getopt(cfg, 'tpm');
-        cfg.tpm = char(cfg.tpm(:));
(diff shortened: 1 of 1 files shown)


schoffelen:
just to be sure, char(cfg.tpm) also behaves as expected when it's a 1xN cell array?

jens-k:
char(cfg.tpm) on
 cfg.tpm = 'one'
 cfg.tpm = {'one'})
 cfg.tpm = {'one', 'two', 'three'}
 cfg.tpm = {'one'; 'two'; 'three'}
...all seem to give the intended result and ft_volumesegment runs without error on my system.

schoffelen:
OK, thanks!

PR #654, opened 2018-01-31
enhancement - append_common appends through chan dimension with 4D di…

dieloz:


robertoostenveld:
looks good. Is this being used or accessible by any higher level FT function?

dieloz:
the only high level functions using append_common are the append functions (ft_appendtimelock/freq/data.m)

robertoostenveld:
So ft_appendfreq now supports the 4D appending?

dieloz:
I tried with my data and it worked. Didn't work for you?

robertoostenveld:
No, I did not try; no reason to be worried. I was just wondering where this might be of use (e.g. on tutorials) and whether we now may be using a work-around, as it was not implemented yet.

dieloz:
I realized because I was analizing an animal dataset on where I was pooling arrays of sensors of closeby areas coming from different fieldtrip structures. I can imagine this is not the most standard dataformat

PR #657, opened 2018-02-05
speedup elec conversion

robertoostenveld:
as discussed with @StolkArjen over email. The struct2double only needs to be done for the data fields, not for the provenance.

### DIFF
--- a/imotions2fieldtrip.m
+++ b/imotions2fieldtrip.m
@@ -100,7 +100,7 @@ for i=1:numel(label)
   end
   
   % try converting the first 20 elements
-  if numel(time)>10
+  if numel(time)>20
     str = dat.table.(label{i})(1:20);
     val = str2double(str);
     if any(~cellfun(@isempty, str) & isnan(val))
--- a/utilities/ft_datatype_sens.m
+++ b/utilities/ft_datatype_sens.m
@@ -161,7 +161,8 @@ switch version
     sens = fixoldorg(sens, false);
     
     % ensure that all numbers are represented in double precision
-    sens = ft_struct2double(sens);
+    % this only affects the top-level fields and does not recurse
+    sens = ft_struct2double(sens, 1);
     
     % in version 2011v2 this was optional, now it is required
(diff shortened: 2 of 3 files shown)


StolkArjen:
This change halved processing time on one dataset, for which many thanks.

Before:
screen shot 2018-02-02 at 12 24 27 pm


After:
screen shot 2018-02-05 at 6 30 52 pm

It's still a lot of time for simply appending data in the channel dimension, but I guess quite a bit of overhead is needed.

robertoostenveld:
A lot of the time spent in ft_senstype can also be avoided if the code would know what the senstype would be. 
Adding elec.senstype=‘ieeg’ after ft_electrodeplacement probably should do the trick.

StolkArjen:
It seems the chantype field gets lost after re-montaging in ft_preprocessing. Will try to find out where this happens.

ft_electrodeplacement used to add this information (using ft_datatype_sens I believe), but it looks like I removed it one time as any potential bug therein would crash it, which is not ideal after the relatively time-consuming electrode localization. But I will add the field programmatically. Before doing so, it'd be great if you could look at / pull in an outstanding PR from Richard and me, as I prefer to work on the latest version of ft_electrodeplacement to avoid merging conflicts.

robertoostenveld:
note the difference between chantype (which is a cell-array, one per channel) and senstype, which is a string like ‘ctf275’ or ‘eeg’. Chantype is in general harder to intuitively determine than senstype. But if senstype is not specified, the code will determine it by looking (i.e. first determining) chantype. See ft_chantype and ft_senstype.

StolkArjen:
Adding chantype for each channel reduced time required for appending data from 350s to 180s. Biggest time reduction was in fileio/private/ft_senstype, from 158 to 8s. Out of the 180s, most time is spent in ft_senstype (95s) and utilities/private/fixoldorg (40s).

Adding senstype to the elec field, whether 'eeg' or 'ieeg', didn't change much.

PR #659, opened 2018-02-08
Various improvements

robertoostenveld:
relevant to mention are the reorganization of the bipolar montage (no functional changes, only more consistent implementation) and the implementation of an asymmetric threshold for onset and offset detection.


robertoostenveld:
the asymmetric threshold is of relevance to @IngridNieuwenhuis, who may want to check out test_ft_artifact_threshold.m

PR #667, opened 2018-02-23
Colormix colorbar

KristijanArmeni:
@schoffelen This should now do for correctly setting the colorbar when 'colormix' is selected. Caxis is set in ft_plot_mesh.

### DIFF
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -1267,9 +1267,6 @@ switch cfg.method
             set(hc, 'YLim', [fcolmin fcolmax]);
         else
             % functional values have been transformed to be scaled
-            set(hc,'ticks',(0:0.1:1));
-            set(hc,'ticklabels',round(100*linspace(fcolmin,fcolmax,numel(get(hc,'ticks'))'))./100);
-            set(hc,'limits',[0 1]);
         end
       else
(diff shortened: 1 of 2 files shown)


schoffelen:
shouldn't the call to ft_plot_mesh in line 1249 of ft_sourceplot get fcolmin/max as clim argument then?

schoffelen:
shouldn't the colorbar be updated in 1269?

KristijanArmeni:
Yes, and it does take the `clim` as `[fcolmin, fcolmax]`. This is l. 1249 currently:

`ft_plot_mesh(surf, 'edgecolor', cfg.edgecolor, 'facecolor', cfg.facecolor, 'vertexcolor', val, 'facealpha', maskval, 'clim', [fcolmin fcolmax], 'alphalim', [opacmin opacmax], 'alphamap', cfg.opacitymap, 'colormap', cfg.funcolormap, 'maskstyle', 'colormix');`

After the call to ft_plot_mesh, the axis of the plot object then has caxis set to [fcolmin fcolmax], so no need to update in 1269 (as per my testing just now).

schoffelen:
sorry, you're right, I did not scroll far enough to the right...

PR #668, opened 2018-02-24
 Update read_brainvision_vhdr.m

jdien07:
handle cases where impedance value is text such as "out of range" or "disconnected" without aborting reading of impedances.  Also changed str2num to str2double.


schoffelen:
Thanks!

PR #677, opened 2018-03-13
added option to reset light position with key press for method = headshape

smg211:
When users are placing electrodes on a headshape mesh, the default light position is directly above the brain, which causes shadows below and on the side of the brain. This can make placing electrodes more difficult when viewing the brain from an angle other than directly above (necessary for more lateral/inferior electrodes). Therefore, I added a hidden keyboard option that allows users to reset the position of the light to the current camera position by pressing the v key, which was previously not assigned. 

I structured the code in a way that would allow for the simple addition of other keyboard options that are specific to the headshape method, such as the w/a/s/d rotate functionality that is described in the command window when the function is called with cfg.method = 'headshape'.

### DIFF
--- a/ft_electrodeplacement.m
+++ b/ft_electrodeplacement.m
@@ -952,6 +952,19 @@ if isempty(key)
   key = '';
 end
 
+if strcmp(opt.method, 'headshape')
+  % some keyboard commands apply only to the headshape method
+  
+  switch key
+    case 'v' % camlight angle reset
+      delete(findall(h,'Type','light')) % shut out the lights
+      camlight; lighting gouraud; % add a new light from the current camera position
+    otherwise
+      % do nothing
+      
+  end % switch key
+end
+
 % the following code is largely shared with FT_SOURCEPLOT
 switch key
   case {'' 'shift+shift' 'alt-alt' 'control+control' 'command-0'}


robertoostenveld:
I think this would be useful for other mesh visualizations elsewhere in FT. Consistency between FT functions is very important and hence should be always in the back of our mind.

Could you make a list of high level FT functions where this might apply as well? Then we can see whether we can port lightswitch option there as well.

StolkArjen:
From skimming over the main folder, besides ft_electrodeplacement;

ft_sourceplot
ft_defacemesh?
ft_interactiverealign?

.. and per lower level functionalities;

ft_plot_mesh
ft_plot_headshape

StolkArjen:
Can you pull this in, @robertoostenveld, and/or follow-up on the implementation in other functions?

robertoostenveld:
testing the code I get 
```
Reference to non-existent field 'clim'.
Error in ft_electrodeplacement>cb_keyboard (line 1044)
    if isempty(opt.clim) 
Error using uiwait (line 81)
Error while evaluating Figure WindowKeyPressFcn.
```
when I press + or -. This is likely not due to the recent changes, but I'll fix it anyway.

robertoostenveld:
the handling of the keyboard keys conditional on the method is largely incorrect. Many keys result in errors and the instructions are incorrect.

+/- does not zoom in/out but changes the clim.
1/2/3 gives an error in headshape mode
r does not remove
w/a/s/d does not rotate

robertoostenveld:
@StolkArjen can you please update the comments/instructions printed on screen in ft_electrodeplacement so that they match what actually happens?

The potential errors in keyboard handling are not specific to this function but also occur in other plotting functions. I did not want to change them everywhere, so I just left it consistent.

robertoostenveld:
@StolkArjen it now states 

```
>> cfg = [];
>> cfg.method = 'headshape';
>> ft_electrodeplacement(cfg, mesh)
Use the mouse to click on the desired position for the electrode
Subsequently use the mouse to click on the corresponding electrode label
Press "r" to delete the last point add
Press "+/-" to zoom in/out
Press "w/a/s/d" to rotate
Press "q" when you are done
```

I believe only 'q' is correct. I suggest to also document 'v'

StolkArjen:
@robertoostenveld, I will update this

PR #678, opened 2018-03-14
excluding cfg.grid from recursive size checking for massive speedup (bug 3419)

Spaak:
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3419


utilities/ft_checkconfig.m
```diff
@@ -725,7 +725,7 @@
 end
 
 % these fields should not be handled recursively
-norecursion = {'event', 'headmodel', 'leadfield'};
```
robertoostenveld: 👍🏻

PR #679, opened 2018-03-14
Bug3417

robertoostenveld:
the data used in the test script was invalid.


test/invalid_bug3417.m
```diff
@@ -0,0 +1,272 @@
+function test_bug3417
```
robertoostenveld: the function name should be updated

PR #681, opened 2018-03-15
FIX: Fix old weight setting

larsoner:
From looking at the code, I *think* that this line should be outside the loop over vertices. But it would be good if someone more familiar with the method could check!


larsoner:
I had an email exchange with Guido, and he confirmed that this fix should be correct. FWIW it doesn't really affect the results, but it's good to have the correct computation in any case.

schoffelen:
Thanks Eric!

PR #684, opened 2018-04-01
EDF Reading Handling of (often occuring) case that Physical Minimum larger than M…

Frederik-D-Weber:
…aximum in header

the error of Physical Minimum larger than Maximum often occurs, 
instead of replacing the Physical Minimum and Maximum with the Digital Minimum larger than Maximum,
the code tries to correct this common error in the header by swiching the Minimums and Maximums in the respective channels.
This handling is not perfect, but under most circumstances leads to the correct signals as they also apear (exaclty) in the EDFbrowser with the correct scaling and polarity of the signals. This handling avoids the more severe case that the scaling of ALL signals are falsely put to the Digital Minimum and Maximum.

### DIFF
--- a/fileio/private/read_edf.m
+++ b/fileio/private/read_edf.m
@@ -160,21 +160,27 @@ if needhdr
     EDF.DigMax = (2^15-1)*ones(EDF.NS,1);
   end
   if (any(EDF.DigMin >= EDF.DigMax))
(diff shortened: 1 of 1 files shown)


Frederik-D-Weber:
...needs to handle potential proper inversion of signal that goes along with handling

PR #687, opened 2018-04-06
added option to define ft_databrowser figure position

ludovicbellier:
Hello,

As I frequently open several ft_databrowser figures on my screen (e.g., to compare different preprocessing strategies on the same data), I was struggling to position them properly. The use of set(gcf, 'Position', [a b c d]) was poorly working when running a script, as it was sometimes called before the the figure was actually displayed, and a pause wasn't always solving it; further, it was involving two displays of the figure (a first one at opening, and a second after defining the new position), thus it was a barely ok solution, but computationally inefficient.
With a very simple edit, one can now call ft_databrowser with cfg.position defined as a classic [left bottom width height] vector, and therefore get a clean figure positioning process.
Also briefly edited the ft_databrowser header to describe this functionality.

Best,
Ludovic

### DIFF
--- a/ft_databrowser.m
+++ b/ft_databrowser.m
@@ -53,6 +53,7 @@ function [cfg] = ft_databrowser(cfg, data)
 %                                 be passed to the selfun (default = 'selected')
 %   cfg.renderer                = string, 'opengl', 'zbuffer', 'painters', see MATLAB Figure Properties. If the databrowser
 %                                 crashes, you should try 'painters'.
+%   cfg.position                = location and size of the figure, specified as a vector of the form [left bottom width height].
 %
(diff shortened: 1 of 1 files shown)


robertoostenveld:
why did you close this?

ludovicbellier:
Oops, that was a mistake, sorry about that! Will correct this ASAP, thanks
for notifying me.

ludovicbellier:
Et voilà =]

robertoostenveld:
This would be interesting to have in all plotting functions. @Spaak do you perhaps know whether something has already been implemented in this direction? I could not find it, except for the positioning of figures at the same place while interactive plotting of channel data.

PR #690, opened 2018-04-10
Updates to external/NIRS (and ft_fetch_header)

Horschig:
some adjustments to our reading functions, and a bugfix for ft_fetch_header, which removed (or did not copy) the opto-structure of the header, thereby leading to an unknown senstype (e.g. by a call to ft_redefinetrial).

I hope our p-files are included in this request, because I do not see them here (but maybe they were already commited earlier by me, because those commits were from July 26 2017).


robertoostenveld:
Thanks. The p-files don't seem to be included in this specific PR, but I do have them in my own copy of the code. I guess they were already merged some time ago.

robertoostenveld:
@marcvanwanrooij, please note that the suggestion from @Horschig is now merged in the HEAD/release version. If all is fine, you would not need a hacked version any more for the tutorial.

PR #691, opened 2018-04-12
fix a crosshair bug in ft_volumerealign top and bottom left plots

ludovicbellier:
Hey there,
While exploring co-registration outputs using the figure produced by ft_volumerealign, I noticed that clicking in the top left or bottom left subplots didn't update the crosshair at the expected location; everything worked fine for the top right subplot though. Took some time to figure out where it was coming from but it happens the fix is extremely simple!
Best,
Ludovic

### DIFF
--- a/ft_volumerealign.m
+++ b/ft_volumerealign.m
@@ -1909,10 +1909,10 @@ if opt.viewresult
 end
 
 if ~isempty(tag) && ~opt.init
-  if strcmp(tag, 'ik')
+  if strcmp(tag, 'ij')
     opt.ijk([1 3])  = round(pos([1 3]));
     opt.update = [1 1 1];
-  elseif strcmp(tag, 'ij')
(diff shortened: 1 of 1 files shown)


ludovicbellier:
Just noticed that while this fix worked great for the co-registration outputs I was looking at, it actually triggers the same bug for ct images. Will thus work further on this before resubmitting.

ludovicbellier:
Here it is! As the crosshair bug is specific to the two-input mode of ft_volumerealign (with "mri" and "target" as inputs), I moved the fix (consisting of switching top left and bottom left subplots' tags (ij <-> ik)) from the cb_getposition callback, which impacts both one- ("mri" only as input) and two-input modes, to the appropriate section in the code which only impacts the two-input mode.

robertoostenveld:
difficult to check for me right now, but I take it you checked it carefully.

PR #696, opened 2018-04-24
implemented helper function to create BIDS sidecar files

robertoostenveld:


robertoostenveld:
The code is not yet 100% complete, but already in a state that others should be able to test it. Hence I have merged it with the master branch.

Notable changes that are still planned are to add support for EEG and iEEG (as discussed with @gpiantoni), to improve the handling of the anatomical MRI sidecar files, and possibly to implement support for fif files.

PR #703, opened 2018-05-03
optimize ft_volumelookup computing time

ludovicbellier:
Hi there,

Currently working on a 250-channel ECoG grid case, I used ft_volumelookup to get each channel's anatomical label, and it took about 1h37min (~23.2 seconds per channel), which was surprisingly long to me. After a two-step optimization, it now takes slightly more than 7 minutes =)

First, the ismember call on line 354 was replaced by a faster implementation (ismember is actually [known to be slow](https://fr.mathworks.com/matlabcentral/answers/7434-find-vector-in-matrix)); running this line was then taking 0.57 seconds instead of ~19 seconds;

However, running this new version of ft_volumelookup was then taking ~6 seconds (thus 25 minutes for my 250 channels), which meant that this line wasn't the only time-consuming step within ft_volumelookup, thus that there was still room for optimization. Indeed, some channel-/point-of-interest-independent steps were making most of this computing time (e.g., ft_checkdata on line 152 was taking 2.6 seconds; a big matrix transposition of line 351 was taking 1.7 seconds): to get labels for my 250 channels, I was thus unnecessarily repeating more than 4.3 seconds steps, for a total of about 18 unnecessary minutes.

Therefore, second optimization step consisted of enabling a multioutput option, considering rows of cfg.roi as independent points of interest, and giving one output by channel; that way, POI-independent steps are done once for all POIs, and this optimized version of ft_volumelookup now takes about 7 minutes to run.

I implemented this option so that previous calls to ft_volumelookup wouldn't be affected by it (default cfg.multioutput is 'no'). In a way, cfg.multioutput is a switch to either see cfg.roi as points defining a region of interest (when equal 'no' => "integrated", single output), or as a list of individual points (when equal 'yes' => same output structure but with as many levels as the number of points provided).

Best,
Ludovic

PS: absolute time values I gave here to illustrate this pull request obviously depend on the resolution of the atlas/volume used, but a general claim would be that this two-step optimization improves computing time of ft_volumelookup by (at least) one order of magnitude.


robertoostenveld:
Speed improvements are welcome! However, with your suggested change test_ft_volumelookup (in the test directory) starts failing 

```
Undefined function or variable 'nSel'.
Error in ft_volumelookup (line 365)
  for iVox = 1:nSel
Error in test_ft_volumelookup (line 43)
mask5     = ft_volumelookup(cfg, mri); 
```

Can you add your usage pattern to the test function and ensure that it does not fail on the older use cases?

Furthermore, it is unclear what the new multioutput option does due to lack of documentation. What would the difference be in the output when it is yes or no?

I think that the overall documentation of the function is poor. It seems that the options can be described as a 2x2 table (2 outputs, each with 2 options for the inputs). Better would be to change it such that the goals are more clear. So rather than "The configuration options for ... " the help  should say "The configuration options to obtain xxx from yyy are ...". Can you update the help? And can you fix querymethod cube/shpere in the help?

ludovicbellier:
The 'nSel' error is now corrected, and test_ft_volumelookup.m is updated with my new usage pattern.

About the documentation, here's the rationale of this multioutput option:
- if you provide several points of interest to the original version of ft_volumelookup.m, these POI are considered as describing a region of interest (hence the field name, cfg.roi) you get only one label structure;
- if you provide several POI to this new version of ft_volumelookup.m, you can keep its previous behavior by choosing cfg.multioutput = 'no', or you can tell the function to consider each POI as independent by choosing cfg.multioutput = 'yes' and then to give you one label per POI (that's especially of interest in the case you want labels for many electrodes, as this is the main factor for speed improvement).

As the two usages are very close, I updated the documentation trying to be the clearest possible, but without repeating a pretty big paragraph which would make the documentation heavier.

ft_volumelookup.m
```diff
@@ -37,15 +37,17 @@
...
 %   cfg.inputcoord          = 'mni' or 'tal', coordinate system of the mri/source/stat
 %   cfg.atlas               = string, filename of atlas to use, see FT_READ_ATLAS
 %   cfg.maxqueryrange       = number, should be 1, 3, 5 (default = 1)
-%   cfg.querymethod         = 'sphere' searches voxels around the roi in a sphere (default)
-%                           = 'cube' searches voxels around the roi in a sphere
-%   cfg.round2nearestvoxel = 'yes' or 'no', voxel closest to point of interest is calculated (default = 'yes')
+%   cfg.querymethod         = 'sphere' searches voxels around the ROI in a sphere (default)
+%                           = 'cube' searches voxels around the ROI in a cube
+%   cfg.round2nearestvoxel  = 'yes' or 'no', voxel closest to POI is calculated (default = 'yes')
+%   cfg.multioutput         = 'no' always outputs one label; if several POI are provided, they are considered together as describing a ROI (default)
```
robertoostenveld: I propose you delete the cfg.multioutput option and rather use cfg.output. It now only has one option, which is silly, but apparently was meant for use cases like this. It could have options cfg.output=single/multiple, where ft_checkconfig with renamedval would be used to replace 'label' by 'single' .

robertoostenveld: so your new functionality would be the result of cfg.output=multiple.

robertoostenveld:
thanks, much appreciated!

PR #704, opened 2018-05-08
resolved some issues in openbci

robertoostenveld:
as detected at the Iklectik performance in London


robertoostenveld:
@stephenwhitmarsh please update to the latest FT version to solve the issues with openbci2ft

PR #707, opened 2018-05-12
add cfg.minqueryrange to enable probabilistic labeling

ludovicbellier:
Using ft_volumelookup to automatically localize ECoG electrodes, I noticed that ambiguous electrodes (e.g., lying between two gyri) almost always ended up with one label only.
Indeed, it's due to how ft_volumelookup searches for labels: it starts at the atlas voxel corresponding to the electrode coordinates provided in input (cfg.roi), and only in case that this voxel doesn't relates to a label, it increases the query range (thus, looks for neighbor voxels' potentially related labels) until it finds a label or it reaches cfg.maxqueryrange, whichever comes first (see lines 374 to 379). Most of the time, the initial voxel does have a label, therefore the search stops, even though immediate neighbors could have a different label. However, in such a case we can expect to record neural activity from both regions, thus it would make sense to get both labels.

To enable this (concretely, to be able to get several nonzero elements in the label.count output), I added cfg.minqueryrange, that you can use to get labels of all voxels contained in a volume (either a cube or a sphere) centered at the provided POI.

p1_labelsmni

Simple example: the electrode designated with the red arrow was automatically labeled as middle temporal gyrus (with a used query range of 1). Setting cfg.minqueryrange to 5, I get 10 voxels labeled as superior temporal gyrus and 8 as MTG. Then, I can either take the max by considering that this electrode is located over STG, or take a probabilistic labeling approach by considering that this electrode records neural activity coming at 56% from STG and 44% from MTG.

### DIFF
--- a/ft_volumelookup.m
+++ b/ft_volumelookup.m
@@ -35,7 +35,8 @@ function [output] = ft_volumelookup(cfg, volume)
 %   cfg.inputcoord          = 'mni' or 'tal', coordinate system of the mri/source/stat
 %   cfg.atlas               = string, filename of atlas to use, see FT_READ_ATLAS
 %   cfg.maskparameter       = string, field in volume to be looked up, data in field should be logical
-%   cfg.maxqueryrange       = number, should be odd (default = 1)
+%   cfg.minqueryrange       = number, should be odd and <= to maxqueryrange (default = 1)
+%   cfg.maxqueryrange       = number, should be odd and >= to minqueryrange (default = 1)
 %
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks for the clear explanation and improvement to the code. It might be valuable to take part of this (and previous) PR comments and make a FAQ out of this, explaining how to look up anatomical or functional labels corresponding with iEEG electrode locations. If you want, just make a new page under www.fieldtriptoolbox.org/faq.

PR #731, opened 2018-06-21
adding mmfmatlabio code

arnodelorme:
1. The MFFMATLABIO code is available at https://drive.google.com/open?id=1PHXdEFmVqDRy04vtOKI9aXn54BHoPcqy
I have it in a private repo, and I am in the process of transferring it to github and version it. It will be GNU GPL3.
I have not added it yet to the Fileio/Fieldtrip repo because it is not 100% ready. However, this will not prevent the code to run if you get the zip file.

2. It is an EEGLAB plugin but it also works as standalone in Fieldtrip. I wanted to have a single code that would work in Matlab, Fieldtrip and EEGLAB to avoid having people branching the code.

3. The file testallfiles will test 27 different MFF files all available. I am attaching the report on all files when using the standalone Fieldtrip/File-io importer/exporter (a couple of files still have problem).
When using Fieldtrip, there is still a problem with importing different sensor types so they can be decoded properly. I will fix it.

https://drive.google.com/drive/folders/13PeOVr67g4tu-svKerrsBCg5fMKlRsBF?usp=sharing

### DIFF
--- a/fileio/ft_read_data.m
+++ b/fileio/ft_read_data.m
@@ -696,6 +696,10 @@ switch dataformat
     % pass the header along to speed it up, it will be read on the fly in case it is empty
     dat = read_mff_data(filename, 'sample', begsample, endsample, chanindx, hdr);
 
+  case 'egi_mff_v3'
+    ft_hastoolbox('mffmatlabio', 1);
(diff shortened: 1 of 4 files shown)


robertoostenveld:
thanks, looking forward to that code being released. At that time it would be good to include it to ft_hastoolbox

PR #740, opened 2018-07-12
fix Bug 3439 (OpenMEEG test script + minor EEG-related issues); fix typo in AAL atlas labels

sarangnemo:
The OpenMEEG test script was not producing expected results. I've updated it to avoid the problematic call to ft_prepare_headmodel (which is not necessary for the working OpenMEEG pipeline originally created by @ddewong and myself). Our pipeline expects outside->inside order of the surfaces, so I have changed those accordingly, and the test script now passes.

**Refs:**
[OpenMEEG mailing list thread](https://lists.gforge.inria.fr/pipermail/openmeeg-info/2018/000254.html) @agramfort 
[FieldTrip bug](http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3439)

**Remains to be done:**
different and seemingly now-broken OpenMEEG pipelines remain in `ft_prepare_headmodel.m` and `ft_compute_leadfield.m`. They call `ft_headmodel_openmeeg.m` and `external/openmeeg/openmeeg_dsm.m` and `external/openmeeg/openmeeg_megm.m`


external/openmeeg/testOpenMEEGeeg.m
```diff
@@ -5,16 +5,16 @@
...
 %% Set the position of the probe dipole
 dippos = [0 0 70];
 
 %% Set the radius and conductivities of each of the compartments
 
 % 4 Layers
-r = [85 88 92 100];
-c = [1 1/20 1/80 1];
+r = [100 92 88 85];
+c = [1 1/80 1/20 1];
```
agramfort: why did you need this?

forward/ft_headmodel_openmeeg.m
```diff
@@ -116,7 +116,7 @@
   end
    
   for ii=1:length(headmodel.bnd)
-    [junk,tname] = fileparts(tempname);
+    [junk,tname] = fileparts(['/tmp/mesh' num2str(ii)]);%; fileparts(tempname);
```
agramfort: this will not work on windows right?

template/atlas/aal/ROI_MNI_V4.txt
```diff
@@ -88,24 +88,24 @@ T2AG	Temporal_Pole_Mid_L	8211
...
+CER7BD	Cerebellum_7b_R	9052
+CER8G	Cerebellum_8_L	9061
+CER8D	Cerebellum_8_R	9062
+CER9G	Cerebellum_9_L	9071
+CER9D	Cerebellum_9_R	9072
+CER10G	Cerebellum_10_L	9081
+CER10D	Cerebellum_10_R	9082
 VER1_2	Vermis_1_2	9100
 VER3	Vermis_3	9110
 VER4_5	Vermis_4_5	9120
```
agramfort: this seems unrelated to the bug fix.

external/openmeeg/testOpenMEEGeeg.m
sarangnemo: "Our pipeline expects outside->inside order of the surfaces, so I have changed those accordingly." :-)

template/atlas/aal/ROI_MNI_V4.txt
sarangnemo: That's the "fix typo in AAL atlas labels" part of the PR.... Ok, it accidentally got selected when pushing this PR, but it needed to happen anyway, so I left it in. :-)

forward/ft_headmodel_openmeeg.m
sarangnemo: Good catch, I didn't mean to leave that hard-coded.

sarangnemo: Fixed in #742

external/openmeeg/testOpenMEEGeeg.m
agramfort: well I think the right fix is to unsure that your pipeline is smart enough to support previously documented code rather than changing the tests :

template/atlas/aal/ROI_MNI_V4.txt
agramfort: @sarangnemo this tells me that merging of PR should not be automatic and done by the person that opens the PR. No need to enforce this but it's good practice to ask for quick review before merging code to master.

sarangnemo: Normally wouldn't do that, but got a (apparently premature!) vacation message from Robert and didn't realize you'd be able to review this here. I acknowledge AAL label update should have been a separate PR, but was consciously submitted in this PR in lieu of digging through git documentation to deselect it.

external/openmeeg/testOpenMEEGeeg.m
sarangnemo: Documented where? The previous code calls a function that tries to rearrange code from outside to inside as well, it happens to also fail at it. Other fieldtrip functions and leadfield_openmeeg.m are also smart enough to rearrange when given anatomical labels (e.g. "scalp", "skull", "brain").

agramfort: well I don't fully see the impact of this change, it just seems inappropriate to change the test to make it pass. test should ensure there is no regression so I prefer to change the functions so the old tests still pass as is.

sarangnemo: Well if every implementation expects outside to inside, or alternatively labels to indicate what each layer corresponds to, then it seems that it's the test that should change. No regression is ensured since the output values match exactly, plus this is what is used in practice for real analyses.

agramfort:
ok fair enough

sarangnemo:
See #743 (move to test directory according to @robertoostenveld's suggestion on bugzilla)

PR #743, opened 2018-07-13
move testOpenMEEGeeg.m to test folder

sarangnemo:
- moved external/openmeeg/testOpenMEEGeeg.m to test/test_openmeeg_eeg.m
- clarified that boundaries are defined outside (scalp) to inside (brain)

### DIFF
--- a/test/test_openmeeg_eeg.m
+++ b/test/test_openmeeg_eeg.m
@@ -104,7 +104,6 @@ function [rdms,mags] = run_bem_computation(r,c,dippos)
     cfg.grid.pos = dippos;
     cfg.grid.unit = 'mm';
     cfg.elec = sens;
-%    grid = ft_prepare_leadfield(cfg);
     grid = ft_prepare_leadfield(cfg);
 
     lf_openmeeg = grid.leadfield{1};
--- a/test/test_openmeeg_eeg.m
+++ b/test/test_openmeeg_eeg.m
@@ -1,6 +1,7 @@
-function testOpenMEEGeeg
+function test_openmeeg_eeg
 
-% TEST testOpenMEEGeeg
+% TEST test_openmeeg_eeg
+% (previously called testOpenMEEGeeg)
 % Test the computation of an EEG leadfield with OpenMEEG
 
 % Copyright (C) 2010-2017, OpenMEEG developers
--- a/test/test_openmeeg_eeg.m
+++ b/test/test_openmeeg_eeg.m
@@ -1,0 +1,132 @@
+function testOpenMEEGeeg
+
+% TEST testOpenMEEGeeg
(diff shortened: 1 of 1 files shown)


test/test_openmeeg_eeg.m
```diff
@@ -0,0 +1,133 @@
...
+    % the order of the layers
+    %cfg.conductivity = c;
+    %vol_bem = ft_prepare_headmodel(cfg, bnd);
+    %vol_bem = rmfield(vol_bem,'mat');
+    
+    cfg.vol = vol_bem;
+    cfg.grid.pos = dippos;
+    cfg.grid.unit = 'mm';
+    cfg.elec = sens;
+%    grid = ft_prepare_leadfield(cfg);
```
agramfort: to remove

test/test_openmeeg_eeg.m
```diff
@@ -0,0 +1,133 @@
...
+    grid = ft_prepare_leadfield(cfg);
+
+    lf_openmeeg = grid.leadfield{1};
+
+    % Rq : ft_compute_leadfield centers the forward fields by default
+    % (average reference)
+    % lf_openmeeg = lf_openmeeg - repmat(mean(lf_openmeeg),size(lf_openmeeg,1),1);
```
agramfort: shall we clean this up?

test/test_openmeeg_eeg.m
```diff
@@ -0,0 +1,133 @@
...
+    vol_bem = [];
+    vol_bem.bnd = bnd;
+    vol_bem.cond = c;
+    vol_bem.type = 'openmeeg';
+
+    % ft_prepare_headmodel has a bug; likely the geom file does not match
+    % the order of the layers
+    %cfg.conductivity = c;
+    %vol_bem = ft_prepare_headmodel(cfg, bnd);
+    %vol_bem = rmfield(vol_bem,'mat');
```
agramfort: ?

test/test_openmeeg_eeg.m
sarangnemo: I'm unaware of the history here, but it may come up again if we consolidate the FT-OM pipelines soon: AFAIK some sections of FieldTrip's forward modeling code assumes average reference for EEG, while others do not.

test/test_openmeeg_eeg.m
sarangnemo: This is a record of a test I made with the ft_prepare_headmodel method, when diagnosing the problem with the original script. Preparing vol_bem in this way (removing the headmodel matrix contained within) also gives the expected results; however, the results are wonky (at least with OpenMEEG 2.3 and 2.4) if the `mat` is retained.

agramfort:
ok make a comment so we don't forget?

PR #745, opened 2018-07-13
om_save_tri: Check and correction for surface normal convention

sarangnemo:
Checks and corrects output meshes for the convention expected by OpenMEEG 2.2. Similar code is scattered in multiple places in OpenMEEG-related functions in FieldTrip, so easiest to consolidate here.

This is inconsequential for v2.3 and up, since these versions appear to internally check/correct the convention anyway. @agramfort 

This will help ensure that the test script test_openmeeg_eeg.m performs as expected across OpenMEEG versions. (It may otherwise fail with v2.2.)


external/openmeeg/om_save_tri.m
```diff
@@ -8,6 +8,14 @@
 
 % Copyright (C) 2010-2017, OpenMEEG developers
 
+% OpenMEEG vs 2.3 and up internally adjusts the convention for surface
+% normals, but OpenMEEG v2.2 expects surface normals to point inwards;
+% this checks and corrects if needed
+ok = checknormals(points,faces);
+if ~ok
+    faces = fliplr(faces);
+end
```
agramfort: can you do this optionally ie add a check_normals parameter to the function that fieldtrip sets to true ? Some people maybe just want to read a tri and write back a tri file to disk, and they would not expect to have their normals automatically flipped.

sarangnemo:
I've reverted this, as it's no longer necessary with my next PR that comprises an overhaul of the OpenMEEG link.

PR #746, opened 2018-07-17
AMP2FT update

fraimondo:
Some time ago I had to use this realtime module and I fixed several issues. Here's the updated code.


robertoostenveld:
It appears that you started off from an older version of the fieldtrip git repository, causing merge conflicts. I have done my best to resolve them.

Your branch now compiles on my computer (macOS High Sierra, 10.13, also with compiler SDK 10.13).  I only still had to make a change in  realtime/src/buffer/src/osx/clock_gettime.h where I changed
```
#ifdef not(__CLOCK_AVAILABILITY)
```
into
```
#if __DARWIN_C_LEVEL < 199309L
```

Could you check whether your branch fraimondo:ampserver compiles for you and perhaps whether it works?

PR #747, opened 2018-07-18
Overhaul of OpenMEEG pipeline

sarangnemo:
The FieldTrip-OpenMEEG pipelines were suffering from duplicated functionality across various functions (see #740 and #673, also [FieldTrip's bugzilla](http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3439) and a [thread on the OpenMEEG mailing list](https://lists.gforge.inria.fr/pipermail/openmeeg-info/2018/000254.html)) .

This PR attempts to remove redundancies and adopt a workflow similar to other head model strategies (i.e., making use of `ft_prepare_headmodel` for sensor-type and position-agnostic information, i.e., the `hminv` matrix output by OpenMEEG). This headmodel structure can then be saved and reused with data from the same participant in the same coordinate space.  Then, `ft_prepare_leadfield` is used for other computations that require sensor type and/or position, ultimately producing the lead field gain matrix.

The code has been tested to yield identical results with both OpenMEEG 2.2 and 2.4 with the EEG test script (`test/test_openmeeg_eeg`). It also yields identical results as the previous Nemo lab pipeline with real MEG data, tested with both OpenMEEG 2.3 and 2.4.

Still remains to be done: the DSM file can be reused if given the same voxel grid, but still need to implement a way to save it somewhere. (The matrix can be >2GB for 5mm voxel grids, so it may not be wise to keep it in memory and pass with the cfg structure.)


forward/ft_sysmat_openmeeg.m
```diff
@@ -0,0 +1,107 @@
...
+%   http://www.biomedical-engineering-online.com/content/9/1/45
+%   doi:10.1186/1475-925X-9-45
+% and
+%   Kybic et al. Generalized head models for MEG/EEG: boundary element method
+%   beyond nested volumes. Phys. Med. Biol. (2006) vol. 51 pp. 1333-1346
+%   doi:10.1088/0031-9155/51/5/021
+%
+% This link with FieldTrip is derived from the OpenMEEG project
+% with contributions from Daniel Wong and Sarang Dalal, and uses external
+% command-line executables. See http://gforge.inria.fr/projects/openmeeg
```
agramfort: can you use this url : http://openmeeg.github.io/

thx

forward/ft_sensinterp_openmeeg.m
```diff
@@ -0,0 +1,127 @@
...
+%   http://www.biomedical-engineering-online.com/content/9/1/45
+%   doi:10.1186/1475-925X-9-45
+% and
+%   Kybic et al. Generalized head models for MEG/EEG: boundary element method
+%   beyond nested volumes. Phys. Med. Biol. (2006) vol. 51 pp. 1333-1346
+%   doi:10.1088/0031-9155/51/5/021
+%
+% This link with FieldTrip is derived from the OpenMEEG project
+% with contributions from Daniel Wong and Sarang Dalal, and uses external
+% command-line executables. See http://gforge.inria.fr/projects/openmeeg
```
agramfort: here also can you use this url : http://openmeeg.github.io/

forward/ft_headmodel_openmeeg.m
```diff
@@ -15,216 +15,196 @@
...
 %   beyond nested volumes. Phys. Med. Biol. (2006) vol. 51 pp. 1333-1346
 %   doi:10.1088/0031-9155/51/5/021
-% 
-% The implementation in this function is derived from the the OpenMEEG project
-%  and uses external command-line executables. See http://gforge.inria.fr/projects/openmeeg
-% and http://gforge.inria.fr/frs/?group_id=435.
+%
+% This link with FieldTrip is derived from the OpenMEEG project
+% with contributions from Daniel Wong and Sarang Dalal, and uses external
+% command-line executables. See http://gforge.inria.fr/projects/openmeeg
```
agramfort: http://openmeeg.github.io/

agramfort:
I cannot take on my machine but I trust your testing skills @sarangnemo 

thanks a lot for refactoring all this.

DanielMiklody:
Hey, cool, thanks a lot! I will test it, soon, have something running at the moment.

DanielMiklody:
Sorry for the late reply, works perfectly for me now with 2.2/2.4

PR #750, opened 2018-07-24
cfg parameter for ft_defaults to override global options

ThreeFive-O:
Based on PR #414 we discussed about an option to override the global options with a cfg parameter. This PR is a very naive approach but gets the work done.


robertoostenveld:
the update for the IBTB documentation is fine, but I don't like your proposed change for the ft_defaults function. I consider a special function to change a global variable to be too obscure. I would rather have that people understand the global variable. I documented it instead.

PR #758, opened 2018-08-03
[nutmegtrip] nmt_polaritytweak: make selection of numSource dimension explicit

2051289v:
if numSources > numTimepoints, length() will select the longer (wrong) dimension. 
s = [numSources, numTimepoints].


sarangnemo:
mange tak!

PR #763, opened 2018-08-06
add: support for Nihon Kohden 1200 data files

StolkArjen:
This PR includes:
- low-level event/hdr/data reading functions
- higher-level integration with ft_read_data/header/event

The functionalities wrap around BrainStorm3 reading functions. Tested on a file locally. Needs testing on multi-epoch NK1200 files, but haven't seen those yet (it currently throws a warning if multiple epochs are detected).


fileio/ft_filetype.m
```diff
@@ -1337,6 +1337,10 @@
   type = 'nihonkohden_m00';
   manufacturer = 'Nihon Kohden';
   content = 'continuous EEG';
+elseif filetype_check_extension(filename, '.EEG')
```
robertoostenveld: there are many more file formats that end with *.eeg or *.EEG, so this is a very poor test. Please look at whether the first N bytes of the (binary?) file are unique.

robertoostenveld:
Please do split the code (as you wrote in your first email) and call it 'brainstorm' rather than NK1200. 

The current code will be much more difficult to maintain in the long run than with the brainstorm files (in_fopen and in_fread) in external/brainstorm. 

Furthermore, this would not have the benefit of other file formats supported by brainstorm.

StolkArjen:
Thanks for the feedback. The following changes have been made:
- all relevant brainstorm functionality is contained by external/brainstorm
- ft_filetype checks for the presence of other files (see screenshot), to dissociate other .eeg from nihonkohden .EEG
- renamed header/data format into 'nihonkohden_eeg'

screen shot 2018-08-07 at 9 28 30 am

robertoostenveld:
thanks. Can you send me an example dataset to add to the regression testing (e.g. over dropbox or wetransfer)?

StolkArjen:
Thanks. I wanted to do this earlier (incl a test function), but the data contains sensitive information

PR #765, opened 2018-08-08
Leadfield

schoffelen:
This PR reflects some work to add functionality to ft_headmovement, and to be able to use this functionality in downstream analysis (requiring the computation of forward models possibly multiple times).

Essentially:
1 ft_headmovement can now also output a struct-array of gradiometer descriptions, based on a kmeans clustering of the HLC reconstructed head position during the measurement. 
2 ft_prepare_leadfield can now compute singleshell headmodel based leadfields much faster (in my specific test case (matlab 2016b, 8196 dipoles) this speeds up the computation with a factor of 6.
3 in order for ft_prepare_leadfield to compute the 'leadfield' in chunks-of-dipoles for the singleshell method, I adjusted also the low-level code contributed by Guido (leadsphere_all), to keep it somewhat memory efficient. The original implementation was extremely greedy, which became apparent when using it on the 8196 dipole example. The function required at some point about 4000mb of memory extra. I have tested the numerical equality between Guido's original and my implementation, and I'm pretty sure that they are equivalent.
4 some small cosmetic changes along the way in functions that I brushed by.
5 (as of yet hidden) option in topoplot_common to plot the topo's in a single figure with subplots, rather than one topo per figure, when multiple inputs are present

### DIFF
--- a/forward/ft_average_sens.m
+++ b/forward/ft_average_sens.m
@@ -44,9 +44,8 @@ nsens  = numel(sens);
 
 % ensure the correct representation
 for i=1:nsens
-  newsens(i) = ft_datatype_sens(sens(i));
+  sens(i) = ft_datatype_sens(sens(i));
 end
-sens = newsens; clear newsens
 
 fiducials = fixpos(fiducials);
 
@@ -74,16 +73,16 @@ else
 end
 
 pos = detrend(sens(1).chanpos, 'constant');
-[u s v]= svd(pos'*pos);
+[u, s, v]= svd(pos'*pos);
 
 % starting with the second PC works a little better (e.g. on BTi)
 x1 = pos*u(:, 2);
 x2 = pos*u(:, 1);
 
 % detemine the indices of three reference sensors that are close to the three principal axes
-[m ind1] = min(x1);
-[m ind2] = max(x1);
-[m ind3] = max(abs(x2 - mean(x2([ind1 ind2]))));
+[~, ind1] = min(x1);
+[~, ind2] = max(x1);
+[~, ind3] = max(abs(x2 - mean(x2([ind1 ind2]))));
 
 % compute the mean of the three reference sensors for the principal axes
 mean1 = [0 0 0];
@@ -206,15 +205,15 @@ switch nfid
     switch afiducials.unit
         case 'mm'
             c = 0.1;
-        case 'cm';
+        case 'cm'
             c = 1;
-        case 'dm';
+        case 'dm'
             c = 10;
         case 'm'
             c = 100;
     end
     
-    [upos, ind] = unique(round(c*afiducials.pos/tolerance), 'rows');
+    [~, ind] = unique(round(c*afiducials.pos/tolerance), 'rows');
     
     afiducials.pos = afiducials.pos(ind, :);
     
--- a/ft_headmovement.m
+++ b/ft_headmovement.m
@@ -1,4 +1,4 @@
-function [grad, grad_orig, grad_avg, trialinfo] = ft_headmovement(cfg)
+function [varargout] = ft_headmovement(cfg)
 
 % FT_HEADMOVEMENT creates a representation of the CTF gradiometer definition
 % in which the headmovement in incorporated. The result can be used
@@ -9,7 +9,7 @@ function [grad, grad_orig, grad_avg, trialinfo] = ft_headmovement(cfg)
 % where the configuration should contain
 %   cfg.dataset      = string with the filename
 %   cfg.trl          = Nx3 matrix with the trial definition, see FT_DEFINETRIAL
-%   cfg.numclusters  = number of segments with constant headposition in which to split the data (default = 12)
+%   cfg.numclusters  = number of segments with constant headposition in which to split the data (default = 10)
 %
 % optional arguments are
 %   cfg.updatesens   = 'yes' (default) or 'no'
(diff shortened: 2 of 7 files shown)


ft_headmovement.m
```diff
@@ -11,12 +11,34 @@
...
+%
+% If cfg.updatesens is specified, the output grad-structure has a 
+% specification of the coils expanded as per the centroids of the position
+% clusters. The balancing matrix is s a weighted concatenation of the
+% original tra-matrix. If cfg.updatesens is false, the output grad is a
+% struct-array with coil and sensor positions as per the clusters'
+% centroids. If cfg.avgovertrial is specified, the clustering of head
+% positions is done on the average headcoil-position per trial.
+%
+% Additional outputs are:
```
robertoostenveld: Multiple low-level outputs is not really FT-style. Looking at ft_regressconfound I see cfg.output=residual/model/beta. Idem for ft_mvaranalysis.

How about adding
cfg.output=original/averaged/split
and adding the trialinfo elsewhere (?)

ft_headmovement.m
```diff
@@ -1,4 +1,4 @@
-function [grad] = ft_headmovement(cfg)
+function [grad, grad_orig, grad_avg, trialinfo] = ft_headmovement(cfg)
```
robertoostenveld: why does this function not accept a data structure (which should contain the HLC channels) as input?

ft_prepare_leadfield.m
```diff
@@ -47,6 +47,22 @@
 %                         whether the lower rank leadfield is projected back onto the original
 %                         linear subspace, or not.
 %
+% Depending on the type of headmodel, some additional options may be
+% specified. 
+%
+% For OPENMEEG based headmodels:
```
robertoostenveld: the option should be called cfg.openmeeg rather than cfg.om, for consistency with ft_prepare_headmodel

ft_prepare_leadfield.m
```diff
@@ -47,6 +47,22 @@
...
+%   cfg.om.batchsize = scalar (default 100e3), number of voxels per DSM batch. 
+%                       Set to e.g. 1000 if not much RAM available
+%   cfg.om.dsm       = 'no'/'yes', reuse existing DSM if provided
+%   cfg.om.keepdsm   = 'no'/'yes', option to retain DSM (no by default)
+%   cfg.om.nonadaptive = 'no'/'yes'
+%  
+% For SINGLESHELL based headmodels:
+%   cfg.om.batchsize = scalar (default 1), but can be 'all'. Number of
```
robertoostenveld: the option should be called cfg.singleshell rather than cfg.om

ft_prepare_leadfield.m
```diff
@@ -178,78 +194,105 @@
   fprintf('calculating leadfield for all positions at once, this may take a while...\n');
   
   if(~isfield(cfg,'om'))
```
robertoostenveld: this should be
cfg.openmeeg = ft_getopt(cfg, 'openmeeg', [])

ft_headmovement.m
schoffelen: Fair point. I will give it some thought. Perhaps, the 'original' can go altogether, and the average/split option can be used. Also, I realize that for extreme consistency (and for ft_analysispipeline to be able to work) we should probably add a cfg field to the grad-structures, in which implicitly (e.g. in cfg.previous.trl in combination with cfg.trials) the 'trialinfo' can be represented.

private/topoplot_common.m
```diff
@@ -52,13 +54,17 @@
 % the call with multiple inputs is done by ft_topoplotIC and recursively by ft_topoplotER/TFR itself
 % the last input argument points to the specific data to be plotted
 if Ndata>1 && ~isnumeric(varargin{end})
+  
   for k=1:Ndata
     
-    if k>1
+    if k>1 && istrue(cfg.newfigure)
```
robertoostenveld: please integrate this as
if k>1
  if istrue(cfg.newfigure)
    ...
  elseif ~istrue(cfg.newfigure)  
    ...
  end
end 

How will this play with interactive figures?

ft_headmovement.m
robertoostenveld: regarding grad.cfg, after ft_electroderealign the output elec does get a cfg.

robertoostenveld:
regarding trialinfo, that usually contains events, not trial numbers per see. Would it not make more sense to indicate the correspondence between a grad and the data using sampleinfo?

Like
 grad(i).cfg.sampleinfo = Mx2

Here is a use case: lets say we specify

```
cfg = [];
cfg.trl = Nx2; say 100 trials, slowly sagging down
cfg.trials = [1:10]; % this would be one cluster of positions, hand-determined here
cfg.output = 'average'
grad = ft_headmovement(cfg)
```

in that case the output grad could be similar as the first grad(1) output that would result from automatic clustering on all trials (with 10 clusters), since that would also put the first 10 trials together in a cluster. 

This separates the clustering from the averaging over the cluster.

ft_prepare_leadfield.m
```diff
@@ -131,6 +131,7 @@
 % check if the input cfg is valid for this function
 cfg = ft_checkconfig(cfg, 'renamed', {'hdmfile', 'headmodel'});
 cfg = ft_checkconfig(cfg, 'renamed', {'vol',     'headmodel'});
+cfg = ft_checkconfig(cfg, 'renamed', {'om',      'openmeeg'});
```
robertoostenveld: 👍🏻

forward/ft_average_sens.m
```diff
@@ -74,16 +73,16 @@
...
 
 % starting with the second PC works a little better (e.g. on BTi)
 x1 = pos*u(:, 2);
 x2 = pos*u(:, 1);
 
 % detemine the indices of three reference sensors that are close to the three principal axes
-[m ind1] = min(x1);
-[m ind2] = max(x1);
-[m ind3] = max(abs(x2 - mean(x2([ind1 ind2]))));
+[~, ind1] = min(x1);
```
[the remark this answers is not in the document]
schoffelen: yes, perhaps I'm overdoing it, but the aim was to get rid of the 'orange bars' in the matlab editor...

private/topoplot_common.m
schoffelen: seems just fine. when selecting some channels in one of the subplots it proceeds to a singleplotER representation.

schoffelen:
OK, after some thought I have decided on the following:

Rather than outputting a grad structure, the function now outputs a data structure, containing the HLC data, or a cell-array of data structures, each of which with a head positions specific grad description. This was the best I could come up with. I did not fancy adding a sampleinfo to a grad structure (because this would call for an extension of the definition of a sens-structure), the sampleinfo is now retained in the data structure.

robertoostenveld:
ok, sounds good. Feel free to merge. It would be nice to have an example page on the wiki that explains it.

PR #755, opened 2018-08-01
import event info for mne -epo.fif

stfnrpplngr:
I am trying to complement the mne-fieltrip integration offering import functionality for Epochs.

It mostly works, but there still is a bug somewhere. In cfg.event every trial appears a second time being labelled as 'trial'. Maybe someone could have a look?

PR #769, opened 2018-08-10
Improve read-in of epoched fif

stfnrpplngr:
Event codes are saved as numeric instead of strings. This is more in line with the general processing procedure because `data.cfg.trl` is appended with a fourth column containing the event codes.
This PR modifies recent PR #755, sorry for bringing it up again.


fileio/ft_read_event.m
```diff
@@ -1548,8 +1548,9 @@
         for i=1:hdr.nTrials 
             event(end+1).type      = 'trial';
             event(end  ).sample    = begsample(i);
-            event(end  ).value     = events_label(events_code == hdr.orig.epochs.events(i, 3), :);
+            event(end  ).value     = hdr.orig.epochs.events(i, 3);
             event(end  ).offset    = -hdr.nSamplesPre;
+            event(end  ).label     = events_label(events_code == hdr.orig.epochs.events(i, 3), :); % is not used in further functions
```
robertoostenveld: this breaks the API and documentation of the function, see http://www.fieldtriptoolbox.org/development/fileio. That is bad,  not only for FT but also for other toolboxes that use fileio (like SPM and EEGLAB).

event.value can be either numeric or strings. If all string-values can be interpreted as numeric, then they should be converted to numbers.

I propose something like

```
if ~any(isnan(cellfun(@str2double, {event.value})))
  % they can all be converted to numeric values
  for i=1:numel(event)
    event(i).value = str2double(event(i).value);
 end
end 
```

robertoostenveld: this be after the for loop, and event.label would not be created.

stfnrpplngr: Then, I would propose to remove all processing of the event labels. I was looking for a way to get the label from MNE-Python into Fieldtrip. The event information in Python saved as dictionary. Matlab reads it as string vector:

`hdr.orig.epochs.event_id = 'cond1:11;cond2:12;cond3:13;cond4:14'`

In my first attempt, I focussed on the label ignoring the numeric code. But that makes further processing invonvenient because of the missing vector in `data.cfg.trl`.

Probably it is best to focus on the numeric values and the user has to know the corresponding mapping.

robertoostenveld:
just for background: multiple external events that relate to the same piece of data (or time segment) are in FT style coded as multiple events. If you for example use ft_read_event on Subject01.ds, you will see many that are in principle duplicates of each other, but represented differently. 
Perhaps that is something to consider here as well, i.e. you might want to map 'cond1:11;cond2:12;cond3:13;cond4:14' onto 4 events like

```
event(1).type = cond1
event(1).value =11
event(2).type = cond2
event(2).value =12
event(3).type = cond3
event(3).value =13
event(4).type = cond4
event(4).value =14
```

where sample, offset and duration would all be the same. Also since trials often have an intrinsic time-axis, I would expect offset to be non-zero. Again: have a look at ft_read_event(Subject01.ds), which is a CTF trial-based (non continuous) dataset. We ignore that in our tutorials and largely pretend that it is a continuous dataset, since that is how we nowadays recommend people to record (no gaps between the trials).

robertoostenveld:
in ft_definetrial you could then specify 

```
cfg = [];
cfg.trialdef.eventtype = 'cond4';
cfg = ft_definetrial(cfg)
data = ft_preprocessing(cfg)
```

Just like how it would work (at least should work, if my memory serves me well) for Subject01.ds by specifying cfg.trialdef.eventtype='trial'.

stfnrpplngr:
That was my first attempt, too :wink:. But, doing this, the prerequesites for line [2153](https://github.com/fieldtrip/fieldtrip/blob/4d5588d50ccee9be78d99410487dffde0a5eb43c/fileio/ft_read_event.m#L2153) are fulfilled. Then the subsequent loop is processed and every trial appears twice in `data.cfg.trl` lacking the eventvalue. Furthermore, by this approach it is not possible to do the same preprocessing in one step to all trials.

schoffelen:
Hi Stefan, sorry for being so extremely slow addressing this. At the moment this PR is still open, and it's been so long, that we don't know anymore whether this still needs to be merged, and if so, how? 
If you think that this still deserves to be merged, could you then check in a clean version of the changes, i.e. without conflicts, so that we can take it from there?

schoffelen:
Hi @stfnrpplngr please advice how to proceed from here? See my previous comment

schoffelen:
Hi Stefan, @stfnrpplngr please advice how to proceed from here? See my previous comment

robertoostenveld:
I had a go at this. Using https://mne.tools/0.15/auto_tutorials/plot_creating_data_structures.html#tut-creating-data-structures and https://mne.tools/0.15/auto_tutorials/plot_object_epochs.html I made an example file.

This resulted in 

```
Error using cat
Dimensions of arrays being concatenated are not consistent.
Error in cell2mat (line 83)
            m{n} = cat(1,c{:,n});
Error in ft_read_event (line 1711)
      events_label = cell2mat(events_id(:, 1)); 
83              m{n} = cat(1,c{:,n});
```

I will post my suggested code improvements in a separate PR.

Issue #774, opened 2018-08-17
test_ricoh started failing because of a difference in selected EEG channels


robertoostenveld:
The dataset used to have 65 EEG channels, now it has 42. This is consistent with hdr.chantype. 

@KiwamuK, can you please have a look at this? It might be that previous channels marked as EEG are now better classifified as EMG, ECG etc. In that case just the test script should be updated.

robertoostenveld:
Oh, please do a "git pull" to get the latest version. I already made a small change to the test script: it is now ft_channelselection with "EEG" rather than "EEG*", since "EEG*" is searching for channels starting with "EEG", which is not the case any more.

KiwamuK:
Thank you for your help. 
Now I have done a pull request for test_ricoh.m. 
I have modified the number of the eeg labels to its correct number 42.

robertoostenveld:
thanks!

PR #775, opened 2018-08-17
Modified test_ricoh.m

KiwamuK:
I have modified the number of the eeg labels to its correct number 42.


robertoostenveld:
This solves #774

PR #779, opened 2018-08-25
fix: potential misordering bug in ft_appensens bug related to matlab's unique operation 

StolkArjen:
Matlab's unique operation automatically sorts the output. However, it does this on the basis of lowest value in the case of digits, and alphabetically in the case of strings. 

ft_appendsens uses the unique operation to find duplicate channels. When those are found, the unique outputs, i.e. channel labels and positions, are at risk of being misordered relative to each other because the chance of them being alphabetically and numerically in agreement is obviously slim.

This fix adds the 'stable' option when calling the unique operation, enforcing the unsorted return of the unique labels and channelpos.


robertoostenveld:
thanks!

PR #784, opened 2018-08-28
clean up automatically added toolbox path at end of FT function

robertoostenveld:
This includes ft_postamble_hastoolbox, which should be called at the end of each main FT function. In conjunction with ft_hastoolbox it cleans up the path in case additional toolboxes were added.

This should a/o facilitate different versions of spm to be used in different sections of a users analysis pipeline.


robertoostenveld:
this should still be implemented in the majority of the FT main functions

PR #787, opened 2018-08-30
enables plotting electrodes using multiple colors

ludovicbellier:
Hi there,

Here's a simple fix to enable plotting electrodes using multiple colors.

I wanted to plot electrodes on a pial surface, using a color code to represent a specific metric. To do this, I originally coded a loop over all electrodes to specify their color individually. Indeed, providing a Nx3 matrix to ft_plot_sens didn't work (N being the number of electrodes), as plot3 only takes a 1x3 array as facecolor.
Replacing plot3 by scatter3 did the trick, and besides simplifying the code a lot (only one call to ft_plot_sens, instead of N), time to generate the full figure is drastically reduced.

Best,
Ludovic

### DIFF
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -335,7 +335,7 @@ switch sensshape
       end
     else
       % the style is not specified, use facecolor for the marker
-      hs = plot3(pos(:,1), pos(:,2), pos(:,3), 'Marker', marker, 'MarkerSize', sensize, 'Color', facecolor, 'Linestyle', 'none');
+      hs = scatter3(pos(:,1), pos(:,2), pos(:,3), sensize.^2, facecolor, marker);
     end
     
   case 'circle'


robertoostenveld:
nice! I only now noticed that the variable "sensize" misses an s. Something for the next time we work on this function.

ludovicbellier:
@StolkArjen Hmm not in an easy way, if you mean specifying both multiple edge colors and multiple face colors. For the record, both the properties you mentioned take as input either 'flat', 'none' or a RGB triplet (or the corresponding string label, e.g. 'r' for [1 0 0]), but not a Nx3 matrix (N RGB triplets, N being the number of channels). When using 'flat', it takes the values in CData, here set as the Nx3 matrix named facecolor. So you could have multiple face colors (from CData/facecolor), and a common edge colors (from an RGB triplet), or vice versa. Besides, the default '.' marker only has an edge, no face, so to do what you want you would need to use the 'o' marker, for example, Anyway, as you see it's not a straightforward thing to do using scatter3, and the easiest way I can think of would be to loop over electrodes. To end up on a positive note, I think that having both multiple face colors and multiple edge colors would be hard on the eyes! =) Might be more readable to use multiple face colors and multiple marker sizes, which is fully doable using scatter3 and thus one call of ft_plot_sens.

@robertoostenveld Yes I noticed that too but thought it was a choice. Will think about it if I word again on this function.

StolkArjen:
Thanks for this elaboration. FYI, I just realized we previously introduced ft_plot_cloud, which is the workhorse for ft_sourceplot when it comes to plotting, shaping, and coloring the electrodes. You could also use it independently, for instance in combination with ft_plot_mesh. Similarly to ft_plot_sens, ft_plot_cloud does not allow controlling edge and facecolors separately, I thought.
 
np-p170458b stolk suppl fig 1-01

PR #798, opened 2018-09-12
small showcallinfo bug fixed

dmatthes1982:
Function ft_topoplotIC – small bug fix → if cfg.showcallinfo is set to no, now really nothing will be displayed in the command window.


robertoostenveld:
thanks

PR #801, opened 2018-09-14
make the order for the forward computation configurable

robertoostenveld:
Both for concentricspheres and for single shell there is an "order" that determines the accuracy. This PR makes it configurable in ft_prepare_leadfield. 

See also http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3453


robertoostenveld:
@simonhomoelle this implements it. Please check the test script for usage.

PR #816, opened 2018-09-21
add events_tsv as a filetype and implemented an event-constructor

schoffelen:


schoffelen:
zoiets?

robertoostenveld:
looks good, although I cannot follow all the heuristics for the low level code. It won't break anything right now. Let's just see whether it works in practice.

PR #824, opened 2018-10-03
ensure consitent walltime and line endings in test scripts

robertoostenveld:


robertoostenveld:
there was one file with DOS line endings, which caused the test batch to abort halfway

PR #825, opened 2018-10-06
fix: this fixes a too many open files error

StolkArjen:
Which would make matlab crash. With the below fix, each opened file is closed after use.


schoffelen:
merci

StolkArjen:
Merci a toi

PR #826, opened 2018-10-07
Bugfix - fixed the problem in Bugzilla bug 3452

NirOfir:
Fixed a bug where the function provides in the feedback the wrong number of selected trials, when logical cfg.trials is used.

[Details here](http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3452)


schoffelen:
thanks.

PR #828, opened 2018-10-08
ENH - removed backward-compatibility breaking call to nargoutchk

schoffelen:


schoffelen:
this should address the issue reported by Guillaume

PR #837, opened 2018-10-10
Cellfunction

schoffelen:


schoffelen:
note that besides fixing the cell functions, it also updates some code where it is used (ft_denoise_tsr)

Issue #831, opened 2018-10-09
Use of tilde to ignore outputs


gllmflndn:
Tilde is used in a few places in `ft_read_header.m` to ignore function outputs:
  https://github.com/fieldtrip/fieldtrip/blob/master/fileio/ft_read_header.m#L444-L475
Would it be possible not to use them for backward compatibility? They are anyway of no use in this context (last outputs of `fileparts`).

robertoostenveld:
On http://www.fieldtriptoolbox.org/development/guidelines/code we state "the ~ output argument is not supported in MATLAB versions < 2009b and should be avoided"

But we also state elsewhere that version up to 5y old are to be supported, and 2009b is now 9 years old. I suspect that causes the tilde to creep into the code by now.

I am fine with trying to keep the tilde out of existing code. I recommend using the variable "dum" instead, as I know that is used elsewhere.

Could you do a PR?

gllmflndn:
I came across an issue with `ft_hastoolbox` and old MATLAB versions that was due to the use of `~`. Looking at the current code, it seems that its use is widespread in the code:

```
data2bids.m:        [~, f, ~] = fileparts(cfg.dataset);
data2bids.m:  [p, f, ~] = fileparts(cfg.outputfile);
data2bids.m:  [p, f, ~] = fileparts(cfg.outputfile);
data2bids.m:  [p, f, ~] = fileparts(cfg.outputfile);
data2bids.m:  [p, f, ~] = fileparts(cfg.outputfile);
data2bids.m:[~, f, ~] = fileparts(f);
ft_prepare_layout.m:      [p, f, ~] = fileparts(cfg.layout);
fileio/ft_read_event.m:    [p, ~, ~] = fileparts(filename);
private/copy_brainvision_files.m:[~, f, ~] = fileparts(newheaderfile);
private/copy_brainvision_files.m:  [~, ff, ~] = fileparts(oldheaderfile); % re-reading as sometimes weird names comes up
private/copy_brainvision_files.m:  [~, ff, ~] = fileparts(oldheaderfile); % re-reading as sometimes weird names comes up
private/copy_ctf_files.m:[~, fo, ~] = fileparts(oldname);
private/copy_ctf_files.m:[~, fn, ~] = fileparts(newname);
private/estimate_fwhm2.m:%   [s,~] = gaussfit3D(dpos(sel,:), Cmat(sel,k));
private/ft_inv.m:    [kappa,~] = ginput(1);
[...]

PR #839, opened 2018-10-10
Remove tilde when used to ignore function outputs

gllmflndn:
As discussed in https://github.com/fieldtrip/fieldtrip/issues/831 for backward compatibility with older MATLAB versions.
`~` is replaced by `dum` or removed when used for last argument of a function (as long as it doesn't affect behaviour for remaining arguments as in `size`, `spm_prep2sn` or `svd`).


robertoostenveld:
thanks. I'll keep an eye open for the regression tests that will run automatically this night.

PR #845, opened 2018-10-17
ENH - solved regression error that was due to eps (1e-15) difference

robertoostenveld:
…in vertices, replaced keyval with ft_getopt, fixed unrelated bug in isalmostequal

### DIFF
--- a/private/isalmostequal.m
+++ b/private/isalmostequal.m
@@ -27,7 +27,7 @@ if nargin==3
   % for backward compatibility
   depth = varargin{1};
 else
-  depth = keyval('depth', varargin);
+  depth = ft_getopt(varargin, 'depth')
   if isempty(depth)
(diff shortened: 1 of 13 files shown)


private/isalmostequal.m
```diff
@@ -110,7 +110,7 @@
     
     if ~isempty(abstol) && any(abs(a-b)>abstol)
       message{end+1} = sprintf('different values in %s', location);
-    elseif ~isempty(reltol) && any((abs(a-b)./(0.5*(a+b)))>reltol)
+    elseif ~isempty(reltol) && any((abs(a-b)./(0.5*abs(a+b)))>reltol)
```
robertoostenveld: when a and/or b would be sufficiently negative, the difference would be negative and hence smaller than reltol

PR #846, opened 2018-10-17
Support for Curry formats based on manufacturer's code

vlitvak:


vlitvak:
Don't know about EOF. Perhaps something to discuss with Reyko via that
support e-mail.

PR #868, opened 2018-11-06
bugfix2673 - ft_selectdata added at the begining to select trials/cha…

dieloz:
…nnels/latency

### DIFF
--- a/ft_rejectvisual.m
+++ b/ft_rejectvisual.m
@@ -204,6 +204,11 @@ elseif (strcmp(cfg.latency, 'poststim'))
   cfg.latency(2) = maxperlength(2);
 end
 
+selcfg = keepfields(cfg, {'trials', 'channel','latency', 'showcallinfo'});
+data = ft_selectdata(selcfg, data);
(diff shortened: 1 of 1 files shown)


schoffelen:
Thanks, jefe

dieloz:
you're welcome!

robertoostenveld:
Hi @dieloz, it seems that this PR broke something, which was just now detected by @ualsbombe in #1152. Could you please have a look at that issue as reported by Lau?

dieloz:
Sorry for beig so late
At some point trial/channel selection had no effect
http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2673#c3
But I failed to see that ft_rejectvisual has its own functions to select data (rejectvisual_trial, rejectvisual_channel and rejectvisual_summary). Makes sense to be that when you make a channel selection, the data output should contain only those selected ones. If this is not the case, let me know and I'll revert the changes given that channel/trial selection for ft_rejectvisual is the exception

Issue #871, opened 2018-11-08
ft_rejectvisual can not handly unequal triallengths


TinekeSnijders:
Using trials of different lengths seems to give a problem in ft_rejectvisual. It wasn’t a problem before (I think in 2016) when I did the same analysis.
When using ft_rejectvisual, in the current fieldtrip version in the output data all trial lengths are reduced to the shortest trial-length, cutting away the remainder of the trial.

schoffelen:
tja, met ft_selectdata_old hadden we dat probleem niet :)

schoffelen:
the internal representation of the time-axes, on which the latency selection is done, is that of a cell-array, intended to represent multiple input datasets, rather than multiple trials within a single dataset. The default behavior of ft_selectdata is to prune (with multiple datasets) the data structures' dimensions such as to return the 'intersect', i.e. the space-time-freq dimensions that all input arguments share. This means that for a 'raw' type data structure the time-axes are pruned prior to the selection of time points as per the definition of cfg.latency. (If cfg.latency = 'all' this step is skipped).

I amended ft_selectdata to treat the datatype = 'raw' as an exception. test_ft_selectdata runs through without problems, but I am not fully sure whether there are potential adverse side effects, although I cannot think of any at this moment.

Will submit this as a PR shortly.

schoffelen:
I think that this has been solved (with some remaining post merge problems addressed)

PR #873, opened 2018-11-09
FIX - output variable length trials with raw data in the input

schoffelen:


schoffelen:
This pertains to issue #871

PR #878, opened 2018-11-16
Read the data once instead of twice

raintwoto:
The m00 file was read twice in the original code. I commented the one in the hdr read.

### DIFF
--- a/fileio/private/read_nihonkohden_hdr.m
+++ b/fileio/private/read_nihonkohden_hdr.m
@@ -89,16 +89,17 @@ hdr.nSamplesPre = 0;
 %ch_names=char(ch_names{1});
 
 
-dc = textscan(fid,'%f',hdr.nChans*hdr.nSamples,'delimiter','\n','headerlines',0);
-df = dc{1};clear dc;
+% dc = textscan(fid,'%f',hdr.nChans*hdr.nSamples,'delimiter','\n','headerlines',0);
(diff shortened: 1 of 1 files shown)


StolkArjen:
Hi Biao, 
Could you be more specific?

raintwoto:
In the original code, in the "read_nihonkohden_hdr.m", it reads the data once and then it read the data the second time in "read_nihonkohden_m00".  It cost much more time for loading the data.

dieloz:
I remember I was not able to textscan the first part of header without going through the entire datafile. I can take a look again next week unless somebody has a good shortcut

StolkArjen:
Thanks, and yes, that would be important to know. Do functions the like of ft_read_header and ft_read_event still work without reading the data in read_nihonkohden_hdr?

dieloz:
I would say they wont work

raintwoto:
It runs quite well in my computer(when I use ft_preprocessing)

robertoostenveld:
I have a better solution which I will commit soon.

robertoostenveld:
@raintwoto please test this change to the code.

raintwoto:
will do that

raintwoto:
> @raintwoto please test this change to the code.

I tested the script and found out the reason for my temp file problem (because of the Chinese character in my username). The script runs very well now

PR #885, opened 2018-11-22
Testbatch failure

schoffelen:
this patch includes multiple small changes that caused a variety of test functions to fail after I reimplemented ft_timelockanalysis.


robertoostenveld:
👍🏻

PR #886, opened 2018-11-24
broadened support for reading impedance values from BrainVision .vhdr files

jdien07:
This update broadens support for reading impedance values from BrainVision .vhdr files by correctly parsing .vhdr files generated by older PyCorder versions (by handling double spaced files) and those generated by a current version of BrainVision Recorder (which uses a different format).


robertoostenveld:
looks good as far as I can tell. Could you send me a test file from the old  PyCorder versions (over dropbox, google drive, wetransfer, or so). That would also be useful for the Python implementation in relation to BIDS.

robertoostenveld:
I understand, no problem.

PR #890, opened 2018-11-27
ENH - added test script for stack trace of error messages

robertoostenveld:
See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3336


robertoostenveld:
@schoffelen using the following code I would say that the error messages are consistent and not confusing. Both show the same (similar) stack trace with three levels.

```
function test_bug3336

% there are two versions: one with error() and one with ft_error()
% these should give the same result

try
  subfunction1
catch
  err1 = lasterror;
end

try
  subfunction2
catch
  err2 = lasterror;
end

assert(~isequal(err1.message, err2.message))
assert(isequal(err1.identifier, err2.identifier))
assert(numel(err1.stack)==3)
assert(numel(err2.stack)==3)

%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% VERSION 2
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

function subfunction1
subsubfunction1

function subsubfunction1
error('SOME:ERROR', 'something went wrong 1');

%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
% VERSION 2
%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%

function subfunction2
subsubfunction2

function subsubfunction2
ft_error('SOME:ERROR', 'something went wrong 2');
```

schoffelen:
Top

PR #891, opened 2018-11-28
adding new functions to transform md website format t pure matlab

aranas:
see also fieldtrip/website issue # 14.

The new function markdown2matlab takes a .md file (ie a tutorial from the new website) and transforms it into a .m file commenting out all markdown syntax and preserving only the matlab code. 
matlab2markdown is the reverse.

Mainly it transforms heading syntax into Matlabs Publish MarkUp syntax and takes care of empty lines around code, as well as structure examples which are also marked as code on the website and the %include functionality.

This is just a first go at this. To have this connect to ft_documentreference and other utilities, we probably should discuss exactly how this conversion will be used and how much it needs to generalize. for now I just implemented it with what I thought was most necessary and tested it on a couple of the already existing tutorial files from the website.

At the moment it is completely disregarding the website specific structures such as header, footer etc. But this should be easy to add.
I'm Happy to hear any feedback of how this needs to be changed & improved to fit the intended purpose.

### DIFF
--- a/utilities/markdown2matlab.m
+++ b/utilities/markdown2matlab.m
@@ -1,0 +1,111 @@
+function markdown2matlab(infile,outfile,varargin)
+
+%This function converts a markdown file to a matlab file, commenting out
+%all text, only leaving the code uncommented. As input it takes the name of
+%the file to be converted. Best is to provide the full filepath, otherwise
+%it will look for the file within the current path
+%In addition to the code in the current markdown file it will also add
+%any external markdown files that are connected via the include syntax.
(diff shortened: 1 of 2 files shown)


aranas:
@robertoostenveld @dieloz

dieloz:
I'll definitely give a try when I'll begin the artifact rejection EEG resting state example/tutorial. This code reminds me the matlab2julia converter (in case you're interested):
http://sciencecow.mit.edu/matlab-to-julia/
http://sciencecow.mit.edu/matlab-to-julia/matlab_to_julia_translator.js

PR #893, opened 2018-11-30
small fix to formatting of monospaced inline code, added support for …

robertoostenveld:
…Jekyll header and for inclusion of figures

@aranas as discussed


robertoostenveld:
I also added the example (now with figure) to the shared dropbox

PR #909, opened 2018-12-17
enh: plot elecs as discs aligned to a headshape

StolkArjen:
This is made with ECoG grids/strips in mind. For instance, using
ft_plot_sens(elecnonstim, 'elecshape', 'disc', 'headshape', hull, 'facecolor', [0 0 0])
one could make plots like these (note how the electrode surfaces are aligned with the outline of cortex):

screen shot 2018-12-17 at 10 24 57 am

PR #913, opened 2018-12-27
 ENH: added disc plotting utility to ft_plot_cloud

StolkArjen:
This pertains to the same functionality as recently added to ft_plot_sens. The reason for also adding it to ft_plot_cloud is that ft_plot_cloud allows automatic coloring (and size-scaling) of the discs relative to the functional data.

### DIFF
--- a/plotting/ft_plot_cloud.m
+++ b/plotting/ft_plot_cloud.m
@@ -1,7 +1,7 @@
 function ft_plot_cloud(pos, val, varargin)
 
-% FT_PLOT_CLOUD visualizes spatially sparse scalar data as points, spheres, or 
-% spherical clouds of points and optionally 2D slices through the spherical 
+% FT_PLOT_CLOUD visualizes spatially sparse scalar data as points, spheres, or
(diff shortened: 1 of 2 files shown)


StolkArjen:
@robertoostenveld or @schoffelen, can you see what's going on here? There doesn't seem to be any problem whatsoever, yet there's this "failing check" concerning just a few lines of documentation of a function.

StolkArjen:
Re-pushed this branch, after a minor change to the documentation. Seems to pass the build test now?

StolkArjen:
@robertoostenveld or @schoffelen, please check and pull this request. It allows ft_plot_cloud to plot electrodes as discs (see https://github.com/fieldtrip/fieldtrip/pull/909) and color those discs based on the functional data.

robertoostenveld:
it would be good to extend the test_ft_plot_cloud function with these additional options.

StolkArjen:
Good point, will do in the context of a planned overview of the current SEEG/ECoG plotting tools in either the plotting tutorial or somewhere else.

PR #914, opened 2019-01-08
improve selection of trials for spike data

robertoostenveld:
plus some other not-so-important changes


robertoostenveld:
@stephenwhitmarsh this is to help you with your question

robertoostenveld:
@stephenwhitmarsh please get the latest version and try
```
data = ft_spikesimulation([])
cfg = [];
cfg.trials = [1 0 0 0 0 0 0 0 0 0 ];
stat = ft_spike_xcorr(cfg, data)
``` 
and you should see that it is possible to run `ft_spike_xcorr` on a single trial.

robertoostenveld:
I guess the issue is that your data is not correctly formatted. I also should admit that the spike code is not as clean as I would like it to be...

stephenwhitmarsh:
Merci!
However, now *ft_spike_plot_raster *will not apply cfg.latency, and instead
will always plot the full trials. Not sure if this bug is introduced just
now, or just because I updated.
It is caused by the following (Line 108), because I give as a third
argument my spike density. IMHO this should not override the cfg, at least
not when the sdf timecourse is larger than the second argument, right?:

% check if a third input is present, and check if it's a timelock structure
if nargin==3
  doTopData = true;
  timelock  = ft_checkdata(timelock, 'datatype', 'timelock', 'hastrials',
'no', 'feedback', 'yes');
  if isfield(timelock,'cfg') && isfield(timelock.cfg, 'latency')
*    cfg.latency = timelock.cfg.latency; *
  end
else
  doTopData = false;
end

robertoostenveld:
please check whether it is this update or not. The only change in ft_spike_plot_raster is in the selection of trials.

I suspect it might be this previous commit fd06844647a3be3a6a01f89c6d29df2a67a7eb85 that caused something to behave different

PR #917, opened 2019-01-09
Update ft_freqbaseline.m

raghav0706:
Having a zscore option would be helpful.

### DIFF
--- a/ft_freqbaseline.m
+++ b/ft_freqbaseline.m
@@ -9,7 +9,7 @@ function [freq] = ft_freqbaseline(cfg, freq)
 %   cfg.baseline     = [begin end] (default = 'no'), alternatively an
 %                      Nfreq x 2 matrix can be specified, that provides
 %                      frequency specific baseline windows.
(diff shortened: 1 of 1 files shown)


robertoostenveld:
can you also add it to the documentation of `ft_freqbaseline` and can you update the documentation of the functions that use `ft_freqbaseline` prior to actual plotting? That would be ft_movieplotTFR, ft_multiplotTFR and ft_singleplotTFR. 

Doing "grep" also reveals ft_multiplotER and ft_singleplotER calling ft_freqbaseline. This applies in the case they receive a TFR input, in which the user can specify a frequency (or frequency range) in which the data is selected (or averaged) and the power time course is subsequently plotted. The cfg.baselinetype option is not documented in the help of those 2 functions, so no reason to update them.

robertoostenveld:
With #918 I only checked `ft_movieplotER`, since it was missing from the grep (which turned out to be OK). I leave the documentation change to you.

robertoostenveld:
please note that you can still change back to this branch, make more changes and push them as well. They will subsequently appear here and I can combine them in one merge.

robertoostenveld:
thanks!

PR #923, opened 2019-01-10
fix for [Bug #3467] ft_spike_plot_raster goes not use correct latency

stephenwhitmarsh:
As requested by @schoffelen


schoffelen:
thanks, could you also update the documentation section (so that this odd behavior is also not mentioned there)?

stephenwhitmarsh:
oops, sure, here it is

schoffelen:
top! merci

PR #929, opened 2019-01-17
Bad channel current density

johalgermissen:
detectchannelnan (new):
- Create new function to detect whether any channels contain NaNs in first sample of first trial; if so, then add to cfg.badchannel 
(to be called in ft_channelrepair and ft_scalpcurrentdensity):

ft_channelrepair (adapted):
- Add detectchannelnan to add channels containing NaNs to bad channels

ft_scalpcurrentdensity (adapted):
- Add detectchannelnan to add channels containing NaNs to bad channels
- Add cfg.badchannels functionality for function spline so as to a) only use "good channels" (where potential is known) for estimation, but b) estimate scd for both good (where potential is known) and bad (where potential is unknown) channel positions
- Methods finite and hjorth throw error in presence of bad channels

test_ft_scalpcurrentdensity (new):
- test methods spline and finite ft_scalpcurrentdensity using data from bug2685
- tests for method spline:
- in absence of bad channels
- in presence of bad channels
- with constant offset
- with channels shuffled
--> always plot butterfly plot
--> always compare to scd in absence of bad channels, throw error if too large discrepancy
- method finite should throw error in presence of bad channels


ft_scalpcurrentdensity.m
```diff
@@ -182,35 +184,56 @@
...
-    % are interested here
-
-    ft_progress(trlop/Ntrials, 'computing SCD for trial %d of %d', trlop, Ntrials);
-    [V2, L2, L1] = splint(elec.chanpos, data.trial{trlop}, [0 0 1], cfg.order, cfg.degree, cfg.lambda);
-    scd.trial{trlop} = L1;
+      % do compute interpolation
+      ft_progress(trlop/Ntrials, 'computing SCD for trial %d of %d', trlop, Ntrials);
+      if ~isempty(cfg.badchannel) % if any bad channels to interpolate: compute scd for all channels
+          fprintf('computing scd also at locations of bad channels');
+          [V2, L2, L1] = splint(goodchanpos, data.trial{trlop}(goodchanidx,:), allchanpos, cfg.order, cfg.degree, cfg.lambda);
```
robertoostenveld: goodchanidx is sorted according to the electrode channels, not sorted according to the data channels. They could be different.

private/detectchannelnan.m
```diff
@@ -0,0 +1,35 @@
+function [badchannels] = detectchannelnan(cfg,data)
```
robertoostenveld: if this is a private function, it should be called by multiple high-level functions. Otherwise it does not make sense to make a separate function out of it.

robertoostenveld: oh, it is now (in this PR) called by two functions. So that is fine

robertoostenveld: the code actually simplifies to a one-liner. No need to make a private function out of it.

private/detectchannelnan.m
```diff
@@ -0,0 +1,35 @@
...
+
+% DETECTCHANNELNAN detects whether one or more channels contain only NaNs
+% and if so adds them to cfg.badchannels
+%
+% Copyright (C) 2003, Robert Oostenveld
```
robertoostenveld: I cannot claim copyright (and certainly not in 2003) for a function that you wrote.

robertoostenveld:
I made some online (i.e. on the github website) changes to the code. I will now check it out and evaluate the test script.

ft_scalpcurrentdensity.m
johalgermissen: Yes, you're right! 
Where I previously defined goodchanpos, let's separately define
goodChanElecidx (to retrieve channel positions of valid channels from the elec file) and goodChanDataidx (for selection of valid channels in the data when computing the spline). So from line 186 onwards:

goodChan = setdiff(data.label,cfg.badchannel);
goodChanDataidx = ismember(data.label, goodChan);
goodChanElecidx = ismember(tmp.label, goodChan);
goodchanpos = tmp.chanpos(goodChanElecidx ,:);
allchanpos = tmp.chanpos;
if isfield(tmp, 'elecpos')
  elec.elecpos = tmp.elecpos;
end
elec.label   = tmp.label;

% compute SCD for each trial
if strcmp(cfg.method, 'spline')
  ft_progress('init', cfg.feedback, 'computing SCD for trial...')
  for trlop=1:Ntrials
      % do compute interpolation
      ft_progress(trlop/Ntrials, 'computing SCD for trial %d of %d', trlop, Ntrials);
      if ~isempty(cfg.badchannel) % if any bad channels to interpolate: compute scd for all channels
          [V2, L2, L1] = splint(goodchanpos, data.trial{trlop}(goodChanDataidx,:), allchanpos, cfg.order, cfg.degree, cfg.lambda);
          scd.trial{trlop} = L2;
      else % if all channels good: just compute scd for input channels, specify arbitrary single channel to-be-discarded for interpolation (saves >50% computation time)
          [V2, L2, L1] = splint(goodchanpos, data.trial{trlop}(goodChanDataidx,:), [0 0 1], cfg.order, cfg.degree, cfg.lambda);
          scd.trial{trlop} = L1;
      end
  end
  ft_progress('close');

private/detectchannelnan.m
johalgermissen: Depends on how elaborate the function should be. My version of detectchannelnan()
- only runs if there are any NaNs detected, which (notably) speeds up things for complete data.
- throws a warning for each channel that is classified as bad, alerting users
- combines the detect channels containing NaNs with any bad channels pre-specified by the user (which might not contain NaNs).

A one-liner without this functionality would of course be possible.
Is there anything more I should/ can do here?

private/detectchannelnan.m
johalgermissen:
```suggestion
% Copyright (C) 2019, Johannes Algermissen
```

johalgermissen:
Anything more I can/should do?

robertoostenveld:
I have it in a local branch on my laptop at the moment for further review and testing and will push it from there. Please don't make any other changes (as otherwise I have to merge them again). So nothing for you to do for now.

robertoostenveld:
ah, I see why the test script did not catch the indexing error; you were shuffling the channel order in the data in the same way as the channel order in the electrodes. I will add another test case.

robertoostenveld:
Note that I am surprised how well it still works if half of the channels are filled with NaNs.

robertoostenveld:
Please also check out the new test script: it has improved visualization of the data. Have a look in the interactive plot at the central channels, and then around 100ms, and you'll see that the CSD nicely resolves two bilateral sources which in the ERP are one big blob.

johalgermissen:
Sorry but where did you locate the new test script?
I called my script test/test_ft_scalpcurrentdensity.m, but I can't find that in the master branch anymore.
There's a script test_pull929.m, which I suspect is the test for this pull request, but it's empty (0 bytes)?

robertoostenveld:
> There's a script test_pull929.m, which I suspect is the test for this pull request, but it's empty

Hmm, how did that happen? That is indeed the one, and the name (for future reference) links to this issue/pr, which points to the 959 issue/pr with additional updates.

robertoostenveld:
that is really weird! I have searched around, and in my MATLAB history (on my local computer) I can see that I have extensively been executing that test_pull929 script for testing and finalizing the implementation.  But neither in https://github.com/fieldtrip/fieldtrip/pull/929/files nor in https://github.com/fieldtrip/fieldtrip/pull/959/files is there any reference to the content of the new file.

I will reconstruct it from the original...

PR #933, opened 2019-01-23
first version of prewhiten function

robertoostenveld:


robertoostenveld:
this replaces #783 which was borked by @schoffelen

PR #934, opened 2019-01-23
Ft inv

schoffelen:


robertoostenveld:
I propose that it also implements `method=moorepenrose` which would do this http://mathworld.wolfram.com/Moore-PenroseMatrixInverse.html and which would not support any of the lambda/kappa/tolerance options. 

I propose that `method=pinv` is renamed into `svd`, since that is the basic technique used. Using this name it is easier to document (and talk) about the other options.

I propose that `method=inv` is renamed into `vanilla` (like @sarangnemo used), indicating that it is the "plain vanilla" implementation in MATLAB. Even then you have many (numerical) sub-methods to compute it, in which we do not want to elaborate upon.

inverse/private/ft_inv.m
```diff
@@ -0,0 +1,144 @@
...
+  end
+  
+  if isempty(kappa) && isempty(tolerance) && interactive
+    figure; plot(1:m, log10(s),'o-');
+    kappa = input('Specify the number of dimensions at which the eigenvalue spectrum will be truncated: ');
+    if isempty(kappa)
+      kappa = m;
+    end    
+  elseif isempty(kappa) && isempty(tolerance)
+    % put the tolerance for the pinv to its (increased by R.O.) default
```
robertoostenveld: possibly add a bit more elaborate explanation, about matlab versions and so

inverse/private/ft_inv.m
```diff
@@ -0,0 +1,144 @@
...
+    % use kappa
+    usekappa = true;
+    usetolerance = false;
+  elseif ~isempty(tolerance)
+    usetolerance = true; % needed for winsorization
+    usekappa = false;
+    kappa = sum(s./s(1) > tolerance);
+  end
+  
+  if kappa==0
```
robertoostenveld: I think it would be good to have this (and/or similar) exceptions further to the top. E.g. empty matrix, 1x1 input matrix, tolerance=inf, etc.

inverse/private/ft_inv.m
```diff
@@ -0,0 +1,144 @@
...
+    ylabel('singular values (log10)');
+    abc = axis;
+    if ~isempty(tolerance)
+      plot([0 m], log10(tolerance).*[1 1], 'k');
+    end
+    plot(kappa.*[1 1], abc([3 4]), 'r');
+  end
+  
+  
+end
```
robertoostenveld: convention is (I would say) not to end a function with an end

inverse/private/ft_inv.m
```diff
@@ -0,0 +1,144 @@
...
+tolerance = ft_getopt(varargin, 'tolerance', []);
+kappa     = ft_getopt(varargin, 'kappa', []);
+lambda    = ft_getopt(varargin, 'lambda', 0);
+feedback  = istrue(ft_getopt(varargin, 'feedback', false));
+interactive = istrue(ft_getopt(varargin, 'interactive', false));
+
+[m, n] = size(A);
+
+if n < m
+  % recurse into ft_inv on the transposed input
```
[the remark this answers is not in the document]
schoffelen: I don't get this. The behavior is consistent with pinv. Actually, the backbone is taken from the matlab pinv.

inverse/private/ft_inv.m
```diff
@@ -0,0 +1,144 @@
...
+    ft_error('regularization is not possible when the matrix is not square');
+  end
+  
+  [U,S,V] = svd(A,0);
+  if m > 1
+    s = diag(S);
+  elseif m == 1
+    s = S(1);
+  else
+    s = 0;
```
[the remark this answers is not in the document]
schoffelen: indeed, that's why it is in the original implementation I think.

schoffelen:
NOTE to selfs: rediscuss the 0'ing of the cross-channeltype covariance in the prewhitening

schoffelen:
@vlitvak , @neurosignal this is the PR that contains the to be pushed changes to the FT code base, concerned with the stuff that was discussed last week. It is not fully tested and complete yet. What I need to think about still is the proper way of handling diagonal loading in combination with kappa truncation in ft_inv

vlitvak:
@schoffelen see http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3447 for Olaf's code that might be relevant.

> @vlitvak , @neurosignal this is the PR that contains the to be pushed changes to the FT code base, concerned with the stuff that was discussed last week. It is not fully tested and complete yet. What I need to think about still is the proper way of handling diagonal loading in combination with kappa truncation in ft_inv

schoffelen:
thanks.

inverse/private/ft_inv.m
```diff
@@ -0,0 +1,179 @@
...
+      s(s./s(1) < tolerance) = tolerance;
+    end
+    S    = diag(1./s);
+    Ainv = V*S*U';
+  otherwise
+    ft_error('Unsupported method for matrix inversion');
+end
+
+if feedback
+  figure; hold on
```
vlitvak: It would be useful to have an option to pass a handle to a figure as we do e.g.  in ft_average_sense. That way I can do the plot in SPM graphics window.

schoffelen: does this still work in newer matlab versions? In the past we used numbers for handles and that broke down when matlab upgraded their graphics handling

vlitvak: I don't know. I haven't had any problems with that but I still use 2014/2015

vlitvak: Guillaume says handles are objects in the newer version so as long as you just handle them as some variable and don't assume they are a number, there shouldn't be a problem.

schoffelen:
merging for now. let's hope that the test batch will catch any obvious problems.

vlitvak:
@schoffelen  I looked at the code of ft_inv and there are some things that I would change, but it would be good if we could agree on those things then we can hopefully share this function between FT and DAiSS.

1) As far as I understood from Olaf, Tichonov regularisation is when you use SVD and add lambda to the original eigenvalues i.e. 1/(s+lambda). When the matrix is full rank that's equivalent to what you call 'tichonov' but it can also be combined with truncation. This is not in any of the methods you implemented. We can add this, just need to agree on terminology. Either we call your 'tichonov' method something like 'ftclassic' and this one 'tichonov' or we call this one something else.

2) The classic FT way of defining lambda is very problematic for maxfiltered data and you didn't fix that. It would make sense to set it at 5% of the values before the cliff but there is no good way to do that at the moment.

3) I would like to be able to set both lambda and kappa by clicking on the plot (ginput). 

4) I think the distinction between kappa and tolerance is confusing. It's all about the same thing: where do you truncate the eigenvalues. 


My suggestion would be to have only lambda and kappa with 'lambdamethod' and 'kappamethod' where either could be 'interactive' and for 'kappamethod' we could have 'tolerance' as one of the options and e.g. kick detection as another.

schoffelen:
w.r.t. point 1: fine with me. Indeed we need to agree on terminology etc. Actually, initially this option was there (i.e. both lambda and kappa/tolerance had an effect in the method 'svd'

w.r.t point 2: FT has to consider more than only maxfiltered data, and the 'X%' has been in for too long in order for it to change behavior I think. For now, my strategy would be to document that the 'X%' will not work for maxfiltered data. A tailored recipe for maxfiltered data should indeed still be constructed (possibly with functionality for automatic cliff detection), but as of yet, the whole ft_inv operation has been aimed to be generic.

w.r.t. point 3: fair request

w.r.t. point 4: Yes, it could be that it is confusing, yet to conflate this into a single parameter (which I consider to be the other option) might not always work. I used to use (in my own code) the distinction between integer value (>1) and <1 non-integer values. The former case implicitly resulting in the parameter to be interpreted as a number of components (as per kappa), the latter case resulting in the parameter to be interpreted as a 'tolerance' value. In my opinion, adding additional key-value pairs for kappamethod makes it only uglier. Would you be less confused if tolerance were renamed into threshold?

vlitvak:
I think lambdamethod and kappamethod would make it possible to accommodate all this (except maybe cases where more than one parameter is necessary to specify the final 'lambda' or 'kappa'. What's nice about them is that they make the implicit explicit to quote Robert unlike these integer vs real tricks that would get people really confused.

PR #936, opened 2019-01-25
FIX: removed preamble and postamble

StolkArjen:
Following a bug reported on the mailing list: 

Hello, Arjen. 

Yes, I did add fieldtrip and its subdirectories but I'm encountering the same error: 

>> addpath C:\Toolbox\fieldtrip-20190113
>> ft_defaults
>> loreta2fieldtrip
다음 사용 중 오류가 발생함: ft_preamble (line 80)
Could not run ft_preamble_callinfo - does not seem to exist

오류 발생: loreta2fieldtrip (line 48)
ft_preamble callinfo
 
Should I install any external toolbox for preamble/postamble or are they already inside the default?

Thanks,
Eunchan

### DIFF
--- a/besa2fieldtrip.m
+++ b/besa2fieldtrip.m
@@ -44,15 +44,6 @@ function data = besa2fieldtrip(input)
 %
 % $Id$
 
-% these are used by the ft_preamble/ft_postamble function and scripts
-ft_revision = '$Id$';
-ft_nargin   = nargin;
-ft_nargout  = nargout;
-
-% do the general setup of the function
-ft_defaults
-ft_preamble callinfo
-
 if isstruct(input) && numel(input)>1
   % use a recursive call to convert multiple inputs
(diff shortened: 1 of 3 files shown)


robertoostenveld:
there is some code that refers to the cfg which I think is now not being used properly. E.g. besa2fieldtrip mentions cfg.filename, but the cfg is not added to the output. Please clean up those parts as well.

StolkArjen:
Removed calls involving cfg in the 3 functions. But please double check spass2fieldtrip as it's not clear to me how important cfg.trl is to the output of this function. Also, I haven't been able to test run these functions beyond removing pre and postamble (there are none).

robertoostenveld:
looks good.

robertoostenveld:
some of the formats (like spass) I am also not able to test myself. Let's make this the users' own responsibility

PR #937, opened 2019-01-28
ENH - ft_icabrowser fixed and improved

dmatthes1982:
Function ft_icabrowser – reference documentation and config handling adapted a little bit into the direction of the fieldtrip code standard → new useful config options implemented (cfg.rejcomp, cfg.blocksize, cfg.powscale, cfg.zlim, cfg.showcallinfo) → BUG with prepraration of layout fixed → usability improved (e.g. reject control button text will switch from Keep to Reject if selected → after pressing next/prev or linear/log the figure will not completly disappear and reconstructed line wise, to avoid giving the impression of a hanging GUI)


schoffelen:
danke schön

PR #939, opened 2019-01-30
Differently localize electrodes with 3d scanner

saurabhsay:
When you select the elctrode positions manually it is time-comsuming to orient the head model to a desired view with just the keys 'w', 'a','s','d', hence I use the matlab figure window 3D rotate functionality. The problem is that when you want to rotate, the place where you point the mouse cursor gets selected as an electrode position. To avoid this I changed the code where the left mouse button does not select the electrode poisitions. The middle button is used for this purpose, but I have commented the line where the right mouse button can as well be configured to be used for this purpose.

### DIFF
--- a/plotting/ft_select_point3d.m
+++ b/plotting/ft_select_point3d.m
@@ -103,7 +103,7 @@ view(az,el);
 while ~done
   k = waitforbuttonpress;
   currkey=get(gcf,'CurrentKey');
-    if k==0 && strcmp(get(gcf,'SelectionType'), 'extend')
+    if k==0 && strcmp(get(gcf,'SelectionType'), 'extend') % 'extend' for middle button, 'alt' for the right button
         [p, v, ~, ~, ~] = sk_select3d(h);
         % a new point was selected
         if nearest
@@ -118,7 +118,7 @@ while ~done
         end
         set(gcf,'SelectionType','normal');
     elseif k==1 && (strcmp(currkey, 'q') || strcmp(currkey, 'r') || strcmp(currkey, '+') || strcmp(currkey, '-') || strcmp(currkey, 'w')|| strcmp(currkey, 'a') || strcmp(currkey, 's')|| strcmp(currkey, 'd'))% You also want to use strcmp here.
-        key = get(gcf,'CurrentCharacter'); % which key was pressed (if any)?
+        key = get(gcf,'CurrentCharacter'); % which key was pressed ?
         if strcmp(key, 'q')
             % finished selecting points
             done = true;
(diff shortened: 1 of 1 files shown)


robertoostenveld:
many computers (especially laptops) don't have a middle mouse button, so I don't think that is a good idea as default.

Could you make a small test script that demonstrates it, starting from high-level functions (i.e. ft_electrodeplacement)?

saurabhsay:
Yes, agreed, the middle mouse botton may not be the best solution therefore in the script there is an option to change this to the right mouse button. I think it is not ideal to click an electrode position just by trying to rotate the head model (easily done with the 3d rotate functionality), specially if hundreds of electrodes are involved where one would like to use this script.
In addition if one may want to use the right mouse button to select electrodes the future additions to this function should explore ways to supress any other menu items to open while right clicking. 
For middle button owners, no issues ofcourse.

Please find the attached test script, it is the same from the tutorial - http://www.fieldtriptoolbox.org/tutorial/electrode/
the model data is also from the same page.

The function ft_electrodeplacement calls the function ft_select_point3d where obviously the changes are made in this suggested commit.

extra: For
[test_script.zip](https://github.com/fieldtrip/fieldtrip/files/2837167/test_script.zip)

 reference I attach another file sk_select_point3d, which is the changed ft_select_point3d file for checking the functionality. To do this line 170 in ft_electrodeplacement will need to be changed from ft_select_point3d to sk_select_point3d.

Horschig:
maybe rotate if you left click and hold shift or ctrl or so? that is also a standard way to do so (but note my confusion about shift and ctrl, because some programs use the one, others the other one - so much for 'standard' way)

saurabhsay:
That's a good point, perhaps the shift/ctrl key in addition to the left mouse click to rotate model is a better way; And just the left mouse click selects the electrode.

schoffelen:
Hi @saurabhsay what is the status of this? Do you know how to proceed based on @Horschig and @robertoostenveld 's inputs?

saurabhsay:
Hi @schoffelen , I really like the idea from @Horschig , and it is easy enough to include this. I can edit and include this functionality in some time. 
Also working on an automatic recognition of the positions using some image processing techniques, however this has been impacted due to the current Covid19 crises. Will come back on it as and when this changes.

p.s. 
The three mouse button solution works for the 'non-Mac' users. I understand that this is not optimal.

schoffelen:
The last ativity on this thread is already > 1.5 year ago. @SteixnerKumar what is the status of this. Has this been resolved for you, or is this easily resolvable? If not, I suggest to close the issue.

SteixnerKumar:
@schoffelen, Thanks for reaching out on this. It is resolved for windows and Linux systems that use the 3 button mouse (the merge request was open for this!). However not for the apple system. I am not currently working on this since I do not use the mac. @Horschig had the idea that will solve this for mac systems as well and would make it consistent over the different OS.
I hope someone wants to implement this? Else, please decide on the PR as it works only on Windows and Linux.

PR #940, opened 2019-02-05
ENH - overhaul of function, to accommodate Vladimir's points: removed…

schoffelen:
… tolerance option, added lambda regularization for svd method, removed tikhonov as an explicit method.


schoffelen:
@vlitvak : something like this? I haven't yet touched the interactive functionality, but rewrote the code such as to get rid of the tolerance option, allow for lambda regularization in combination with kappa, and removed the method 'tikhonov'

vlitvak:
Yes, that part is fine, just still not sure where something like cliff detection would fit. Also in the interactive part you seem to override the user-specified value. I'm trying to refactor the test script in the meantime so good that we are doing complementary things.

robertoostenveld:
I am also working on this at the moment. I propose that you merge and then leave me to it for some time...

PR #941, opened 2019-02-06
cleaned up ft_inv implementation

robertoostenveld:
as discussed over slack with @schoffelen 

@vlitvak , could you also have a look at this updated implementation (and updated help)?

### DIFF
--- a/private/ft_inv.m
+++ b/private/ft_inv.m
@@ -64,7 +64,7 @@ function Y = ft_inv(X, varargin)
 %
 %    FieldTrip is distributed in the hope that it will be useful,
 %    but WITHOUT ANY WARRANTY; without even the implied warranty of
-%    MERCHANTABILITY or FITNESS FOR X PARTICULAR PURPOSE.  See the
+%    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 %    GNU General Public License for more details.
 %
 %    You should have received a copy of the GNU General Public License
--- a/private/ft_inv.m
+++ b/private/ft_inv.m
@@ -1,39 +1,58 @@
-function Ainv = ft_inv(A, varargin)
+function Y = ft_inv(X, varargin)
 
-% FT_INV computes a (regularized) inverse of a matrix.
+% FT_INV computes a matrix inverse with optional regularization.
 %
 % Use as
-%  Ainv = ft_inv(A, ...)
-% where optional additional arguments should be defined as key-value pairs.
+%  Y = ft_inv(X, ...)
 %
-% method    = 'svd', 'vanilla', 'moorepenrose', or 'winsorize',
-%               the default method is 'svd', and the matrix is regularized
-%               with a weighted identity matrix before inversion, using  
-%               a truncated svd, fixing the number of singular values
-%               according to kappa (see below) before reassembling
-%               the inverse. A non-zero lambda can be specified for
-%               regularization. If the method specified is 'vanilla', a normal 
-%               inv() is computed. If the method specified is 'moorepenrose',
-%               a Moore-Penrose pseudoinverse is computed. If the method 
-%               specified is 'winsorize', an svd based inverse is computed, based
-%               on the original singular values up to the kappa'th, but in addition the 
-%               the remaining singular values lambda are replaced by the value according
-%               to lambda.
-% kappa     = scalar integer, reflects the ordinal singular value at which
-%               the singular value spectrum will be fixed. A value <=0 will
-%               result in an all-zeros output matrix. Can be 'interactive'
-% lambda    = scalar, or string (expressed as a percentage), specifying the
-%               regularization parameter for Tikhonov regularization in the svd method,
-%               or the replacement value for winsorization. Lambda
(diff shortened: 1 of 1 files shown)


private/ft_inv.m
```diff
@@ -1,39 +1,58 @@
...
+%
+% 'moorepenrose' - the Moore-Penrose pseudoinverse is computed, no regularization is
+% applied.
+%
+% 'tsvd' - this results in a pseudoinverse based on a singular value decomposition,
+% truncating the singular values according to either kappa or tolerance parameter
+% before reassembling the inverse.
+%
+% 'tikhonov' - the matrix is regularized according to the Tikhonov method using the
+% labmda parameter, after which the truncated svd method (i.e. similar to MATLAB
```
schoffelen: lambda, rather than labmda

private/ft_inv.m
```diff
@@ -45,127 +64,175 @@
 %
 %    FieldTrip is distributed in the hope that it will be useful,
 %    but WITHOUT ANY WARRANTY; without even the implied warranty of
-%    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+%    MERCHANTABILITY or FITNESS FOR X PARTICULAR PURPOSE.  See the
```
schoffelen: did you do a search replace all for the input argument? X -> A

robertoostenveld: well caught. Yes, my bad.

schoffelen:
looks good! feel free to merge.

vlitvak:
This only addresses my first comment but not the others. I still don't see how this interface will be extended for e.g. cliff detection.

2) The classic FT way of defining lambda is very problematic for maxfiltered data and you didn't fix that. It would make sense to set it at 5% of the values before the cliff but there is no good way to do that at the moment.

3) I would like to be able to set both lambda and kappa by clicking on the plot (ginput). 

4) I think the distinction between kappa and tolerance is confusing. It's all about the same thing: where do you truncate the eigenvalues.

robertoostenveld:
The goal for now was not to fix any issues, just to make a consistent implementation that is backward compatible. Cliff detection relates to the kappa parameter. I can image extensions such as kappa=‘auto’, or kappa=‘cliffdetect’. 

2. Data specific defaults are better set elsewhere: in documentation or in the high-level function, where the user can find them and where they can be part of computational provenance)

3. Indeed: have been thinking about that as well. Feel free to send a PR that implements it.

4. What do you propose?

vlitvak:
The problem with kappa=‘cliffdetect’ is that it does not allow for specifying extra parameters for cliff detection. 

2. The problem is not with documentation but with the fact that currently there is no way to specify lambda other than as a number of as a percentage of the mean trace whereas in the elekta case it would make sense to specify it as a percentage of mean non-truncated trace. Even if you know that, there is no way to do it other than by hand. 

4. I think using tolerance is just one of the possible ways of specifying kappa so it should be at the same level as e.g. cliff detection. My suggestion was to have 'lambda',  'lambdamethod', 'kappa' and 'kappamethod' but that would still assume that there is at most one parameter for any method. Alternatively, you could allow passing an additional struct with parameters or maybe two structs: one for lambda and one for kappa.

robertoostenveld:
2. *after* prewhitening (i.e. in the beamformer) the lambda value should be nicely behaved, since the eigenvalues should be ~1. Prior to prewhitening I don’t think that lambda should be used. So I propose ft_denoise_prewhiten with kappa, and then beamforming with lambda.

3. Lambda and kappa are themselves parameters, both are used differently for the different inverse methods. The lambda and kappa parameter themselves are not methods, but there can be methods for determining the actual parameter. Can we list the different methods for the parameters?

lambdamethod = absolute, interactive, percentageofmeaneigenvalue, percentageofmaximumeigenvalue
kappamethod = absolute, interactive, percentageofnumber, autocliff, lcurve

others?

vlitvak:
2. That would assume that everyone will pre-whiten, but I don't expect that to happen because in FT that has not been an established practice an not everyone has a baseline. 

3. That would work as long as you only need a single parameter. I'm fine with that just warning that at some point we might run into a method that requires two parameters. I would have 'threshold' as a kappamethod and then kappa would be 'tolerance'. I would also say that when both lambda and kappa are specified kappa takes precedence i.e. the eigenvalues are first truncated and then whatever was specified to get lambda is done.

robertoostenveld:
2. for eleka data everyone should prewhiten. I won’t support people that want to do it differently, in that case they are on their own. I will support people that need to prewhiten but don’t have a noise covariance, e.g. by having a template noise covariance (with 1 and 1/58th as relative scaling factors).

3. at a certain point we will run into methods that needs 3 or 4 or N parameters. Or another toolbox. Or lengthy computations. Libraries have been filled with books on matrix inversion, I don’t claim to be an authority here. Just trying to move forward, once step at a time.

vlitvak:
OK. I'll send an e-mail regarding my conclusion from the 'mystery' analysis. The bottom line is I don't see how whitening makes any difference there except for combining the two channeltypes. For that yes I agree that the default covariance should also help.

3. If you are happy with my other suggestions I'm happy to assume a single parameter.

PR #948, opened 2019-02-13
FIX #947 - revert the state specifically for the warning message

### LINKED ISSUE #947: ft_warning off does not retain state correctly
@dmatthes1982 wrote by email

> today I've encountered a problem within the function ft_warning or rather ft_notification.
> 
> I was calling the following commands
> 
> ft_warning off; ft_topoplotIC(cfg, data); ft warning on;
> 
> Contrary to my expectations, ft_plot_topo was throwing a warning message. After closer examination, I found out, that one source of these problem is e.g. in line 364-366 in topoplot_common.
> 
> ws = ft_warning('off', 'FieldTrip:getdimord:warning_dimord_could_not_be_determined');
> [data] = ft_selectdata(tmpcfg, data);
> ft_warning(ws);
> 
> After executing the first command, ws is a structure and contains only the old state of 'FieldTrip:getdimord:warning_dimord_could_not_be_determined'. When I later execute the third command, ft_default.notifications.warning will be overwritten with the content of ws, all other former content (e.g. 'all', 'off') of this structure (ft_default.notifications.warning) is deleted.
> 
> I assume that the aim of the third line is to only reset the entry of 'FieldTrip:getdimord:warning_dimord_could_not_be_determined' not to overwrite the whole ft_default.notifications.warning structure.
> 
> Now, I'm not sure which fix you would prefer.
> 
> One could either modify the third line of this last code snippet to ft_warning(ws.state, ws.identifier) or change the handling of structures in ft_notifcation line 154-157 or change the return value of line one in the code snippet. The function ft_notification should then return the whole content of ft_default.notifications.warning and not only the old state of the entry to be changed.

robertoostenveld:
that is specified, do not overwrite the full state

### DIFF
--- a/test/test_issue947.m
+++ b/test/test_issue947.m
@@ -1,0 +1,92 @@
+function test_issue947
+
+% WALLTIME 00:10:00
+% MEM 2gb
+
+% the warning state is kept in the global ft_default structure
+global ft_default
+
+
+%%
+
+ft_default = [];
+
+% both should show
+ft_warning('a:a', 'a')
+ft_warning('b:b', 'b')
+
+%%
+
+ft_warning('off', 'a:a')
+
+% only b should show
+ft_warning('a:a', 'a')
+ft_warning('b:b', 'b')
+
+%%
+
+ft_default = [];
+
+ft_warning('off')
+
+% none of them should show
+ft_warning('a:a', 'a')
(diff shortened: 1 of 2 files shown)


robertoostenveld:
See #947

PR #949, opened 2019-02-13
implement cfg.reproducescript option

robertoostenveld:
for re-creating a simplified but still full analysis script during the execution of the users own (possibly convoluted) analysis script.


robertoostenveld:
@matsvanes this is the code that relates to reproducing a full analysis.

@Spaak proposed to also write a png image upon saving the figure, which I think is a good (and simple to implement) idea. I will still add that.

PR #955, opened 2019-02-15
Extent ft_prepare_mesh to create individualised meshes on the basis of volume conduction models and head surface information

simonhomoelle:


robertoostenveld:
why did you close this?

PR #959, opened 2019-02-20
compute current density also for bad channels

robertoostenveld:
This includes (and replaces) #929


robertoostenveld:
@johalgermissen this pull request includes the additional changes that I made to the code. It was quite a puzzle to get the code robust to random orders of the data channels and electrodes, but in the end it looks quite clean again. Please check it out. I will now merge it with the release version (which means that it will also end up on home/common and and on the ftp). You can "checkout master" on your local copy again and "git pull upstream master" to get the update, and discard your local branch.

Issue #899, opened 2018-12-06
Issues with compat folder when wrongly added to the search path


gllmflndn:
Hi,

it has been brought to my attention that adding FieldTrip _with subfolders_ to the search path wreaks havoc on recent MATLAB versions as core functionalities such as `edit` or `open` stop working. The underlying cause is that the `contains` implementation is not complete and compatible. An a minima fix could be this:

```
Index: compat/matlablt2016b/contains.m
===================================================================
@@ -17,6 +17,10 @@
 if nargin<3
   str = 'IgnoreCase';
 end
+if exist('contains','builtin')
+    tf = builtin('contains',s, pattern, str, boolean);
+    return;
+end
 if ~strcmpi(str, 'ignorecase')
   error('incorrect third input argument, can only be ''IgnoreCase''');
 end

```

I know that FieldTrip's subfolders should not be added to the search path (same rule applies for SPM), that this is mentioned explicitly in the documentation:
```
% This is a compatibility function that should only be added to the path on
% MATLAB versions prior to 2016b.
```
and that FieldTrip displays a warning if this is detected but MATLAB crashes with its most basic functionalities as soon as the wrong folders are included and before FieldTrip being used (i.e. before any warning can be displayed). For example, using the MATLAB interface, just changing directory to `fieldtrip/compat/matlablt2016b/` and double-clicking on a file to open it will fail with error `PATTERN must be a string scalar or character vector`.

I am not sure that the comfort brought by using `contains` (and related methods of the `string` class) is currently worth the disruptions the required extra compatibility layer can create.

robertoostenveld:
So you are saying that https://github.com/fieldtrip/fieldtrip/blob/71c87539a884265741b1742eef4f2d3835c39b20/ft_defaults.m#L363
does not work (in time)? Bummer.

The number of functions is relatively limited: 

```
mac011> cd /Users/roboos/matlab/fieldtrip/compat
mac011>  ls matlablt201*/*.m | wc -l
      16
```

and I suppose your solution would work for all of them, right? Should we try to make a generic solution, like `compat/matlablt2016b/contains.m` (and similar) doing the check like this

```
function varargout = compatfunction(varargin)

if exist(mfilename,'builtin')
  varargout{:} = builtin(mfilename, varargin{:})
else
  varargout = feval(mfilename, varargin{:})
end
```
where the actual drop-in replacement would go to `compat/matlablt2016b/private`?

gllmflndn:
[...]

PR #960, opened 2019-02-20
proposed solution for addpath(genpath(...)) problems

robertoostenveld:
as brought up by @gllmflndn in issue #899


robertoostenveld:
there are 16 functions 
```
mac011> ls matlablt201*/*.m | wc -l
16
```

that should get this piece of code. Partially it is identical for all, but the actual function call (with input and output args) would have to be customized. Still quite doable.

robertoostenveld:
these 5 would have to be dealt with differently 

```
>> fn = {
'iscolumn'
'isrow'
'ismatrix'
'narginchk'
'isequaln'
'istable'
'webread'
'weboptions'
'webwrite'
'endsWith'
'contains'
'startsWith'
'newline'
'pad'
'isfolder'
'isfile'
};

for i=1:numel(fn)
  if ~exist(fn{i}, 'builtin')
    disp(fn{i});
  end
end

istable
webread
weboptions
webwrite
pad
```

although ... I also realize that Mathworks is sometimes (over versions) changing a function into a builtin.  So it should not only test for builtins.

compat/matlablt2016b/contains.m
```diff
@@ -7,6 +7,33 @@
 % This is a compatibility function that should only be added to the path on
 % MATLAB versions prior to 2016b.
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% see https://github.com/fieldtrip/fieldtrip/issues/899
+
+if exist(mfilename, 'builtin') || strncmp(mfilename('fullpath'), matlabroot, length(matlabroot))
```
robertoostenveld: the 2nd part does not work like this

robertoostenveld:
how do you like this @gllmflndn?

gllmflndn:
Thanks @robertoostenveld for looking into this.

I think your changes would help minimise disruptions. When testing using the (unusual) approach described in #899:

> For example, using the MATLAB interface, just changing directory to `fieldtrip/compat/matlablt2016b/` and double-clicking on a file to open it will fail with error `PATTERN must be a string scalar or character vector`.

I notice a number of things:

- `ft_warning` is not in the path so an error still appears. Maybe the compatibility functions should not call any `ft_` functions?
- `rmpath` displays a warning that the directory to be removed is not in the path so there could be a check that it is explicitly in the path first, teamed with `~isdeployed`.

I would agree this is a peculiar corner case but it would be great to have a solution as robust as possible. There will anyway be a number of warnings being displayed:

> Warning: Function xxxx has the same name as a MATLAB builtin. We suggest you rename the function to avoid a potential name conflict.

robertoostenveld:
thanks for the good suggestions, I will follow that up and also implement it for the remaining 14 functions.

PR #963, opened 2019-02-21
reconstructed content of the test_pull929.m  test script

robertoostenveld:
which somehow went missing. 

See https://github.com/fieldtrip/fieldtrip/pull/929#issuecomment-465984106 in issue #929

### DIFF
--- a/test/test_pull929.m
+++ b/test/test_pull929.m
@@ -1,0 +1,156 @@
+function test_pull929
+
+% MEM 1000mb
+% WALLTIME 00:01:00
+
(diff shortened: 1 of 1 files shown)


robertoostenveld:
@johalgermissen have a look at it now. Note that I changed the plotting, since I actually want to see the ERP and SCD topographically. Please do note that you can drag-select and click in the multiplot and singleplot figures to make selections, which are subsequently plotted in more detail. Without these topographic plots you cannot see how nicely it resolves the two bilateral blobs in the data, even with only 50% of the channels.

PR #965, opened 2019-02-22
improve the plotting of labels in relation to the sensors

robertoostenveld:


robertoostenveld:
this also includes improvements and corrections to the computation of lead fields (i.e. potential distributions) for electric current monopoles in a infinite and halfspace conductive medium.

PR #967, opened 2019-02-26
Set up continuous integration with travis

robertoostenveld:
using the Mathworks cloud service for testing


robertoostenveld:
the CI tests are not running successfully since the FieldTrip repository is too large. But technically it is working.

PR #970, opened 2019-02-27
fixed regression errors that I introduced 

robertoostenveld:
with the cleanup/renaming of the low-level infinite and halfspace forward models


robertoostenveld:
ajeto

schoffelen:
buurman

PR #973, opened 2019-02-27
improve names and handling of large cfg elements

robertoostenveld:
... such as vol/headmodel, grid/sourcemodel, elec/grad/opto, layout, headshape, etc. This is needed for the reproducescript option that we are working on with @matsvanes and @Spaak. 

We now have ft_fetch_xxx functions for some fields, and ft_prepare_xxx functions for some other fields. The handling should be made more consistent, so that the input cfg.xxx fields can be read from disk (and written, when `reproducescript` is used).


robertoostenveld:
I have now consistently changed vol into headmodel, and grid into sourcemodel. However, there are still quite some regression errors with the test batch that need to be resolved. 

Also something to deal with is compatibility with public function names (ft_read_vol etc.)

robertoostenveld:
I have added backward compatibility wrapper functions for those public functions that were renamed. The test batch is now running without errors (mostly).

robertoostenveld:
Saving of large cfg fields is now implemented for elec, grad, opto, layout, headmodel, sourcemodel, headshape and event.

robertoostenveld:
loading of cfg.mri should probably work, but still needs to be implemented for cfg.neighbours, cfg.events and  possibly for some other "large" config fields. 

Other than that (which can wait a bit), all tests run though and the code should be significantly cleaner now than it was yesterday morning. Please keep your eyes open for bugs and your ears for users reporting problems.

matsvanes:
If any structure in the cfg is a [0x0 struct] instead of [], printstruct (reproducescript line 33) crashes. I'm not sure what causes this issue, but I think it has something to do with replacing large cfg fields by inputfiles. In the beamformer tutorial, during ft_sourceanalysis, and then during reproducescript>printstruct (line 99)  val.grid.cfg.headmodel.cfg.grad gives [] before commit 4a9005998ffcd240485e8147c9b4b22b6a03f103 (which puts large cfg fields in inputfile) and [0x0 struct] after the commit.

schoffelen:
Perhaps it would make sense to try and make printstruct robust for empty struct inputs?

matsvanes:
Good point. Just created a pull request which does this.

PR #984, opened 2019-03-08
ENH - reorganised the code a bit. Moved many of the cfg options to cf…

schoffelen:
…g.mvpa

### DIFF
--- a/ft_statistics_mvpa.m
+++ b/ft_statistics_mvpa.m
@@ -1,6 +1,6 @@
 function [stat, cfg] = ft_statistics_mvpa(cfg, dat, design)
 
-% FT_STATISTICS_MVPA performs multivariate pattern classification 
+% FT_STATISTICS_MVPA performs multivariate pattern classification
 % on the data. If the data has not been averaged over time, classification
 % is performed separately for every time point. Additionally, searchlight
 % analysis (classification for each channel/voxel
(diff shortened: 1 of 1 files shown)


schoffelen:
@treder I started to rename the cfg a bit, (move stuff to cfg.mvpa).
Also, I removed the cfg.std option, because it can be given back to the caller function at no extra cost, right?

Also, I realise that with the is_data_3d argument you have so far only considered time domain data? What if the dim does not exist, or if it is time/freq, or freq data, or 3D source dipole positions? Also, I am not sure whether cfg.dim is guaranteed to exist

treder:
re:cfg.mvpa looks good! And you're right, std does not incur extra cost other than cluttering the stat.metric so happy to ditch it.

treder:
re:data_3d   indeed, for now I tried to get the interface right for ft_timelockstatistics [esp everything covered in the tutorial]. Source data would hopefully not be a problem (voxels could be treated like channels including the searchlight). 

It gets trickier with freq and time-freq data. MVPA-Light does not deal with 4D data at this stage (and I do not see how to build it into the toolbox without getting messy), so it might be better to build this into ft_timelockstatistics. It guess it woud involve looping across one of the dimensions.

treder:
sorry I meant ft_statistics_mvpa  (not ft_timelockstatistics). Does ft_statistics_mvpa 'know' whether it has been called by timelockstats, freqstats or sourcestats ? How would you know the dimensions if there's no dim field? 

For classification it's also a bit unclear what are the features? E.g. for time-freq data, do you want to classifiy for each time point and each frequency separately (yielding a 2D map of metrics), or only classify every time point and use the different frequencies as features? It seems we'll need extra cfg parameters here.

schoffelen:
All good questions! indeed, now I think of it, I believe that cfg.dim might always be present. 

I think it is indeed good to aim at getting it working for space-versus-time data first. 

In terms of treating the different dimensions of the features space: could this be dealt with somehow by creatively designing the nb matrix?

treder:
hmm. The nb matrix really only used in mv_searchlight so not sure if this works. 

Another possibility could be specifying a new field cfg.features (or cfg.mvpa.features) where in the dimensions are put that serve as features. For instance, cfg.features = {'freq' 'chan'} would perform classification across time (if there's a time dimension) and build a feature vector from both freq and chan.

PR #987, opened 2019-03-12
ENH – optimization regarding runtime

dmatthes1982:
Function ft_fetch_data – ENH optimization regarding runtime. The zeros-function needed a lot of time in case of large dataset and allocated a lot of memory, which afterwards was never used.

### DIFF
--- a/utilities/ft_fetch_data.m
+++ b/utilities/ft_fetch_data.m
@@ -125,18 +125,22 @@ if trlnum>1
     else
       trlbeg = trlbeg - begsample + 1;
       trlend = trlend - begsample + 1;
+      offset = 0;
+      overlap = 0;
       if trlbeg < 1
+        offset = 1 - trlbeg;
         trlbeg = 1;
       end
-      if trlend > endsample
-        trlend = endsample;
+      if trlend > buflen
+        overlap = trlend - buflen;
+        trlend = buflen;
       end
       % make vector with 0= no sample of old data, 1= one sample of old data, 2= two samples of old data, etc
       count(trlbeg:trlend) = count(trlbeg:trlend) + 1;
       % make vector with 1's if samples belong to trial 1, 2's if samples belong to trial 2 etc. overlap/ no data --> Nan
       trialnum(trlbeg:trlend) = trllop;
       % make samplenum vector with samplenrs for each sample in the old trials
-      samplenum(trlbeg:trlend) = 1:trllen(trllop);
+      samplenum(trlbeg:trlend) = (1 + offset):(trllen(trllop) - overlap);
     end
   end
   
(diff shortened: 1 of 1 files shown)


dmatthes1982:
In addition to this change I have a second question: This function is always running into the if case in the "get the options"-section, but the else path is a little faster. Why do you always prefer the if case?

schoffelen:
The code looks good (from reading), Daniel. Can you confirm that you have ensured that the behaviour of the function has not changed?

robertoostenveld:
please use the `test_ft_fetch_data.m` script, which you can find in fieldtrip/test

dmatthes1982:
Thanks for your hint @robertoostenveld. I found a mistake and I will fix it as soon as possible.

dmatthes1982:
Now, test_ft_fetch_data quits says only

avg. #samples: 1199.00 +/- 0.00
Elapsed time is 3.533915 seconds. 

It seems to work.

robertoostenveld:
I had not looked into the script and I did not know that the test script was mainly for speed. It would actually be interesting to know how fast it was prior to your change. 

Furthermore, since `test_ft_fetch_data.m` does not  test many different cases, could you please also test these other scripts:

```
mac011> grep -l ft_fetch_data *
grep: private: Is a directory
test_bug1667.m
test_bug1735.m
test_bug2269.m
test_bug3267.m
test_bug3379.m
test_ft_fetch_data.m
test_issue968.m
test_pull810.m 
```

and see which ones you could also test? Some of them will be reading data from disk, which you don't have and therefore cannot test. But some of them are with simulated data in memory.

dmatthes1982:
Today, I checked the performance increase. The script ft_test_fetch_data shows no significant improvement. Before the change, it says 3.87 sec. After the change it says 3.77 sec.

But I have a dataset with a length of 71 Minutes, consisting of 25 trials each with a duration of around three minutes. I want to subsegment these trials into subtrials with a length of 1 second and by creating an overlap of 50%. When I now run ft_redefinetrial, which calls ft_fetch_data, the computation time is ca. 45 seconds before changing the code and 4.6 seconds afterwards. In a later step I'm additionally running ft_rejectartifact on the subsegmented data. Here a got at least an improvement from 166 seconds to 97 seconds. So this change in ft_fetch_data helps me a lot.

I will test the other scripts in the following days and keep you updated.

robertoostenveld:
thanks for testing the speed. 

Good that it did not get slower;  since that is what I am worried about: that the implementation can be optimized for one use-case, but subsequently gets worse for another use case.  

Please do also test the others (at least the ones you can test without having the data)

robertoostenveld:
Since it does not get slower in the test scripts, and does get faster in your specific case, I consider it a general improvement.  I tested the whole batch of test scripts against your pull request, all tests work fine.

PR #988, opened 2019-03-13
ENH - ft_icabrowser

dmatthes1982:
Function ft_icabrowser – ENH → computation time of power spectrum smoothing optimized → two new buttons implemented (first page, last page)


schoffelen:
thanks!

PR #993, opened 2019-03-16
specify neighbours using ft_prepare_neighbours struct

treder:
- renamed cfg.nb to cfg.neighbours
- a FT neighbourhood structure can be passed , it is then internally converted to a matrix using the FT channelconnectivity function


treder:
btw bit off-topic but thanks for the work on ft_statistics_mvpa, we're getting there. 
Integrating multi-dimensional data (eg time-frequency) and arbitray dimension orders will take a while, it might need a more general implementation of mv_searchlight. I guess specifying multiple neighbourhoods (eg in space, time, freq) have been integrated in your cluster stats already so could have a look there.

robertoostenveld:
sorry for this taking some time to get merged.

treder:
thanks & no worries!

PR #1007, opened 2019-03-18
Update ft_singleplotER.m

matsvanes:
remove inputfile and reproducescript fields from configuration before calling ft_timelock/freq baseline

### DIFF
--- a/ft_singleplotER.m
+++ b/ft_singleplotER.m
@@ -273,15 +273,16 @@ end
 
 % apply baseline correction
 if ~strcmp(cfg.baseline, 'no')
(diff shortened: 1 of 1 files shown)


ft_singleplotER.m
```diff
@@ -273,15 +273,16 @@
 
 % apply baseline correction
 if ~strcmp(cfg.baseline, 'no')
+  tmpcfg = removefields(cfg, {'inputfile', 'reproducescript'});
```
robertoostenveld: it would be better to explicitly keep the fields that the baseline function expects, rather than excluding the ones it should not have. In the current code (also with your change) there are still many cfg options that end up in the baseline function which should not be there

PR #1008, opened 2019-03-18
Update ft_singleplotTFR.m

matsvanes:
remove inputfile and reproducescript fields from configuration before calling ft_freqbaseline

### DIFF
--- a/ft_singleplotTFR.m
+++ b/ft_singleplotTFR.m
@@ -200,11 +200,12 @@ end
 
 % Apply baseline correction:
 if ~strcmp(cfg.baseline, 'no')
+  tmpcfg = removefields(cfg, {'inputfile', 'reproducescript'});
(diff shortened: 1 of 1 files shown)


ft_singleplotTFR.m
```diff
@@ -200,11 +200,12 @@
 
 % Apply baseline correction:
 if ~strcmp(cfg.baseline, 'no')
+  tmpcfg = removefields(cfg, {'inputfile', 'reproducescript'});
   % keep mask-parameter if it is set
   if ~isempty(cfg.maskparameter)
     tempmask = data.(cfg.maskparameter);
   end
-  data = ft_freqbaseline(cfg, data);
+  data = ft_freqbaseline(tmpcfg, data);
```
robertoostenveld: as elsewhere: better specify baseline-specific fields to keep.

PR #1013, opened 2019-03-19
extend cfg.reproducescript functionality to detect identical data variables

Spaak:
As discussed with @robertoostenveld.

I considered detecting identity via variable names, which are available for input arguments through builtin `inputname`. However, there is no analogous mechanism for retrieving output argument names, which would thus need to go through `dbstack`, which is probably a bad idea. Furthermore, variables with identical names can of course nonetheless be modified in the user's script outside of FieldTrip, also making reliance on variable names a bad idea.

I therefore chose to implement the check using hashes in a separate file `hashes.mat` in the reproducescript folder. Note that the hash is computed for the data structure as a whole, which thus also includes `cfg.previous` and `cfg.callinfo` etc., which guarantees a proper data structure match if the hashes match.

Additionally, I moved the functionality for storing large cfg fields to disk to a helper function, thus reducing duplication between `ft_postamble_savevar` and `ft_postamble_savefig`. @robertoostenveld there were (already) some slight differences between the corresponding snippet in `ft_postamble_savevar` and `ft_postamble_savefig`, which I now harmonized. There was an explicit check on size in `ft_postamble_savefig` (but not in `ft_postamble_savevar`), which I now removed. I think storing these variables to disk should not depend on size at this point in the code, but maybe I missed something, so please check :)


Spaak:
@matsvanes your input here would also be welcome.

Spaak:
Also tagging @schoffelen; since development on reproducescript is quite active at the moment, would be good to merge this soon to avoid conflicts (just had to resolve one because of #1015).

robertoostenveld:
I think it would indeed be a good idea to have it on a feature branch, since I suspect there will be more changes to come and it is risky (and messy) to have them in the release every time. I will make a reproducescript branch, can you please resubmit this PR to that one?

robertoostenveld:
actually I realize that I can just as easy merge this PR with the new reproducescript branch myself.

robertoostenveld:
mac011> git push --set-upstream upstream reproducescript
Counting objects: 46, done.
Compressing objects: 100% (46/46), done.
Writing objects: 100% (46/46), 5.98 KiB | 1.49 MiB/s, done.
Total 46 (delta 35), reused 0 (delta 0)
remote: Resolving deltas: 100% (35/35), completed with 11 local objects.
   182816f96..4207bef85  reproducescript -> reproducescript
Branch 'reproducescript' set up to track remote branch 'reproducescript' from 'upstream'.

robertoostenveld:
@Spaak @matsvanes @schoffelen please do follow up commits and PRs to the `reproducescript` branch

PR #1014, opened 2019-03-19
Fixed ft_databrowser

matsvanes:
bugfix in preparing layout when cfg.layout if specified.


matsvanes:
@robertoostenveld this should fix test_bug182, test_bug2316 and test_bug2559.

PR #1022, opened 2019-03-22

matsvanes:
If cfg fields (headmodel, sourcemodel, neighbours) contain input file names, load these and replace the cfg field by the actual data (instead of leaving the file name in). This is adapted now in ft_freq/timelockstatistics, ft_megplanar and ft_sourceanalysis. Other fields (like elec) and functions (which ones?) have not yet been tested.
Additionally, excluding cfg.inputfile from cfg inputs to  secondary fieldtrip calls.


schoffelen:
Uhm, ook al zegt ie dat ie geen conflicten heeft met de base branch, ben ik een beetje huiverig om met deze PR 7831 changed files te mergen. Zou je je veranderingen aub in een schone PR kunnen doen?

PR #1025, opened 2019-03-23
Add an fopen_or_error() function with error checking

apjanke:
It's inconvenient to check `fopen()`'s return values for error indicators. The various FieldTrip functions use different idioms for doing so, and most occurrences of `fopen()` that I see aren't checked for errors at all.

This could be made easier and more robust by adding an `ft_fopen()` wrapper function that does the error checking for you, and turns error-indicating status codes into raised Matlab exceptions.


robertoostenveld:
I don't mind a general error handling of fopen, but I don't want that to be in the API that is exposed to the FT toolbox users: people might start using it for other purposes and then we have to maintain it in the next decades.

Better is to have it in the respective private directories (and using bin/synchronize-private.sh). 

Also, better is to call it `fopen_or_error` or so, which indicates better what the function does.

fileio/ft_fopen.m
```diff
@@ -0,0 +1,20 @@
+function fid = ft_fopen(filename, mode, varargin)
```
robertoostenveld: Please call it fopen_or_error and move it to the private directories.

apjanke: Done.

fileio/ft_read_event.m
```diff
@@ -429,9 +429,10 @@
     event = read_ah5_markers(hdr, filename);
     
   case 'brainvision_vmrk'
-    fid=fopen(filename,'rt');
-    if fid==-1
-      ft_error('cannot open BrainVision marker file')
+    try
+      fid = fopen_or_error(filename,'rt');
+    catch err
+      ft_error(sprintf('cannot open BrainVision marker file: %s', err.message);
```
robertoostenveld: this error would not happen right? It should be caught in the open_or_error, so the try-catch can be removed.

robertoostenveld: I just checked, the error message in fopen_or_error suffices.

fileio/private/np_read_splitted_fileinfo.m
```diff
@@ -228,9 +225,10 @@
 if strcmp(pa,''),
     pa=pwd;
 end
-fid=fopen([pa filesep fn(1:14) '.EE_'],'r');
-if fid==-1,
-    ft_error('Unable to read setup (marker 70) in *.EE_ file.');
+try
+    fid=fopen_or_error([pa filesep fn(1:14) '.EE_'],'r');
+catch err
+    ft_error(sprintf('Unable to read setup (marker 70) in *.EE_ file: %s', err.message));
```
robertoostenveld: this is a very specific error message, I propose that this stays as it is

fileio/private/read_biosemi_bdf.m
```diff
@@ -60,8 +60,9 @@
   cname=computer;
   if cname(1:2)=='PC' SLASH=BSLASH; end;
 
-  fid=fopen(FILENAME,'r','ieee-le');
-  if fid<0
+  try
+    fid=fopen_or_error(FILENAME,'r','ieee-le');
+  catch err
     fprintf(2,['Error LOADEDF: File ' FILENAME ' not found\n']);
     return;
```
robertoostenveld: I don't see why it would return here. This would only cause an error elsewhere (since the return value is not defined). I propose to remove the try-catch.

fileio/private/read_brainvision_vhdr.m
```diff
@@ -128,83 +128,88 @@
...
-            end
-            if ~feof(fid)
-                tline='';
-                while ~feof(fid) && isempty(tline)
-                    tline = fgetl(fid);
+try
+    fid = fopen_or_error(filename, 'rt');
+catch err
+    % quash
+    % TODO: Are we sure we want this to just silently return, instead of raising
```
robertoostenveld: this should not have a try-catch, it should simply error inside fopen_or_error. Returning nothing does not make sense.

fileio/private/read_ctf_res4.m
```diff
@@ -41,12 +41,10 @@
...
 
-fid = fopen(fname,'r','ieee-be');
-
-% Check if header file exist
-if fid == -1
-  errMsg = strcat('Could not open header file:',fname);
-  ft_error(errMsg);
+try
+  fid = fopen_or_error(fname,'r','ieee-be');
+catch err
```
robertoostenveld: it should not have a try-catch, but simply error in fopen_or_error

fileio/private/read_mat.m
```diff
@@ -31,18 +31,19 @@
 %
 % $Id$
 
-f=fopen(fn);
-if (f==-1)
-  fprintf('\nCannot open %s\n\n', fn);
+try
+f=fopen_or_error(fn);
+catch err
```
robertoostenveld: better not try-catch here, just rely on the error in fopen_or_error

fileio/private/read_mclust_t.m
```diff
@@ -16,9 +16,11 @@
 for iF = 1:nFiles
 	tfn = tfilelist{iF};
 	if ~isempty(tfn)
-		tfp = fopen(tfn, 'rb','b');
-		if (tfp == -1)
+    try
+ 		  tfp = fopen_or_error(tfn, 'rb','b');
+    catch err
```
robertoostenveld: I cannot judge this one, so better leave it as is

robertoostenveld:
if you make the final changes, I will pull the branch and test it on our compute cluster with all scripts in fieldtrip/test.

apjanke:
Thanks!

PR #1028, opened 2019-03-25
Add "contributing" doco and .editorconfig

apjanke:
This PR adds `.editorconfig` and `.github/CONTRIBUTING.md` to make it easier for contributing developers to discover FieldTrip's code style.

The `.github/CONTRIBUTING.md` file will be presented to users automatically by GitHub the first time they make a bug report or pull request. It has a brief summary of FieldTrip's code style, and pointers to the contributors' documentation on the FieldTrip website.

The `.editorconfig` will cause [supporting text editor programs](https://editorconfig.org/#download) to automatically configure line and indentation formatting to match FieldTrip's preferred style. EditorConfig is not supported by Matlab itself, but many other popular text editors support it, so it will work for many users that use an external text editor.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# EditorConfig file for FieldTrip code style
+# (see https://EditorConfig.org for info)
+
+root = true
+
+[*]
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.m]
+charset = utf-8
+indent_style = space
+indent_size = 2
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,0 +1,25 @@
+Contributing to FieldTrip
+=========================
+
+## Community interaction
+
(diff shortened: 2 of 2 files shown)


Spaak:
While I would prefer British English (and Oxford commas!), I think at some point we decided prose should be American English, which we have subsequently not been enforcing at all ;) (Think of e.g. `cfg.neighbours` etc.) This sounds like an executive decision for @robertoostenveld . Otherwise looks good to me!

robertoostenveld:
Eelke is right, we did indeed decide for American spelling. See http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=841. Updating (old) code is hard (especially function names), but wherever possible (e.g. new code and documentation) we should stick to it.

robertoostenveld:
Btw, I did not know the phrase Oxford comma, but found it [here](https://www.grammarly.com/blog/what-is-the-oxford-comma-and-why-do-people-care-so-much-about-it/). I guess I have been trained (~30y ago) in not using it. But I actually never explicitly thought about it and don't have a preference.

apjanke:
In that case: https://github.com/fieldtrip/website/pull/111. This is what tricked me in to thinking you used British English.

Issue #1027, opened 2019-03-24
Automatic MEX compilation is getting paths wrong


apjanke:
I'm trying to run FieldTrip under GNU Octave. When I do so, I get a bunch of `mkoctfile` errors.

```
$ octave-default
GNU Octave, version 6.0.0
[...]
octave:2> addpath ~/opp/fieldtrip/fieldtrip-20190322
octave:3> ft_defaults
warning: strmatch is obsolete; use strncmp or strcmp instead
[...]
octave:5> run_trials
Loaded EEGTrial support functions and helpers from /Users/janke/local/repos/oh-oh-thesis
creating sourcemodel based on inward-shifted brain surface from volume conductor model
using electrodes specified in the configuration
warning: trying to compile MEX file from /Users/janke/opp/fieldtrip/fieldtrip-20190322/private/solid_angle.c
 In '/Users/janke/opp/fieldtrip/fieldtrip-20190322/private/solid_angle.m' at line 95
 In '/Users/janke/opp/fieldtrip/fieldtrip-20190322/private/bounding_mesh.m' at line 73
 In '/Users/janke/opp/fieldtrip/fieldtrip-20190322/private/find_outermost_boundary.m' at line 45
 In '/Users/janke/opp/fieldtrip/fieldtrip-20190322/private/headsurface.m' at line 197
 In '/Users/janke/opp/fieldtrip/fieldtrip-20190322/ft_prepare_sourcemodel.m' at line 557

clang: error: no such file or directory: '../../src/geometry.c'
clang: error: no input files
warning: mkoctfile: building exited with failure status
clang: error: no such file or directory: '../../src/solid_angle.c'
clang: error: no input files
warning: mkoctfile: building exited with failure status
clang: error: no such file or directory: 'solid_angle.o'
clang: error: no such file or directory: 'geometry.o'
warning: mkoctfile: building exited with failure status
warning: trying to compile MEX file from /Users/janke/opp/fieldtrip/fieldtrip-20190322/private/solid_angle.c
clang: error: no such file or directory: '../../src/geometry.c'
clang: error: no input files
warning: mkoctfile: building exited with failure status
[...]
```

It looks like the stub functions that auto-compile the MEX files are getting the source code paths wrong. For example, [here's `private/solid_angle.m`](https://github.com/fieldtrip/fieldtrip/blob/fa1592bb1576b7ad0defc06a388773e5ac24f290/private/solid_angle.m#L89-L105).

```
[...]

PR #1029, opened 2019-03-25
bugfix - private/solid_angle.m stub: fix source paths and refactor

apjanke:
Here's an approach to fixing https://github.com/fieldtrip/fieldtrip/issues/1027. I only did it to one of the MEX stub files, so you can see how you like this approach before applying it to all of them.

It fixes the `mex` compilation error by correcting the paths to the source files. I assume the bad paths were left over from an older refactoring of the code, or from when `solid_angle.m`, which exists in many different `private/` directories, was copied from one to the other without the code being adjusted.

This also refactors the code to:
  * reduce the number of places with hardcoded constants, making it easier to copy and paste from one stub function to another, and
  * reduce the difference between the Windows and Unix code paths

With this new form of the function, the only things inside the function body that need to change when you copy and paste it from one place to another are 
  * the path to `src`
  * the list of object files/.c compilation units it depends on (e.g. `{'geometry'}`, here)

It also cleans up the intermediate .o/.obj files instead of leaving them as droppings.

### DIFF
--- a/private/solid_angle.m
+++ b/private/solid_angle.m
@@ -86,24 +86,36 @@ function [varargout] = solid_angle(varargin)
 pwdir = pwd;
 
 % determine the name and full path of this function
-funname = mfilename('fullpath');
-mexsrc  = [funname '.c'];
-[mexdir, mexname] = fileparts(funname);
+funfile = mfilename('fullpath');
+[mexdir, mexname] = fileparts(funfile);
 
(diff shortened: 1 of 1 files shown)


apjanke:
Closing due to https://github.com/fieldtrip/fieldtrip/pull/1032#issuecomment-476297462.

PR #1030, opened 2019-03-25
documentation - fix "juist" typo in code comments

apjanke:
This just fixes a typo that appears in comments in several MEX stub function files.

### DIFF
--- a/@config/private/deepcopy.m
+++ b/@config/private/deepcopy.m
@@ -60,7 +60,7 @@ catch
 end
 
 if success
-  % execute the mex file that was juist created
+  % execute the mex file that was just created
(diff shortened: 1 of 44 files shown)


Horschig:
just leaving a comment here because I saw this: I am not sure this is actually a typo. 'Juist' is the Dutch word for 'correct', and seeing that the comment is after an `if success` statement, I think it is supposed to be a pun? Could fit @schoffelen's style, so maybe he is to blame ;)

schoffelen:
Hehehe, no, not me. I think this is an unintentional typo. The Dutch also has 'zojuist' which in meaning is closer to 'just'.

PR #1033, opened 2019-03-25

matsvanes:
If cfg fields (headmodel, sourcemodel, neighbours) contain input file names, load these and replace the cfg field by the actual data (instead of leaving the file name in). This is adapted now in ft_freq/timelockstatistics, ft_megplanar and ft_sourceanalysis. Other fields (like elec) and functions (which ones?) have not yet been tested.
Additionally, excluding cfg.inputfile from cfg inputs to secondary fieldtrip calls.

### DIFF
--- a/private/topoplot_common.m
+++ b/private/topoplot_common.m
@@ -342,10 +342,11 @@ if ~strcmp(cfg.baseline, 'no')
   if ~isempty(cfg.maskparameter)
     tempmask = data.(cfg.maskparameter);
   end
+  tmpcfg = removefields(cfg, 'inputfile');
   if strcmp(xparam, 'time') && strcmp(yparam, 'freq')
-    data = ft_freqbaseline(cfg, data);
+    data = ft_freqbaseline(tmpcfg, data);
   elseif strcmp(xparam, 'time') && strcmp(yparam, '')
-    data = ft_timelockbaseline(cfg, data);
+    data = ft_timelockbaseline(tmpcfg, data);
   end
   % put mask-parameter back if it is set
   if ~isempty(cfg.maskparameter)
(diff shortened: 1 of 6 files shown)


robertoostenveld:
Rather than loading the neighbours from disk at various places, it should be dealt with on one place using ft_prepare_neighbours. 

That means that the cfg.template field in ft_prepare_neighbours should be renamed into cfg.neighbours (using ft_checkconfig for backward compatibility). And whenever neighbours are needed, something like this should be done

    tmpcfg = keepfields(cfg, {'neighbours', ...})
    cfg.neighbours = ft_prepare_neighbours(tmpcfg) 
  
 Please check how ft_prepare_layout is being used inside other functions.

and in ft_prepare_neighbours it should check on cfg.neighbours being a filename, and if so, use loader to read the layout from disk.

private/topoplot_common.m
```diff
@@ -425,10 +426,11 @@
   nplots = numel(xlims)-1;
   nyplot = ceil(sqrt(nplots));
   nxplot = ceil(nplots./nyplot);
+  tmpcfg = removefields(cfg, 'inputfile');
   for i=1:length(xlims)-1
     subplot(nxplot, nyplot, i);
-    cfg.xlim = xlims(i:i+1);
-    ft_topoplotTFR(cfg, data);
+    tmpcfg.xlim = xlims(i:i+1);
```
robertoostenveld: this is something you may want to copy over in your next PR

utilities/private/ft_preamble_init.m
```diff
@@ -133,7 +133,7 @@
     % with cfg.reproducescript functionality, throw error to make them
     % mutually exclusive
     if (isfield(cfg, 'inputfile') && ~isempty(cfg.inputfile)) || ...
-       (isfield(cfg, 'outputfile') && ~isemty(cfg.outputfile))
+       (isfield(cfg, 'outputfile') && ~isempty(cfg.outputfile))
```
robertoostenveld: this should certainly be copied over

PR #1041, opened 2019-03-26
fix - minor bugfixes artinis reading files

Horschig:
most importantly, preventing a crash when reading in data that
contains non-NIRS data (EEG, trigger devices, etc).
Also including some more fixes, mostly due to syncing the FT
functions with the official artinis oxysoft2matlab functions


schoffelen:
Danke schön

PR #1042, opened 2019-03-26
Fix leadfield_simbio

pierremegevand:
Corrected a small mistake in the code when initializing the leadfield matrix. That did not lead to an error, but initialized the matrix to incorrect dimensions.


schoffelen:
Top! Thanks.

PR #1045, opened 2019-03-27
added the correct diameter to CTF275

ualsbombe:
Follow-up
1. Does the ctf151 have the same diameter?
2. What about the neuromag sizes?

If you have these sizes, I can update and make a pull request

### DIFF
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -182,7 +182,7 @@ if isempty(senssize)
     case 'neuromag122'
       senssize = 35; % FIXME this is only an estimate
     case 'ctf151'
-      senssize = 15; % FIXME this is only an estimate
+      senssize = 20;
     case 'ctf275'
       senssize = 18;
     otherwise
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -184,7 +184,7 @@ if isempty(senssize)
     case 'ctf151'
       senssize = 15; % FIXME this is only an estimate
     case 'ctf275'
-      senssize = 15; % FIXME this is only an estimate
+      senssize = 18;
     otherwise
       if strcmp(sensshape, 'sphere') || strcmp(sensshape, 'disc')
         senssize = 4; % assuming spheres are used for intracranial electrodes, diameter is about 4mm


schoffelen:
ft_hastoolbox('mne', 1);
  [ftver, ftpath] = ft_version;
  def = mne_load_coil_def(fullfile(ftpath, 'external', 'mne', 'coil_def.dat'));

{def.description}'

Although I think that we still can make the distinction between 20 and 18 mm for CTF.

schoffelen:
@robertoostenveld from the documentation on /opt/32bit etc I think I can conclude that the coil diameter for ctf151 is 20 mm, rather than 18. This would be at odds with the Matti's coildef, which does not make a distinction between the different type of ctf systems. Do you think that this would lead to large inaccuracies in forward modeling when using coilaccuracy?

plotting/ft_plot_sens.m
```diff
@@ -184,7 +184,7 @@
     case 'ctf151'
       senssize = 15; % FIXME this is only an estimate
```
robertoostenveld: @ualsbombe can you change this to 1.999 for the CTF151

robertoostenveld:
I had an independent look at it. From 
```
>> hdr = ft_read_header('Subject01.ds')
>> sqrt(hdr.orig.res4.senres(100).area(1)/pi)*2
ans =
    1.9990
```
I would say that the CTF151 coil diameter is 1.999

and for
```
>> hdr = ft_read_header('SubjectRest.ds')
>> sqrt(hdr.orig.res4.senres(100).area(1)/pi)*2
ans =
    1.8000
```

which is a CTF275 recording, it is 1.8.

plotting/ft_plot_sens.m
ualsbombe: Just to be clear - that would be 19.99 in the code, right, because everything is in millimetres, I assume?

schoffelen:
yes.

plotting/ft_plot_sens.m
robertoostenveld: yes. And I propose you make it 20, because that is probably according to design and the difference might be a rounding off error. Round numbers are easier to remember.

ualsbombe: I've changed it to 20 now and pushed to my fork.

Issue #1024, opened 2019-03-23
How do you run the tests?


apjanke:
I'd like to make a PR for FieldTrip and run all the tests before submitting it so I can check that I haven't broken anything. How do you run the full test suite? I don't see anything in the README or the [FieldTrip documentation](http://www.fieldtriptoolbox.org/documentation/) on it.

apjanke:
Oh. I found [this message on the mailing list](https://mailman.science.ru.nl/pipermail/fieldtrip/2010-December/003334.html), which says there's no full test suite run, because it just takes too long. Is this still the case?

schoffelen:
Yes, it is still the case that the execution of the full set of test functions takes a long time. Some functionality to run parts of it (assuming there's no data to be loaded from our local file system) is documented here:

http://www.fieldtriptoolbox.org/development/dashboard/

robertoostenveld:
Oh man, @apjanke is asking a lot of good but undocumented questions ;-)

robertoostenveld:
See https://github.com/fieldtrip/website/commit/07305a6cf6e919989d47a125abb84f6f63d04c5e

Note that I am also piloting the MATLAB CI service, see https://github.com/fieldtrip/playground and https://github.com/fieldtrip/fieldtrip/blob/master/.travis.yml. Some of the tests depend on data that cannot be shared (too large, and/or no permission for sharing). Those can also not run on the MATLAB CI service. The `ft_test` function can be used to filter those out using the `dccnpath no` option.

schoffelen:
Yes, but those questions are very relevant ones, indeed!

apjanke:
Ah, cool.

Is the source code and documentation for the http://dashboard.fieldtriptoolbox.org/api API server available somewhere? I see `contrib/MOxUnit_fieldtrip` but that doesn't look like it has the API serving parts. Is that [`dashboard.js` over in the `fieldtrip/dashboard` repo](https://github.com/fieldtrip/dashboard/blob/master/node/dashboard.js)?

 I'm getting timeouts when retrieving test results and wondering if knowing the API better could help me understand them.

```
>> ft_test report
Error using webread (line 122)
The connection to URL 'http://dashboard.fieldtriptoolbox.org/api/' timed out after 30 seconds. Set options.Timeout to a higher value.
Error in ft_test_report (line 58)
  result = webread([url query], options);
Error in ft_test (line 104)
    ft_test_report(varargin{:}); 
[...]

PR #1049, opened 2019-03-27
enhancement - add option to return 'ft_test report' results in a table

apjanke:
Addresses https://github.com/fieldtrip/fieldtrip/issues/1024#issuecomment-477354678.

After:

```
>> rslt = ft_test('report', 'fieldtripversion', 'cef3396');
>> class(rslt)
ans =
    'table'
>> summary(rslt)
Variables:
    matlabversion: 2480×1 cell array of character vectors
    fieldtripversion: 2480×1 cell array of character vectors
    branch: 2480×1 cell array of character vectors
    arch: 2480×1 cell array of character vectors
    hostname: 2480×1 cell array of character vectors
    user: 2480×1 cell array of character vectors
    passed: 2480×1 logical
        Values:
            True       2448  
            False        32  
    runtime: 2480×1 double
        Values:
            Min             0  
            Median         10  
            Max         11893  
    functionname: 2480×1 cell array of character vectors
>> 
```

If you do not capture the output, behavior does not change.

### DIFF
--- a/utilities/ft_test.m
+++ b/utilities/ft_test.m
@@ -1,4 +1,4 @@
-function ft_test(varargin)
+function out = ft_test(varargin)
 
 % FT_TEST performs selected FieldTrip test scripts or reports on previous test
 % results from the dashboard database.
@@ -58,6 +58,11 @@ function ft_test(varargin)
 %   hostname         = string
 %   user             = string
 %
+% Optionally, you may capture the output to get the results as a Matlab table
+% array, in which case they are not automatically displayed.
+%   rslt = ft_test('report', 'fieldtripversion', 'cef3396');
+%
+%
 % ========= Comparing tests =========
 %
 % To print a table comparing different test results, you would do
@@ -101,11 +106,17 @@ switch (varargin{1})
   case 'moxunit_run'
     ft_test_moxunit_run(varargin{:});
   case 'report'
-    ft_test_report(varargin{:});
+    if nargout == 0
+      ft_test_report(varargin{:});
+    else
+      out = ft_test_report(varargin{:});
+    end
   case 'compare'
     ft_test_compare(varargin{:});
   otherwise
     ft_error('unsupported command "%s"', varargin{1})
 end % switch
 
-
+if nargin == 0
+  clear out
+end
--- a/utilities/private/ft_test_compare.m
+++ b/utilities/private/ft_test_compare.m
@@ -81,7 +81,7 @@ for i=1:numel(functionname)
 end % for each of the features
 
 % convert the struct-array to a table
-table = struct2table(summary);
+table = struct2tablestrs(summary);
 fprintf('%s\n', table{:});
 
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
--- a/utilities/private/ft_test_report.m
+++ b/utilities/private/ft_test_report.m
@@ -82,6 +82,10 @@ if ~showdate
 end
 
 % convert the struct-array to a table
-table = struct2table(result);
-fprintf('%s\n', table{:});
+if nargout == 0
+  tbl = struct2tablestrs(result);
+  fprintf('%s\n', tbl{:});
+else
+  result = struct2table(result);
+end
 
--- a/utilities/private/struct2tablestrs.m
+++ b/utilities/private/struct2tablestrs.m
@@ -1,13 +1,13 @@
-function table = struct2table(s)
+function table = struct2tablestrs(s)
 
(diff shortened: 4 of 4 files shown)


utilities/private/ft_struct2table.m
```diff
@@ -1,13 +1,13 @@
-function table = struct2table(s)
+function table = ft_struct2table(s)
```
robertoostenveld: Please rename this back; I prefer private functions not to be prefixed with ft\_

It makes the high-level code more messy, and they should not interfere with users' functions anyway.

test/mesh27cube.ply
```diff
@@ -0,0 +1,235 @@
+ply
```
robertoostenveld: this .ply file is the result of a test (I don't know which one), and does not belong in the repository.

It would be good to find out which test creates it and to change it such that it is created in tempdir instead.

test/mesh1tet.ply
```diff
@@ -0,0 +1,17 @@
+ply
```
robertoostenveld: this file does not belong here

test/mesh1cube.ply
```diff
@@ -0,0 +1,23 @@
+ply
```
robertoostenveld: this file does not belong here

utilities/private/ft_struct2table.m
apjanke: It needs to be renamed to something, because it is masking Matlab's own `struct2table`, which is needed for the table conversion.

How about `struct2tablestrs`?

test/mesh27cube.ply
apjanke: Removed all the .ply files.

Perhaps `*.ply` should be added to `.gitignore`, for convenience?

utilities/private/ft_struct2table.m
robertoostenveld: hmm, I did not notice that name clash before. But on https://nl.mathworks.com/help/matlab/ref/struct2table.html I see that it already exists since 2013b...

Oh, red flag: I realize that I am using them both. The one in utilities/private is only used in ft_test_xxx, all other uses of struct2table is using the Mathworks version.

Looking at my own struct2table I also find it a rather limited function, it only serves for pretty-printing. I suspect that at the time of implementation (which may predate 2017) I was not willing to use the MATLAB table object yet. Now I think it actually makes more sense to use an actual table object, since those display nicely.

Could you check whether we could get rid of my struct2table and simply use a proper table?

apjanke: Uh oh.

Could you check whether we could get rid of my struct2table and simply use a proper table?

We could do this, except for one thing: when you `disp(t)` a table, it includes ugly quotes around all the cellstrs:

```

>> rslt = ft_test('report', 'fieldtripversion', 'cef3396');
>> disp(rslt(1:10,:))
    matlabversion    fieldtripversion     branch       arch        hostname        user      passed    runtime        functionname     
    _____________    ________________    ________    _________    ___________    ________    ______    _______    _____________________
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c046'    'roboos'    true         2       'test_csp'           
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c041'    'roboos'    true         1       'test_bug840'        
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true         6       'test_bug1149'       
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c037'    'roboos'    true         4       'test_bug1315'       
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true         7       'test_bug1242'       
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c041'    'roboos'    true         1       'test_bug1881'       
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c047'    'roboos'    true         4       'test_ft_plot_lay'   
       '2016b'          'cef3396'        'master'    'glnxa64'    'dccn-c041'    'roboos'    true         1       'test_bug1910'       
```

But that's not a blocker. We could just replace your struct2table with a custom pretty-printing function that accepts Matlab `table` objects as input and produces the same output. That would be a good way to do it - table objects are really nice, so it'd be good to settle on passing them around instead of custom display structures.

apjanke: I have reverted the renames of the other uses of `struct2table` so they now keep calling Matlab's `struct2table`.

apjanke:
Yay!

Fair warning: I'm a total table-head. I think tables (aka [relations](https://www.amazon.com/Introduction-Database-Systems-8th/dp/0321197844)) are one of the most power basic data structures we have at our disposal, and I'm going to try to talk you in to using them everywhere.

PR #1053, opened 2019-03-28
improve consistency of figure decoration

robertoostenveld:
- all figures now use cfg.renderer
- all figures have a FieldTrip menu with about and pipeline
- all figures have a title based on the input variable name

I implemented this in relation to http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2652, where I wanted to check the provenance through the FieldTrip menu item.


robertoostenveld:
especially the figure names are a bit tricky, hence I will first run the test batch on it.

PR #1057, opened 2019-03-28
added returnerror=immediate/final/no

robertoostenveld:
this will cause ft_test to give an error in case one of the script fails.


robertoostenveld:
@apjanke please notice these changes as well. This is it from my side for now.

apjanke:
Looks good.

Issue #1051, opened 2019-03-28
ft_test should use proper tables throughout


robertoostenveld:
and the utilities/private/struct2tablestrs.m should be converted into a pretty-printing  `printable` function, a bit like FT `printstruct` but without the need that it returns an executable string.

See also #1049

robertoostenveld:
Although, I don't know whether we really need printtable. Looking at the table output of ft_test, I get this 
```
>> disp(result)
    matlabversion    fieldtripversion     branch       arch        hostname        user      passed    runtime                    functionname                 
    _____________    ________________    ________    _________    ___________    ________    ______    _______    _____________________________________________

       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c046'    'roboos'    true          1      'test_ft_spike_maketrials'                   
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          0      'test_benchmark_ft_derivative'               
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          0      'test_benchmark_ft_baselinecorrect'          
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          0      'test_benchmark_ft_detrend'                  
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c046'    'roboos'    true          2      'test_bug1207'                               
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c046'    'roboos'    true          3      'test_ft_spike_rate'                         
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          2      'test_bug1618'                               
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          2      'test_csp'                                   
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          4      'test_ft_plot_topo3d'                        
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c036'    'roboos'    true          5      'test_ft_plot_headmodel'                     
       '2016b'          '8bf2326'        'master'    'glnxa64'    'dccn-c046'    'roboos'    true          7      'test_bug1404'                               
[...]

PR #1062, opened 2019-03-29
ENH - return results as proper MATLAB table

robertoostenveld:
as discussed with @apjanke  in https://github.com/fieldtrip/fieldtrip/issues/1051#issuecomment-477829265


apjanke:
👍

PR #1065, opened 2019-04-01
Icosahedron

schoffelen:
OK, this is to get the hang of multiple file editing. I will still need to check a bit, and run some tests before I will merge


schoffelen:
I have ran ft_test('run', 'dependency', 'ft_prepare_mesh'); 

It runs through apart from one test function that misses a required toolbox on my path, which does not have anything to do with my changes.

PR #1066, opened 2019-04-01
 Added documentation to fittemplate method in ft_prepare_mesh, and updated testscript for bug3473

simonhomoelle:


ft_prepare_mesh.m
```diff
@@ -22,12 +22,24 @@
...
 %                     points, or a structure with a single or multiple boundaries
 %
 % For method 'cortexhull' you should not give input data, but specify
-%   cfg.headshape   = sting, filename containing the pial surface computed by freesurfer recon-all
+%   cfg.headshape   = string, filename containing the pial surface computed by freesurfer recon-all
+%
+% For method 'fittemplate' you should specify
+%   cfg.headshape   = a filename containing headshape
+%   cfg.template    = a filename containing headshape
+%   Example use:
```
robertoostenveld: it is not customary to have examples in FieldTrip help, we have the website for that. Furthermore, this makes the examples even more a mess (by having them on multiple locations)

ft_prepare_mesh.m
```diff
@@ -22,12 +22,24 @@
...
+% For method 'headshape' you should specify
 %   cfg.headshape   = a filename containing headshape, a Nx3 matrix with surface
 %                     points, or a structure with a single or multiple boundaries
 %
 % For method 'cortexhull' you should not give input data, but specify
-%   cfg.headshape   = sting, filename containing the pial surface computed by freesurfer recon-all
+%   cfg.headshape   = string, filename containing the pial surface computed by freesurfer recon-all
+%
+% For method 'fittemplate' you should specify
+%   cfg.headshape   = a filename containing headshape
```
robertoostenveld: it does not explain which one is fitted to which

ft_prepare_mesh.m
simonhomoelle: Ok, I will remove the example.

ft_prepare_mesh.m
simonhomoelle: I was just wondering where to explain this. Would it make sense to put the description of the method to line 35 (so replacing the example)?

simonhomoelle: @robertoostenveld what do you think about this suggestion?

simonhomoelle:
@robertoostenveld I updated and tested test_bug3473 on the cluster. It runs through now.

ft_defacevolume.m
```diff
@@ -16,7 +16,7 @@
 % If you specify the box method, the following options apply
 %   cfg.translate  = initial position of the center of the box (default = [0 0 0])
 %   cfg.scale      = initial size of the box along each dimension (default is automatic)
-%   cfg.translate  = initial rotation of the box (default = [0 0 0])
+%   cfg.rotate     = initial rotation of the box (default = [0 0 0])
```
robertoostenveld: please also update ft_defacemesh, it has the same typo

ft_prepare_mesh.m
robertoostenveld: an example is not explanation. Just add a sentence like 

"with this method you are fitting the headshape from the configuration to the template; the resulting affine transformation is applied to the input mesh (or set of meshes), which is subsequently returned as output variable.

ft_prepare_mesh.m
robertoostenveld: the example is still here...

test/test_bug3473.m
```diff
@@ -1,10 +1,11 @@
 function test_bug3473
 
-% WALLTIME 00:10:00
+% WALLTIME 00:30:00
 % MEM 2gb
 
-% DEPENDENCY ft_prepare_mesh 
+% DEPENDENCY ft_prepare_mesh
 
+addpath(genpath(dccnpath('/home/common/matlab/fieldtrip/data/test/bug3473/cpd')));
```
robertoostenveld: should this not be handled by ft_hastoolbox?

ft_prepare_mesh.m
simonhomoelle: Added this documentation

test/test_bug3473.m
simonhomoelle: As we use external software for this, I add it manually

robertoostenveld: ft_hastoolbox is meant for handling external software. It would be enough if it would just give an error and point people to the URL to download the software.

simonhomoelle: Ok this is what it currently is doing!

PR #1067, opened 2019-04-01

matsvanes:


robertoostenveld:
I am looking at this now. I think that ft_prepare_neighbours requires some more work to make it more consistent.

I will write a test script that goes over each of the input fields

PR #1069, opened 2019-04-03
enh: added disc plotting utility to test battery, and cylinder subfunction to plot_sens and plot_cloud

StolkArjen:
This fixes an issue introduced by one of my own functions being on the path. To this end, I have included a cylinder generation tool as a subfunction in ft_plot_sens and ft_plot_cloud, which is used to create disc-shaped electrodes that are aligned to the mesh (think of an ECoG grid/strip array).

Second, this PR includes updates to the test functions corresponding to the above functions.

### DIFF
--- a/plotting/ft_plot_cloud.m
+++ b/plotting/ft_plot_cloud.m
@@ -707,7 +707,7 @@ else % plot 3d cloud
       Fn = Fn * (1/sqrt(sum(Fn.^2,2))); % normalize
       
       % create disc aligned with the headshape (ideally, a hull)
-      [X,Y,Z] = cylinder2([rmax(n) rmax(n)],[Fn(:,1) Fn(:,2) Fn(:,3)], 100);
+      [X,Y,Z] = cylinder([rmax(n) rmax(n)],[Fn(:,1) Fn(:,2) Fn(:,3)], 100);
       X(1,:) = X(1,:)+pos(n,1); Y(1,:) = Y(1,:)+pos(n,2); Z(1,:) = Z(1,:)+pos(n,3);
       t = rmax(n)/10; % add thickness (outward), X(2,1)-X(1,1) etc.
       X(2,:) = X(1,:)-t*Fn(:,1); Y(2,:) = Y(1,:)-t*Fn(:,2); Z(2,:) = Z(1,:)-t*Fn(:,3);
@@ -805,3 +805,48 @@ else % plot 3d cloud
     caxis(gca, clim);
   end
 end
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% SUBFUNCTION return a cylinder with radius r, direction d, and n points
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+function [XX,YY,ZZ] = cylinder(r,d,n)
+
+% basic cylinder
+r = r(:);
+d = d(:);
+m = length(r);
+theta = (0:n)/n*2*pi;
+sintheta = sin(theta); 
+sintheta(n+1) = 0;
+x = r * cos(theta);
+y = r * sintheta;
+z = (0:m-1)'/(m-1) * ones(1,n+1);
+
+% rotation axis and angle
+d0      = [0,0,1];
+rotaxis = cross(d0,d);
+if norm(rotaxis)==0
+  rotaxis = [1,0,0];
+end
+rotaxis = rotaxis/norm(rotaxis);
+angle   = -atan2(norm(cross(d0,d)),dot(d0,d));
+
+% rotation quaternion
+q(1:3,1) = rotaxis*sin(angle/2);
+q(4,1)   = cos(angle/2);
+
+% rotation matrix
+Q   = [0, -q(3), q(2);q(3), 0, -q(1);-q(2), q(1), 0];
+C   = eye(3)*(q(4)^2-q(1:3)'*q(1:3))+2*q(1:3)*q(1:3)'-2*q(4)*Q;
+
+% generate cylinder
+x1      = reshape(x,1,size(x,1)*size(x,2));
+y1      = reshape(y,1,size(y,1)*size(y,2));
+z1      = reshape(z,1,size(z,1)*size(z,2));
+M       = (C*[x1;y1;z1])';
+X1      = M(:,1);
+Y1      = M(:,2);
+Z1      = M(:,3);
+XX      = reshape(X1,size(x,1),size(x,2));
+YY      = reshape(Y1,size(y,1),size(y,2));
+ZZ      = reshape(Z1,size(z,1),size(z,2));
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -349,7 +349,7 @@ switch sensshape
     
   case 'circle'
     plotcoil(pos, ori, [], senssize, sensshape, 'edgecolor', edgecolor, 'facecolor', facecolor, 'edgealpha', edgealpha, 'facealpha', facealpha);
-
+    
   case 'square'
     % determine the rotation-around-the-axis of each sensor
     % only applicable for neuromag planar gradiometers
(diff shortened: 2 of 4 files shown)


plotting/ft_plot_cloud.m
```diff
@@ -805,3 +805,48 @@ function ft_plot_cloud(pos, val, varargin)
     caxis(gca, clim);
   end
 end
+
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+% SUBFUNCTION return a cylinder with radius r, direction d, and n points
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+function [XX,YY,ZZ] = cylinder(r,d,n)
```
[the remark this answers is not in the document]
robertoostenveld: probably you want to combine that with the private rotate and translate functions, and use ft_warp_apply. All of them are already in plotting/private. See ft_plot_dipole for an example.

plotting/ft_plot_sens.m
```diff
@@ -423,13 +423,13 @@
       ori(i,:) = Fn;
       
       % create disc aligned with the headshape (ideally, a hull)
-      [X,Y,Z] = cylinder2([senssize/2 senssize/2],[ori(i,1) ori(i,2) ori(i,3)], 100);
+      [X,Y,Z] = cylinder([senssize/2 senssize/2],[ori(i,1) ori(i,2) ori(i,3)], 100);
```
robertoostenveld: also here: use the existing cylinder function

test/test_ft_plot_cloud.m
```diff
@@ -75,6 +76,9 @@
 figure; ft_plot_cloud(pos, val, 'mesh', {mesh1, mesh2}, 'facealpha', 0, 'edgealpha', [.5 1], 'edgecolor', {'r', 'b'});
 figure; ft_plot_cloud(pos, val, 'mesh', {mesh1, mesh2}, 'facealpha', 0, 'edgealpha', [.5 1], 'edgecolor', {[1 0 0], [0 0 1]});
 
+% plot electrodes as discs
```
robertoostenveld: why do you want to plot electrodes with ft_plot_cloud and not with ft_plot_sens?

Does this make the electrodes translucent or so?

StolkArjen: ft_plot_cloud is a helper function for ft_sourceplot and plots the electrodes as various shapes including clouds, discs, etc and color-codes those shapes based on the functional data.

plotting/ft_plot_cloud.m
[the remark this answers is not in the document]
StolkArjen: Hadn't seen it, but will check whether it supports the same functionality

test/test_ft_plot_cloud.m
robertoostenveld: the difference between ft_plot_cloud and ft_plot_sens should remain clear. I don't want confusing/conflicting plotting implementations in both of  them. Clouds (like puffs of smoke) don't have a direction, electrodes do.

schoffelen:
Uhmm, spuit 11: it seems that Arjen's cylinder function does not return a mesh, so it behaves different from the cylinder function in fieldtrip (which is nominated to be renamed into mesh_cylinder, as per PR1073)

test/test_ft_plot_cloud.m
robertoostenveld: plotting scalar values per channel, linked to the geometrical representation of that channel (e.g. as a disk for EEG, cylinder for sEEG, square for neuromag MEG), should be implemented on a single place.

3D rendering of geometry (cortex, scalp) together with scalar values on specific locations (e.g. at the electrodes or magnetometers) is a relevant feature to have, but should be designed in a wider perspective than just iEEG.

robertoostenveld:
> Uhmm, spuit 11: it seems that Arjen's cylinder function does not return a mesh, so it behaves different from the cylinder function in fieldtrip (which is nominated to be renamed into mesh_cylinder, as per PR1073)

what does it return?

schoffelen:
I don't know, but something called X, Y, and Z, which is one output argument more than a pos/tri pair

test/test_ft_plot_cloud.m
StolkArjen: If I am interpreting the above correctly, do you think we need a separate ft_plot_disc functionality? In a way, 'disc' is an elecshape so it could belong in ft_plot_sens, which already supports headshapes as inputs. It belongs less in ft_plot_cloud given that the discs do have a direction (being aligned with the headshape). However, ft_plot_cloud is the closest we got to plotting functional data in electrodes/clouds/etc -shape.

StolkArjen:
> I don't know, but something called X, Y, and Z, which is one output argument more than a pos/tri pair

JM right, 'cylinder' gives the 3d coordinates of the cylinder. Maybe we could relabel it into 'disc' then to more clearly distinguish it from the mesh cylinder functionality.

robertoostenveld:
A cylinder consists of a surface. Do X, Y, Z, describe the vertices, or are they somehow used to describe surface elements?

schoffelen:
the matlab cylinder returns the x/y/z coordinates that describe the top and bottom circle it seems. Then the thing can be visualized with the matlab mesh function, which creates the surface on the fly.

robertoostenveld:
In that case pos=[X(:) Y(:) Z(:)]. We don't need another cylinder implementation, please use the one that we have and `ft_warp_apply`.

StolkArjen:
Okay, let me look into replacing the current functionality by that one on the first next opportunity. If you could advise on the other thread in this PR, that would be appreciated too.

robertoostenveld:
the other being that ft_plot_cloud should not be replicating ft_plot_sens? That is hard, since the use case you have in mind is still not clear to me. 

ft_plot_sens can already plot colored electrodes ('point', 'circle', 'square', 'sphere', or 'disc'). It seems to me that extending it around line 379 would be easy. The code there already seems to use the same MATLAB "sphere" as the "cylinder" that you were using in your code (without the triangulation.

StolkArjen:
The use case is to give functional data to ft_sourceplot, or ft_plot_cloud directly, which will then plot and color the electrode discs based on the data and settings including clim. Technically, it might be possible to wrap  Ifdaround ft_plot_sens by getting the handles for each disc and coloring them after plotting. I would have to look into that too, as I vaguely recall it wasn't that straightforward. Will let you know. If that works, however, we might even take out the sphere and other overlapping plotting functionalities between ft_plot_sens and ft_plot_cloud (e.g., the sphere), in order to keep ft_plot_cloud dedicated to just cloud plotting.

robertoostenveld:
Plotting electrodes is what ft_plot_sens is for. Please remove the functionality from ft_plot_cloud (which is for puffs). And I don’t see a reason to wrap the two aroudn each other. Better wrap ft_sourceplot around them individually.

StolkArjen:
That makes sense. Given the size of the overhaul (note to myself: reorganization of plotting tools and cylinder generation), I will continue this project when there is a large enough window.  

Thinking along, plotting the electrodes as discs along with the brain/pial mesh requires specification of the cortical hull (for aligning the discs with, or some head shape in the case of scalp-EEG) to ft_sourceplot, besides the brain/pial mesh. This would possibly require a separate cfg input argument for the hull (which has a smooth outline as compared to the folded brain), e.g. cfg.outline, cfg.headshape, cfg.hull, cfg.align_the_discs_to_this_surface, you get the point. Any thoughts on that?

robertoostenveld:
I have started a branch https://github.com/robertoostenveld/fieldtrip/tree/plotcloud for improving ft_plot_cloud. So far I mainly I updated the handling of units inside the function, making the cloud density consistent with the input. 

Other options than cloudtype=cloud need to be tested and possibly redistributed to other functions (e.g. mesh intersections are to be plotted with ft_plot_mesh). This also means that ft_sourceplot should not simply dump everything into ft_plot_cloud, but should call multiple ft_plot_xxx functions to build the complete visualization.

StolkArjen:
For starters, I think it'd be best to see if we can move the disc plotting utility (providing the most realistic representation of an ECoG/EEG electrode) to ft_plot_sens, and then have ft_sourceplot call ft_plot_sens and use its handle to color the disc electrodes afterward. This would require an additional cfg.hull/cfg.outline input argument to ft_sourceplot (for aligning the discs with). From there, it's a smaller step to move other parts of ft_plot_cloud over, although it's a 1000 line beast made of cloud-specific functionality. You can close this PR if you agree.

robertoostenveld:
agreed. I will close this one and we should start on other PRs to get the functionality implemented in ft_plot_sens.

StolkArjen:
Oki doki, would you mind committing the updates contained by your plotcloud branch?

PR #1075, opened 2019-04-04
Issue with preproc.m when using cfg.implicitref

rtheze:
bugfix - fixed a problem line 260. When calling **cfg.implicitref**, the implicit ref is added to the list of channels **label** - a Nchan x 1  cell array - but the output **label** becomes 1 x Nchan cell array, whereas it should be Nchan x 1 cell array. This can be solved by adding an apostrophe operator to transpose the resulting array.


robertoostenveld:
thanks

PR #1077, opened 2019-04-08
ENH: built in support for IRASA in ft_freqanalysis

StolkArjen:
Irregular-resampling auto-spectral analysis (IRASA, Wen & Liu, 2016) extracts the fractal/arrhythmic component from the time domain signal by stretching and extending it multiple times with irregular amounts prior to taking the FFT. Essentially, this moves rhythmic content up and down the power spectrum with various distances, and the median power spectrum then gives the fractal/arrhythmic component. See below for a simulation.

The code, in ft_specest_irasa, is based on ft_specest_mtmfft, preserving as many functionalities as possible to keep it consistent with the rest of the specest module (e.g., for scaling up to tfr purposes). The tapers, Hannings with variable resampling-based lengths, are computed once for speedup purposes.

I plan to further develop the below script into a wiki example script. This example/test script will also include fitting of Gaussians to estimate (individual) rhythm frequency bands. 

One thing that still puzzles me, however, is the overall offset/downshift of the fractal component relative to the original power spectrum. I am pretty sure this is not because of the irasa vs. mtmfft implementation since I checked and because my previous (all-in-one) implementation occasionally showed the same offset. It's also less obvious with fewer trials, e.g. set rpt = 1 in the below simulation. Any thoughts on this would be _much appreciated_. 

 

% simulate 100 trials of arrhythmic and rhythmic data
for rpt = 1:100
  N = 1000;
  t = 1 : N;
  k = (1 : N/2)';
  kt = k*t;
  theta = unifrnd(0,2*pi,N/2,1);
  Ck = repmat(sqrt(1*k.^(-1)*N),1,N);
  frac_sim = sum(Ck.*cos(2*pi*kt/N - repmat(theta,1,N)));
  
  tim = (1:N)/N;
  osci_sim = cos(2*pi*15*tim)*50;
  mixed_sim = frac_sim+osci_sim;
  
  sim.trial{1,rpt} = mixed_sim;
  sim.time{1,rpt} = tim;
  sim.label{1} = 'sim';
end

% fft
cfg = [];
cfg.foilim = [1 200];
cfg.taper = 'hanning';
cfg.method = 'irasa';
frac = ft_freqanalysis(cfg, sim);
cfg.method = 'mtmfft';
orig = ft_freqanalysis(cfg, sim);

cfg = [];
cfg.parameter = 'powspctrm';
cfg.operation = 'x2-x1';
osci = ft_math(cfg, frac, orig);

% plot
figure; loglog(frac.freq, frac.powspctrm)
hold on; loglog(orig.freq, orig.powspctrm)
legend('fractal', 'original')

figure; plot(osci.freq, osci.powspctrm)
legend('oscillatory')

Screen Shot 2019-04-07 at 8 13 14 PM

### DIFF
--- a/specest/ft_specest_irasa.m
+++ b/specest/ft_specest_irasa.m
@@ -1,0 +1,210 @@
+function [spectrum,ntaper,freqoi] = ft_specest_irasa(dat, time, varargin)
+
+% FT_SPECEST_IRASA performs Irregular-Resampling Auto-Spectral Analysis
+% (Wen et al., 2016) to estimate the powerspectral arrythmic component in the
+% time-domain signal. This arrythmic component can be used to identify
+% individual rhythmic components in the powerspectrum (see Stolk et al., 2019)
+%
+% Use as
+%   [spectrum,ntaper,freqoi] = ft_specest_irasa(dat,time...)
+% where
+%   dat        = matrix of chan*sample
+%   time       = vector, containing time in seconds for each sample
+%   spectrum   = matrix of taper*chan*freqoi of fourier coefficients
+%   ntaper     = vector containing number of tapers per element of freqoi
+%   freqoi     = vector of frequencies in spectrum
+%
+% Optional arguments should be specified in key-value pairs and can include
+%   taper      = 'dpss', 'hanning' or many others, see WINDOW (default = 'dpss')
+%   pad        = number, total length of data after zero padding (in seconds)
+%   padtype    = string, indicating type of padding to be used (see ft_preproc_padding, default: zero)
+%   freqoi     = vector, containing frequencies of interest
+%   tapsmofrq  = the amount of spectral smoothing through multi-tapering. Note: 4 Hz smoothing means plus-minus 4 Hz, i.e. a 8 Hz smoothing box
+%   dimord     = 'tap_chan_freq' (default) or 'chan_time_freqtap' for memory efficiency (only used when variable number slepian tapers)
+%   polyorder  = number, the order of the polynomial to fitted to and removed from the data prior to the fourier transform (default = 0 -> remove DC-component)
+%   taperopt   = additional taper options to be used in the WINDOW function, see WINDOW
+%   verbose    = output progress to console (0 or 1, default 1)
+%
+% See also FT_FREQANALYSIS, FT_SPECEST_MTMFFT, FT_SPECEST_MTMCONVOL, FT_SPECEST_TFR, FT_SPECEST_HILBERT, FT_SPECEST_WAVELET
+
+% Copyright (C) 2019, Arjen Stolk
+%
+% This file is part of FieldTrip, see http://www.fieldtriptoolbox.org
+% for the documentation and details.
+%
+%    FieldTrip is free software: you can redistribute it and/or modify
+%    it under the terms of the GNU General Public License as published by
+%    the Free Software Foundation, either version 3 of the License, or
+%    (at your option) any later version.
+%
+%    FieldTrip is distributed in the hope that it will be useful,
+%    but WITHOUT ANY WARRANTY; without even the implied warranty of
+%    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+%    GNU General Public License for more details.
+%
+%    You should have received a copy of the GNU General Public License
+%    along with FieldTrip. If not, see <http://www.gnu.org/licenses/>.
+%
+% $Id$
+
+% these are for speeding up computation of tapers on subsequent calls
+persistent previous_argin previous_tap
+
+% get the optional input arguments
+taper     = ft_getopt(varargin, 'taper'); if isempty(taper), ft_error('You must specify a taper'); end
+pad       = ft_getopt(varargin, 'pad');
+padtype   = ft_getopt(varargin, 'padtype', 'zero');
+freqoi    = ft_getopt(varargin, 'freqoi', 'all');
+tapsmofrq = ft_getopt(varargin, 'tapsmofrq');
+dimord    = ft_getopt(varargin, 'dimord', 'tap_chan_freq');
+fbopt     = ft_getopt(varargin, 'feedback');
+verbose   = ft_getopt(varargin, 'verbose', true);
+polyorder = ft_getopt(varargin, 'polyorder', 0);
+tapopt    = ft_getopt(varargin, 'taperopt');
+
+% Set IRASA resampling factors
+hset = 1.1:0.05:1.9;
+nhset = length(hset);
+
+if isempty(fbopt)
+  fbopt.i = 1;
+  fbopt.n = 1;
+end
+
+% throw errors for required input
+if isempty(tapsmofrq) && (strcmp(taper, 'dpss') || strcmp(taper, 'sine'))
+  ft_error('you need to specify tapsmofrq when using dpss or sine tapers')
+end
+
+% this does not work on integer data
+dat = cast(dat, 'double');
+
+% Set n's
+[nchan,ndatsample] = size(dat);
+
+% This does not work on integer data
+if ~isa(dat, 'double') && ~isa(dat, 'single')
+  dat = cast(dat, 'double');
+end
+
+% Remove polynomial fit from the data -> default is demeaning
+if polyorder >= 0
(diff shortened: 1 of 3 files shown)


robertoostenveld:
looks good. Could you add the example code to test/test_ft_specest_irasa, and if possible also add the corresponding code that uses ft_freqanalysis?

StolkArjen:
That's the plan. Do you know if we have a tool that can simulate a signal with 1/f content? ft_connectivitysimulation can generate oscillations but not 1/f, I thought.

schoffelen:
I have a pinknoise function, which attempts to generate 1/f in various ways.

robertoostenveld:
perhaps something to add to ft_freqsimulation for testing?

StolkArjen:
If you don't mind adding your pinknoise function to ft_freqsimulation @schoffelen, I'll use that in the example/test script.

specest/ft_specest_irasa.m
```diff
@@ -0,0 +1,213 @@
...
+%   taper      = 'dpss', 'hanning' or many others, see WINDOW (default = 'hanning')
+%   pad        = number, total length of data after zero padding (in seconds)
+%   padtype    = string, indicating type of padding to be used (see ft_preproc_padding, default: zero)
+%   freqoi     = vector, containing frequencies of interest
+%   tapsmofrq  = the amount of spectral smoothing through multi-tapering. Note: 4 Hz smoothing means plus-minus 4 Hz, i.e. a 8 Hz smoothing box
+%   dimord     = 'tap_chan_freq' (default) or 'chan_time_freqtap' for memory efficiency (only used when variable number slepian tapers)
+%   polyorder  = number, the order of the polynomial to fitted to and removed from the data prior to the fourier transform (default = 0 -> remove DC-component)
+%   taperopt   = additional taper options to be used in the WINDOW function, see WINDOW
+%   verbose    = output progress to console (0 or 1, default 1)
+%
```
robertoostenveld: Can you add a reference to the original paper here? And please also add it to https://github.com/fieldtrip/website/blob/master/references_to_implemented_methods.md

StolkArjen:
I plan to finalize this during the weekend: @schoffelen, it's not a priority but if you were planning on adding in your pink noise functionality, I could wait

StolkArjen:
Oki doki, this one is good to go. Created an example/test script and added the paper to the list of references. I will ask the original authors to double check my implementation once this is pulled in.

StolkArjen:
Output of the test script:

[tessie.pdf](https://github.com/fieldtrip/fieldtrip/files/3099751/tessie.pdf)

jens-k:
Are you sure the function returns the oscillatory component and not the fractal component?

I used it on ony of my datasets and compared it to the result of the code provided by the authors of the original study (https://purr.purdue.edu/publications/1987/1). That code gives me what I would have expected while the Fieldtrip version gives me a very fractal-looking estimate.

Here a comparison of 
left: the output of freqanalysis (.powspctrm) with cfg.method = 'irasa';
center: the fractal component as returned from the reference code;
right: the oscillatory component as returned from the reference code.

image

cfg = [];
cfg.foi = 3:.2:22;
cfg.method = 'irasa';
cfg.channel = 'C3';
freq = ft_freqanalysis(cfg, data)

jens-k:
Also, have you thought of returning both the fractal and the oscillatory component? 
This way one can calculate the ratio between the two, which can correct for the fact that also deviations from the fractal component are stronger at longer frequencies. In above dataset (sleep recordings) this ratio shows a clear (and expected) peak in the sleep spindle band.

image


(Shout out to Til Ole Bergmann for coming up with this.)

robertoostenveld:
@StolkArjen can you look into this? If needed, please open a new issue...

StolkArjen:
Indeed, cfg.method = 'irasa' gives the fractal component, not the oscillatory component (can't find where it incorrectly said that?).

Robert and I discussed the option of outputting both at the same time, but that would require extra fields in the output (which might not be compatible with downstream functionality), require calculating the regular power spectrum on the fly (code of which is already implemented in mtmfft), and make the application less transparent (in contrast to separate calls to irasa and mtmfft and then their subtraction - see example/test code in this PR).

jens-k:
Mmh, seems I just assumed the output would be the oscillatory component and read the docu accordingly. Sorry! So the user is expected to calculate the mixed spectrum and the fractal component separately and do the correction by hand?
What about adding another parameter that specifies the desired output (fractal, oscillatory, percent change)?

Thanks for implementing this by the way!

StolkArjen:
Appreciate your appreciation 👍 

Note that IRASA stands for irregular-resampling auto-spectral analysis, which is what it does, i.e. estimating/extracting the fractal component. Adding an extra output option would require running the analysis twice if one wanted to obtain both the fractal and oscillatory component (given a single .powspctrm field in the output).

jens-k:
Ok, I understand. Intuitively, however, I'd say that most people want to use IRASA to get the oscillatory component, don't you think? For example, because a) it's a great way to detect peaks in the spectrum, like the spindle peak I postet above, and b) it might be a way to normalize power estimates for across-subjects comparisons. Anyways, just a thought.

StolkArjen:
Thanks for the suggestion. Were you able to detect that spindle component you showed with the current implementation?

robertoostenveld:
I propose to add to the documentation that the user should use `ft_math` to get the spectral component, by subtracting the fractal spectrum from the mixed spectrum..

jens-k:
Here is some code to get from the fieldtrip implementation to the oscillatory component and the percent change. The used data (1.8 MB) is here: https://www.dropbox.com/s/jq4swggo0onwv5m/RC101_clean.mat?dl=0

```
load RC101_clean.mat
cfg = [];
cfg.foi = 3:.2:22;
cfg.method = 'irasa';
cfg.channel = 'C3';
frac = ft_freqanalysis(cfg, data); % get fractal component
subplot(4,1,1), plot(frac.freq, frac.powspctrm)

cfg.method = 'mtmfft';
cfg.taper = 'hanning';
mixed = ft_freqanalysis(cfg, data); % get mixed spectrum 
subplot(4,1,2), plot(mixed.freq, mixed.powspctrm)

cfg = [];
cfg.parameter = 'powspctrm';
cfg.operation = 'subtract';
osci = ft_math(cfg, mixed, frac); % subtract fractal component to get to oscillatory component
subplot(4,1,3), plot(osci.freq, osci.powspctrm)

cfg.operation = 'divide';
perc = ft_math(cfg, osci, frac); % calculate percent change
subplot(4,1,4), plot(perc.freq, perc.powspctrm)
```

image

StolkArjen:
You successfull reproduced the example script. :) Is that 4th figure supposed to be the same as the one you sent earlier, based on the original implementation and with the sharp peak around 14 Hz? Or is this different data?

jens-k:
Ah, example script, now I saw it too...

It shows the same phenomeon but in different data. Here is a direct comparison of the 'reference' implementation (top) and the fieldtrip implementation (bottom), based on the data from my last post. Not sure where the slight differences come from (both use a hanning window before the fft).

image

StolkArjen:
Yes, this looks more like expected. There are still minimal differences between the original and my implementation, in particular, related to sub-segmenting step A of the algorithm described in Wen & Liu, Brain Topogr 2016. In the original implementation, this step changes the frequency resolution whereas in the current fieldtrip implementation that frequency resolution is directly based on the specified window size. As employed in this PR https://github.com/fieldtrip/fieldtrip/pull/1092, I am adding a warning until further notice, depending on discussion with the original authors about the necessity/confounding effects of step A.

jens-k:
Ok excellent, seems everything works as desired then, sorry for the trouble. 

Where is the documentation Robert mentioned gonna be? I'd be glad to help.

schoffelen:
> 
> 
> I plan to finalize this during the weekend: @schoffelen, it's not a priority but if you were planning on adding in your pink noise functionality, I could wait

Hier spuit 11. I think that it would be good to investigate the shared functionality/options in ft_dipolesimulation/ft_connectivitysimulation/ft_freqsimulation. I would assume that most of them have an option to add noise to it. Perhaps we could consider a generic noise function with a method flag that can be 'white', 'pink', 'brown' etc. My pinknoise version has a few 'methods' implemented, which range from IIR-filtering of randn, to inverse fft-ing of a specified 1/f^something passband, to an AR-generative model

PR #1078, opened 2019-04-08
add 'grid' and 'vol' to the recursesize list.

matsvanes:
Old style cfg options with large fields were not yet saved by reproducescript. Now grid and vol are added to the recursesize list and saved if large. Probably these two options are used most often instead of sourcemodel and headmodel, because they were changed recently. If it turns out that other fields need this functionality, we will have to add them to this list, but for now just these two are fine (in consultation with @robertoostenveld)


robertoostenveld:
There was a conflict due to the list being sorted differently in yours than in the shared branch. I have resolved it. Please update your copy of the reproducescript branch to prevent further conflicts.

matsvanes:
I just encountered another, but similar problem with cfg.matrix in ft_math. I tried adding it to the recursesize list as well, but the issue is that cfg.callinfo.usercfg.matrix is cleared by checkconfig.

robertoostenveld:
I think another issue is that cfg.matrix is not documented. I did actually not know about this option, but see it was added by @schoffelen. I have just submitted a PR to resolve this.

PR #1080, opened 2019-04-08
improved handling of units for plotting of clouds

robertoostenveld:
as requested in https://github.com/fieldtrip/fieldtrip/pull/1069#issuecomment-480960566

I have not reviewed whether the functionality as used in http://www.fieldtriptoolbox.org/tutorial/human_ecog/#seeg-data-representation still works. That shoudl first be checked prior to merging this into the master (i.e. release) branch.


StolkArjen:
The code and variable name changes look good time. Cannot test it as I'm traveling. The only critical change seems to be that scaling factor.

robertoostenveld:
the code was internally rescaling input positions, but not input parameters. In general, users are required to specify parameters that are consistent with input data. Now the "unit" parameter only affects the defaults (which are specified in "mm", but will be numerically scaled to the right magnitide). I will test and if ok, merge.

StolkArjen:
I vaguely recall that might have been because of ft_scalingfactor occasionally failing to determine the correct scaling (in the past). For instance, when there was just one depth probe with close by electrodes, it would over/underestimate the units.

robertoostenveld:
Please note that I have removed the cloudtype `point` and `disk` option. It will give an error if you still try to use them. 

I have tested the cloud plotting for `cloud` and `surf` using http://www.fieldtriptoolbox.org/tutorial/human_ecog/#interactive-plotting and it still works.

StolkArjen:
Fine about disc. We did, however, present the point as a cloudtype in https://static-content.springer.com/esm/art%3A10.1038%2Fs41596-018-0009-6/MediaObjects/41596_2018_9_MOESM1_ESM.pdf

robertoostenveld:
I don't find `point` in http://www.fieldtriptoolbox.org/tutorial/human_ecog or in the test script.... ah, it is in the supplement as figure without code. We should look at that, but also at the surf: it looks different in the figure than what I recall from the screen. Too bad that we don't have code to reproduce the figure...

Screen Shot 2019-04-18 at 08 58 27

StolkArjen:
There’s the full tutorial example script (suppl data 1) but it needs the tutorial data (zenodo). It’s probably better to keep the point cloudtype in plot cloud. Unlike disc, which is about electrodes, point scales with effect sizes (as all cloudtypes do)

StolkArjen:
> I don't find `point` in http://www.fieldtriptoolbox.org/tutorial/human_ecog or in the test script.... ah, it is in the supplement as figure without code. We should look at that, but also at the surf: it looks different in the figure than what I recall from the screen. Too bad that we don't have code to reproduce the figure...
> 
> Screen Shot 2019-04-18 at 08 58 27

The code is in Step 56 (likely needs a few more preceding steps), and the full pipeline is here: https://static-content.springer.com/esm/art%3A10.1038%2Fs41596-018-0009-6/MediaObjects/41596_2018_9_MOESM3_ESM.pdf (needs the Zenodo data). 

However, much easier even, there's a test_ft_plot_cloud for creating the figures, in particular testing the 'point' cloudtype (which is probably now lacking from ft_plot_cloud).

PR #1086, opened 2019-04-18
Update for field selection in ft_freqstatistcs

rtheze:
bugfix -  ft_freqstatistics has the cfg option 'select' and 'nanmean' but line 117 the function keepfields creates a temporary cfg for input to ft_selectdata without those fields. As a consequence ft_selectdata uses  the "default" options ('intersect' instead of 'union' for the field "select" and 'no' instead of 'yes' for the field "nanmean").


robertoostenveld:
thanks. I will also update ft_timelockstatistic and ft_sourcestatistics accordingly.

PR #1096, opened 2019-04-28
fix: added back in point cloud-method

StolkArjen:
This fixes the issue at https://github.com/fieldtrip/fieldtrip/pull/1080, in which the 'point' cloud-method was removed.

### DIFF
--- a/plotting/ft_plot_cloud.m
+++ b/plotting/ft_plot_cloud.m
@@ -1,6 +1,6 @@
 function ft_plot_cloud(pos, val, varargin)
 
-% FT_PLOT_CLOUD visualizes spatially sparse scalar data as points, spheres, discs, or
+% FT_PLOT_CLOUD visualizes spatially sparse scalar data as points, spheres, or
 % spherical clouds of points and optionally 2D slices through the spherical clouds
 %
 % Use as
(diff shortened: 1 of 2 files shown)


robertoostenveld:
please implement this in ft_sourceplot using ft_plot_sens, not by passing all cfg options from ft_sourceplot to ft_plot_cloud

StolkArjen:
Should the few hundred lines of code (144-635) that implement the color and size scaling (based on functional data) as well as the slicing of meshes (for plotting clouds on) also be copied over to ft_plot_sens then? It seems optimal to have the point cloudtype in ft_plot_cloud where it uses the same "preparations" as the cloud and sphere cloudtypes. Otherwise, there would be the same functionality in different places.

robertoostenveld:
Plotting of values at sensors (as points, disks, squares) is already available in ft_plot_sens, including sensor orientation, which is useful for (new) cylinders. That code should be reused rather than reimplemented.

StolkArjen:
You're describing the disc implementation, which uses orientations and cylinders, was rightfully removed from ft_plot_cloud, and is implemented in ft_plot_sens. This PR is about the plotting data clouds (point-shaped, far-right in below image), which scales and colors clouds based on the functional data, requires most of the code contained in ft_plot_cloud, and is now (permanently) lost. 


https://user-images.githubusercontent.com/899043/56342378-21c5d200-61b8-11e9-9588-a98e47a14a63.png

Issue #1055, opened 2019-03-28
Feature request: superlets


DanBenHa:
Hi,

I recently came upon this manuscript: https://doi.org/10.1101/583732 

Are there plans to implement this superlet decomposition in FieldTrip?

Best wishes,
Daniel

robertoostenveld:
We have no plans for this here in Nijmegen, but we do welcome contributions from your side. 
See http://www.fieldtriptoolbox.org/contribute/#contribute-code.

PR #1103, opened 2019-05-03
enhancement: spectral estimation using superlets

DanBenHa:
Added superlet estimation as a sub-method of the 'wavelet' method. Calculation according to Moca et al. 2019 (https://doi.org/10.1101/583732). This is related to #1055 .

I hope this is useful and that I complied with all the guidelines, since this is my first pull request.

Best wishes,
Daniel

### DIFF
--- a/ft_freqanalysis.m
+++ b/ft_freqanalysis.m
@@ -564,13 +564,13 @@ for itrial = 1:ntrials
         end
 
         % compute superlets
-        spectrum = NaN(1,length(cfg.foi),length(cfg.toi));
+        spectrum = NaN(nchan,length(cfg.foi),length(cfg.toi));
         % index of 'freqoi' value in 'options'
         idx_freqoi = find(ismember(options(1:2:end), 'freqoi'))*2;
         foi = options{idx_freqoi};
         for i_f = 1:length(cfg.foi)
           % collext individual wavelets' responses per frequency
-          spec_f = NaN(cfg.sl_order(i_f), 1, length(cfg.toi));
+          spec_f = NaN(cfg.sl_order(i_f), nchan, length(cfg.toi));
           opt = options;
           opt{idx_freqoi} = cfg.foi(i_f);
           % compute responses for individual wavelets
@@ -578,7 +578,7 @@ for itrial = 1:ntrials
             [spec_f(i_wl,:,:),~,toi] = ft_specest_wavelet(dat, time, 'timeoi', cfg.toi, 'width', cycles{i_f}(i_wl), 'gwidth', cfg.gwidth, opt{:}, 'feedback', fbopt);
           end
           % geometric mean across individual wavelets
-          spectrum(1,i_f,:) = prod(spec_f, 1).^(1/cfg.sl_order(i_f));
+          spectrum(:,i_f,:) = prod(spec_f, 1).^(1/cfg.sl_order(i_f));
         end
         clear spec_f
       end
--- a/ft_freqanalysis.m
+++ b/ft_freqanalysis.m
@@ -106,10 +106,18 @@ function [freq] = ft_freqanalysis(cfg, data)
 %   cfg.foilim     = [begin end], frequency band of interest
 %   cfg.toi        = vector 1 x numtoi, the times on which the analysis
 %                    windows should be centered (in seconds)
-%   cfg.width      = 'width', or number of cycles, of the wavelet (default = 7)
+%   cfg.width      = 'width', or number of cycles, of the wavelet or base wavelet for 
+%                    superlets(default = 7)
 %   cfg.gwidth     = determines the length of the used wavelets in standard
 %                    deviations of the implicit Gaussian kernel and should
 %                    be choosen >= 3; (default = 3)
+%   cfg.superlets  = 'no', 'additive', 'multiplicative' (default = 'no')
+%                    If set to value other than 'no' performs superlet spectral estimation 
+%                    according to Moca et al. 2019 (https://doi.org/10.1101/583732), i.e.
+%                    frequency-wise combination of responses from Morlet wavelets of varying
+%                    cycle width using the geometric mean.
+%   cfg.sl_order   = vector 1 x numfoi, superlet order, i.e. number of combined wavelets, for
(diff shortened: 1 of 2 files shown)


ft_freqanalysis.m
```diff
@@ -106,10 +106,18 @@
...
 %                    windows should be centered (in seconds)
-%   cfg.width      = 'width', or number of cycles, of the wavelet (default = 7)
+%   cfg.width      = 'width', or number of cycles, of the wavelet or base wavelet for 
+%                    superlets(default = 7)
 %   cfg.gwidth     = determines the length of the used wavelets in standard
 %                    deviations of the implicit Gaussian kernel and should
 %                    be choosen >= 3; (default = 3)
+%   cfg.superlets  = 'no', 'additive', 'multiplicative' (default = 'no')
+%                    If set to value other than 'no' performs superlet spectral estimation 
+%                    according to Moca et al. 2019 (https://doi.org/10.1101/583732), i.e.
```
schoffelen: Hmmm, so this is so far published only as a preprint? Do you know whether it is accepted as a peer-reviewed paper somewhere?

ft_freqanalysis.m
```diff
@@ -106,10 +106,18 @@
 %   cfg.foilim     = [begin end], frequency band of interest
 %   cfg.toi        = vector 1 x numtoi, the times on which the analysis
 %                    windows should be centered (in seconds)
-%   cfg.width      = 'width', or number of cycles, of the wavelet (default = 7)
+%   cfg.width      = 'width', or number of cycles, of the wavelet or base wavelet for 
+%                    superlets(default = 7)
 %   cfg.gwidth     = determines the length of the used wavelets in standard
 %                    deviations of the implicit Gaussian kernel and should
 %                    be choosen >= 3; (default = 3)
+%   cfg.superlets  = 'no', 'additive', 'multiplicative' (default = 'no')
```
schoffelen: I would like to suggest to organize the code (and options) slightly differently.
Rather than having cfg.superlets = 'no', 'additive', 'multiplicative', I'd recommend to have 

cfg.method = 'superlet';

and to deal with this as a separate method in the switch-boxes below.

In other words, keep the handling of the 'normal'  wavelet as is, and handle the superlets in a separate case.

For the option-handling, I'd make a substructure (cfg.superlet) that handles the method specific options. Admittedly, this is not how ft_freqanalysis deals with the options of the methods (this has grown historically), but this is how less ancient (or more often updated) fieldtrip functions deal with it. For instance, cfg.<somemethod>.someparameter in ft_sourceanalysis.

So, you would need cfg.superlet.sl_order, and cfg.superlet.combine (or so) to be defined.

ft_freqanalysis.m
```diff
@@ -531,7 +541,47 @@
...
+      else
+        % calculate number of wavelets and respective cycle width dependent on superlet order
+        % equivalent one-liners: 
+        %   multiplicative: cycles = arrayfun(@(order) arrayfun(@(wl_num) cfg.width*wl_num, 1:order), cfg.sl_order,'uni',0)
+        %   additive: cycles = arrayfun(@(order) arrayfun(@(wl_num) cfg.width+wl_num-1, 1:order), cfg.sl_order,'uni',0)
+        cycles = cell(length(cfg.foi),1);
+        for i_f = 1:length(cfg.foi)
+          frq_cyc = NaN(1,cfg.sl_order(i_f));
+          if strcmp(cfg.superlets, 'multiplicative')
+            for i_wl = 1:cfg.sl_order(i_f)
```
schoffelen: As mentioned above, I'd make this as a separate case

schoffelen:
Thanks for doing this by the way, and congratulations with your first PR!

robertoostenveld:
I agree to the suggestions made by @schoffelen. Having it as separate method facilitates documentation and future code maintenance.

ft_freqanalysis.m
DanBenHa: Yes, it's only a preprint so far. According to the senior author it's been submitted to some journal.
I'm not affiliated with the authors in any way, so I can't really say more about this.

DanBenHa:
Thanks for the suggestions! I actually first thought about implementing it as a separate method, but since it is basically a subtype of wavelet analysis, I did it as a submethod. I do see the advantages that a separate method has and will implement it like that.

DanBenHa:
Sorry for the delay.
I reorganised the code as suggested and added a test function.

DanBenHa:
Seems like the automatic checks have failed because of a timeout. Can someone help me here? Is it because I didn't rebase my new commits on the current master branch?

robertoostenveld:
Please don't bother with the automatic checks. It is something that we are plotting at Mathworks - using their cloud to run unit tests - but it turns out that the virtual machine time is limited to 20 minutes. A sizable (and variable) amount of that is already taken by transferring the repository from github to Mathworks, so there is very little actual compute time left, if any. The script that executes is this https://github.com/fieldtrip/fieldtrip/blob/master/test/test_matlab_ci.m

schoffelen:
Apparently I self assigned myself for a review. What's the status of ths Daniel? If you think it's good to go, could you add to the first lines of the test functions  a specification of the required resources for function execution (memory, walltime), and dependencies? This is needed for proper automatic deployment. 
See the functions in fieldtrip/test for an example

DanBenHa:
Hi Jan-Mathijs,

I'll add the code that you requested in the next few days.

Best wishes,
Daniel

schoffelen:
@DanBenHa I lost track of this recently, for which I apologize. It seems that the PR is close to be mergeable, but it would still require some action on your end.

DanBenHa:
@schoffelen Apologies from my side, too. I also kind of lost track of it and didn't realise that it's such a small piece that was still missing.
On my machine the test runs in less than 10 seconds and doesn't seem to require more than a few MB, but I guess the values I put in are safe for the automatic testing.

DanBenHa:
FIY, the method has been published: https://www.nature.com/articles/s41467-020-20539-9

The PR still says that changes are requested but I think they have been addressed.

schoffelen:
OK, op hoop van zegen, then. Thanks for your contribution and your patience!

PR #1106, opened 2019-05-07
Update crossvalidator.m

matsvanes:
Remove resetting of random seed: it is not optionable in this script, and the option is already available in ft_timelockstatistics.

### DIFF
--- a/external/dmlt/+dml/crossvalidator.m
+++ b/external/dmlt/+dml/crossvalidator.m
@@ -204,9 +204,7 @@ classdef crossvalidator
       end
       
       if strcmp(obj.type,'split'), obj.folds = 1; end
-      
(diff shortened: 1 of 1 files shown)


matsvanes:
Affirmative. If cfg.randomseed is specified in calling ft_timelockstatistics, the random number generator is fixed.

schoffelen:
Perhaps could you leave a comment in the code, specifying that this is a fieldtrip-version specific change (i.e. something like: the fixing of the random seed has been removed here in the fieldtrip version, the randomseed is dealt with higher up in the code hierarchy)

matsvanes:
Yep, that's what I did. I just _removed_ the feedback to the user that said the randomseed was fixed, because that's  not the case anymore.

schoffelen:
Oh, sorry, then I misread.

PR #1108, opened 2019-05-08
fixing bug when reading in curry data from current working directory

Spaak:
This came up at the Oxford workshop.


Spaak:
Hang on with the merge, something strange happened with some special characters (which I didn't touch). Any idea how best to fix that?

PR #1111, opened 2019-05-09
Update to eeglab2fieldtrip.m

arnodelorme:
I rewrote a lot of this function to support returning header, data and events as for the Fieldtrip import function. It is more robust and supports better the interface between EEGLAB and Fieldtrip.

Also the function in Fieldtrip was creating a conflict with the current function in EEGLAB.


robertoostenveld:
I find the output now rather confusing. The new header and event outputs correspond to the `ft_read_xxx` output, which according to https://youtu.be/7B4rDZYwQLM?t=4189 I consider low-level functions.  The previous output was a high-level fieldtrip data structure, like documented in [ft_datatype_raw](https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_datatype_raw.m) etc.  

Also inconsistent: In case of fieldtripbox=component, the output `header` is a fieldtrip [component](https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_datatype_comp.m) data structure. In case of fieldtripbox=preprocessing, the output `dat` is a 2D or 3-Darray which corresponds to ft_read_data.

Note that in variable naming and documentation I always (try to) make the distinction between `dat` as a numeric 2D or 3D array, and `data` as a structure that passes validation in ft_checkdata.

I think it makes more sense to change it like 
````
Output according to high-level fieldtrip data structures:

[data] = eeglab2fieldtrip(EEG, 'preprocessing')
[comp] = eeglab2fieldtrip(EEG, 'componentanalyis')
[timelock] = eeglab2fieldtrip(EEG, 'timelockanalysis')

and output according to low-level fieldtrip ft_read_xxx functions:

[hdr, dat, evt] = eeglab2fieldtrip(EEG, 'read')
```

arnodelorme:
Yes, we could do that. Ideally the function that import EEGLAB function in File-IO would use this function to avoid code duplication.

PR #1117, opened 2019-05-18
fixed problem writing the physical min and max to the header

robertoostenveld:
also write data in one-second blocks rather than one-sample blocks


robertoostenveld:
there were some unintended commits in here as well

PR #1120, opened 2019-05-22
improve data2bids using cfg.method and by supporting participants/scans

robertoostenveld:
This PR includes two large changes: 

The first is that cfg.bidsroot, cfg.sub, cfg.ses etc are added to allow dynamic construction of the cfg.outputfile and construction of the output directory with the participants.tsv and scans.tsv in the right place. 

The second is that I added cfg.method to explicitly distinguish between decorate/convert/copy, which seem the three major "things" that data2bids does. The old option cfg.keepnative introduces in #966 is now obsolete, but should keep working (using checkconfig). But please be aware....

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1,9 +1,7 @@
 function cfg = data2bids(cfg, varargin)
 
 % DATA2BIDS is a helper function to convert MEG, EEG, iEEG or anatomical MRI data to
-% the Brain Imaging Data Structure. This function starts from an existing data file on
-% disk or MATLAB memory, converts/writes the data file if needed and creates the
-% required sidecar files.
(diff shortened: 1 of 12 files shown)


robertoostenveld:
FYI, by using cfg.bidsroot, cfg.sub, cfg.ses, etcetera, it is now not required any more to use BASH and Python tools. There are some use cases (renaming a CTF dataset) that still require it, but quite a lot of cases can be handled from within MATLAB.

I will update http://www.fieldtriptoolbox.org/example/bids_eeg/ and possibly also http://www.fieldtriptoolbox.org/workshop/madrid2019/convert_bids/ to reflect the new functionality.

PR #1125, opened 2019-05-24
bugfix: fixed 'zscore' baseline correction in ft_freqbaseline

simonruch:
Fixes:
- 'zscore' was listed as valid baselinetype, but was not listed as valid option in the line of code that checked the validity of the input. => fixed
- The following code was used to perform the zscore-correction: "data=(data-meanVals)./stdVals;" but "stdVals" was not defined. => fixed
- 'vssum' is implemented as baselinetype but is not listed as valid type in the description => fixed


simonruch:
The nanflag in std was introduced in R2015a. Will change the code to use nanstd

schoffelen:
Thanks!

PR #1127, opened 2019-05-27
fixed ' from ’ to '

behinger:


schoffelen:
Oy, thanks. I realize that the change I made has never been communicated to the freesurfer folks. Do you think we should do this?

PR #1134, opened 2019-06-05
 enh: added support for postprocessing with the mars toolbox

StolkArjen:
@schoffelen, I've added the above support to ft_volumenormalise, transferring code from ft_volumesegment. 

One point of difference between the two implementation sites is that in ft_volumenormalise the original code for spmmethod = 'new' is:
spm_preproc_write8(params, zeros(6,4), [0 0], [0 1], 1, 1, bb, cfg.downsample);
.. whereas in ft_volumesegment it is:
spm_preproc_write8(p, [ones(6,2) zeros(6,2)], [0 0], [0 1], 1, 1, nan(2,3), nan);

The vector of ones and zeros determine the 'tc' variable in SPM but there's too little documentation on this variable to easily grasp its consequences. Any thoughts on whether this is important?

### DIFF
--- a/ft_volumenormalise.m
+++ b/ft_volumenormalise.m
@@ -245,7 +245,7 @@ if ~isfield(cfg, 'spmparams')
     %VF         = spm_vol([cfg.intermediatename '_anatomy' ext]);
     flags.nits = 0; % put number of non-linear iterations to zero
     params     = spm_normalise(VG, VF(1), [], [], [], flags);
-  elseif strcmp(cfg.spmversion, 'spm12') && strcmp(cfg.spmmethod, 'new')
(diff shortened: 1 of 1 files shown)


schoffelen:
At some point I think I understood this functionality (by reading the code: hint-hint :))
I believe that the 0/1 in the columns determine which type of volume is written. It's about writing the image in original space, normalised space, and the forward/inverse warp fields I think. Let me know if I need to look along.

StolkArjen:
Thanks, from the few annotations in those functions I think you're right in that they pertain to writing the native and warped images. Since we only need the deformation field in ft_volumenormalise, I have set those flags to zero.

PR #1138, opened 2019-06-08
enh: added support for point cloudtypes to ft_sourceplot

StolkArjen:
This PR builds into ft_sourceplot the ability to plot and scale electrode markers using the functional data, a functionality previously supported by ft_plot_cloud. @robertoostenveld, please have a look and adjust if necessary on the basis of what you had in mind. It's possible to additionally replace the call to plot3 by a call to ft_plot_sens but I'm afraid that would make it unnecessarily slow (and less robust). 

Screen Shot 2019-06-07 at 11 39 05 PM
left to right: cloudtype = 'cloud', cloudtype = 'point', cloudtype = 'surf' (see also supplemental fig of the ecog tutorial paper)

Please change the location of the ecog tutorial data in the test script as requested.


robertoostenveld:
looks good, thanks!

PR #1139, opened 2019-06-08
improve handling of dipole getting outside of source compartment

robertoostenveld:
@britta-wstn some improvements were needed to better deal with the dipole jumping outside of the head


inverse/dipole_fit.m
```diff
@@ -291,6 +295,11 @@
   lf = lf * ori;
 end
 
+if ~isinside
```
robertoostenveld: @arnodelorme I think that this change in https://github.com/fieldtrip/fieldtrip/pull/1139 may be relevant for you as well. Please check the whole code in the error function.

PR #1140, opened 2019-06-11
FIX #3475 - improve plotting of electrode labels for linear arrays

robertoostenveld:


robertoostenveld:
This addresses http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3475 as reported by @StolkArjen

StolkArjen:
Appreciated!

StolkArjen:
This fix works well generally. It's only in instances when the elec structure contains two or less electrodes that the delaunay projection fails. Such instances occur when plotting subsets of electrodes as when plotting statistically significant channels or a depth probe on which many electrodes were discarded. Would it be possible if we expanded line 300 to read
if ~any(isnan(pos(:))) && size(pos,1) >= 3 
This would force the function into the sphere fitting part for arrays smaller than 3 (not sure if this okay).

StolkArjen:
Hang on, that's weird. Trying to replicate the issue, it actually works.

elec.elecpos = [1 1 1];
elec.chanpos = elec.elecpos;
elec.label = {'1'};
elec.unit = 'mm';
ft_plot_sens(elec,'label','on')

I'll see if I can find the erroneous situation again...

robertoostenveld:
I think I really fixed all cases you mentioned, also single channel and multiple channels that are all on the same place (although I can still think of fringe situations where the current implementation might not work). 

There are two things to be aware of: first is the `orientation` of the channel, which is used for offsetting the label but also for plotting circles and squares. The second is the `offset`, which determines how far the label is displaced along the orientation direction.  

For the orientation it now only reverts to the sphere fitting if there is a nan in the channel positions (don't know when that actually would happen). In case either triangulation or speherefitting fails, it now orients the channel along the z-axis (see line 328 in ft_plot_sens). The offset is now also more robust, see line 426 and the lines just before.

robertoostenveld:
I have been thinking about plotting cylinders (for contacts along a sEEG shaft), for which the current code would not work. In that case you would want the cylinder orientation to be along the shaft, but the label displacement perpendicular to the shaft. I don't think that this could be done fully automatic; to realize it we would probably have to add an input argument. Cylinders would not only be applicable to sEEG, but (short) cylinders could also be used for scalp EEG.

robertoostenveld:
And now that I am at it: I realize that the current implementation is likely to fail for curved shafts, or corkscrew shafts, or linear shafts with some noise on the positions. To work automatically for those,  some non-zero tolerance should be used around line  38 in `plotting/private/projecttri.m`, i.e. when nearly linear it should not try to triangulate.

StolkArjen:
It was really hard to replicate, but I found such a situation. The case has two electrodes with positions that someone produce the following error scenario. I am saying 'somehow' because I cannot replicate it by arbitrarily choosing two electrode positions.

Screen Shot 2019-06-12 at 11 48 34 PM

Adding the "... && size(pos,1)>=3" statement solved the issue. I can't oversee whether that interferes with what you have in mind more generally. But I'll briefly create a PR before I jump in bed in case you'd like to make use of it.

robertoostenveld:
can you check why the rank detection earlier on in the code fails to deal with this?

StolkArjen:
ft_plot_sens(tmp, 'label', 'on') 

(copying the content of elecpos/chanpos does not replicate the issue, hence I'm uploading the file). 

[tmp.mat.zip](https://github.com/fieldtrip/fieldtrip/files/3284580/tmp.mat.zip)

As for those depth shafts, I have code somewhere to draw Behnke-Fried electrodes (might have emailed you before).

robertoostenveld:
I don't see any problem with tmp.

Screen Shot 2019-06-13 at 09 16 33

StolkArjen:
Hmm weird. Running it on Matlab 2017b, mac OS

>> load('tmp.mat')
>> ft_plot_sens(tmp)
Error using delaunay
At least three input points must be provided to define a triangular mesh.

Error in projecttri (line 83)
    tri = delaunay(prj(:,1), prj(:,2));

Error in ft_plot_sens (line 303)
    tri = projecttri(pos, 'delaunay');

StolkArjen:
Will debug tomorrow - good night

robertoostenveld:
if there are only two vertices, the rank should be 1.

StolkArjen:
Somehow in my case, rank.m's s(2) is larger than the tolerance (6.7 e-15 vs. 5.3 e-15), making it conclude the rank is 2. It's probably got something to do with inaccuracy in rounding or representation. If I manually copy the content of tmp in projecttri over to a new variable (A), the rank of that new variable is 1. Indeed, this new variable A is now centered around zero between the two data points (after demeaning at lines 34-36 in rank.m) whereas the original tmp variable is not.

sprintf( '%0.20f', tmp(1,1)) = '-0.32309100255496758791'
sprintf( '%0.20f', tmp(2,1)) = '0.32309100255496048248'

sprintf( '%0.20f', A(1,1)) = '-0.32309999999999999831'
sprintf( '%0.20f', A(2,1 ) = '0.32309999999999999831'

Why this does not happen on your computer is beyond me

robertoostenveld:
No idea, but I think it demonstrates that it is not robust yet. Especially with curved shafts or electrodes with some noise in their position. For those your suggested fix also does not work.  

I think that point, line, 2d plane, 3d surface and ideally also 3d scattered (e.g. the case with sEEG electrodes along multiple shafts, in a random order) should be determined in another way. Or perhaps have the user pass the geometry as option...

StolkArjen:
"(e.g. the case with sEEG electrodes along multiple shafts, in a random order)"
Yes, in the intracranial field, group electrode plots are going to be more common and there's no clear structure in the location of electrodes originating from different subjects. Could we not use the original implementation (plotting labels where the electrodes are) for these cases, assuming we can identify those use cases?

robertoostenveld:
Yes, I think that would be best. Perhaps

    ft_plot_sens(elec, 'labelpos', xxx, 'labeldist', yyy)

where xxx and yyy can be auto (as is now the case) but also explicitly specified to deal with the other cases. This would encourage us also to distinguish between the orientation of the channel (which was initially only conceived for a MEG sensor), and the orientation along which the label gets displaced. The default would remain that the label displacement is along the orientation of the sensor.

StolkArjen:
My quick two cents would be that the function works robustly as before, while fancier label positioning would be more appropriate for certain use cases in which the user explicitly requests it.

PR #1142, opened 2019-06-13
fix: skip delaunay for elec arrays smaller than 3

StolkArjen:
As discussed in PR https://github.com/fieldtrip/fieldtrip/pull/1140

### DIFF
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -298,7 +298,7 @@ else
 end % if istrue(individual)
 
 if isempty(ori) && ~isempty(pos)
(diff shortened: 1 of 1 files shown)


plotting/ft_plot_sens.m
```diff
@@ -298,7 +298,7 @@
 end % if istrue(individual)
 
 if isempty(ori) && ~isempty(pos)
-  if ~any(isnan(pos(:)))
+  if ~any(isnan(pos(:))) && size(pos,1)>=3
```
robertoostenveld: this should not be needed, projecttri deals with this.

PR #1147, opened 2019-06-19
implement support for the XDF format, similar as for variable sampling-rate EDF datasets

robertoostenveld:


robertoostenveld:
@schoffelen this should result in an error for `test_issue789.m` in the dashboard batch this evening, since the low-level mex file is not yet compiled for linux (on which the batch runs), only for maci64.

That is in itself an interesting test case: the changes should not end up in the release until that `test_issue789.m` passes.

PR #1149, opened 2019-06-20
ENH - flip the dipole moment and orientation with +1/-1 

robertoostenveld:
to make it more consistent with the overall distribution.

This cane up in discussions with @arnodelorme for https://sccn.ucsd.edu/wiki/A08:_DIPFIT#Advanced_source_reconstruction_using_DIPFIT.2FFieldtrip

I know that @schoffelen has been doing this in other contexts, not only for 3D but also cortical sheets. Jan-Mathijs, could you check whether this makes sense, or whether something similar is already in the code somewhere or in the documentation somewhere?


ft_sourcedescriptives.m
```diff
@@ -679,7 +701,20 @@
     end
     ft_progress('close');
   end
-
+  
+  if flipori
```
robertoostenveld: this is where the actual action is happening.

Issue #1128, opened 2019-05-29
Loading discontinuous Neuralynx data


ludovicbellier:
Hi there,

While I recently tried to align Neuralynx data with data simultaneously recorded using a Nihon Kohden system, I stumbled upon the discontinuous data loading error that has previously been reported (https://github.com/fieldtrip/fieldtrip/pull/58) and documented (http://www.fieldtriptoolbox.org/faq/discontinuous_neuralynx/).

After further investigation, I found out that the Neuralynx system is saving data as a series of 512-sample batches and that when user hits pause or stop buttons, the recording stops at sample i from the current 512-sample batch, and the remaining samples (i+1 to 512) are a copy of samples i+1 to 512 of the previous batch. These repeated samples are marked as invalid somewhere in the ncs structure.
Besides, in the dataset I explored, I found that all ncs files had the same number of breaks (i.e., of 512-sample batches containing invalid samples), at the same timing, whatever their sampling rate.

Here's a very simple code snippet that loads Neuralynx data as epochs:
```
% Load Neuralynx data as continuous
cfg = [];
cfg.dataset = 'IRxx/macro/';
dataNLX = ft_preprocessing(cfg);

% Get epoch boundaries from low-level Fieldtrip functions
ncs = read_neuralynx_ncs('IRxx/macro/elec_y.ncs');
idx = find(ncs.NumValidSamp<512);
nValSamp = ncs.NumValidSamp(idx);
nTrials = length(idx);
trl = zeros(nTrials, 2);
trl(:, 2) = ((idx-1)*512 + nValSamp)';
trl(2:end, 1) = (idx(1:end-1)*512+1)';

% Apply these boundaries to epoch data
cfg = [];
cfg.trl = trl;
dataNLX = ft_redefinetrial(cfg, dataNLX);
```

Then, each epoch can successfully be aligned with the Nihon Kohden data using cross-correlation (see the following figure; top row shows raw cross-correlation coefficients for each of the four epochs; CS for Clinical System/Nihon Kohden; SU for Single-Unit/Neuralynx system).

![CS-SUalignment](https://user-images.githubusercontent.com/28165157/58596296-68cbcc00-8228-11e9-89c7-3c135e77f519.png)

[...]

PR #1155, opened 2019-06-28
load discontinuous Neuralynx data as epochs

ludovicbellier:
Hi there,

As discussed on issue #1128, here's a way to automatically load discontinuous Neuralynx data as epochs (and to avoid getting repeated, artifactual samples in our data due to the pause/stop built-in feature).

Best,
Ludovic

### DIFF
--- a/fileio/private/read_neuralynx_ds.m
+++ b/fileio/private/read_neuralynx_ds.m
@@ -157,17 +157,42 @@ if needhdr
     end
   end % if ftype_ncs
 
+  % take into account the pause/stop feature + epoch data if discontinuous recording
+  ncs = read_neuralynx_ncs(fname{1});
+  idxPauses = find(ncs.NumValidSamp < recordsize);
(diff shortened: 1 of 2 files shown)


robertoostenveld:
One solution would be an implementation similar to https://github.com/fieldtrip/fieldtrip/blob/master/trialfun/ft_trialfun_brainvision_segmented.m
which is also a file format that can have segments of different length. 

But I believe best would be to use the existing ft_read_event output and add the segments as "events" (with the appropriate event.duration). That should allow you to do something similar to

```
cfg = [];
cfg.dataset = xxx
cfg.trialfun = 'ft_trialfun_general';
cfg.trialdef.eventtype = 'segment';
cfg = ft_definetrial(cfg)  % this adds cfg.trl, only for the segments

data = ft_preprocessing(cfg); % this reads the segments
```

This is also how "trial" events are being read from some files, such as the CTF format. E.g. look at Subject01.ds, which has 3-second (900 sample) segments.

PR #1156, opened 2019-07-01
data written to edf was incorrect 

robertoostenveld:
When reading from EDF with a blocksize of 1 sample, and writing it back to EDF with a blocksize of 1 second, I realized that the data is transposed. This would not have mattered with the previous implementation (one sample per block), but now resulted in corrupt data.


robertoostenveld:
I have fixed it in the code. Since EDF remains an instable format in our code, I will add a test script that does a round-trip test, i.e., read-weite-read and check before and after.

Issue #1159, opened 2019-07-02
Error in ft_senstyp for datatype "comp"


mcvinding:
Error in _ft_senstyp_ when reading component data, e.g. for plotting ICA components with _ft_topoplotIC_ or _ft_databrowser_. The error appears to be in reading and processing the "balance" field in the grad structure related to unmixing the components.

Error message:

```
> Error using ft_datatype_sens (line 283)
> unexpected number of coils (510) contributing to channel MEG0111 (1)
> 
> Error in undobalancing (line 58)
> sens = ft_datatype_sens(sens);
> 
> Error in ft_senstype (line 283)
>       sens = undobalancing(sens);
```

The error appeared in both my own data recorded with a Neuromag MEG and the CTF tutorial data from the FieldTrip tutorials. I have attached a grad structure from the Neuromag data.

[grad.zip](https://github.com/fieldtrip/fieldtrip/files/3350259/grad.zip)

schoffelen:
This is probably the same issue as issue #980

robertoostenveld:
I renamed and copied the test file to our file server

```
mac011> mv grad.mat issue1159.mat
mac011> scp issue1159.mat mentat001:/home/common/matlab/fieldtrip/data/test
```

When I try a direct replication, it now works for me

```
>> load issue1159.mat
>> ft_senstype(grad)
ans =
    'neuromag306'
```

But I think this morning you showed that after ft_componentanalysis it did not have teh `type` field (which I suppose you added by hand). The following is what works for me to reproduce the error that you report as well:


```
>> ft_senstype(rmfield(grad, 'type'))
undoing the comp balancing for the gradiometer definition
Error using ft_datatype_sens (line 283)
unexpected number of coils (510) contributing to channel MEG0111 (1)
Error in undobalancing (line 58)
sens = ft_datatype_sens(sens);
Error in ft_senstype (line 283)
      sens = undobalancing(sens); 
```

robertoostenveld:
I made a test script  and added it to a branch on my own https://github.com/robertoostenveld/fieldtrip copy of the repository

mac011> git commit test/test_issue1159.m 
[issue1159-undobalancing 98fec29af] ENH - added test script to reproduce the problem
 1 file changed, 20 insertions(+)
 create mode 100644 test/test_issue1159.m

mac011> git push --set-upstream origin issue1159-undobalancing
Counting objects: 78, done.
Delta compression using up to 4 threads.
[...]

Issue #980, opened 2019-03-01
ft_rejectcomponent error


dnacombo:
Just after computing components, I try to reject components and get the following error.
```
cfgica              = [];
cfgica.method       = 'runica';
cfgica.numcomponent = 50;
comp             = ft_componentanalysis(cfgica, data);

cfg                           = [];
cfg.component                 = [1 10];
data_corrected                = ft_rejectcomponent(cfg, comp, data );

undoing the comp balancing for the gradiometer definition
Error using ft_datatype_sens (line 279)
unexpected number of coils (510) contributing to channel MEG0111 (1)
Error in undobalancing (line 58)
sens = ft_datatype_sens(sens);
Error in ft_senstype (line 283)
      sens = undobalancing(sens);
Error in ft_datatype_sens (line 151)
ismeg = ft_senstype(sens, 'meg');
Error in ft_datatype_raw (line 112)
      data.grad = ft_datatype_sens(data.grad);
Error in ft_datatype_comp (line 110)
      raw = ft_datatype_raw(raw, 'version', rawversion, 'hassampleinfo', hassampleinfo, 'hastrialinfo', hastrialinfo);
Error in ft_checkdata (line 266)
  data = ft_datatype_comp(data, 'hassampleinfo', hassampleinfo);
Error in ft_rejectcomponent (line 90)
  comp    = ft_checkdata(comp, 'datatype', 'comp'); 

```
Nota:
```
data.grad
ans = 
  struct with fields:

     balance: [1×1 struct]
     chanori: [306×3 double]
     chanpos: [306×3 double]
    chantype: {306×1 cell}
    chanunit: {306×1 cell}
     coilori: [510×3 double]
     coilpos: [510×3 double]
    coordsys: 'neuromag'
       label: {306×1 cell}
         tra: [306×510 double]
        type: 'neuromag306'
        unit: 'cm'

comp.grad
ans = 
  struct with fields:

       balance: [1×1 struct]
       chanori: [50×3 double]
       chanpos: [50×3 double]
       coilori: [510×3 double]
       coilpos: [510×3 double]
      coordsys: 'neuromag'
         label: {50×1 cell}
           tra: [50×510 double]
          unit: 'cm'
    chanposold: [306×3 double]
      labelold: {306×1 cell}
    chanoriold: [306×3 double]
```

schoffelen:
[...]

PR #1160, opened 2019-07-02
resolve error in undobalancing as reported in #1159 and #980

robertoostenveld:
The issue reported in #1159 is quite a tricky issue, since chanunits are not kept in ft_componentanalysis. Consequently undobalancing results in the units as unknown, and ft_checkdata_sens failed to deal with the update of the scaling.


robertoostenveld:
the commit contains multiple times the same chance, since I executed `bin/synchronize-private.sh`. The main change is that the scaling of the sensor array is not updated any more in case either `amplitude` or `distance` are unknown.

PR #1164, opened 2019-07-03
do not use ~ to ignore function outputs, see #831

robertoostenveld:


robertoostenveld:
this does not address all occurrences mentioned in #831 yet, but merging this will cause `bin/synchronize-private.sh` to be executed, which will at least propagate the changes to all copies of the same files. After that we'll have to look again.

PR #1171, opened 2019-07-19
FIX: use a check for private function that is compatible with GNU Octave

nno:
In the directory 'compat/octave/', the isfile and isfolder functions use the expression:

cellfun(@isempty, regexp(which(mfilename, '-all'), fullfile('private', mfilename))))

which in does not work in Octave, as the regex function returns an empty array (not a cell).

The attached code seems to fix this issue. 

Note that I found the isfile.m and isfolder.m in other compat/* subdirectories, so for consistency it may be good to use the same expression there as well. (I would be happy to change the PR and change those files as well).


nno:
An example where this breaks code:

https://travis-ci.org/nno/CoSMoMVPA/builds/561038055

failure: cellfun: C must be a cell array
  isfolder:12 (/home/travis/build/nno/fieldtrip/compat/octave/isfolder.m)
  ft_version:64 (/home/travis/build/nno/fieldtrip/utilities/ft_version.m)
  ft_preamble_init:50 (/home/travis/build/nno/fieldtrip/utilities/private/ft_preamble_init.m)
  ft_prepare_neighbours:1 (/home/travis/build/nno/fieldtrip/ft_prepare_neighbours.m)
  ft_preamble:83 (/home/travis/build/nno/fieldtrip/utilities/ft_preamble.m)
  ft_prepare_neighbours:71 (/home/travis/build/nno/fieldtrip/ft_prepare_neighbours.m)
  test_meeg_chan_neighbors>test_meeg_neighbors:100 (/home/travis/build/nno/CoSMoMVPA/tests/test_meeg_chan_neighbors.m)

schoffelen:
It seems indeed that the addition of the 'match' argument in regexp does not change behavior in case it's called from matlab. It would be great if you could also update the others indeed.

nno:
I marked the function as Work In Progress, because after applying the changes to other functions it seems it may lead to infinite recursion when using older functions (e.g. ismatrix.m in matlablt2010b). Not sure if that is because of conflicting names with builtin functions

@schoffelen: could you please advice on how to check / test that changes in this PR do not break existing functionality?

schoffelen:
Not sure. What's the recursion that one ends up in? Wouldn't this also be a problem for the octave equivalents?

schoffelen:
if these issues occur only in the functions that support the extremely (i.e. >5 years) old matlab versions, we could also consider not to update those...

nno:
@schoffelen: one case may be in matlab2010b/isrow.m, which calls ``fileparts``. If fileparts calls ``isrow`` then you get infinite recursion. But this may only be an issue when using more recent versions of Matlab to test older versions, as in 2010b there cannot have been a fileparts implementation that uses isrow. 
As far as I could tell, other functions seem to work fine. Is there a way to test this for different Matlab versions? Do you think more testing is required to be sufficiently confident this PR won't break existing functionality?

nno:
ping @schoffelen: any ideas with respect to my questions above?

robertoostenveld:
@schoffelen is on holiday, but I am back - so I suggest I take over.

Looking at the issue, it seems that it is not due to `regexp` itself, since that also returns a array on MATLAB 

```
>> regexp('I would like my reg to ex, eh, ax', 'x')
ans =
    25    33
>> regexp('I would like my reg to ex, eh, ax', 'y')
ans =
    15
>> regexp('I would like my reg to ex, eh, ax', 'z')
ans =
     []
```  

This is the same output as I see with Octave 4.0.1 (sorry, latest version installed here).

The reason for it being a cell is that the input (from `which`) is a cell, and I see that 

```
octave:3> a = which('ls', '-all')
a = /opt/octave/4.0.1/share/octave/4.0.1/m/miscellaneous/ls.m
octave:4> a = which('ls')
a = /opt/octave/4.0.1/share/octave/4.0.1/m/miscellaneous/ls.m
```
returns a string in both cases, whereas on MATLAB it returns a cell in case of `-all`.

robertoostenveld:
https://octave.sourceforge.io/octave/function/which.html confirms that the `-all` option is ignored. Should that be reported somewhere for octave? Even if the option does not apply, formatting the output as cell array would make it more compatible.

robertoostenveld:
> if these issues occur only in the functions that support the extremely (i.e. >5 years) old matlab versions, we could also consider not to update those...

This code was added upon explicit request by @gllmflndn to ensure that SPM (which calls FieldTrip) works on even older MATLAB versions

robertoostenveld:
I believe the recursion only happens in case you run `isrow` while the present working directory is `fieldtrip/compat/matlablt2010b`. In that case it tries to remove it from the path, but cannot solve the issue of it being in the pwd. So that does not need to be addressed to solve the specific issue.

robertoostenveld:
This code serves to check that people have the wrong directory on the path. In this case MATLAB users could also have compat/octave on the path, and Octave users could have compat/matlabltXXXX on the path. 

 I prefer not to mess with the regexp code, that is difficult already as it is - the first part translates more or less into "startsWith", but cannot use "startsWith", the second translates into "contains", and again cannot use "contains" itself - i.e. it looks for an alternative that starts with the matlabroot, but does not contain the specific function in a private directory.

Instead, could you change it into the following (i.e. add a few lines prior to the if-statement) and test it?

```
alternatives = which(mfilename);
if ~iscell(alternatives)
  % this is needed for octave, see https://github.com/fieldtrip/fieldtrip/pull/1171
  alternatives = {alternatives};
end

if exist(mfilename, 'builtin') || any(strncmp(alternatives, matlabroot, length(matlabroot)) & cellfun(@isempty, regexp(alternatives, fullfile('private', mfilename))))
  % remove this directory from the path
  ...
```

The change should be in all compat/matlabltXXXX functions and in all compat/octave functions. I also see that `compat/octave/contains.m` still needs to be made robust to deal with the [original issue](https://github.com/fieldtrip/fieldtrip/issues/899).

robertoostenveld:
The easy solution turned out to be to use `strfind` rather than `regexp`, combined with the section to ensure that `alternatives` is a cell-array.

PR #1177, opened 2019-07-31
added standard deviation method to ft_sourceparcellate

ualsbombe:


ualsbombe:
To ft_sourceparcellate that is

robertoostenveld:
hmm, you don't want this any more?

robertoostenveld:
ah, I now see #1191, which replaces this.

PR #1180, opened 2019-08-03
fix: cloud colorscale

StolkArjen:
This fixes a coloring issue in ft_plot_cloud and ft_sourceplot. Now the entire color range of the colorbar is used.


StolkArjen:
Trying another PR as this one's "checks fail"

PR #1181, opened 2019-08-06
[FIX] typo stcmp -> strcmp

gdemarchi:
Fixed a small typo ('stcmp' instead of 'strcmp’) that prevented data2bids to run when providing a MATLAB ‘table' instead of a Nx3 matrix as cfg.events.trl.


robertoostenveld:
thanks!

PR #1192, opened 2019-08-15
New obj reader

neurofractal:
Following on from discussion on PR
https://github.com/fieldtrip/fieldtrip/commit/d073bb2de8c2aa61a37eeb97b99c177e15985911

This PR implements a new .obj reader from “Boffin Blogger”, which is
much faster than the previous read_wobj function. The function also now
ignores ft_struct2double for .obj files with colour information, in
order to preserve the int8 format. Works well for data acquired from a
structure sensor (@Macquarie-MEG-Research).

### DIFF
--- a/external/wavefront/read_obj_new.m
+++ b/external/wavefront/read_obj_new.m
@@ -1,0 +1,72 @@
+function [vertex, faces, texture, textureIdx] = read_obj_new(filename)
+
+% Faster .obj reader for headshape information acquired using a structure
+% sensor
+
+% from http://boffinblogger.blogspot.com/2015/05/faster-obj-file-reading-in-matlab.html
+% modified output structure and added scan for texture and textureIdx to handle vt
+% Modifications by Robert Oostenveld and Robert Seymour
+
+fid = fopen(filename);
+if fid<0
+    error(['Cannot open ' filename '.']);
+end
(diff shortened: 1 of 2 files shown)


schoffelen:
Thanks Rob, don't worry about the CI checks having failed. I will merge soon. What's the rationale for keeping the int8 class for the .obj ?

neurofractal:
Hi JM... it seems that plotting the color of the mesh (using ft_plot_mesh) requires the uint8 class (at least for MATLAB 2016b). I didn't want to mess around too much with the color plotting in that function. 

Happy to change if necessary...

schoffelen:
Ah, I see. But I don't see why this would only be an issue for the wobj's...
In any case, rgb plotting indeed either is a uint8 thingy (with values [0, ..., 255]), but can also be double. In that case the values should be in the range of [0 1]. In other words, an alternative solution could be to handle the color values in the 'wobj' case, by checking whether they are integers in the range of [0 255], in which case they should be divided by 255, so that the 'wobj' specific handling of ft_struct2dboule at the end of the function is not needed anymore. 
May I suggest this course of action?

neurofractal:
Please see https://github.com/fieldtrip/fieldtrip/pull/1192/commits/86bb261f54e16e2de29d13847615ac4b4ad92bbb

PR #1195, opened 2019-08-18
fix: fixed a scaling issue that occurred with padding

StolkArjen:
The issue occurred only when padding was applied. Now the PSD extracted from the padded data is appropriately scaled (using the endnsample variable)


StolkArjen:
For comparison, the dashed lines indicate the original and fractal PSD (grey and black) extracted using the original IRASA technique. The solid lines indicate their counterparts extracted using the FT-based implementation of IRASA.

Screen Shot 2019-08-18 at 2 39 55 PM

PR #1199, opened 2019-08-24
make sure LF backprojection not used with SAM

AdoNunes:
SAM methods, except the spinning, do not support back-projected LFs and wrong weights will be estimated. 

This commit makes sure that LFs are not rank deficient, if so it will recalculate them.


schoffelen:
Thanks. Before merging, could you please update the number of spaces-per-tab indent from 4 to 2, as per the convention used in the FieldTrip code base?

schoffelen:
Top, thanks.

PR #1205, opened 2019-08-27
FIX #1203 - improve electrode and coordsystem handling for BIDS

### LINKED ISSUE #1203: data2bids does not deal with EEG electrodes properly
It should be able to specify electrodes for EEG and iEEG as cfg.electrodes (which would be "tsv"-style already), as cfg.elec, or in the input data. Furthermore, if electrodes are specified, coordsystem should also be specified.

robertoostenveld:
this also includes support for video data (along the lines of the recently added EMG support).


robertoostenveld:
als dat maar goed komt...

PR #1208, opened 2019-08-29
fixed bugs in oxy4 support (fix #1309) and fix - restricting endsample to max. available data size 

### LINKED ISSUE #1309: ft_datatype_sens renames NIRS opto fields, but other code expects old fieldnames
In line 166 ff, ft_dataytpe_sens renames some fields in data.opto, especially opto.transceiver to opto.transmits. Code in other functions, however, like ft_nirs_transform_ODs, do access opto.transceiver.

Proposed fix: 
As discussed a couple of months ago, change the fieldnames in all functions.

Horschig:


robertoostenveld:
Hi Jorn,

Assuming that read_artinis_oxy4.m is called by ft_read_data.m, which is called by ft_preprocessing, it would result in a data structure in which the `trialinfo` is inconsistent with the actual data (since the high level functions don't know about the missing samples).

Instead of this solution, I propose to pad the data with nan. We also use that elsewhere (e.g. ft_fetch_data) and `ft_artifact_nan` combined with `ft_rejectartifact` can be used to detect and remove the nans.

Horschig:
good point, I implemented this in commit a41fbf97f37852ddf052ecbe54ebbd68d97386e3

Horschig:
Hey Robert, can you merge this? I got some complaints that conversion does not work, but I could not reproduce the issues on my computer. I just now realize that this merge has never been completed, which explains why others had issues but I didn't.

robertoostenveld:
Hi Jorn, I have merged this in the master branch. The test-and-release process is explained on http://www.fieldtriptoolbox.org/development/dashboard, hopefully it will be in the release soon as well, but that takes at least a day

PR #1209, opened 2019-08-29
Update xdf2fieldtrip so that it also return events

varghese-peter:
The function now reads the event data and puts it in a fieldtrip compatible event structure

### DIFF
--- a/xdf2fieldtrip.m
+++ b/xdf2fieldtrip.m
@@ -49,7 +49,6 @@ ft_hastoolbox('xdf', 1);
 
 % read all streams
 streams = load_xdf(filename);
-
 iscontinuous = false(size(streams));
 % figure out which streams contain continuous/regular and discrete/irregular data
 for i=1:numel(streams)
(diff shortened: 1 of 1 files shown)


xdf2fieldtrip.m
```diff
@@ -116,6 +152,7 @@
   data{i}.label = hdr.label;
   data{i}.time = {streams{i}.time_stamps};
   data{i}.trial = {streams{i}.time_series};
+  data{i}.event = event;
```
robertoostenveld: This is not where other functions would the events. In case the event structure is passed from the workspace into other FieldTrip functions, if is specified as `cfg.event`. But it is never part of the data. See https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_datatype_raw.m for the specification of this data structure.

Better would be to extend the API that is exposed to the user from 
```
data = xdf2fieldtrip(filename, ...)
```
into 
```
[data, event] = xdf2fieldtrip(filename, ...)
```

xdf2fieldtrip.m
```diff
@@ -65,6 +64,43 @@
...
+for i=1:length(streams)
+    if (strcmp(streams{i}.info.type,'EEG'))
+        EEG_t1= str2double(streams{i}.info.first_timestamp);
+        EEG_srate=streams{i}.info.effective_srate;
+        EEG_info=[EEG_t1 EEG_srate];
+    end
+end
+
+if isempty(EEG_info)
+  ft_error('no EEG streams were selected');
```
robertoostenveld: do you want an error in case the data only contains non-EEG data, e.g. only eye tracking?

xdf2fieldtrip.m
```diff
@@ -65,6 +64,43 @@
...
+        try
+            events = struct('sample', [], 'offset', [], 'duration', num2cell(ones(1, length(streams{i}.time_stamps))),...
+                'type', cellstr(repmat({'Marker'}, 1, length(streams{i}.time_stamps))), 'value', ' ', 'timestamp', []); 
+            for k=1:length(streams{i}.time_stamps)
+                if iscell(streams{i}.time_series)
+                    events(k).value = streams{i}.time_series{k};
+                else
+                    events(k).value = num2str(streams{i}.time_series(k));
+                end
+                events(k).sample = round((streams{i}.time_stamps(k)- EEG_t1)*EEG_srate);
```
robertoostenveld: event samples are expressed relative to the EEG sampling rate. What if there are two EEG streams with different sampling rates? Or what if another data stream (e.g. audio or high speed eyetracker) has a higher sampling rate?  The resampling of all streams further down uses the highest sampling rate, not per see the EEG sampling rate. 

samples as specified in the event structure should match that of the resampled+appended data.

schoffelen:
Hi @varghese-peter just pinging you to check in. What is the status of this? It seems that, although still some work would be needed, not much more needs to be done here. I think that most importantly this PR requires some test data, and some principled solution to deal with data that contains time series which are sampled at different frequencies. The latter cannot be represented in Fieldtrip in a single data structure.

schoffelen:
It looks as if this is a stale PR, @varghese-peter, since you are the main stakeholder here, is it OK to close this one?

schoffelen:
xdf2fieldtrip has been recently changed, it is not clear whether those changes pertain to the issue that this PR attempted to fix, but since this one has not seen any activity for a long time, I am closing it, since it is unrealistic that it will be followed up on.

robertoostenveld:
I looked into this, and noticed that @varghese-peter did make some changes and commented in the commit. Those commit comments do not directly appear here in the issue discussion, hence they went unnoticed for me. He wrote:

> I have incorporated the changes suggested by Robert. Please review.
> Please find a sample dataset here:https://cloudstor.aarnet.edu.au/plus/s/5E9qcAuXC6sZ32K 

The suggested change to "have xdf2fieldtrip return events" does make sense to me, and with the attached example (which I downloaded already) that can be implemented and/or tested.  Let me follow up on this one and get it implemented...

schoffelen:
OK, I was just poking around a bit, and tried to play marie condo...

schoffelen:
no rush here I'd say.

PR #1212, opened 2019-09-03
Broaden permitted channel names

jens-k:
..for NSX datasets.

Non a-z channel names lead to unexpected behavior. For example, a channel name 'A1.elec021 lfp' would lead to two cells per channel ({elec021} and {lfp}), which means that a) port information is discarded and b) two cells leads to an error when creating the chaninfo table.

Code works for my example but I am not sure why the regexp was so strict or why a regular expression was necessary in the first place. So please double check.


robertoostenveld:
looks fine to me, I don't have data to test this with.

PR #1218, opened 2019-09-18
add support for BIDS data represented in tsv/json files  

robertoostenveld:
This is used for `physio.tsv` and `stim.tsv` in the current BIDS specification, but I am also using it for eyetracker and motion capture data. See for example <http://www.fieldtriptoolbox.org/example/bids_eyetracker/>

If the data is accompanied by an `events.tsv` file, that can also be read using FT_READ_EVENT.

### DIFF
--- a/fileio/ft_read_event.m
+++ b/fileio/ft_read_event.m
@@ -1,9 +1,8 @@
 function [event] = ft_read_event(filename, varargin)
 
-% FT_READ_EVENT reads all events from an EEG/MEG dataset and returns
-% them in a well defined structure. It is a wrapper around different
-% EEG/MEG file importers, directly supported formats are CTF, Neuromag,
-% EEP, BrainVision, Neuroscan, Neuralynx and Nervus/Nicolet.
+% FT_READ_EVENT reads all events from an EEG, MEG or other time series dataset and
+% returns them in a common data-independent structure. The supported formats are
+% listed in the accompanying FT_READ_HEADER function.
 %
 % Use as
 %   [event] = ft_read_event(filename, ...)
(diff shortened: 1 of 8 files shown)


robertoostenveld:
hmm, apparently something else snuck in here as well: this PR also improves the reading of SMI eyetracker data. This was implemented initially by @dieloz but apparently never completed, i.e. the FT_READ_DATA did not read the actual time courses. I have checked the implementation, extended and improved it where needed. I have only tested it with a limited number of files from the DCCN MR labs.

fileio/ft_read_event.m
```diff
@@ -1051,25 +1049,10 @@
...
-    % the start of the data acquisition
-    for i=1:length(timestamp)
-      event(end+1).type       = 'Trigger';
-      event(end  ).sample     = (timestamp(i)-hdr.FirstTimeStamp)/hdr.TimeStampPerSample + 1;
-      event(end  ).timestamp  = timestamp(i);
-      event(end  ).value      = value(i);
-      event(end  ).duration   = 1;
-      event(end  ).offset     = 0;
+    chanindx = find(strcmp(hdr.label, 'Trigger'));
+    event = read_trigger(filename, 'header', hdr, 'dataformat', dataformat, 'begsample', flt_minsample, 'endsample', flt_maxsample, 'chanindx', chanindx, 'detectflank', detectflank, 'trigshift', trigshift);
```
robertoostenveld: rather than using custom code for SMI triggers, this uses the generic READ_EVENT function. The underlying data had to be changed a bit, so that the trigger channel remains part of the data matrix.

PR #1221, opened 2019-09-20
improve implementation of SAM beamformer, and add possibility to compute pseudoZ statistic for beamformers

AdoNunes:
There was a SAM BF method that was not working, with undefined variables and a comment stating that the code did not work.


AdoNunes:
I did some simulations, in the fixedori 'Robert' method, the orientations are not being estimated properly.

robertoostenveld:
@AdoNunes is present here at the MNE hackaton and will work on this now.

Adonay, have a look at test_bug1049, or test_issue1198, or test_bug1599

AdoNunes:
@robertoostenveld I have included a script to test the orientations estimated by the different SAM methods. Currently, the robert method works only if the LFs are reduced, but the other methods do not work if the LF are reduced and back-projected (a PR was done previously and now it checks that the LF are not back-projected).

schoffelen:
Hi Adonay, I just merged a pull request that streamlined the beamformer code for the lcmv, dics and pcc methods (#1346). It would be nice to also update the SAM code if possible. What is the status of this?

AdoNunes:
@schoffelen  I had a look and basically, in this PR I added an option to input a noise covariance and get the most salient direction based on the ratio signal/noise "pseudo" power, and called 'moiseev'. This ratio provides a pseudo-z.

There are two previous methods, 'garteth', takes the ratio signal/squared signal and the results will be wrong if the LF rank is reduced to two, and 'Robert' which it seems it will give wrong orientations as I think I showed in the file test_sam.py, if the LF rank is not reduced. 

I think it would be important to at least have a sanity check and give an error if the user has a rank 3 with 'robert' method or 2 with 'garteth'. 

Let me know if you want some changes in this PR or have comments.

schoffelen:
Yes, I saw this in the code as well. In its current state, there's a conflict in ft_sourceanalysis (most likely caused by this function having changed since you branched it off of the master branch). This needs to be fixed first in this pull request before we can merge it.

ft_sourceanalysis.m
```diff
@@ -821,10 +821,16 @@
 
   if hasbaseline
     % baseline and active are only available together for resampling purposes,
+    % or as a noise covariance for SAM beamforming
     % hence I assume here that there are multiple trials in both
-    baseline.avg   = baseline.avg(datchanindx,:);
-    baseline.cov   = baseline.cov(:,datchanindx,datchanindx);
-    baseline.trial = baseline.trial(:,datchanindx,:);
+    if isfield(baseline, 'avg')
```
schoffelen, line 826: I think that this can be done in a more principled way, e.g. by looking at the 'covdimord' it can be determined whether the noise covariance has its original trials still present.
Also, from the lower level code, it seems that beamformer_sam assumes the baseline covariance to be a square matrix, thus it will fail if the baseline covariance actually contains repetitions.

schoffelen:
Sorry Adonai, I now totally messed up this PR. I was trying to make some improvement suggestions to your changes, but now there's a lot of mess added in. I will try to fix this as soon as possible.

schoffelen:
This PR is to be closed. We continue working on this in #1443

AdoNunes:
> Sorry Adonai, I now totally messed up this PR. I was trying to make some improvement suggestions to your changes, but now there's a lot of mess added in. I will try to fix this as soon as possible.

These were a lot of commits hehe I will work on #1443

schoffelen:
Thanks much!

PR #1222, opened 2019-09-20
Adding fast interactive source (surface) plot

Spaak:
This is working well on my end, but probably still needs some polishing; e.g. not all documentation is ready yet. I thought I'd file a pull request nonetheless, because I'd be interested to hear what @robertoostenveld thinks about some of the choices I made (adding two new files to `external/fileexchange`, putting the implementation class in `plotting`, using object-oriented code in the first place, etc.)

(Btw getting rid of the object-oriented code is nearly impossible without severely sacrificing functionality, clarity, and performance, so I hope this is fine.)

The new entry high-level function is `ft_sourceplot_interactive`, the real work is done by `plotting/ft_interactivesourceviewer`. I considered putting the latter in `private`, but private folders cannot contain class definitions.


Spaak:
Also interested to hear @schoffelen 's thoughts. BTW, would it make sense to create a branch for this in the main project?

external/fileexchange/subtightplot.m
```diff
@@ -0,0 +1,67 @@
+function h=subtightplot(m,n,p,gap,marg_h,marg_w,varargin)
```
robertoostenveld: this should come with a subtightplot.license file, since contributions on file exchange have different licenses.

external/fileexchange/brewermap.m
```diff
@@ -0,0 +1,528 @@
+function [map,num,typ] = brewermap(N,scheme)
```
robertoostenveld: this one is not needed, it is already present in fieldtrip/external/brewermap. It is also recognized by and handled in `ft_hastoolbox`.

plotting/ft_interactivesourceviewer.m
```diff
@@ -0,0 +1,343 @@
+classdef ft_interactivesourceviewer<handle
```
robertoostenveld: I don't see a better place for this file, but do propose to change its name. For high-level functions we tend to use ft_functionname, whereas low-level functions usually have multiple underscores. In this case I propose to start with ft_plot_xxx

Furthermore, I wonder whether it is "source" specific. That is a high-level concept and can be volumetric or surface-based. I suspect that this is more an interactive "mesh" plotter. A "mesh" is lower level and also used for other data (e.g. ft_plot_topo3d, ft_plot_mesh, or ft_meshrealign). Conceptually I think that a "source" can be a "mesh", but not the other way around; however, it is not always clear (considerer "raw+comp", "timelock+comp" and "freq+comp"). In this case I think the class expects a mesh with data, where data can have multiple dimensions.

How about calling it ft_plot_meshinteractive?

private/struct2keyval.m
```diff
@@ -0,0 +1,21 @@
+function keyval = struct2keyval(s)
```
robertoostenveld: this duplicates `ft_keyval2cfg` which you can find in fieldtrip/utilities.

robertoostenveld:
I added some first comments to be addressed, just upon browsing the code on github. I did not run it yet, so expect that a 2nd round is needed. I would like the PR to include an inspect_xxx.m script, or an inspect_pull1222 (pointing to this online PR).

robertoostenveld:
Btw, I don't see the advantage of having a branch in fieldtrip/fieldtrip. Right now @schoffelen and I can also push directly (when needed) to the newsurfaceviewer branch on Spaak/fieldtrip.

plotting/ft_interactivesourceviewer.m
Spaak: Agreed, I'll call it `ft_plot_mesh_interactive`. Indeed the higher-level function is requesting 'source+mesh' data from ft_checkdata.

private/struct2keyval.m
Spaak: The one I wrote goes the other way; I'll rename `struct2keyval` to `ft_cfg2keyval` and put it in fieldtrip/utilities instead.

Spaak:
Thanks for the quick feedback. I think it's now ready for a test drive.
(Sorry about the deletion/addition of the basically the entire file in commit 57a33c7, that was due to some line endings I converted to Unix-style.)

> Right now @schoffelen and I can also push directly (when needed) to the newsurfaceviewer branch on Spaak/fieldtrip.
Cool, I didn't know that was possible!

Spaak:
Anything more that needs to be done here? Or ready for merge @robertoostenveld ?

robertoostenveld:
hé, ik was net aan het reviewen! Maar zag er wel goed uit hoor ;-)

schoffelen:
Eelke sprak me erover aan bij de koffieautomaat, en het leek me wel een leuke functie om te showcasen in Parijs. Scheelt jou weer werk nu. Alleen jammer van het object-georienteerde gedeelte ... :)

PR #1230, opened 2019-10-11
fixing bug1152

ualsbombe:


robertoostenveld:
Looks good to me.
 
@ualsbombe have you tested this with different datasets (not only planar and mags, but also with EEG?)

schoffelen:
Ah, I think I clicked the wrong button. I approve of the change, but I would like to see some additional changes in the documentation and the cfg handling.

ualsbombe:
The change is no good - cfg.channel has no function - no matter what you put in - all channels are shown - including trigger channels. Anyway to revert it?

schoffelen:
Hi @ualsbombe Lau, What's the status of this? I don't know in whose court the ball is, could you have quick look at it to get it resolved?

ualsbombe:
Hi Jan-Mathijs

By mistake, I created two pull requests for this. The other one (with the correct solution) was pulled some days ago (https://github.com/fieldtrip/fieldtrip/pull/1281).

I'll delete this one accordingly.

All the best

Lau

PR #1238, opened 2019-10-17
improve specification of colors when plotting lines for channels and/or conditions

### LINKED ISSUE #714: Allow plotting of multiple channels using ft_singleplotER
As it came up in the NIRS hackathon, hereby a request to have the channel averaging in ft_singleplotER optional, i.e. as a cfg-field (could e.g. be done analogue to ft_databrowser with cfg.viewmode = 'butterfly' or 'average' or 'vertical', where 'average' is the default). I see several ways this could be implemented, but all of them require a substantial amount of work (given the small added feature), and would therefore categorize this as a 'mid effort, low priority' (for NIRS users it is probably high priority). Note that someone of us here might pick this up in future if not being worked on from your side (over the next 2 years I have to say).

robertoostenveld:
closes #714

### DIFF
--- a/ft_connectivityplot.m
+++ b/ft_connectivityplot.m
@@ -62,8 +62,9 @@ if ft_abort
 end
 
 % check if the input cfg is valid for this function
-cfg = ft_checkconfig(cfg, 'renamed', {'zparam', 'parameter'});
-cfg = ft_checkconfig(cfg, 'renamed', {'color',  'graphcolor'}); % to make it consistent with ft_singleplotER
+cfg = ft_checkconfig(cfg, 'renamed', {'zparam',     'parameter'});
+cfg = ft_checkconfig(cfg, 'renamed', {'color',      'linecolor'});
+cfg = ft_checkconfig(cfg, 'renamed', {'graphcolor', 'linecolor'});
 
 % set the defaults
 cfg.channel   = ft_getopt(cfg, 'channel',   'all');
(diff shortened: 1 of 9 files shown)


robertoostenveld:
this improves the handling of the user-specified channel colors in case a single dataset is supplied to ft_databrowser (which already supported different colors per channel) and ft_multiplotER (which now also supports different colors per channel)

In case multiple datasets/conditions are specified, the behavior is the same as before, i.e. different colors per conditions. You cannot mix colors over channels in case of multiple conditions.

I also decided to not implement multiple-channel display in ft_singleplotER, since the averaging and the masking/highlighting interferes with it. Instead, please use the cfg.layout='butterfly' option with ft_multiplotER if you want to plot them over each other. That also supports different colors as in

```cfg = [];
cfg.layout = 'butterfly';
cfg.linecolor = 'rgbymc';
cfg.colorgroups = 'sequential';
figure; ft_multiplotER(cfg, cond1);
 ```
Screenshot 2019-10-17 at 14 58 58

helenacockx:
I've tested the 3 functions (ft_singleplotER, ft_multiplotER and ft_databrowser) out on an fNIRS dataset and this seems to work! You can find a simplified dataset (timelock_fingertapping.mat), the used layout (layout.mat) and the script (issue_1238.m) attached as an example.
[issue_1238.zip](https://github.com/fieldtrip/fieldtrip/files/3749750/issue_1238.zip)

PR #1244, opened 2019-10-22
added cfg.templatecoordsys to ft_volumenormalize

mcvinding:
Option to specify a different coordinate system for normalized MRI. Only overwrites the default. Otherwise will be 'spm' as default and will be forced to be 'spm' if using any SPM template.

### DIFF
--- a/ft_volumenormalise.m
+++ b/ft_volumenormalise.m
@@ -107,6 +107,7 @@ cfg.keepinside       = ft_getopt(cfg, 'keepinside',       'yes');
 cfg.keepintermediate = ft_getopt(cfg, 'keepintermediate', 'no');
 cfg.nonlinear        = ft_getopt(cfg, 'nonlinear',        'yes');
 cfg.smooth           = ft_getopt(cfg, 'smooth',           'no');
+cfg.templatecoordsys = ft_getopt(cfg, 'templatecoordsys', 'spm');
 
 % check that the preferred SPM version is on the path
 ft_hastoolbox(cfg.spmversion, 1);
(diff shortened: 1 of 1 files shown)


ft_volumenormalise.m
```diff
@@ -137,6 +138,7 @@
     if strcmpi(cfg.spmversion, 'spm2'),  cfg.template = fullfile(spmpath, filesep, 'templates', filesep, 'T1.mnc'); end
     if strcmpi(cfg.spmversion, 'spm8'),  cfg.template = fullfile(spmpath, filesep, 'templates', filesep, 'T1.nii'); end
     if strcmpi(cfg.spmversion, 'spm12'), cfg.template = fullfile(spmpath, filesep, 'toolbox',   filesep, 'OldNorm', filesep, 'T1.nii'); end
+    cfg.templatecoordsys = 'spm';
```
robertoostenveld: it is not needed to specify it here right? It is already specified on line 110, with a default value of 'spm'

robertoostenveld:
line 126 and 128 also need to be updated, since the input MRI should be (approximately) coregistered with the template.

ft_volumenormalise.m
mcvinding: This will override the cfg.templatecoordsys provided when using spm's templated to make sure it is consistent with what is done.

mcvinding:
On line 126+128: _ft_convert_coordsys_ goes from various MEG coordinate systems to acpc but not the other way around. It gives warning if trying to convert to spm. If we provide the default output coordinate system as 'spm' it will convert the inout volume to acpc, but give a waring every time it runs.´

robertoostenveld:
It might be that ft_convert_coordsystem needs further changes. But the existing default behaviour (as used in the documentation on the website) will still work fine, so further changes only relate to actual extensions of functionality.

Issue #1235, opened 2019-10-15
`optopos` and `fiberpos` (and label) should be named consistently 


robertoostenveld:
@helenacockx discovered that the artinis code uses one, whereas field trip main uses the other.

@Horschig is there a reason to deviate from the [initial design](http://www.fieldtriptoolbox.org/development/project/nirs/#layouts)?

robertoostenveld:
Hi @Horschig, do you want me to fix this in the Artinis code?

helenacockx:
I've noticed that **opto2lay** (from **ft_prepare_layout**) is also asking for `opto.fiberlabel`, so this should maybe be implemented too?

robertoostenveld:
I see that the code that uses `fibers` in ft_nirs_prepare_ODtransformation dates back to 2016-10-07, which might very well be prior to us settling on the documented opto definition.

robertoostenveld:
this has been resolved with #1245

PR #1245, opened 2019-10-22
use `opto` instead of `fiber`, see #1235

robertoostenveld:


robertoostenveld:
if @Horschig agrees, then this closes #1235

Horschig:
The problem seems to lie a bit deeper. The header defines fiberpos, fibertype and fiberlabel as standard fields. Those would need to be changes as well for consistency.

I think the renaming from opto* to fiber* came about in the discussion on bugzilla (http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2753#c9). The issue is that a physical laser can be split so that it is attached to different sites on the head. Internally by the electronics, it is still regarded as one optode, and as such it is regarded as one datastream. However, for the user it is two optodes. Hence, we made the distinction between the 'fiber', representing what is physically on the head of the subject, and the optode which coincides with the eletronics of the device.

It's all a matter of nomenclature. I am actually not happy anymore with the name fiber*, because the same concept can hold for LED-based systems. I think re-naming it opto* is fine. Users just call it optode anyway and not fiber. But if we decide to rename/revert then it should be consistent in all the code. Agree?

robertoostenveld:
I did not realize that the initial reading code (in the p-files) already starts with `fiber` rather than `opto`, I thought it was just some old code remnants in a few files. I now reread parts of the bugzilla discussion, glad that we still have that! 

I don't have a strong personal preference but in the meetings with researchers "optodes" seem to be more commonly used to describe the "things" that we place as individual units on the head. Those "things"  have a position and a label on a sticker. And as I understand it, they can comprise multiple fibers, each transmitting at a different wavelength.

Let's go for optodes and use the definition as now in `ft_dataype_sens`
```
% The structure for NIRS channels contains
%   sens.optopos        = contains information about the position of the optodes
%   sens.optotype       = contains information about the type of optode (receiver or transmitter)
%   sens.chanpos        = contains information about the position of the channels (halfway transmitter and receiver)
%   sens.tra            = NxC matrix, boolean, contains information about how receiver and transmitter form channels
%   sens.wavelength     = 1xM vector of all wavelengths that were used
%   sens.transmits      = NxM matrix, boolean, where N is the number of optodes and M the number of wavelengths per transmitter. Specifies what optode is transmitting at what wavelength (or nothing at all, which indicates that it is a receiver)
%   sens.laserstrength  = 1xM vector of the strength of the emitted light of the lasers
```

For backward compatibility with existing scripts and data stored as mat files on disk we also should use `ft_datatype_sens` to convert `fiber` on the fly. That should be called by `ft_checkdata` on all input data structures. So initially it should be possible to keep the p-files as they are, and only update the higher level FT functions.

robertoostenveld:
I will update ft_datatype_sens and will make one (possibly two) test scripts from the current tutorials, to ensure that they keep running as they are.

Horschig:
True in the metainfotohdr.p there is indeed the fiber* definition hard-coded (even still reminiscent commented code definint opto.transmits).

I think the comment in ft_datatype_sens was just not updated in the end, because it still contains a descirption of sens.transmits and sens.laserstrength. Those two did not make it into the final release ;) they can thus be removed.

robertoostenveld:
Right now sens.transmits is still in there (I actually just now renamed sens.tranceiver into sens.transmits), but it is not a boolean as specified on http://www.fieldtriptoolbox.org/development/project/nirs/. Can you define it?

Also, I now notice that sens.DPF is optionally used in ft_nirs_prepare_ODtransformation, but I just removed it in ft_datatype_sens. Can you define it?

Also sens.laserstrength is still in there. Should it be removed?

Please take either the definition [here](http://www.fieldtriptoolbox.org/development/project/nirs/#header) or the one in ft_datatype_sens as in this PR, and extend it the way you think it should.... You can post it here, I can integrate it in the PR. Anything that is not defined/documented to the end-user, should be removed.

robertoostenveld:
I am able to run the complete singlechannel and most of the multichannel tutorial, but run into a problem towards the end of the multichannel. There is some renaming of channels that I don't understand and that the tutorial code actually gives an error on.

@helenacockx or @Horschig  or @marcvanwanrooij could you clarify what is happening there?

helenacockx:
This is because the labels of the layout are changed from 'O2Hb' to 'functional', but the timelockDEV.label has still the 'O2Hb' labels. Or you should rename those to 'functional' too, or you should just leave the labels of the layout 'O2Hb'. I think that the last option is best. I am not sure why you would like to change the label to 'functional', because 'O2Hb' and 'functional' do not exactly mean the same.

marcvanwanrooij:
Agreed. 

If I am not mistaken, the functional component should not be in the multichannel tutorial (possibly a remnant from my analyses of the data). 

The code for a 'functional' and a 'systemic' component should still be correctly implemented (and yes, it should not replace other channel labels), and verified.

robertoostenveld:
Although not all questions wrt the `opto` definition have been resolved, I will merge this now to ensure that we can all move ahead.

Horschig:
Reminder for myself that many of the artinis\ functions are still working with fiberlabel and therefore do not work right now. I need to change this.

PR #1251, opened 2019-11-06
[ENH] LCMV beamformer: automatically compute noise for NAI weight normalization

britta-wstnr:
If the LCMV beamformer is run with ``cfg.weightnorm = 'nai'`` but without explicitly setting ``cfg.projectnoise = 'yes'``, the code crashes with ``Undefined function or variable 'noise'.``

This PR makes the noise computation in the NAI case automatic.
Furthermore, it adds ``weightnorm`` to the documentation where it was missing so far.

### DIFF
--- a/inverse/beamformer_lcmv.m
+++ b/inverse/beamformer_lcmv.m
@@ -168,8 +168,9 @@ else
   tmplambda = 0;
 end
 
-if projectnoise
+if projectnoise || strcmp(weightnorm, 'nai')
     % estimate the noise level in the covariance matrix by the smallest singular (non-zero) value
+    % always needed for the NAI weight normalization case
     noise = svd(Cy);
     noise = noise(rankCy);
     % estimated noise floor is equal to or higher than lambda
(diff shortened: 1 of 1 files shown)


inverse/beamformer_lcmv.m
```diff
@@ -168,8 +169,9 @@
   tmplambda = 0;
 end
 
-if projectnoise
+if projectnoise || strcmp(weightnorm, 'nai')
```
britta-wstnr: @robertoostenveld Here, I am not sure, whether it would be necessary to inform the user via a warning, that noise is computed.

schoffelen: I think it is not necessary. Also, the noise is computed and used for the weight normalisation, but not outputted

schoffelen:
Related to this: as far as I know, the other beamformer implementations don't support 'weightnorm' as an option, although  there's no real reason for this. Would you be game for implementing this in beamformer_dics (and beamformer_pcc, not sure whether possible) as well?

inverse/beamformer_lcmv.m
britta-wstnr: That makes sense!

britta-wstnr:
I can have a look at the DICS, but I have not much insight into the inner workings of a PCC beamformer ... The DICS should be straight forward enough - time being the only constraint ;-)

PR #1262, opened 2019-11-21
feed the option: tolerance down to ft_selectdata via ft_timelockanalysis

bmaess:
Dear Fieldtrip-Team,
I'd like to suggest to use 1/samp_freq as a reasonable tolerance value when comparing different time vectors over trials. Using 10^-5 for all sampling rates does not make much sense to me. However, to be able to configure this parameter, it should not be removed from the cfg struct.
best,
Burkhard

### DIFF
--- a/ft_timelockanalysis.m
+++ b/ft_timelockanalysis.m
@@ -136,7 +136,7 @@ end
 
 % compute the covariance matrix, if requested
 if computecov
-  tmpcfg = keepfields(cfg, {'trials', 'channel'});
+  tmpcfg = keepfields(cfg, {'trials', 'channel','tolerance'});
   tmpcfg.latency = cfg.covariancewindow;
   datacov = ft_selectdata(tmpcfg, data);
   % restore the provenance information
@@ -191,7 +191,7 @@ if computecov
 end
 
 % select trials and channels of interest
-tmpcfg = keepfields(cfg, {'trials', 'channel', 'latency'});
+tmpcfg = keepfields(cfg, {'trials', 'channel', 'latency','tolerance'});
 data   = ft_selectdata(tmpcfg, data);
 % restore the provenance information
 [cfg, data] = rollback_provenance(cfg, data);


schoffelen:
Thanks Burkhard,
It sounds reasonable to want to have the tolerance parameter to be passed on to ft_selectdata by the higher level fieldtrip functions. Whether or not 1/fs is appropriate for latency selection, that's up for discussion. I would argue that this depends on the application :). 

Anyway, I suggest that this then should be built into all instances where a high-level fieldtrip function (i.e. in the fieldtrip main directory) is calling ft_selectdata with the possibility to select tolerance-supporting quantities (it seems that pos and freq are candidates, besides latency). Also, I think that in that case also the help of the respective functions needs a little bit of editing.

If I grep for ft_selectdata, I get the following list:
JM-laptop:fieldtrip jansch$ grep -l ft_selectdata *.m
ft_appenddata.m
ft_appendsource.m
ft_channelnormalise.m
ft_channelrepair.m
ft_clusterplot.m
ft_componentanalysis.m
ft_connectivityanalysis.m
ft_connectivityplot.m
ft_crossfrequencyanalysis.m
ft_denoise_dssp.m
ft_denoise_pca.m
ft_denoise_prewhiten.m
ft_denoise_synthetic.m
ft_denoise_tsr.m
ft_detect_movement.m
ft_eventtiminganalysis.m
ft_freqanalysis.m
ft_freqdescriptives.m
ft_freqstatistics.m
ft_globalmeanfield.m
ft_headmovement.m
ft_math.m
ft_megplanar.m
ft_megrealign.m
ft_multiplotER.m
ft_multiplotTFR.m
ft_mvaranalysis.m
ft_preprocessing.m
ft_qualitycheck.m
ft_redefinetrial.m
ft_rejectvisual.m
ft_removetemplateartifact.m
ft_resampledata.m
ft_scalpcurrentdensity.m
ft_singleplotER.m
ft_singleplotTFR.m
ft_sourceanalysis.m
ft_sourcedescriptives.m
ft_sourcegrandaverage.m
ft_sourceplot.m
ft_sourceplot_interactive.m
ft_sourcestatistics.m
ft_statistics_stats.m
ft_timelockanalysis.m
ft_timelockgrandaverage.m
ft_timelockstatistics.m
imotions2fieldtrip.m

With a text editor the changes can be easily made.

@robertoostenveld, what do you think?

bmaess:
# I have answered via email - but this does not show up here ... so this was the email:
Hi Jan-Mathijs & Robert,

o.k. I see. If it is more than the time axis which uses this tolerance parameter, you might consider renaming it, e.g.
time_tol
freq_tol
pos_tol
ori_tol
... ?
I got stuck with the current convention filtering out the tolerance value and setting it to 10^{-5}. This is why, I modified the code of ft_timelockanalysis.m
After updating from an older version (20180308) to the current version 20191118 - the ft_timelockanalysis did not work anymore. Time axes between trials were too different :(
Adding so many changes without testing is probably not the best idea, isn't it? What would you suggest - will you introduce the modality specific tolerances with specific names? Shall I withdraw my PR?

best,
Burkhard

schoffelen:
Thanks Burkhard,

Well, these days we have a testing framework in place that is ran upon each change to the code. Only when no new failures show up, the updated code is allowed to show up in the release version. Although it's not necessarily directly geared towards testing the change, the total number of different scenarios is sufficiently generic to provide some protection.

Currently, in ft_selectdata, no distinction is made between freq/latency/pos, the tolerancy number is just a scalar. I think it's fine for now to leave it like that.

robertoostenveld:
I think it is indeed a good idea to make the numerical tolerance user configurable.

> Currently, in ft_selectdata, no distinction is made between freq/latency/pos, the tolerancy number is just a scalar. I think it's fine for now to leave it like that.

I would not know how to interpret this number. In some cases it can be relative, but in some cases not. E.g. time is usually sampled linearly, frequency might be sampled logarithmically, and the sampling of positions might be irregular (cortical sheet, or single dipole) which makes it even harder to find the "typical scale". For `cfg.timetolerance` I would say it should be in seconds (since all with time is in seconds), for `cfg.freqtolerance` it should be in Hz, and `cfg.postolerance` should be in units in which the positions are specified(?)

bmaess:
Hi again,
defining the tolerance as a relative or as an absolute value is both reasonable in its ways. The tolerance value for comparing the different axes, however, is possibly easier to understand when given as absolute values. As I stated in my first comment - for a time axis - it is reasonable to take the inverse of the sampling rate as a tolerance threshold. Otherwise, you may observe, what I have observed: Fieldtrip finds differences between time axes were no difference should be! The same is true for freqtolerance. Note, that freq. sampling rates could also be even at the logarithmic scale. So, one solution would be to let the user decide, or the other let fieldtrip decide in a data driven way. Only in the latter, fieldtrip would be able to handle linear and logarithmic scales correctly. So maybe, this is rather what you would prefer? It seems to me that in fieldtrip you try to do everything for the user ;)
BTW, I identified the tolerance also as a problem in respect to the minimum norm solutions. Here, the tolerance is probably best defined as an relative value. But, the current code sets the tolerance to 10^-12 and this is much too small for the neuromag data if you have used maxfilter (SSS correction). In my case, I get reasonable solutions for the simple source space model if the tolerance to set to 10^-6. The simple source space is just the shrunken version of the inner skull surface.

robertoostenveld:
for a concrete implementation it would help to have a piece of example code that demonstrates how the user would specify the tolerance for source positions, for frequencies and for time. That could e.g. be added as a test script in fieldtrip/test.

bmaess:
o.k. - I could write a script test/test_pull1262.m in a similar way as you have the other test scripts. I would also add a block of MEG data to demonstrate the problem with the too low tolerance value for my time scales between trials?

bmaess:
Just, a short question - do you expect me to do as suggested three days ago, or do you prefer something else?

robertoostenveld:
Hi @bmaess this has been on hold for some time, but I hope you are still interested in pushing this forward. 

Could you first extend your PR with a test_pull1262 script that shows the effect (i.e. two simulated timelock structures with a time axis that is slightly offset between them). Let's first review how that would work for ft_selectdata on timelock structures, and from there extend it to the others (as already identified by @schoffelen)

bmaess:
o.k. - I almost forgot this issue already. Please allow a few days before I can answer it.

robertoostenveld:
I think it would be good to get this - rather minor - pull request out of the way and merged. 

I will have a look at some of the functions where it is easy to pass `cfg.tolerance` but cannot guarantee that I'll change it throughout.

robertoostenveld:
let's wait and see whether the [testing](http://www.fieldtriptoolbox.org/development/testing/) passes and this makes it into the release.

PR #1265, opened 2019-11-25
fix for the recursive call issue

vlitvak:


schoffelen:
Sorry for being so slow on this @vlitvak. The change looks good

PR #1270, opened 2019-12-05
Appenddata fs

abush84:
Modified ft_appenddata to use sampling rate from header (hdr.Fs) if available, instead of using deprecated field fsample.

### DIFF
--- a/ft_appenddata.m
+++ b/ft_appenddata.m
@@ -110,12 +110,15 @@ cfg.tolerance  = ft_getopt(cfg, 'tolerance', 1e-5);
 isequaltime  = true;
 isequallabel = true;
 issamelabel  = true;
-isequalfsample = true;
+isequaltrial = true;
+isequalfreq  = true;
 for i=2:numel(varargin)
   isequaltime  = isequaltime  && isequal(varargin{i}.time , varargin{1}.time );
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks for the suggested change. Actually, although the plan was to deprecate `fsample`, in reality it never really happened since too much code depends on it. On https://github.com/fieldtrip/fieldtrip/blob/561b85333cfbd87e945e1106e41d93b8e8be068d/utilities/ft_datatype_raw.m#L121 you can see that `fsample` is also added when not present, and   `ft_datatype_raw` should always be called when raw data enters a high-level FT function. 

In the specific case of `ft_appenddata`, I would  expect that https://github.com/fieldtrip/fieldtrip/blob/561b85333cfbd87e945e1106e41d93b8e8be068d/ft_appenddata.m#L102 ensures that `fsample` is present.  

But in ft_appenddata it cannot be assumed that `hdr.Fs` exists or that it is correct, since the data might have been resampled prior to the call to `ft_appenddata`. That is why we never (or hardly ever) rely on `hdr` fields. Perhaps you may want to suggest an improvement to `ft_datatype_raw` for using the header field to improve the estimate/determination of the sampling frequency.

PR #1271, opened 2019-12-05
Neuroomega nsx

abush84:
Several fixes and updates to read NeuroOmega and Blackrock's NSx files.
(Add ?w=1 to github's URL to see the diff with whitespace ignored)


schoffelen:
Dear Alan,

Thanks and sorry for our slow response in this. The changes look good. Do you happen to have some small test data to share, which (in combination with a test function) can be used to test (and ensure) functionality.

Also, could you have a look at the functions' help and expand where possible? For instance, ft_read_header mentions BlackRock and Neuroomega in the option chantype, but not downstream, where it mentions the supported dataformats.

Also, to be considered: the 'getting started' section on the website is very outdated for blackrock (http://www.fieldtriptoolbox.org/getting_started/blackrock/), and does not contain an entry for neuroomega. It might be worthwile to update this.

And finally, are there any native *.mpx file readers around for matlab, and do you think there is interest of directly reading in the data into fieldtrip, rather than going through the Map File Converter?

abush84:
Dear Jan, 

Sorry for taking so long to get back to you. I can provide a small file for testing purposes. Where should I upload it? I couldn't find any test data directory within the fieldtrip repo. Regarding the test function, I'll copy the structure from an existing function in the test directory. Any other suggestions or pointer as to how I should write it?

There are no native *.mpx readers in matlab or any other language. This is a proprietary format from AlphaOmega, and they are not willing to share the format specification, so for the moment we are stuck with the Map converter.

schoffelen:
Dear Alan,
Thanks. We have a collection of test data files on our local file server; the overall collection is quite sizeable, so it's not part of the release package. We use these data for local daily testing, to ensure that, while the code develops, things keep working. Please see http://www.fieldtriptoolbox.org/faq/how_should_i_send_example_data_to_the_developers/ for some pointers about how to share data. Otherwise, you can also use the data sharing facility of your institution.

abush84:
Dear Jan, 

I updated the documentation of ft_read_header and did a couple of minor changes. You can find NeuroOmega and ns2 files for testing [here](https://www.dropbox.com/sh/ct1kshsrnsjmlp8/AACpUKzFVfy-CbIm7E6SkCdQa?dl=0). I also included a test_pull1271.m file, but you will have to modify the path for it to work on your system. 

Best, 
Alan

schoffelen:
Hi Alan,
Thanks once more. Sorry to keep nagging about this. Although in principle your changes are fine, I am wondering whether 'in the grand scheme of things' these changes are most robustly represented in ft_read_header, rather than somewhere else. 
We have (in addition to ft_read_header etc), a bunch of low-level fileformat-specific helper functions that contain a lot of low-level intelligence. In light of the current PR I am thinking of ft_chantype, ft_senstype, ft_senslabel in particular. For some other fileformats, for instance, the hdr.chantype field is filled by a call to ft_chantype. Thus, I was wondering whether your chantype handling code could be relatively straightforwardly moved to ft_chantype?

abush84:
Hi Jan, 

Yes, I see your point, but I'm having a little trouble identifying exactly how this would work for AlphaOmega files. This system is an intraoperative clinical equipment that records from a bunch of sensors, including microelectrodes for single/multi unit recordings, macroelectrodes for LFPs, EMG, analog channels, timestamps for digital triggers, etc. Interestingly, it records each data stream at a different sampling rate, which complicates things. Additionally, it calculates 'derived' signals, for instance it does some spike sorting, different frequency-filtered version of the raw channels, etc. Another complication is that some channel names reflect the clinical setup and can change between recording sessions.

What would be a reasonable behavior for ft_senslabel in this case? The available channels can change a lot between different files, so it is hard to come up with a predefined set of sensor labels. ft_senslabel should have access to the header to return a useful set of labels, but then I will not be able to use the function from ft_read_header. 

I could expand ft_chantype and ft_senstype for NeuroOmega files, but don't think I'll used those functions from ft_read_header either. 

Thanks,
Alan

robertoostenveld:
My apologies that I don't have more time to look into this, but I want to make some general comments here: 

The code of ft_read_header, data and event should be maintained/extended such that the interface remains consistent over multiple data formats. See http://www.fieldtriptoolbox.org/development/module/fileio/. This is needed to maintain internal compatibility, but also with EEGLAB and SPM. 

The code that is suggested with the following documentation in the help of ft_read_header should go:

```
% Use cfg.chantype='chaninfo' to get hdr.chaninfo table. For BlackRock
% specify decimation with chantype:skipfactor (e.g. cfg.chantype='analog:10')
```

The output hdr cannot have more fields than it has now according to the API specification. Dataformat-specific stuff should go into `hdr.orig`. 

Options for selecting channels differently than through `chanindx` should be implemented in a consistent fashion over multiple file formats. The EDF implementation can act as a guideline here.

abush84:
I modified the functions to conform to the API specification. Now all NeuroOmega and Blackrock specific info is returned in hdr.orig.

fileio/ft_read_event.m
```diff
@@ -1836,38 +1836,53 @@
...
+    end
+
     rx=regexp(fields_orig,'^CDIG_IN_{1}(\d+)[a-zA-Z_]*','tokens');
     dig_channels=unique(cellfun(@(x) str2num(x{1}), [rx{:}]));
-    
-    event.type=[]; event.sample=[]; event.value=[];
+
     if ~ismember(detectflank,{'up','down','both'})
-      ft_error('incorrect specification of ''detectflank''');
+      ft_error('incorrect specification of cfg.detectflank. Use up, down or both');
```
robertoostenveld, line 1861: the help of this function should not refer to how the cfg option is called in ft_preprocessing, ft_databrowser, or elsewhere in high-level functions. In this function the option is simply called 'detectflank'

robertoostenveld:
Hi Alan, please see the [PR](https://github.com/Brain-Modulation-Lab/fieldtrip/pull/1) that I just made against your branch. It now looks good to me.

abush84:
Hi Robert, I accepted the PR, should be good to go. 
Thanks for working on this.

abush84:
Corrected error messages.

schoffelen:
OK, I am about to merge this, since there's nothing to be done here. Robert's comments have been addressed, as far as I can see. The only thing that remains, is to think a bit about the large number of files (although the size seems to be OK) in the NPMK toolbox. A large part of this seems to be in .svn/* so not functional at all. We could consider removing this from the version that is packaged with fieldtrip

abush84:
Hi Jan.I just removed the .svn directories from the NPMK toolbox. Let me know if there is anything else to change before we are good to merge.

schoffelen:
Thanks @abush84. I will merge soon, but not today, since we are ironing out some ripples in the code base, and we want to make a guaranteed stable new release version. Once this is in place, we are allowed to make random merges again :). Stay tuned, I think tomorrow or otherwise Thursday. Thanks for your perseverance and patience.

schoffelen:
note for self. Need to update the test files still, but let's not be hold off by this detail.

PR #1272, opened 2019-12-05
ft_resampledata now returns 0 as extrapolation value

abush84:
Specified the 'extrapolation' parameter for interp1 to ensure data is not extrapolated. This can happen when cfg.time is used and times are given outside the range of the current data. The added parameter enforces that all values outside the data range are set to zero. 
(Add ?w=1 to github's URL to see the diff with whitespace ignored)

### DIFF
--- a/ft_resampledata.m
+++ b/ft_resampledata.m
@@ -150,14 +150,14 @@ if usefsample
   %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
   % resample/downsample based on new sampling frequency
   %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-  
+
   ntr = length(data.trial);
   nchan  = numel(data.label);
-  
+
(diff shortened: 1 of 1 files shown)


schoffelen:
Sorry for our slow response here. Thanks to the current global state of affairs I have some opportunity to clean up the list of open PRs.
- I guess that your reasoning was that extrapolation in itself is a bad idea, because it leads to the edges most of the time looking ugly.
-But why is '0' than the right extrapolation value, and not nan?
-I would go for Nan by default, and make an option cfg.extrapval
@robertoostenveld what do you think?

robertoostenveld:
Both nan and 0 would lead to problems if the data were subsequently filtered, baseline corrected or otherwise (pre)processed. I would expect an extrapolated value to be closer to the hypothetical value of the data.
 
But if any, I would go for nan, since in that case the downstream problems would become visible (i.e. the nan would spread out), whereas with 0 the user would not become aware of possible problems. 

I think this is to be considered in a similar way as an artifact, i.e. there is a section in the data that is not trustworthy. We have `ft_artifact_nan` to detect nans and we have in `ft_rejectartifact` the option `cfg.artfctdef.value` to replace nans with a value of choice.

schoffelen:
Good points. My experience with extrapolation is that typically the data 'explodes', if the length of the to-be-extrapolated segment is more than a few samples, which makes this behavior also rather useless.

schoffelen:
Regarding @robertoostenveld last point: perhaps related to this are the different ways in which padding can be done...

PR #1274, opened 2019-12-06
ENH - updated in order to get a match between Eric's expectations of …

schoffelen:
…the output frequencies and the actual output frequencies

### DIFF
--- a/contrib/misc/ft_laggedcoherence.m
+++ b/contrib/misc/ft_laggedcoherence.m
@@ -122,23 +122,19 @@ oddnsamples = mod(nsamplespertimwinall,2) == 1;
 nsamplespertimwinall = nsamplespertimwinall(oddnsamples);
 freqall = freqall(oddnsamples);
 timepertimwinall = nsamplespertimwinall/Fs;
-% % Find the best matching frequencies
-% freq = [];
-% t_ftimwin = [];
-% nsamplespertimwin = [];
-% for foiind = 1:length(cfg.foi)
-%   absdiff = abs(freqall-cfg.foi(foiind));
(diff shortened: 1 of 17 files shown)


schoffelen:
Hi @hadrianswall, I have updated the heuristic to determine the t_ftimwin, and the correspondingly allowed toi bins per frequency, in order to match the expectations between the specified cfg.foi, and the actual output of ft_laggedcoherence. As discussed in the FAQ and in person, the leading quantity that eventually determines the returned frequencies, is the cfg.pad parameter. Since in the original implementation the leading parameter was the t_ftimwin (with an imperative odd number of samples fitting exactly three cycles of oscillation) I think that the expected integer frequency bins can never be achieved to begin with. So, this means, that a small compromise is made, so that the t_ftimwin is (if necessary) slightly longer than 3 cycles, the effect of which is taken care of by the Hanning taper. 
The test_laggedcoherence (now in fieldtrip/test) at least returns the expected frequency bins. I also think that the toi's are selected correctly, avoiding data overlap.

robertoostenveld:
is this ready to be merged?

schoffelen:
according to me: yes, but still waiting for Eric.

robertoostenveld:
@hadrianswall can you please look at this? If not, I will close the pull request due to lack of interest.

schoffelen:
OK, the PR is getting a bit messy, due to me fooling around to re-create this branch locally. Eric and I discussed the issue, and we have now agreed on a version of the code that we both can live with for now. I will close this PR, and PR a fresh branch that correctly reflects only the changes made.

PR #1275, opened 2019-12-06
Fixing issue with empty tissue label

arnodelorme:
This occurs for example when ft_read_atlas is used to read Talairach-labels-2mm.nii (since the labels are stored in a separate text file)


schoffelen:
I am not familiar with the Talairach-labels-2mm.nii file. I could only find a zipped version of it on neurovault, and not the associated text file. Please provide with pointers to where the data are to be grabbed from, and some example code to demonstrate the issue, and the fix. Also, why is there a commit pertaining to ft_version in this PR?

schoffelen:
OK, I found a file in the fsl release.
When I do ft_read_atlas('Talairach.xml') I get a result.
So, what problem does this PR intend to solve?

utilities/ft_version.m
```diff
@@ -141,9 +141,12 @@
   
 else
   % get it from the Contents.m file in the FieldTrip release
-  a = ver(ftpath);
-  ftver = a.Version;
-  
+  if ~isdeployed
```
schoffelen, line 144: This difference seems a bit out of place for this PR, probably it needs to be removed.

schoffelen:
Hi @arnodelorme could you provide some code that replicates your problem. Also, see my earlier comment about the respective atlas that seems to cause the issue for you.

schoffelen:
Hi @arnodelorme, I'd like to resolve this issue. To me, it's still unconfirmed. Could you provide us with some directions here? Otherwise, I will close this PR for now.

schoffelen:
In the absence of any input, I will close for now.

PR #1277, opened 2019-12-12
Update in ft_redefinetrial to enable overlap

rtheze:
So far the option allow for overlap is not possible because the option is not retrieved and not taken into account by ft_fetch_data. With this modification, adding cfg.overlap = true will enable it when calling ft_redefintrial. Apparently this was to prevent some issues, but when you need to manually add some buffer that will be further removed, for instance, then it can become useful to have data overlap.

### DIFF
--- a/ft_redefinetrial.m
+++ b/ft_redefinetrial.m
@@ -239,6 +239,9 @@ elseif ~isempty(cfg.trl)
   % ensure that sampleinfo is present, otherwise ft_fetch_data will crash
   data = ft_checkdata(data, 'hassampleinfo', 'yes');
   
+  % check for overlap info, otherwise ft_fetch_data will ignore it
+  overlap = cfg.overlap;
+  
   dataold = data;   % make a copy of the old data
   clear data        % this line is very important, we want to completely reconstruct the data from the old data!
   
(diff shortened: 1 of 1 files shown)


schoffelen:
As can be read from the help section, the argument cfg.overlap is a scalar value between 0 and 1, and has been implemented in combination with cfg.length. 

The current change intends to extend the functionality of the redefinition when using cfg.trl, but 
seems to treat the cfg.overlap option as a boolean 0/1. This is inconsistent and confusing.

Is there a conceivable reason why the current default behavior ft_fetch_data does not allow for overlap at all? Wouldn't this be at the user's responsibility and shouldn't the 'allowoverlap' argument be true by default? @robertoostenveld what do you think?

My suggestion would be to change the call to ft_fetch_data to allow for overlap, and NOT use the cfg.overlap option to optionally allow for this. If there is a need to configure this, I'd create a new option, e.g. cfg.allowoverlap as a boolean. Also, this should be documented properly in the help section of the function. Also, a snippet of test code would be useful, to demonstrate that the new functionality works properly, and that it does not have any unwanted side effects.

schoffelen:
I suggest to close this one, and optionally create a new issue/PR if it is thought to be useful to follow up on my suggestions (i.e. add functionality to ft_fetch_data, and not define a new, method dependent, functionality of the cfg.overlap option in ft_redefinetrial

PR #1293, opened 2020-01-31
Correct a channel label in easycapM1_neighb.mat

rq-Chen:
The channel PO8 was incorrectly labeled as PO6 in the original `easycapM1_neighb.mat`, inconsistent with `easycapM1.mat`.

`ft_neighbourplot(cfg)` with `cfg.method = 'template'; cfg.layout = 'easycapM1.mat'; cfg.template = 'easycapM1_neighb.mat';` would give a figure like this:

Original

I changed all `'PO6'` to `'PO8'` in the `label` and `neighblabel` fields and now it's correct:

New


schoffelen:
Indeed, it seems that the corresponding layout contains the label PO8, rather than PO6. This seems to be consistent with the documentation from easycap.

schoffelen:
Thanks, I will merge this so that I can test it. Seems OK

rq-Chen:
You're welcome.

schoffelen:
And apologies for taking us so long.

schoffelen:
Note that PRs get merged into the master branch, and will only move to the release branch if our nightly run batch of test functions runs through smoothly. This may occasionally go wrong due to other changes to the code, so it could take a little bit of time before the change will become visible to all.

rq-Chen:
No problem! And thank you for the useful toolbox!

Issue #856, opened 2018-10-23
improve transformation between source models in MNI and individual coordinates 


robertoostenveld:
As discussed with @ualsbombe:

Proposed change
* make mri.cfg.final more consistent with mri.initial and mri.params (after ft_volumenormalise)
* update ft_transform_geometry such that it also can deal with nonlinear transformations (from params)
* implement ft_transform_sourcemodel

robertoostenveld:
@mcvinding let's use this issue to keep some notes of what we have been discussing. 

One idea was to extend ft_transform_geometry to also work on MRI. There were other concrete ideas as well...

xiaochenzheng:
Hey Rob, I would like to use ft_transform_geometry on my MRI, could you please help me out? Thanks!

robertoostenveld:
The pull request #1296 does not address all ideas yet, but is a step to make coordinate transformations more transparent and symmetric.

robertoostenveld:
On the website http://www.fieldtriptoolbox.org/faq/how_are_the_different_head_and_mri_coordinate_systems_defined/ I just noticed the use of `ft_estimate_units`. That is a low-level function that converts a single number (the approximate size, e.g. 10) to a string (e.g. 'cm'). it should be made in a proper low-level function and hidden from the user.

robertoostenveld:
Although #1296 improves the initial part of the conversion between coordinate systems, there is still work to be done on `ft_volumenormalise`. Right now it is hard-coded for non-MNI to MNI conversions.

PR #1296, opened 2020-02-05
improve conversion between coordinate systems

robertoostenveld:
The goal is to make all conversions more transparent and to reduce the number of functions involved. 

See #856


robertoostenveld:
with this PR the following functions are more consistent
- ft_determine_coordsys
- ft_convert_coordsys
- ft_determine_units
- ft_convert_units
- ft_transform_geometry

robertoostenveld:
this should also make ALIGN_NEUROMAG2ACPC and ALIGN_CTF2ACPC obsolete. The affreg/normalise code has moved one level up (to ft_convert_coordsys) and the xxx2yyy transforms are now hardcoded in ft_convert_coordsys. 

Also ALIGN_FSAVERAGE2MNI shoudl now be obsolete, the hardcoded transform has also moved to ft_convert_coordsys.

utilities/ft_hastoolbox.m
```diff
@@ -162,6 +162,7 @@
   'MVPA-LIGHT'    'see https://github.com/treder/MVPA-Light'
   'XDF'           'see https://github.com/xdf-modules/xdf-Matlab'
   'MRTRIX'        'see https://mrtrix.org'
+  'BAYESFACTOR'   'see https://klabhub.github.io/bayesFactor'
```
robertoostenveld: sorry, this change should not have been included in this PR. But it is harmless, so please ignore.

robertoostenveld:
I ran this branch with the dashboard, there are some regression errors. 

```
test_bug1828.txt:=== test_bug1828 FAILED in 1 seconds
test_bug1832.txt:=== test_bug1832 FAILED in 2 seconds
test_bug2990.txt:=== test_bug2990 FAILED in 10 seconds
test_bug3435.txt:=== test_bug3435 FAILED in 3 seconds
test_ft_volumenormalise.txt:=== test_ft_volumenormalise FAILED in 11 seconds
test_issue856.txt:=== test_issue856 FAILED in 7 seconds
test_spm_ft_integration.txt:=== test_spm_ft_integration FAILED in 3 seconds
```
 
I will try to fix these. Hopefully it would not affect the functional design.

robertoostenveld:
There is apparently a long-standing issue in `align_fsaverage2mni`: it contains a 3x4 transform matrix, which means that the output transform has never been correct (that was my bad in https://github.com/fieldtrip/fieldtrip/pull/453). 

See http://freesurfer.net/fswiki/CoordinateSystems

I will fix it. However, this does not solve all regression errors.

robertoostenveld:
there are still some issues, specifically with `tal` and `ras`. I should discuss these with @schoffelen 

But already a lot of conversions work, including nice graphical feedback like this (SubjectCMC, from CTF to ACPC):

Screenshot 2020-02-06 at 17 09 03

PR #1304, opened 2020-02-14
improve partialization of CSD matrix prior to connectivity computation 

robertoostenveld:
The issue was reported by Raul in Leuven as 

> I was interested on computing the Phase Locking Value between two relevant brain sources. However it was suggested to partialise the contribution of the stimulus to the two signals. In this case the stimulation was a sinusoid of a frequency of 9.766 Hz evoking a brain response at the same frequency. After computing the plv in FieldTrip most of the computed values are Nans. The input to ft_connectivityanalysis was obtained as shown below. I had a look at the cross-spectrum and could not see anything 'strange' at the frequency bin of interest.

```
cfg = [];
cfg.method = 'mtmfft';
cfg.output = 'fourier';
cfg.taper = 'dpss';
cfg.tapsmofrq = 1; % i.e. no real smoothing
cfg.foilim = [1 100];
[Data_freq] = ft_freqanalysis(cfg, Data);
cfg = [];
cfg.method = 'plv';
cfg.partchannel = 'Stimuli';
Data_stat = ft_connectivityanalysis(cfg, Data_freq);
```

### DIFF
--- a/connectivity/ft_connectivity_corr.m
+++ b/connectivity/ft_connectivity_corr.m
@@ -114,22 +114,20 @@ if ~isempty(pchanindx) && isempty(powindx)
   newsiz = siz;
   newsiz(2:3) = numel(chan); % size of partialised csd
   
-  A  = zeros(newsiz);
+  A = zeros(newsiz);
   
-  % FIXME this only works for data without time dimension
-  if numel(siz)==5 && siz(5)>1, ft_error('this only works for data without time'); end
-  for j = 1:siz(1) %rpt loop
-    AA = reshape(input(j, chan,  chan, : ), [nchan  nchan  siz(4:end)]);
-    AB = reshape(input(j, chan,  pchan,: ), [nchan  npchan siz(4:end)]);
-    BA = reshape(input(j, pchan, chan, : ), [npchan nchan  siz(4:end)]);
-    BB = reshape(input(j, pchan, pchan, :), [npchan npchan siz(4:end)]);
-    for k = 1:siz(4) %freq loop
-      %A(j,:,:,k) = AA(:,:,k) - AB(:,:,k)*pinv(BB(:,:,k))*BA(:,:,k);
+  for j = 1:siz(1) % loop over rpt
+    AA = reshape(input(j, chan,  chan, : ), [nchan  nchan  prod(siz(4:end))]); % fold freq_time into one dimension
+    AB = reshape(input(j, chan,  pchan,: ), [nchan  npchan prod(siz(4:end))]);
+    BA = reshape(input(j, pchan, chan, : ), [npchan nchan  prod(siz(4:end))]);
+    BB = reshape(input(j, pchan, pchan, :), [npchan npchan prod(siz(4:end))]);
+    for k = 1:prod(siz(4:end)) % loop over freq or freq_time
       A(j,:,:,k) = AA(:,:,k) - AB(:,:,k)/(BB(:,:,k))*BA(:,:,k);
     end
   end
   input = A;
   siz = size(input);
+  
 elseif ~isempty(pchanindx)
   % linearly indexed crossspectra require some more complicated handling
   if numel(pchanindx)>1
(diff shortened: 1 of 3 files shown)


robertoostenveld:
running the example, I noticed something strange

>> Data_stat = ft_connectivityanalysis(cfg, Data_freq);
the call to "ft_selectdata" took 0 seconds and required the additional allocation of an estimated 0 MB
Warning: could not determine dimord of "crsspctrm" in:

        label: {2×1 cell}   		2 CHANNELS HERE 
         freq: [1×102 double]
    cumsumcnt: [32×1 double]
    cumtapcnt: [32×1 double]
          cfg: [1×1 struct]
       dimord: 'rpt_chan_chan_freq'
    crsspctrm: [32×3×3×102 double]	3 CHANNELS HERE


This rpt_chan_chan_freq is an intermediate representation that is constructed on fly, since it is easier to do the PLV computation on the CSD matrix. 

The conversion happens on line 410
          [data, powindx, hasrpt] = univariate2bivariate(data, 'fourierspctrm', 'crsspctrm', dtype, 'channelcmb', cfg.channelcmb, 'keeprpt', normrpt);

but there it is still ok (after checking with the debugger).

Somewhere between 504-537 one label is removed, but the CSD matrix still stays the same. On line 625 it continues…
There I see that data.(inparam) became 5D with [1    32     3     3   102], whereas dimord is still 4D.

This error happens at line 570, where ft_selectdata gets (rightfully I would say!) confused about the input it receives. As a consequence, it does not average (which it should).

robertoostenveld:
the suggested code change keeps the data structure better intact, i.e. no internal inconsistencies any more. However, the output is still incorrect, since 

```
>> Data_stat = ft_connectivityanalysis(cfg, Data_freq)
the call to "ft_selectdata" took 0 seconds and required the additional allocation of an estimated 0 MB
the call to "ft_selectdata" took 0 seconds and required the additional allocation of an estimated 0 MB
the call to "ft_connectivityanalysis" took 1 seconds and required the additional allocation of an estimated 1 MB
Data_stat = 
  struct with fields:

        label: {3×1 cell}
       dimord: 'chan_chan_freq'
    plvspctrm: [2×2×102 double]
         freq: [1×102 double]
          dof: 32
          cfg: [1×1 struct]
```
the output only contains the 2x2 channels, whereas `label` is now 3x1.

robertoostenveld:
@schoffelen could you have a look at my changes? Your original code is not always clear to me, but I think that I removed (in a general way) the time restriction, and I think that the channel label handling now in general is better. However, at the end of the top-level function the output channel labels still need to be properly assigned.

schoffelen:
I think the changes look overall good to me, although it is not clear (from reading the discussion and the changes in the code) what the original problem was, and how Robert's changes have fixed this. The connectivity code is very hairy indeed (in particular the bookkeeping of the data), and it is worth considering a total makeover. With the datatypes well defined these days it should be more easy to deal with than when the code was originally written. Although the chancmb_ vs chan_chan_ representation will probably still be a cause for headaches.

The incorrect output label list is introduced in line 526/527, where originally the data.label was replaced by keepchn, but now it reads data.label(keepindx) = keepchn. Obviously, ideally the lower level code (which is doing the partialisation) should know about the labels and return the updated list, but back in the days that the code was designed we thought it a good idea to keep the lower level functions as dumb as possible.

robertoostenveld:
is there a test script that goes through the partialisation for all methods that support it? E.g. check that the output is consistent in size, that partialised values are smaller (or at least different) than non-partialised ones, etc?

robertoostenveld:
line 526/527 was the place of the original problem, since the data structure became internally inconsistent (3 channels in the CSD, 2 in the label), causing ft_selectdata to return the wrong results, wreaking total havoc further down. As long as it is still passed to other FT functions, the data structure should always remain internally consistent. 

At the end it now does 
```
stat = keepfields(data, {'label', 'labelcmb', 'grad', 'elec'});
```
Would it not be better to only keep grad, elec, opto, freq and time, and add all other fields (including the updated names for the channels) explicitly to the output? What is otherwise the point of changing the labels?

schoffelen:
No, I don't think so. test_ft_connectivityanalysis calls ft_connectivity once, with a non empty partchannel in the cfg

schoffelen:
I agree. This relates to the general strategy to create the output from scratch, rather than pruning the input data object half-heartedly

PR #1329, opened 2020-03-06
ENH - added all possible combinations of generic axis orientations (w…

schoffelen:
…ith unspecified origin)


schoffelen:
Hi Robert,
I implemented a generic conversion between coordinate systems in which only the axis directions are specified (with a,p,l,r,s,i). I think it is correct, but could you have a look to be sure?  The critical line is 498 in ft_convert_coordsys, where I am 99% sure that it is correct in the ordering of order_from and order_to

schoffelen:
Ah, good point. The from "with-origin-specified-and-implicitly-known-axis-orientations" to "without-origin-specified" should be possible indeed.

robertoostenveld:
how do you like these transformations? The 1st index (row) is "from", the 2nd index is "to". Anything that is not dark-blue is possible (light blue, green and yellow are the different ways the transformation was constructed).
 
Screenshot 2020-03-06 at 15 12 00

schoffelen:
Nice!

schoffelen:
Did you adjust your local version of the code already, or just colored your ideal matrix?

robertoostenveld:
I will send you a PR on your branch

schoffelen:
I think it is now properly merged from my end. Shall I merge this one into fieldtrip/master?

robertoostenveld:
yes, looks good. Sorry for the confusion.

schoffelen:
Next step would be to automatically decode the axis orientations from the dicom/nifti headers so that the coordsys can be educatedly guessed and added to the output of ft_read_mri

robertoostenveld:
perhaps we can use https://deepmind.com for that

PR #1330, opened 2020-03-08
Fixes #1312

### LINKED ISSUE #1312: [reproducescript] issue with comment "new input file"
If reproducescript writes a cfg-field to disk, which is used by a next fieldtrip call, a comment is written in the generated script, indicating that is a new file that enters the pipeline. This should be removed. The comment should only be written when there is a change to the data outside of fieldtrip @Spaak.

Spaak:
For reproducescript functionality, large cfg-fields are no longer stored as an input file (i.e. with `_input_` in their filename), but as a `_largecfginput_` file. This prevents them from being explicitly named in the comments by `reproducescript`.

@matsvanes


Spaak:
PS: Apologies for some weird git issues:

1. I have no idea why in the PR overview all my commits are listed as coming from 27 Nov 2019? I suspect a Github bug, since the Commits tab shows everything as it should be.

2. The two commits actually made on 27 Nov 2019 are irrelevant and superceded by later merge commits. All actual files changed pertain to the relevant issue.

robertoostenveld:
@matsvanes this is now merged

PR #1331, opened 2020-03-09
more than one channel to repair, allow entering them as a row vector.

dnacombo:
Allowing that code to run.
```
cfg = [];
cfg.badchannel = {'Fp2', 'AF8', 'FTT9h'};
cfg.neighbours = neighbours;
cfg.layout = lay;
data_clean = ft_channelrepair(cfg,data);

```
This was crashing because at modified location, badchannel was catenated as a row vector with a colon vector of missing channels.

### DIFF
--- a/ft_channelrepair.m
+++ b/ft_channelrepair.m
@@ -159,7 +159,7 @@ if ~isempty(cfg.missingchannel) && strcmp(cfg.method, 'weighted')
 end
 
 % get selection of channels that are missing and/or bad
-cfg.missingchannel = cat(1, cfg.missingchannel(:), cfg.badchannel);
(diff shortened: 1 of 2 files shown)


robertoostenveld:
Could you add a test script as `test_pull1331.m` that demonstrates (with the current code) that it fails and shows (with the new code) that it works? It is fine to base this off from Subject01.ds or similar, or a mat file with preprocessed data from the ftp server. We can update the test script so that the right demo data is loaded from the right place. 

The test script helps us to ensure that the new functionality is regression tested and not accidentally removed or broken in the future.

schoffelen:
Thanks, Max.

schoffelen:
for reference: I updated the test function so that it runs on our filesystem

Issue #1335, opened 2020-03-12
ft_prepare_layout does not convert 3D optode positions into 2D layout


helenacockx:
When creating a layout for nirs optodes based on their 3D optode positions, you got the following error: 
> Unable to perform assignment because the size of the left side is 1-by-3 and the size of the right side is 1-by-2.
> Error in ft_prepare_layout (line 1122)
>     layout.pos(end+1,:) = [min(pos(:,1))-width/2 min(pos(:,2))-height/2];

This is because 3D positions are not converted to 2D positions prior to generating a layout. 
You can find an example file here: 
[opto.zip](https://github.com/fieldtrip/fieldtrip/files/4324247/opto.zip)

robertoostenveld:
I see in http://www.fieldtriptoolbox.org/tutorial/nirs_multichannel/ (search for ft_layoutplot) that it is used differently. When specifying the oxy3 file, ft_read_sens returns positions (from the xml file?) that are 2d. This apparently never has worked for a 3D optode specification.

robertoostenveld:
Please see the new test script. 

I can explain the extra options (projection, viewpoint) later in person, or see http://www.fieldtriptoolbox.org/tutorial/layout/#creating-a-layout-from-sensor-positions for an example on a MEG sensor array and http://www.fieldtriptoolbox.org/tutorial/human_ecog/ for an ECoG example (see figure following step 49). I don't think that they are very relevant right now, but might be with a bilateral optode array that is very low (e.g. over temporal) and hence not rendered optimally if you project it pancake style along the z-axis.

PR #1336, opened 2020-03-12
convert 3d optode positions to 2d optode positions

helenacockx:
added a line to convert 3d positions to 2d positions.


robertoostenveld:
although this might work for the issue identified in #1335, it is not a generic solution. The core problem is that ft_read_sens on an oxy3 file returns optode positions that are 2D, whereas they should be 3D (according to the specification in ft_datatype_sens).

I will close this issue, and instead make a PR that 
- updates ft_datatype_sens to ensure that positions are 3D (i.e. add z=0) 
- extend ft_prepare_layout for fNIRS so that it work with 3D optode positions (also when all z=0)

PR #1340, opened 2020-03-13
improve handling of short NIRS channels

### LINKED ISSUE #1339: short channels in Artinis oxy3 file are not represented correctly
@helenacockx noticed this when trying to make a layout

```
Unable to perform assignment because the indices on the left
side are not compatible with the size of the right side.
Error in ft_prepare_layout>opto2lay (line 1436)
  layout.pos(i, :) = opto.optopos(rxid, :)/2 +
  opto.optopos(txid, :)/2;
Error in ft_prepare_layout (line 628)
    layout = opto2lay(sens, sens.label, cfg.rotate,
    cfg.projection, cfg.viewpoint); 
```

The problem is that opto.optolabel and optopos contain duplicates. _"Fysiek (dus in de cap) is het maar 1 optode. Schematisch stellen ze die voor als 2 optodes om het verschil tussen korte en lange kanalen weer te geven."_

I think this is also relevant for @Horschig to be aware of.

robertoostenveld:
This fixes #1339

### DIFF
--- a/fileio/ft_read_header.m
+++ b/fileio/ft_read_header.m
@@ -2630,9 +2630,8 @@ if checkUniqueLabels
     for i=1:hdr.nChans
       sel = find(strcmp(hdr.label{i}, hdr.label));
       if length(sel)>1
-        % there is no need to rename the first instance
-        % can be particularly disruptive when part of standard MEG
-        % or EEG channel set, so should be avoided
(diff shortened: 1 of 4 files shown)


robertoostenveld:
the processing of the short channels is now only marginally improved, but this is as far as I can get it.

Issue #1334, opened 2020-03-12
selection of reference channels failed for combined MEG-EEG data


awollbrink:
for a combined MEG-EEG dataset recorded with the CTF 275 sensor system the channel selection for extracting the reference channels fails.
Applying ft_preprocessing with cfg.channel = {'megref'} results in a data structure containg the 29 reference sensors plus some EEG sensors which channel labels start with the letter 'P'.
I am using Fieldtrip revision 93dd0f61b on a Ubuntu 18.04 system.
For further verification attached you can find an tar archive containing a data .mat file and a matlab script.
[IBB-MEEG.zip](https://github.com/fieldtrip/fieldtrip/files/4324224/IBB-MEEG.zip)

schoffelen:
Thanks Andreas,
We'll check whether we can confirm the issue with a more recent version of the code :)

schoffelen:
OK, confirmed.

schoffelen:
selection fails in ft_selectdata, line 651, where ft_channelselection is called

schoffelen:
I notice that the logic is not fully robust to begin with, because ft_selectdata makes an exclusive selection of going into the 'grad' or 'elec' mode, whereas both a 'grad' and 'elec' could be present in the data. 
Yet, in addition, apparently the automatic selection of the megref labels from a list of labels that contain, besides the ctf sensor names EEG channel labels is insufficiently robust in ft_channelselection

schoffelen:
Ah, I think that this is caused by the EEG channels in your data to have 'proper names' rather than the generic EEG123 labels.

in case of ctf data, fieldtrip identifies all channelnames that start with P as reference

schoffelen:
Ah, I think that this is caused by the EEG channels in your data to have 'proper names' rather than the generic EEG123 labels.

in case of ctf data, fieldtrip accepts all channels that start with 'P' as meg reference channel

schoffelen:
https://youtu.be/gg0I9rskXTk

schoffelen:
I see two things that can be improved:
* make ft_channelselection a bit smarter with a hard coded list of reference channel names
* make the handling of the selection in ft_selectdata more robust, i.e. treat both grad and elec as a separate case (operating on the subset of channels in the varargin.label that are of generic meg or eeg chantype.

@robertoostenveld, any quick thoughts?

robertoostenveld:
[...]

PR #1342, opened 2020-03-14
Selectdata

schoffelen:
This PR addresses issue #1334, which has been reported by @awollbrink

### DIFF
--- a/test/test_issue1334.m
+++ b/test/test_issue1334.m
@@ -1,0 +1,14 @@
+function test_issue1334
+
+% MEM 2gb
+% WALLTIME 00:10:00
+
+% DEPENDENCY ft_selectdata ft_channelselection
+
+load(dccnpath('/home/common/matlab/fieldtrip/data/test/test_issue1334.mat'));
+
+cfg = [];
+cfg.channel={'megref'}; % reference data
+
+dataRef = ft_preprocessing(cfg, data);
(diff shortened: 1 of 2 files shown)


schoffelen:
I thought I already linked it here, but there's an open bugzilla bug relating to the meta-issue at hand: http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2028

I think indeed it is worthwile to make ft_channelselection a bit more clever, in that it allows for key-value pairs that pre inform the function about the chantype or something like that.

Yet for now, to keep things moving forward, I think that the current PR is a good enough bandaid to roll out (given that it now at least also deals correctly with situations with both a grad and elec present in the data structure). Shall we merge this one, and follow the ft_channelselection issue up in a separate issue?

robertoostenveld:
agreed!

robertoostenveld:
this PR caused a regression error with `test_tutorial_multivariateanalysis`, due to ft_channelselection in that script searching for `MLO', but senstype being unknown. Consequently `findmlo` was true, but `labelmlo` was not defined (since that is only defined for senstype=ctf).

I will add a few lines to the `test_issue1342` script

robertoostenveld:
I resolved the error, it was enough to initialize all channel lists to an empty cell-array. Note that it was not introduced by this PR, but only surfaced due to this PR.

PR #1343, opened 2020-03-16
Update ft_chanunit.m

matsvanes:
input.chanunit might not exist (see line 118); replace by chanunit (which does exist).


robertoostenveld:
there is another `input.chanunit` on line 176, should that also be changed?

matsvanes:
You are right. Already changed it now.

PR #1346, opened 2020-03-17
Weightnorm

schoffelen:
new attempt for a clean PR


schoffelen:
The goal of this PR is to have more consistent implementation of the different beamformer formulations, both code-wise and in terms of supported functionality. The direct inspiration was to have similar (to LCMV) weight normalisation options in DICS/PCC. Note: SAM will not be touched for the time being.

inverse/beamformer_dics.m
```diff
@@ -108,6 +108,7 @@
 projectnoise   = istrue(projectnoise);
 fixedori       = istrue(fixedori);
 dofeedback     = ~strcmp(feedback, 'none');
```
robertoostenveld, line 110: things like dofeedback here are deprecated as feedback style. Rather we should use ft_debug, ft_info, ft_notice, ft_warning.

inverse/beamformer_dics.m
```diff
@@ -49,7 +50,7 @@
 % is specified, its orientation will be used and only the strength will
 % be fitted to the data.
 
-% Copyright (C) 2003-2008, Robert Oostenveld
+% Copyright (C) 2003-2020, Robert Oostenveld and Jan-Mathijs Schoffelen
```
robertoostenveld, line 53: 👍🏻

inverse/beamformer_dics.m
schoffelen: hmm, you are right. I merged your latest version into this PR just earlier, and thought that I caught all of this by addressing the conflicts. Always a joy, addressing conflicts.

inverse/beamformer_dics.m
```diff
@@ -81,28 +82,32 @@
 powmethod      = ft_getopt(varargin, 'powmethod');  % the default for this is set below
 realfilter     = ft_getopt(varargin, 'realfilter'); % the default for this is set below
 subspace       = ft_getopt(varargin, 'subspace');
+
 % these settings pertain to the forward model, the defaults are set in compute_leadfield
```
robertoostenveld, line 86: what about backproject, weight?

schoffelen: Yes, well there indeed I might have missed it while fixing the conflicts, but I didn't find it back in the code from master that I merged into the PR

robertoostenveld: you are right, these two are in the help but not in the actual code. My bad!

schoffelen: OK, I have now added weight and backproject as options for ft_compute_leadfield, using the same heuristic as in ft_prepare_leadfield, for passing on the options to the lower level function. While I was at it, I also added a similar heuristic for the options that go into ft_inv

inverse/beamformer_dics.m
schoffelen: dofeedback is nowhere used. also, the feedback option in the low-level functions relate to the old-style ft_progress. Let's not touch this right now, I removed the lines in the code where dofeedback was defined.

schoffelen:
argh, I pushed a change into this PR, which shouldn't be there. For now I keep it in. Note that it pertains to a speed improvement of unrelated code, I don't want to spend time right now to fix this.

schoffelen:
Apart from my stupid commit mistake, I'd appreciate an extra couple of eyes (depending on who rises to the challenge: @robertoostenveld @britta-wstnr @sarangnemo) for my latest changes into beamformer_lcmv, which were aimed at making a clearer distinction between a situation where precomputed filters are present (or not). Also, the flow of the code has changed a bit w.r.t. handling of the fixedori argument. Also, it now implements correctly the different weight normalisation options, which were originally contributed by @sarangnemo, but only with a scalar beamformer in mind. However, in the original implementation no check was performed whether or not the user had specified fixedori along. The vector versions of these weight normalisations are slightly different in their mathematics (as per Sekihara's book), which has now been fixed in the code. @britta-wstnr is doing something similar for MNE-python it seems. Anecdotally, the vector implementation of the unitnoisegain constraint is prone to numerical misbehaviour if one of the elements of the normalisation vector goes to zero (which probably (among others) happens when the radial dipole component coincides with one of the x/y/z axes).

inverse/beamformer_lcmv.m
```diff
@@ -191,143 +205,189 @@
...
+        else
+          % compute the matrix that is used for scaling of the filter's rows, as per eqn. 4.83
+          denom = pinv(lf' * invC * lf);
+          gamma = denom * (lf' * invC_squared * lf) * denom;
+          
+          % compute the spatial filter, as per eqn. 4.85
+          filt = diag(1./sqrt(diag(gamma))) * denom * lf' * invC;
+        end
+      case 'arraygain'
+        % filt*lf = ||lf||, applies to scalar leadfield, and to one of the possibilities of the vector version, eqn. 4.75
```
britta-wstnr, line 362: What if an already normalized lead field is supplied?

schoffelen: if that's the case, or when the on-the-fly computed leadfield has been normalised then it's a bit 'double up' (proper Dutch). Although the lf normalisation in ft_compute_leadfield is per a frobenius norm, yet here it's using another norm...

britta-wstnr: (I looked for a safety net for this, but maybe I overlooked it in the code)

schoffelen: Oh, and thanks for assigning yourself as a reviewer, I didn't manage to do so: for some reason github couldn't find you in the reviewer box

britta-wstnr: If a normalized lead field has a flag (`cfg.normalize = 'yes'` or something) it could be caught. I am just thinking that it is maybe not obvious for everyone that normalizing a lead field and an 'array gain beamformer' is the same, so this could be a not so uncommon outcome.

britta-wstnr: He, I think that happened automatically when I left this review! Might not be in the list because I have not contributed that much here (yet ...).

inverse/beamformer_lcmv.m
```diff
@@ -38,16 +38,21 @@
 %  'weightnorm'       = normalize the beamformer weights,                can be 'no', 'unitnoisegain' or 'nai'
```
britta-wstnr, line 38: I guess the list needs to be extended by `'arraygain'`

britta-wstnr: If you want to expose the option, that is - might be just for testing for now?

schoffelen: Yes, indeed.

inverse/beamformer_lcmv.m
```diff
@@ -191,143 +205,189 @@
...
+          gamma = denom * (lf' * invC_squared * lf) * denom;
+          
+          % compute the spatial filter, as per eqn. 4.85
+          filt = diag(1./sqrt(diag(gamma))) * denom * lf' * invC;
+        end
+      case 'arraygain'
+        % filt*lf = ||lf||, applies to scalar leadfield, and to one of the possibilities of the vector version, eqn. 4.75
+        lfn  = lf./norm(lf);
+        filt = pinv(lfn' * invC * lfn) * lfn' * invC; % S&N eqn. 4.09
+
```
britta-wstnr, line 365: Does this hold in the vector case? I tried a quick check in MATLAB and I don't think the constraints (eq. 4.63-4.65 in Sekihara & Nagarajan 2008) are fulfilled.

I checked the weightnorm vector case as well (both against book and quick example) and that looks good to me!

schoffelen: so your final verdict is that this is resolved, right?

britta-wstnr: The weight norm vector case yes.
About the array gain vector case I am not sure (that's where I referred to the equations).

I think we are using different terminology and I wasn't clear in my last comment, sorry! For me weight norm = unit noise gain, but here weight norm is the option that can _also_ carry array gain.

So I checked unit noise gain vector and that looks alright (both compared to formula and checking constraints on toy example) , but when I checked array gain vector as it is implemented above, I am not sure the constraints hold.

schoffelen: Ah, I see. Thanks for the clarification. Indeed weightnorm is used here as the option name, that implements the different flavours of weight normalisation. This naming was not invented by me. 
Anyway, the arraygain normalisation for the vector case is taken from equation 4.75

inverse/beamformer_lcmv.m
sarangnemo: Hi, glad to see this getting attention again! Tomorrow is a holiday in Denmark so I'll take a deeper look early next week.

inverse/beamformer_lcmv.m
britta-wstnr: Ah okay, I assumed the first option for vector beamformers. For this I can confirm that the _correct_ constraint (now eq. 4.74) holds for my toy example :) 
Maybe you can add the vector case equation number to the comment? (Right now it only has the scalar case eq. number -- might save future confusion over the different vector formulas, might also just be nitpicky.)

schoffelen: I like nitpicky, yet the formula number is already there: in line 362

britta-wstnr: Ha, I overlooked that with `% S&N eqn. 4.09` being the comment right behind the implementation - my bad!!

inverse/beamformer_lcmv.m
```diff
@@ -191,143 +207,189 @@
...
+ 
+        otherwise
+          % compute the leadfield for the optimal dipole orientation that maximizes spatial filter output
+          % subsequently the leadfield for only that dipole orientation will be used for the final filter computation
+          % filt = pinv(lf' * invC * lf) * lf' * invC;
+          % [u, s, v] = svd(real(filt * C * ctranspose(filt)));
+          % in this step the filter computation is not necessary, use the quick way to compute the voxel level covariance (cf. van Veen 1997)
+          [u, s, v] = svd(real(pinv(lf' * invC *lf)));
+          maxpowori = u(:,1);
+          eta = s(1,1)./s(2,2); % ratio between the first and second singular valu
```
britta-wstnr, line 324: This is new, no? Just curious whether this will perform more stable? (Taking the ratio).
Also, you are missing an "e" in the comment (more nitpicks!).

schoffelen: You mean the 'eta' variable? This is just a ratio that is not used as an activity estimate, but can be inspected to get a feel for how justified it is to assume a fixed orientation at the location of interest. The closer to 1, the more arbitrary this assumption is.

britta-wstnr: Ah, you renamed the actual orientation to `maxpowori`, I see. But interesting, you can basically plot that throughout the brain and see how good/bad it is to force optimal orientation?

inverse/beamformer_lcmv.m
```diff
@@ -191,143 +207,189 @@
...
+          filt = pinv(sqrt(noise * lf' * invC_squared * lf)) * lf' *invC; % based on S&N eqn. 4.08
+        else
+          ft_error('vector version of nai weight normalization is not implemented');
+        end
+      case 'unitnoisegain'
+        % filt*filt' = I
+        % Unit-noise gain minimum variance (aka Borgiotti-Kaplan) beamformer
+        % below equation is equivalent to following:  
+        % filt = pinv(lf' * invC * lf) * lf' * invC; 
+        % filt = filt/sqrt(filt*filt');
```
britta-wstnr, line 352: Curious here: have you confirmed this for the vector case? We tried in Python and could not get the output to be equivalent with normalizing the unit gain beamformer as described above (`filt_UNG = filt_UG / sqrt(filt_UG @ filt_UG.T)` ), we assume that it was something about the rotation of the output that was off, but have not looked in it further as we initially further pursued the `sqrtm()` approach over Sekihara.

inverse/beamformer_lcmv.m
schoffelen: yes indeed something like that. I have fixed the typo and also allow for an 'eta' computation for the other fixed orientation computation methods.

inverse/beamformer_lcmv.m
schoffelen: I don't know, haven't tried. Yet, I think that this inspired me to use the sqrtm. Note that currently, the vector case is dealt with if ~fixedori

britta-wstnr:
LGTM! 
Keeping notes: I compared the formulas to Sekihara and verified that the constraints hold for the newly implemented vector versions of unit-noise-gain and array-gain beamformers on a toy example.

schoffelen:
> LGTM!
> Keeping notes: I compared the formulas to Sekihara and verified that the constraints hold for the newly implemented vector versions of unit-noise-gain and array-gain beamformers on a toy example.

Hehe, I needed to look this one up. The urban dictionary lets me choose between (among others):
1) Let's get this merged
2) Literally garbage to me
3) Let's get the mayonnaise

I take the liberty of interpreting this as a '1'.

Thanks for your input! I will make the corresponding changes to the 'dics' code, and look into how to roll this out for pcc as well.

britta-wstnr:
Another one, which I refer to: "Looks good to me".

Feel free to ping me if you want more input on the DICS etc!

PR #1349, opened 2020-03-18
Fixes bug with linestyle cell array only plotting last style in array

### LINKED ISSUE #928: Bug with linestyle cell array only plots last style in array
I've found and solved this bug that makes plots only show the last linestyle in the cell array. I've edited 3 of the files to solve the bug but have no permission to push them to the repo. You can find the code here:

[correctedFunctions.zip](https://github.com/fieldtrip/fieldtrip/files/2759744/correctedFunctions.zip)

eurunuela:
Closes #928

### DIFF
--- a/plotting/ft_plot_vector.m
+++ b/plotting/ft_plot_vector.m
@@ -397,18 +397,18 @@ switch highlightstyle
     elseif ischar(color) && numel(color)==nline
       % plot each line with its own color
       for i=1:size(vdat,1)
-        h(i) = plot(hdat, vdat(i,:), style, 'LineWidth', linewidth, 'Color', color(i), 'markersize', markersize, 'markerfacecolor', markerfacecolor);
+        h(i) = plot(hdat, vdat(i,:), style{i}, 'LineWidth', linewidth, 'Color', color(i), 'markersize', markersize, 'markerfacecolor', markerfacecolor);
       end
     elseif isnumeric(color) && size(color,1)==nline
       % the color is specified as Nx3 matrix with RGB values for each line
       for i=1:size(vdat,1)
-        h(i) = plot(hdat, vdat(i,:), style, 'LineWidth', linewidth, 'Color', color(i,:), 'markersize', markersize, 'markerfacecolor', markerfacecolor);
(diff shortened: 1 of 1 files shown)


eurunuela:
Given that it's been over a year since I wrote the fix and fieldtrip has gone through some changes in the code, I believe this PR should be enough to fix the bug.

Could you please confirm @schoffelen ?

schoffelen:
hmmm, I am not sure whether this is going to work. By default, style will be a '-', i.e a string, indexing a cell-array is going to fail in that case. Also you may need to consider how the higher level functions are calling ft_plot_vector, is the 'style' argument always defined? If so, what's the datatype (string/cell)

robertoostenveld:
After having a quick look at this, I think that part of the problem might be due to the code being inconsistent. Some plotting options are applied to all lines, whereas others are applied to each line that is plotted. I think that for each of the options with"see MATLAB standard line properties" in the help, it should be clarified whether and/or how that option can be specified to deal with multiple lines that are drawn at once. 

Also it should be checked how it is being used in `ft_multiplotER` and `ft_singleplotER`.

@eurunuela could you make a short test script (e.g. with some constructed timelock structure) that demonstrates the problem (with the current code) and shows how the new code solves it? We can then extend that test script to ensure that nothing of the currently expected functionality breaks.

eurunuela:
I've just checked `ft_singleplotER` and `ft_multiplotER` and I believe the bug has already been fixed.

When I opened the issue the call to ft_plot_vector was like this: `'style', cfg.linestyle`. However, it is now done like this: `'style', cfg.linestyle{i}`, which IMO has already fixed the bug.

eurunuela:
See [ft_singleplotER](https://github.com/fieldtrip/fieldtrip/blob/7ca740cca0397d0dd58d9a1a453647778aee9538/ft_singleplotER.m#L463) for instance.

robertoostenveld:
I did indeed recently work on this (~2 months ago or so) in relation to plotting different data on top of each other - fNIRS oxytocin's and deoxyhemoglobin channels, and that in different conditions. 

If it now works for you, please close this PR

eurunuela:
As I said in #928, I don't use fieldtrip and wrote a quick fix for a colleague of mine as he kept seeing this bug.

I don't have any data to test with as I work on the development of methods for fMRI. So I'll close this PR as you were able to work on this with no issues.

PR #1350, opened 2020-03-18
[reproducescript] save state of random number generator

matsvanes:
I'm not sure why the first 2 commits are also included in the PR..


robertoostenveld:
looks harmless to me, and I'll merge. 

But it would be good to have a test script that would check randomseed to be used properly, i.e. that `ft_timelocksimulation` gives the same result when it should (same seed), and a different when not.

PR #1351, opened 2020-03-19
Issue1167

schoffelen:


schoffelen:
This is a placeholder for issue1167. The problematic example data has disappeared, and can be added to the function once it resurfaces.

PR #1356, opened 2020-03-21
updated 3d slicer

raintwoto:
I have implemented the ft_read_sens file and ft_filetype, however, I was stucked at 'ft_convert_units' in the 'ft_write_sens' since I did not understand the units there and why it must be mm in other types.


fileio/private/read_nihonkohden_hdr.m
```diff
@@ -0,0 +1,105 @@
+function [hdr] = read_nihonkohden_hdr(filename)
```
schoffelen: Why is this function part of the PR? Is this function referenced in ft_read_header and or ft_read_data?

schoffelen:
Why do you need to update ft_write_sens? Also, if the units are unknown, then I guess that they should remain unknown. ft_convert_units sometimes tries to do an educated guess as to what the units are. Yet, this does not depend on the file type / format.

fileio/private/read_nihonkohden_hdr.m
raintwoto: Sorry, that is from my last PR.....

raintwoto:
> Why do you need to update ft_write_sens? Also, if the units are unknown, then I guess that they should remain unknown. ft_convert_units sometimes tries to do an educated guess as to what the units are. Yet, this does not depend on the file type / format.

I think you are right. I don't think we need the ft_write_sens too.

schoffelen:
OK, thanks I will merge. Do you happen to have an example file that can be used for testing?

PR #1362, opened 2020-03-26
Fix #1361

### LINKED ISSUE #1361: [reproducescript] ft_prepare_sourcemodel can't handle char in cfg.sourcemodel
ft_prepare_sourcemodel can handle a cfg that contains the fields 'mri' and 'sourcemodel', but not when cfg.sourcemodel is a string (these are hard-coded to be mutually exclusive, line 254). This causes an error when cfg.sourcemodel is a file on disk, created by reproducescript.

matsvanes:
if cfg.sourcemodel is based on a file, it is not mutually exclusive with the other options for constructing a sourcemodel.

### DIFF
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -251,7 +251,7 @@ if basedonmni
 end
 
 % these are mutually exclusive
-if sum([basedonresolution basedongrid basedonpos basedonshape basedonmri basedonvol basedoncortex basedonmni basedonfile])~=1
+if sum([basedonresolution basedongrid basedonpos basedonshape basedonmri basedonvol basedoncortex basedonmni])~=1
   ft_error('incorrect cfg specification for constructing a sourcemodel');
 end
 
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -9,25 +9,26 @@ function [sourcemodel, cfg] = ft_prepare_sourcemodel(cfg)
 % where the details of the configuration structure determine how the source
 % model will be constructed.
 %
-% A source model can be constructed based on
-%   - regular 3D grid with explicit specification
-%   - regular 3D grid with specification of the resolution
-%   - regular 3D grid, based on segmented MRI, restricted to gray matter
-%   - regular 3D grid, based on a warped template grid, based on the MNI brain
-%   - surface mesh based on the brain surface from the volume conduction model
-%   - surface mesh based on the head surface from an external file
-%   - cortical sheet that was created in MNE or Freesurfer
-%   - using user-supplied source positions, which can be regular or irregular
+% % The different options for constructing a source model are
+%   cfg.mode =   'basedongrid'        regular 3D grid with explicit specification
+%                'basedonpos'         regular 3D grid with specification of the resolution
+%                'basedonshape'       surface mesh based on inward shifted head surface from external file
+%                'basedonmri'         regular 3D grid, based on segmented MRI, restricted to gray matter
+%                'basedonmni'         regular 3D grid, based on a warped template grid, based on the MNI brain
+%                'basedoncortex'      cortical sheet from external software such as Caret or FreeSurfer, can also be two separate hemispheres
+%                'basedonresolution'  regular 3D grid with specification of the resolution
+%                'basedonvol'         surface mesh based on inward shifted brain surface from volume conductor
+%                'basedonfile'        the sourcemodel should be read from file
 % The approach that will be used depends on the configuration options that
 % you specify.
 %
-% Configuration options for generating a regular 3D grid
+% The following parameters apply to cfg.mode='basedonresolution'
 %   cfg.xgrid      = vector (e.g. -20:1:20) or 'auto' (default = 'auto')
 %   cfg.ygrid      = vector (e.g. -20:1:20) or 'auto' (default = 'auto')
 %   cfg.zgrid      = vector (e.g.   0:1:20) or 'auto' (default = 'auto')
 %   cfg.resolution = number (e.g. 1 cm) for automatic grid generation
 %
-% Configuration options for predefined source positions
+% The following parameters apply to cfg.mode='basedonpos'
 %   cfg.sourcemodel.pos        = N*3 matrix with position of each source
 %   cfg.sourcemodel.inside     = N*1 vector with boolean value whether position is inside brain (optional)
 %   cfg.sourcemodel.dim        = [Nx Ny Nz] vector with dimensions in case of 3D grid (optional)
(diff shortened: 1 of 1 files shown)


robertoostenveld:
if cfg.sourcemodel is a file, then it would already have source positions. How would cfg.mri (pointing to a file) be used in the code, since that is used for segmenting and making yet other positions?

matsvanes:
I see, this fix might solve my particular case, but can cause other issues. 
But when cfg.sourcemodel is a file and therefore has source positions, it should not be mutually exclusive with cfg.isfield(cfg.sourcemodel, 'pos'), which is the case now.

robertoostenveld:
your suggestion is not a robust change. If someone in the future would turn around the section in the code where  "if basedonpos" is handled with the section where "if basedonfile" is handled, it will break (from a user's perspective, i.e. it will all of a sudden do the other thing). 

There really should be only one "basedon" option, which can be compared to `cfg.method` in other functions. Ideally it would be rewritten such that it would have an explicit cfg.method, rather than implicitly guessing it from the input settings. If there were a cfg.method, the main section of the code would be 

```
switch cfg.method
  case 'basedonpos'
    ...
  cased 'basedonfile'
    ...
```

and the two would also be mutually exclusive.

matsvanes:
@robertoostenveld I made this an option and replace al 
```
if basedon
...
end
```
parts by
```
switch cfg,.method
  case 'basedonX'
  case 'basedonY'
```
becasue all basedon options are mutually exclusive.

Note that now the order of determining which basedon is correct, is relevant. Another alternative is to check whether each basedon-option is true with more extensive if-statements:
`basedonY=isfield(...) && ~basedonX`

robertoostenveld:
could you try it with the changes that I just suggested. Mainly it is some reorganization of the documentation, making the format more consistent with e.g. ft_electrodeplacement. And I renamed `cfg.mode` into `cfg.method`

matsvanes:
It all works! :)

schoffelen:
can this be merged?

matsvanes:
Yes

ft_prepare_sourcemodel.m
```diff
@@ -166,96 +170,84 @@
...
+    cfg.method = 'basedonpos'; % using user-supplied positions, which can be regular or irregular
+  elseif ~isempty(cfg.headshape)
+    cfg.method = 'basedonshape'; % surface mesh based on inward shifted head surface from external file
+  elseif isfield(cfg, 'mri') && ~(isfield(cfg, 'warpmni') && istrue(cfg.warpmni))
+    cfg.method = 'basedonmri'; % regular 3D grid, based on segmented MRI, restricted to gray matter
+  elseif isfield(cfg, 'mri') &&  (isfield(cfg, 'warpmni') && istrue(cfg.warpmni))
+    cfg.method = 'basedonmni'; % regular 3D grid, based on warped MNI template
+  elseif isfield(cfg, 'headshape') && (iscell(cfg.headshape) || any(ft_filetype(cfg.headshape, {'neuromag_fif', 'freesurfer_triangle_binary', 'caret_surf', 'gifti'})))
+    cfg.method = 'basedoncortex'; % cortical sheet from external software such as Caret or FreeSurfer, can also be two separate hemispheres
+  elseif isfield(cfg, 'resolution') && ~basedonmri && ~basedonmni
```
schoffelen, line 186: I might be mistaken, but the 'basedonsomething' arguments here don't exist, so why are they included in the elseif statements? Do they need to be replaced by someting more informative?

matsvanes: You are correct, I missed this. I think these statements can be removed entirely, because the statement can't even be reached if cfg.method already is 'basedonmri', or 'basedonmni'
```suggestion
  elseif isfield(cfg, 'resolution')
```

schoffelen:
Just checking, have you grepped the functions in fieldtrip/test/ for the occurrence of ft_prepare_sourcemodel, and ran the corresponding functions? Just as a first check whether this will pass the test-batch at all? Or shall we do a 'god zegene de greep' and see what tonight's scheduled test run will give us?

matsvanes:
I have not done this. I just looked at the folder and saw that many of them use ft_prepare_sourcemodel. I don't know how to easily test them all, so I suggest we let the scheduled test run do its job tonight.

schoffelen:
it should not be too difficult. 

ft_test would be your friend in this case

schoffelen:
Well, I am going to merge, but if the sh* hits the fan, I know whom to blame :)

schoffelen:
FYI @matsvanes the PR was faulty, and caused an error in a range of test functions:
f5ff765/test_bug1298.txt:=== test_bug1298 FAILED in 6 seconds
f5ff765/test_bug131.txt:=== test_bug131 FAILED in 19 seconds
f5ff765/test_bug1806.txt:=== test_bug1806 FAILED in 14 seconds
f5ff765/test_bug1870.txt:=== test_bug1870 FAILED in 3 seconds
f5ff765/test_bug2148.txt:=== test_bug2148 FAILED in 34 seconds
f5ff765/test_bug367.txt:=== test_bug367 FAILED in 7 seconds
f5ff765/test_bug950.txt:=== test_bug950 FAILED in 8 seconds
f5ff765/test_example_megrealign.txt:=== test_example_megrealign FAILED in 8 seconds
f5ff765/test_ft_megplanar.txt:=== test_ft_megplanar FAILED in 9 seconds
f5ff765/test_ft_megrealign.txt:=== test_ft_megrealign FAILED in 6 seconds
f5ff765/test_tutorial_beamformingextended.txt:=== test_tutorial_beamformingextended FAILED in 36 seconds

I did not check all of them, but with a fix (which has already been pushed) the (ones I looked into) earlier failing test functions now seem to run through again.

robertoostenveld:
I had a more detailed look in the failing test scripts, they show this error

```
test_bug1298.txt:incorrect cfg specification for constructing a sourcemodel
test_bug131.txt:incorrect cfg specification for constructing a sourcemodel
test_bug1806.txt:incorrect cfg specification for constructing a sourcemodel
test_bug1870.txt:incorrect cfg specification for constructing a sourcemodel
test_bug2148.txt:incorrect cfg specification for constructing a sourcemodel
test_bug367.txt:incorrect cfg specification for constructing a sourcemodel
test_bug950.txt:incorrect cfg specification for constructing a sourcemodel
test_example_megrealign.txt:incorrect cfg specification for constructing a sourcemodel
test_ft_megplanar.txt:incorrect cfg specification for constructing a sourcemodel
test_ft_megrealign.txt:incorrect cfg specification for constructing a sourcemodel
```

robertoostenveld:
I will schedule the complete MATLAB batch once more, in about 2 hours it should be completed. I am not sure whether you can then access the logs on /home/mrphys/roboos/fieldtrip/dashboard/logs and then latest. If not, I will have a look later today. Or we just wait for the automatic email tomorrow morning.

robertoostenveld:
some first results show that not all errors are resolved

```
Error using ft_prepare_sourcemodel (line 684)
Reference to non-existent field 'pos'.
Error in test_tutorial_beamformingextended (line 90)
```

schoffelen:
I suggest to not be total 'curling parents' and have Mats have a look at it. :)

matsvanes:
> some first results show that not all errors are resolved
> 
> ```
> Error using ft_prepare_sourcemodel (line 684)
> Reference to non-existent field 'pos'.
> Error in test_tutorial_beamformingextended (line 90)
> ```

I can't access the other log files @robertoostenveld. Can you copy them to /home/common/temporary? I'm fixing the error above, but probably there are more?

schoffelen:
no, only the tutorial still fails. the rest of the failing functions were fixed with my fix

matsvanes:
I fixed this in #1370

Issue #1260, opened 2019-11-19
componentanalysis/rejectcomponent messes up the balance for combined EEG+MEG


robertoostenveld:
@mcvinding identified that the output of a pipeline with combined EEG+MEG only contained a grad, and that the EEG and MEG channels were mixed in that grad structure. 

The following test script already demonstrates that there is something wrong with the handling of the `elec` and `grad`. Probably we should consider the handling of `opto` as well.

```
cfg = [];
cfg.dataset = 'tactile_stim_raw_tsss_mc.fif';
cfg.trl = [1 1000 0];
data_all = ft_preprocessing(cfg);

cfg.channel = 'EEG*';
data_eeg = ft_preprocessing(cfg);

cfg.channel = 'MEG*';
data_meg = ft_preprocessing(cfg);

cfg.channel = 'MEGMAG';
data_mag = ft_preprocessing(cfg);

cfg.channel = 'MEGPLANAR';
data_planar = ft_preprocessing(cfg);


%%

cfg = [];
cfg.method = 'pca';
data_all_comp = ft_componentanalysis(cfg, data_all);
data_eeg_comp = ft_componentanalysis(cfg, data_eeg);
data_meg_comp = ft_componentanalysis(cfg, data_meg);
data_mag_comp = ft_componentanalysis(cfg, data_mag);
data_planar_comp = ft_componentanalysis(cfg, data_planar);

%%

cfg = [];
cfg.component = 1:10;
cfg.updatesens = 'yes';
data_all_comp_rej = ft_rejectcomponent(cfg, data_all_comp);
data_eeg_comp_rej = ft_rejectcomponent(cfg, data_eeg_comp);
data_meg_comp_rej = ft_rejectcomponent(cfg, data_meg_comp);
data_mag_comp_rej = ft_rejectcomponent(cfg, data_mag_comp);
data_planar_comp_rej = ft_rejectcomponent(cfg, data_planar_comp);

assert(isfield(data_all_comp_rej, 'elec'));
assert(isfield(data_all_comp_rej, 'grad'));

assert(isfield(data_eeg_comp_rej, 'elec'));
assert(isfield(data_eeg_comp_rej, 'grad'));

assert(isfield(data_meg_comp_rej, 'elec'));
assert(isfield(data_meg_comp_rej, 'grad'));

assert(isfield(data_mag_comp_rej, 'elec'));
assert(isfield(data_mag_comp_rej, 'grad'));

[...]

PR #1363, opened 2020-03-26
ENH - don't deal exclusively with grad  if also an elec is present

schoffelen:
This PR is a placeholder to deal with an issue reported by Mikkel.

Merging this in its current state will for sure crash test_ft_componentanalysis because the data on disk will be different from the computed data, due to the handling of the sensor array information


schoffelen:
this relates to issue #1260

Thinking it over, it is probably difficult to correctly deal with, because in principle a sequence of componentanalysis/rejectcomponent with updatesens is true with mixed EEG/MEG data yields grad/elec arrays with mixed in EEG/MEG data. Dealing with this correctly in the forward computation is probably not straightforward.

schoffelen:
However, having said this, I think that it is reasonable to expect correct functional code if a mixed channel data structure is subjected to componentanalysis with cfg.channel = 'MEG' or 'EEG'

schoffelen:
although this will end up being much more work than anticipated, I think I'd like to build in some better control (and support) for different chantypes in componentanalysis/rejectcomponent. My suggestion will be to have the same cfg option as in ft_denoise_prewhiten, cfg.split which determines how the unmixing is determined, which can be either across different channel types, or per specified channel type (where it makes sense to e.g. define cfg.split = {'MEG' 'EEG'} or perhaps also {'MEGMAG' MEGGRAD' 'EEG'} or so. The consequence of this would be that the updatesens results in internally consistent grad/elec arrays (because the off-diagonal blocks of unmixing coefficients will be zero by construction), which will not be the case when the unmixing is done across channel types. In the latter case, we could throw an error/warning if updatesens = true. @robertoostenveld @natmegsweden what do you think?

schoffelen:
Or perhaps we should constrain ourselves here to 'senstypes' rather than 'chantypes'

robertoostenveld:
once we are at it, I think we should do chantypes and not only senstypes. Splitting between meggrad and megmag makes sense, and consistency with ft_denoise_prewhiten would be valuable. I could imagine this being a suitable project for a hackathon session.

schoffelen:
Most of it is now in place, the only thing left to do - for consideration - is to deal with the elec/grad. Currently, the sensor information is discarded.

Issue #1365, opened 2020-03-31
ft_denoisePCA fails using per-trial correction mode


awollbrink:
the ft_denoisePCA function shows a faulty behavior when applying it to trial data and using the per-trial correction mode
using the following code snip with MEG data (273 channels in 'data' and 27 references in 'dataRef') the function returns different number of channels in 'dataCorr' and 'dataCorrPerTrial' depending on how cfg.pertrial was chosen ('no' -> 273 channels, 'yes' -> 27 channels)

I am using Fieldtrip revision 93dd0f61b on a Linux PC.

##### begin of code snip ########
%% denoise PCA
cfg=[];
cfg.channel={'MEG'};

cfg.pertrial= 'yes';
dataCorrPerTrial = ft_denoise_pca(cfg, data, dataRef);

cfg.pertrial= 'no';
dataCorr = ft_denoise_pca(cfg, data, dataRef);

schoffelen:
I cannot confirm this using the Subject01.ds data and the latest ft version on the github repo (master branch).

Please: 
1) verify if the problem persists using an up-to-date version of the code

If so:
2) provide a test function + data to reproduce your problem

awollbrink:
you can download test data from https://uni-muenster.sciebo.de/s/toh0edyZR046udJ
and use the following code

load('testDataDenoisePCA.mat', 'data', 'dataRef');

%% denoise PCA
cfg=[];
cfg.channel={'MEG'};

cfg.pertrial= 'yes';
dataCorrPerTrial = ft_denoise_pca(cfg, data, dataRef);

cfg.pertrial= 'no';
dataCorr = ft_denoise_pca(cfg, data, dataRef);

schoffelen:
Why do you input the reference data twice? 

The function is intended to work either
1. with a single data argument, containing both the channels-of-interest and the 'reference channels', then as specified as cfg.refchannel
2. with 2 data arguments, the first with the channels-of-interest and the second containing the reference channels.

Apparently, something goes allegedly (because you haven't reported back whether this persists in a more recent FT version) and silently wrong if the user by mistake includes a duplicate of the reference channels.

schoffelen:
Thanks for reporting this Andreas, I added a warning statement in the code (among others) to make users aware of the fact that the reference data was duplicated in the 2 data arguments, and made a small change that prevented fieldtrip to output the 'wrong' channels in that case.

robertoostenveld:
[...]

PR #1366, opened 2020-04-01
ENH - made function robust when pertrial=true and user inputs duplica…

schoffelen:
…ted reference channel data


schoffelen:
This addresses issue #1365, reported by Andreas Wollbrink, thanks for reporting

PR #1367, opened 2020-04-01
mi gcmi

schoffelen:


schoffelen:
I just managed to rebase my local branch (which branched off from fieldtrip master about 4 years ago). I want to merge this experimental functionality into the main branch, just to avoid all the work that went into this getting lost in the near future.

robertoostenveld:
The number of changed files (=9) is not so large. I propose that you copy the files with differences to a separate location, switch to master, and then copy them back. And then commit on master. That would keep the commit history clean.

robertoostenveld:
I think it is otherwise unlikely that (in case needed in the future) we would be able to make sense out of all of these 85 commits that went into it

schoffelen:
there should also be a way to squash these commits into a single one, right?

robertoostenveld:
yes, that should indeed also be possible

schoffelen:
shoot, I almost managed...

schoffelen:
I give up. Rebasing is just a beast. I think the commit history is now clean enough...

robertoostenveld:
5 commits and 8 files changed is not bad, right? I would be fine if you merge this. I did not check the code in detail, but take your word on it being ok.

schoffelen:
pinging @robince here. Don't feel obliged to start following/reading/reviewing all this, but my intention is to merge this PR relatively soon, just in order to not to have all the work done so far to be lost to oblivion. It took me quite some time with squashing/rebasing/etc. to get this poorly maintained branch unconflicted again. My idea would be to just 'release' it and work on improvements, if needed in a new branch. What I have implemented is to get a working version of mi/di/dfi in ft_connectivityanalysis (so not in an event-related setup, which is the direction into which you have been working, as per ft_statfun_gcmi). The idea is to be able to estimate from 'continuous data', mi (as a function of lag between the time series), di (lag vs. lag, source to target conditioned on target), dfi (feature to target, conditioned on source, lag vs. lag), and co-information ( MI('target',feature) + MI('source',feature) - MI('target+source',feature), lag vs. lag).

The computations are performed in ft_connectivity_mutualinformation. If all signals are univariate, the covariance is precomputed between all signal pairs in the data only once, which requires bypassing the mi_gg/cmi_ggg functions. In that case, the entropies etc. are computed in a subfunciton.

schoffelen:
Merging for now. Will work on documentation soon.

PR #1368, opened 2020-04-04
ENH: added electrode disc plotting utility to ft_plot_sens

StolkArjen:
This PR addresses a long anticipated utility for ft_plot_sens, i.e. plotting (ECoG) electrodes as discs. As discussed elsewhere (https://github.com/fieldtrip/fieldtrip/pull/1069/files), one of the challenges was to replace the previous method for generating the cylinder (the "disc") by matlab's built-in version. 

A remaining challenge still is to employ ft_warp_apply in the rotation (and translation) of the cylinder so that it is aligned with the headshape (@robertoostenveld: we'd need a transformation matrix based on the operations done in lines 458-480). This could make the code more concise and readable.

Use as:
ft_plot_sens(elec, 'elecshape', 'disc', 'headshape', hull)
   where headshape is used for aligning the electrode discs to. the headshape itself is not plotted.

Example figure, showing electrodes plotted alongside a pial surface:
Screen Shot 2020-04-04 at 10 40 18 PM

The overarching goal is to make 1) an example script for the wiki and 2) to add calls to ft_plot_sens from ft_sourceplot in order to allow plotting of functional data on the discs (with activity color-coded).

### DIFF
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -401,7 +401,7 @@ switch sensshape
     R = senssize/2; % convert coilsenssize from diameter to radius
     hold on
     for i=1:size(pos,1)
-      hs = surf(rsp*X+pos(i,1), R*Y+pos(i,2), R*Z+pos(i,3));
+      hs = surf(R*X+pos(i,1), R*Y+pos(i,2), R*Z+pos(i,3));
       set(hs, 'EdgeColor', edgecolor, 'FaceColor', facecolor, 'EdgeAlpha', edgealpha, 'FaceAlpha', facealpha);
     end
   
--- a/plotting/ft_plot_sens.m
+++ b/plotting/ft_plot_sens.m
@@ -25,8 +25,9 @@ function hs = ft_plot_sens(sens, varargin)
 % The following options apply to EEG electrodes
 %   'elec'            = true/false, plot each individual electrode (default = false)
 %   'orientation'     = true/false, plot a line for the orientation of each electrode (default = false)
-%   'elecshape'       = 'point', 'circle', 'square', or 'sphere' (default is automatic)
+%   'elecshape'       = 'point', 'circle', 'square', 'sphere', or 'disc' (default is automatic)
 %   'elecsize'        = diameter of the electrodes (default is automatic)
+%   'headshape'       = headshape, required for elecshape 'disc'
 % The following options apply to NIRS optodes
 %   'opto'            = true/false, plot each individual optode (default = false)
 %   'orientation'     = true/false, plot a line for the orientation of each optode (default = false)
@@ -96,6 +97,7 @@ coilsize        = ft_getopt(varargin, 'coilsize');  % default depends on the inp
 elec            = ft_getopt(varargin, 'elec', false);
 elecshape       = ft_getopt(varargin, 'elecshape'); % default depends on the input, see below
 elecsize        = ft_getopt(varargin, 'elecsize');  % default depends on the input, see below
+headshape       = ft_getopt(varargin, 'headshape', []); % for elecshape 'disc'
 % this is for NIRS optode arrays
 opto            = ft_getopt(varargin, 'opto', false);
 optoshape       = ft_getopt(varargin, 'optoshape'); % default depends on the input, see below
@@ -131,7 +133,7 @@ style           = ft_getopt(varargin, 'style');
 marker          = ft_getopt(varargin, 'marker', '.');
 
 % this is simply passed to ft_plot_mesh
-if strcmp(sensshape, 'sphere')
+if strcmp(sensshape, 'sphere') || strcmp(sensshape, 'disc')
   edgecolor     = ft_getopt(varargin, 'edgecolor', 'none');
 else
   edgecolor     = ft_getopt(varargin, 'edgecolor', 'k');
@@ -189,8 +191,8 @@ if isempty(senssize)
     case 'ctf275'
       senssize = 18;
     otherwise
-      if strcmp(sensshape, 'sphere')
-        senssize = 4; % assuming spheres are used for intracranial electrodes, diameter is about 4mm
+      if strcmp(sensshape, 'sphere') || strcmp(sensshape, 'disc')
+        senssize = 4; % assuming spheres/discs are used for intracranial electrodes, diameter is about 4mm
       elseif strcmp(sensshape, 'point')
         senssize = 30;
       else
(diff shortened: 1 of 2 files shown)


robertoostenveld:
the actual estimation of the sensor orientation (needed for MEG, but also for EEG) is not part of the current `ft_plot_sens` code. Rather, it uses the `coilori` and `elecori` that are present in the sensor structure. Wyy not use that here as well? Or why not estimate and support the electrode orientation estimation also for coilshape=circle?

A disk is basically a very flat cylinder, right? How do you specify the height and diameter of the cylinder? Can you include a test script that also shows how the proposed changes work for plotting sEEG electrodes (i.e. the contact surfaces along the shaft) as cylinders? 

Also NIRS optodes could be plotted as cylinders, not as flat as EEG electrodes but with an orientation orthogonal to the scalp surface. Can you also demonstrate how that would work?

The code for rotating a 3D representation (coilshape=circle or square) is already implemented. I don't think it makes sense to reimplement it differently for coilshape=disk.

StolkArjen:
> the actual estimation of the sensor orientation (needed for MEG, but also for EEG) is not part of the current ft_plot_sens code. Rather, it uses the coilori and elecori that are present in the sensor structure. Wyy not use that here as well?

The orientation is not fixed as it depends on what you use for a reference at the time of plotting, e.g. the pial surface, the cortical hull. Though, it's possible to outsource this to say a ft_estimate/determine_orientation function, and then to append an elecori field to the elec structure.

> A disk is basically a very flat cylinder, right? How do you specify the height and diameter of the cylinder? Can you include a test script that also shows how the proposed changes work for plotting sEEG electrodes (i.e. the contact surfaces along the shaft) as cylinders?

The thickness of the disc/cylinder is proportional to the radius, the latter of which is specified by elecsize/senssize. Indeed, plotting sEEG shafts is a future spin-off that would build on the current disc utility.

> Also NIRS optodes could be plotted as cylinders, not as flat as EEG electrodes but with an orientation orthogonal to the scalp surface. Can you also demonstrate how that would work?

Not with the current code, but that would definitely be another potential spin-off. 

> The code for rotating a 3D representation (coilshape=circle or square) is already implemented. I don't think it makes sense to reimplement it differently for coilshape=disk.

Agreed, and I think this is the main remaining challenge (besides maybe oursourcing elec.elecori estimation). Currently, angle estimation (of the norm vector and the cylinder axis) and rotation happens at lines 448-471. Ideally, this is reduced to a call to ft_warp_apply (it wasn't trivial to me).

StolkArjen:
Data for simulation:

http://www.arjenstolk.nl/hull.mat
http://www.arjenstolk.nl/elec.mat

ft_plot_sens(elec, 'elecshape', 'disc', 'headshape', hull);

robertoostenveld:
The code needs to be made more generic to be included in FieldTrip. It should not only deal with one sensor type (ECoG electrodes), but also sEEG and scalp EEG. Otherwise we will have extra work in the future and will have to make non-compatible changes to the code that will break scripts that rely on this.

robertoostenveld:
Can you add the following to fieldtrip/test?

```
function inspect_issue1368

% WALLTIME 00:10:00
% MEM 2gb
% DEPENDENCY ft_plot_sens

load(dccnpath('/home/common/matlab/fieldtrip/data/test/issue1368/hull.mat'))
load(dccnpath('/home/common/matlab/fieldtrip/data/test/issue1368/elec.mat'))

ft_plot_sens(elec, 'elecshape', 'disc', 'headshape', hull);
```

StolkArjen:
> The code needs to be made more generic to be included in FieldTrip. It should not only deal with one sensor type (ECoG electrodes), but also sEEG and scalp EEG. Otherwise we will have extra work in the future and will have to make non-compatible changes to the code that will break scripts that rely on this.

Agreed on generalizability. The same utility would be directly applicable to scalp-EEG. However, sEEG is a different beast, and would require further development (which I won't be able to complete in the near future). fNIRS, I don't know, but if you had a headshape for aligning the optodes with, that would work too.

The two milestones as I see it are: 1) implementing the local norm vector estimation in a dedicated function a la ft_determine_orientation, in order to add a .ori field to the sens structure (this is trivial for me, if you agree with), and 2) doing the rotation relatively to the rotation axis and angle of the cylinder with the local norm vector. This second part should be outsourced to ft_warp_apply, but defining the respective transformation matrix wasn't trivial to me.

robertoostenveld:
1) I moved the code for estimating the sensor orientation based on the headshape to the code which was already there, i.e. in the earlier part of the function. 

2) I simply reused the existing code for rotating/scaling/translating. 

Furthermore, I also updated the plotting of spheres, such that it also uses the same code (even though rotations don't make a difference here).

StolkArjen:
Clean solution. The part where you magically rotate the meshes is beyond my paygrade.

I quickly checked and can confirm it works as expected:
Screen Shot 2020-04-08 at 9 43 29 PM

This should make nice material for an example plotting script.

StolkArjen:
For historical reasons, this hellraiser brain constituted the very first attempt at the current disc plotting utility:

Screen Shot 2018-12-05 at 11 26 10 PM

robertoostenveld:
adding cylinders (for sEEG contacts) to the plotting function would be easy, estimating their direction (represented in `elecori`) would computationally not be hard: simply demean all positions of all  contacts on a single shaft, do an svd, and then take u(:,1). 

But it is not so easy to figure out where that orientation estimation would have to be done. 

Note that for the  disks it is now offset in one direction (away from the surface), whereas for sEEG contacts it should be offset in two directions (i.e. the position of the electrode is the middle, not one of the sides).

robertoostenveld:
> Note that for the disks it is now offset in one direction 

that is how it should be ... you may actually want to check this: increase the disk thickness and it should be easy to see

StolkArjen:
> that is how it should be ... you may actually want to check this: increase the disk thickness and it should be easy to see

Indeed, I had confirmed this already for myself.  

> adding cylinders (for sEEG contacts) to the plotting function would be easy, estimating their direction (represented in elecori) would computationally not be hard: simply demean all positions of all contacts on a single shaft, do an svd, and then take u(:,1).
> 
> But it is not so easy to figure out where that orientation estimation would have to be done.

Another complication is that the shafts are flexible and sometimes do end up bent following implantation. This wasn't an issue for the rotating brain movie (I hand-coded the depth shaft) on https://knightlab.neuro.berkeley.edu, which gives an idea of what I have in mind for implementing in ft_plot_sens.

PR #1370, opened 2020-04-06
deal with incorrect specification of cfg.method

matsvanes:
If the method is unsopported, remove it from the cfg and determine the method automatically, based on the rest of the input.

### DIFF
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -169,6 +169,14 @@ if isfield(cfg, 'resolution') && isfield(cfg, 'zgrid') && ~ischar(cfg.zgrid)
   ft_error('You cannot specify cfg.resolution and an explicit cfg.zgrid simultaneously');
 end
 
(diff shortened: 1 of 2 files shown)


schoffelen:
Thanks, I replaced your suggestions by a call to ft_checkconfig, with 'allowedval' functionality. Also, I initialize the cfg.method to be []; 
Can you pull this locally, and quickly check a few test functions that depend on ft_prepare_sourcemodel, to confirm that it's OK now?

matsvanes:
for one, the test function that my implementation was fixing, now gives an error. In my implementation a warning is given when a option is specified that is not allowed; now it gives an error..

schoffelen:
So, why would this be a problem? This suggests an incorrect specification of the method in the test function, which is due to the cfg not being re-created empty after the call to ft_prepare_headmodel with cfg.method = 'singleshell';

I suggest to fix this in the test function. 

I agree that ft_checkconfig might be a bit 'ijverig', but I think that throwing a warning and just emptying the method is hiding away the user error.

PR #1371, opened 2020-04-06
[reproducescript] write 'figure;' before plotting functions (fix #1369)

### LINKED ISSUE #1369: [reproducescript] error saving figure when not started with empty figure
When plotting functions are run with reproducescript enabled, the corresponding line in the standerdized script (script.m) can produce an error (something about a corrupt file). This can occur when the plotting function plots in a figure that was not cleared before the function was ran.

matsvanes:


utilities/private/isplottingfunction.m
```diff
@@ -0,0 +1,25 @@
...
+% is used for the cfg.reproducescript functionality. It compares the input
+% function name with the list of known FieldTrip plotting functions and
+% returns 1 if it is a plotting function, and 0 otherwise.
+
+% determine FieldTrip path
+ftpath = fileparts(which('ft_defaults'));
+
+functionlist = [];
+% browse the main FieldTrip directory
+functionlist = [functionlist; dir([ftpath,'/', '*plot*.m'])];
```
schoffelen: these types of commands are not guaranteed to be robust for all operating systems. Windows has a '\' rather than a '/'. It could be that more recent versions of matlab know how to deal with this, but more robust would be to use fullfile, rather than the [somestring, '/', someotherstring] syntax

schoffelen: Note that I took the liberty of making this change directly into this PR

schoffelen:
can this be merged?

utilities/private/isplottingfunction.m
robertoostenveld: there is also the MATLAB function `filesep`

schoffelen: I know, but that leads to ugly code

schoffelen: don't forget strcat

PR #1373, opened 2020-04-09
read hdr information from nirs.json files

helenacockx:
I've updated the events_tsv.m function such that headers of nirs.json files will also be recognized.


robertoostenveld:
do you also want to add the update to `readtable` such that it only uses tabs and no spaces as separators?

helenacockx:
I thought to overcome this by deleting the spaces in my own table. If we want to update it such that it only uses tabs as separators we should either have a look at the detectImportOptions or change the opts.delimiter manually.

PR #1376, opened 2020-04-17
improve ft_sourceanalysis flexbility to work with pre-computed leadfields

vitpia:
The current version of ft_sourcenanalysis recomputes the leadfields (lines 373 to 403). It ignores the fact that the input may already contain pre-computed leadfields. 
Suggested changes would best be done in "prepare_headmodel" (called in line 354), which could prepare more things than what it currently does (according to @robertoostenveld).

@mcpiastra and I will prepare a test script. 
@schoffelen, do you know from the top of your head which test script is particularly good and elaborate, so we can use as a basis to introduce the types of inconsistencies we need to test?


schoffelen:
Hi Vitoria, perhaps best to close this PR and open a new one after you created a new branch locally (and pushed it to your fork), because strictly you don't want to merge the master branch into the release branch.

schoffelen:
Also, I never have this problem, so I don't understand well why you happen to run into this. Could it be that you use cfg.grid, rather than cfg.sourcemodel?

vitpia:
> 
> 
> Also, I never have this problem, so I don't understand well why you happen to run into this. Could it be that you use cfg.grid, rather than cfg.sourcemodel?

(I was quite sure I had done this the wrong way... Thanks! I still need to figure out how to do it first.)
For now: No, I'm using cfg.sourcemodel. The issue is that we compute FEM and the leadfields outside of FT and once I run ft_sourceanalysis, it enters the else in line 387 and ft_prepare_sourcemodel is called again. Robert, Maria Carla and I discussed it and he suggested I open this issue.

schoffelen:
So, actually it's a matter to finally get robust support for FEM computations within FT then :)?

robertoostenveld:
hi @vitpia, thanks for this first attempt! There are two technical git things to resolve. 

1) you started off from an older version of your vitpia:master branch, you better make a separate branch for your specific changes 
2) you sent the pull request against fieldtrip:release, that should be fieldtrip:master

Please see http://www.fieldtriptoolbox.org/development/git/ and try once more

PR #1377, opened 2020-04-17
improve ft_sourceanalysis flexbility to work with pre-computed leadfields

vitpia:
The current version of ft_sourcenanalysis recomputes the leadfields (lines 373 to 403). 
It ignores the fact that the input may already contain pre-computed leadfields. So for any leadfield computed in a non-standard way (for example FEM computed externally, but the issue is not unique to FEM).
Suggested changes would best be done in "prepare_headmodel", which could have everything needed set, before it goes on with the inverse solution (suggestion by @robertoostenveld).

Test script will be updated here soon.
@mcpiastra and V.Piai will prepare a test script.


schoffelen:
allow me to chime in here, to try and push this forward. I see a lot of finetuning to the test script. What is now the status and the plan of action? Is the test script functional? Define functional: does it demonstrate the current limitations? If so, what are the next steps to be taken?

mcpiastra:
@vitpia and I worked on the test script and it would be great to have feedback on it. We highlighted the current limitations by printing error messages. Please let us know what else we'd need to adjust further. Many thanks!

robertoostenveld:
One error mentioned in the code related to this

```
cfg.rawtrial = 'yes';
cfg.sourcemodel = grideeg;
cfg.sourcemodel.filter = sourcedics3d1.avg.filter;
ft_sourceanalysis(cfg, EEG_freq);
```

The user-error here is that `cfg.sourcemodel` in first instance is created with 92 channels, and then the 64 channel filter is added. So it is internally not consistent. A correct way of doing it is 

```
cfg.rawtrial = 'yes';
cfg.sourcemodel.pos = sourcedics3d1.pos; 
cfg.sourcemodel.inside = sourcedics3d1.inside; 
cfg.sourcemodel.label = sourcedics3d1.avg.label; 
cfg.sourcemodel.filter = sourcedics3d1.avg.filter;
ft_sourceanalysis(cfg, EEG_freq);
```

The `avg` sub-structure layer in between is supposed to disappear at a certain moment, all fields should ideally be represented at the same level.

robertoostenveld:
note that I am currently working on the code, so please don't change it yourself.

robertoostenveld:
The second error noticed in the code relates to precomputed leadfields and EEG dipole fitting. That causes an error because of the channel mismatch between the data (64 channel) and the electrodes and leadfields (92 channels). That error is not detected in the MEG case since there the precomputed leadfields match the data. But it would also happen there if one of the channels were disabled.

Issue #1380, opened 2020-04-19
Typo line 214 in Preproc - Median Filter


pcardone95:
Hi,

I want to report a bug for the [**preproc**](https://github.com/fieldtrip/fieldtrip/blob/master/private/preproc.m) script. On line 214, it checks whether a value was assigned to cfg.medianfilter with [**ft_getop**](https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_getopt.m), as shown in the following

https://github.com/fieldtrip/fieldtrip/blob/d5740e3ee6de8913246e735ed4de28c8ce1e1800/private/preproc.m#L214

However, as you can see the _"key"_ value in [**ft_getop**](https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_getopt.m) has typo (an added space, 'medianfilter ' instead of 'medianfilter'), resulting in default assignment of "no" to cfg.medianfilter, independently to whether it was assigned or not.

Thank you very much for your time!

robertoostenveld:
thanks, well spotted!

Although it is easy to fix for me, could you please try to fix it yourself and send us a pull request?  On http://www.fieldtriptoolbox.org/development/git/ we have a tutorial for that. 

What you can also do is use the Github interface: if you go to https://github.com/fieldtrip/fieldtrip/blob/master/private/preproc.m and then click on the pen symbol in the upper right

![Screenshot 2020-04-19 at 10 07 50](https://user-images.githubusercontent.com/899043/79682818-a614c000-8225-11ea-9b3a-130cde869193.png)
 
You can then edit the file directly in your browser. Github will subsequently guide you through the process of forking it (so that you have your own copy of the repo), committing the change, and suggesting the change in the form of a pull request.

pcardone95:
Thank you very much for the suggestion, if I had known before I would have communicated it directly via the Github - my bad. 
Have a nice day! :)

PR #1381, opened 2020-04-19
Typo line 214

pcardone95:
There was a typo on line 214 with medianfilter.  ft_getop key was set to "medianfilter " (with space) instead of "medianfilter"

Here is the reported [issue](https://github.com/fieldtrip/fieldtrip/issues/1380)


robertoostenveld:
thanks! We are looking forward to your future contributions.

PR #1392, opened 2020-05-04
Fixing #1391 (rejectvisual rejecting wrong channels/trials)

### LINKED ISSUE #1391: ft_rejectvisual with summary mode rejects wrong channel if cfg.channel is set
See the following snippet. When executing `ft_rejectvisual`, select the second channel in the display (label 'c', values 3). I'd expect `datout` to have channel labels 'a' and 'd' ('b' no longer there because it was not selected in `cfg.channel`, 'c' no longer there because it was toggled in the GUI). Instead, channels 'a' and 'c' are present! Thus channel d was removed.

```
%% create data

data = [];
data.label = {'a' 'b' 'c' 'd'}';

% ensure the channels are distinguishable by their data (helps in the
% GUI)
dat = ones(4,1000);
dat(2,:) = 2*dat(2,:);
dat(3,:) = 3*dat(3,:);
dat(4,:) = 4*dat(4,:);

data.trial = {dat dat};
data.time = {1:1000 1:1000};

%% check it

cfg = [];
cfg.channel = {'a' 'c' 'd'};
cfg.method = 'summary';
cfg.metric = 'max';
datout = ft_rejectvisual(cfg, data);

% Now I select the second channel to reject, which corresponds to channel
% label 'c' (due to the selection above). Weirdly, datout now has channels
% a and c, so channel d was removed!
```

Spaak:
Fixes #1391


Spaak:
Note this PR also updates the x/ylabels in the panels of `rejectvisual_summary` to make explicit that they refer to linear indices in the selection of channels/trials. (I.e. if `cfg.channel`/`cfg.trials` is used, in general these indices will not match the order of the original data anymore.) Not directly related to the issue being fixed here, but might be helpful nonetheless.

robertoostenveld:
I reviewed the code and tested the user interface with the script. It now all works, as in: "the output of method=summary is not incorrect/unexpected anymore". 

However, I do find the behavior still confusing with the two levels of selection interacting. This is not how I originally designed the function, but apparently this is what it became after the involvement of various people. For reference, this is the list that contributed to this function according to `git log`: Diego, Eelke, Jan-Mathijs, Jim, Johanna, Jorn, Keith, Lau, Robert, Roemer, Saskia, Tim. 

Let me ask for opinions (also from @schoffelen): 

Originally I had in mind that this function is to be used for selecting a few trials and/or channels to be removed from a large number. It was possible to preselect them using, but there was only one selection. Then it was extended to allow the selection to be optionally kept, but there was still one selection. 

Now it appears as if there are two selections, one for the channels/trials to be shown and the other for the channels/trials to be removed. I suspect this was introduced to deal with cases where large numbers of channels had to be de-selected in advance (i.e. repeated application to remove trials on the basis of variance in neuromag planar and mag channels). Rather than using ft_selectdata (2x) and ft_rejectvisual (2x) and ft_appenddata, the neuromag use case (hiding 204 while showing 102) is now the dominant one. @Spaak implemented this correctly by gving the axes elaborate labels as you see here.
Screenshot 2020-05-12 at 14 49 31

How would you feel about reverting the whole selection-in-selection and simply treating cfg.channel and cfg.trials in the same was as the user-selection in the gui?

robertoostenveld:
with Eelke's version and

```
cfg = [];
cfg.trials = [1 2 4] ; % 'all';
cfg.channel = [1 2 4] ; % 'all';
cfg.metric = 'max';
cfg.method = 'channel';
datout = ft_rejectvisual(cfg, data); % don't make any additional selection, just quit
```

I do get a bug:

Screenshot 2020-05-12 at 14 53 30

The last trial displayed is trial 3 (in the selection) and trial 4 (in the data).

robertoostenveld:
If we were to change this back to "selections-in-the-cfg work just like selections-in-the-gui" then we probably would have to update some examples/tutorials. And users would have to change their scripts...

Spaak:
> I do get a bug:
> The last trial displayed is trial 3 (in the selection) and trial 4 (in the data).
This is the same 'bug' as the one I 'fixed' by updating the labels in summary mode, correct? I.e. if the label were to read "trial 003 (in selection)" then it would be consistent behaviour. Or am I missing an additional bug?

robertoostenveld:
no, you are right.

Spaak:
> Now it appears as if there are two selections

I think there is now still only one selection. If you specify `cfg.channel` and/or `cfg.trials`, then the output data will only have those channels/trials that were selected. So, removing them either by selecting in the interface or by omitting them from the `cfg` has the same effect: they won't end up in the output. But I might be misunderstanding the general concern here.

In any case I agree that this functionality is potentially (very) confusing. If we want to fix this confusion, one good approach might be to remove the `cfg.channel` and `cfg.trials` options altogether.

robertoostenveld:
I suspect that some people right now don't feel comfortable that channels that they want to keep (e.g. 204 planars, when checking the magnetometer channels) are called "BAD".

If we were to rename GOOD/BAD into KEEP/REMOVE or SELECT/UNSELECT, perhaps people would already consider the use of the function differently.

robertoostenveld:
> In any case I agree that this functionality is potentially (very) confusing. If we want to fix this confusion, one good approach might be to remove the cfg.channel and cfg.trials options altogether.

That is very drastic. But what if it SELECT/UNSELECT is used consistently in the cfg settings, in the GUI, and in the feedback printed at the end?

robertoostenveld:
I recall in the past that I explained that there are two ways to use the function. 
1. ft_selectdata (2x) and ft_rejectvisual (2x) and ft_appenddata
2. ft_rejectvisual with keeptrial=yes and keepchannel=yes (repeatedly, with different selections) and then ft_rejectartifact (for the bad trials)  plus ft_selectdata (with a setdiff, for the channels)

robertoostenveld:
showing unselected channels/trials as white rather than blue (default for nan) would also be an improvement.

Spaak:
> I recall in the past that I explained that there are two ways to use the function.

And I can add that I usually do 3. call ft_rejectvisual a few times (with different filter options (for e.g. muscles) and channels (for e.g. eyetracker)), with the default keep*=no, then at the end store `dataout.label` and `dataout.trialinfo(:,1)` (containing unique trial IDs in my case) so I can always apply the "keep" selection using `ft_selectdata` on the original data.

> But what if it SELECT/UNSELECT is used consistently in the cfg settings, in the GUI, and in the feedback printed at the end?

Yes, I agree that would be a good change.

In addition to that, I think it would be desirable to be able to have the GUI refer to trials by their indices as they occur in the data. For channels the analogous functionality would be easy to implement (simply only use channel *labels* and not indices). For trials it might be a bit trickier to do in a neat way, given the implementation with only `tmpdata` passed on to the lower level function.

robertoostenveld:
I just rediscovered the option `cfg.viewmode = 'hide';` which hides deselcted channels. It was apparently implemented by @Horschig in 2011, and updated by me in 2014.

I think it does exactly what we are trying to solve all the time. It even hides the indices from the trial and channel axes when they would be interpreted as inconsistent (aka as "in selection").

robertoostenveld:
> I just rediscovered the option `cfg.viewmode = 'hide';` which hides deselcted channels. It was apparently implemented by @Horschig in 2011, and updated by me in 2014.

it is an "Undocumented option".... ohhh too bad! Had we documented it, we could have saved quite some headaches.

Spaak:
> 

> I just rediscovered the option `cfg.viewmode = 'hide';` which hides deselcted channels. It was apparently implemented by @Horschig in 2011, and updated by me in 2014.
> 
> I think it does exactly what we are trying to solve all the time. It even hides the indices from the trial and channel axes when they would be interpreted as inconsistent (aka as "in selection").

Hmm but `rejectvisual_summary` now has no way of knowing which channels/trials were deselected based on `cfg` options, as it doesn't have the original data anymore, only `tmpdata` for which those `cfg`-options have been applied. (But I don't have Matlab running at the moment, so maybe I'm mistaken!)

robertoostenveld:
Shall we plan some time on zoom to look at this together?

schoffelen:
Spuit 11 staat standby, maar wacht even de ontwikkelingen af, OK?

Spaak:
> Spuit 11 staat standby, maar wacht even de ontwikkelingen af, OK?

Prima :) Robert en ik hebben een Zoom vandaag om 13u, als je mee wil doen laat weten en dan mail ik je de link.

schoffelen:
Mail voor de zekerheid maar, al heb ik zelf om 13.00 al een skype afspraak zijn. Ik kan evenwel altijd later nog inhaken, zoals een goede spuit 11 betaamt.

robertoostenveld:
het was een kort overleg, met als uitkomst dat we idd het oude gedrag (dus eenduidige selectie in de GUI en in de cfg) terug willen. Ik ben ondertussen de wijzigingen aan het doorvoeren.

schoffelen:
As long as the two of you agree, I'm fine. Does this mean that Jörn's undocumented behavior is going to be retracted?

schoffelen:
OK, as far as I know, Robert's PR to this PR needs to be merged, and then this one can be merged, right?

Spaak:
Yes, you're right! I was expecting to get an email when Robert would file his PR, but apparently won't get one. Only prompted by your message do I see that the PR is ready. Will get to it later today.

Spaak:
I'll merge Robert's PR into this one, but then I'll add some minor changes myself before marking this one as ready to merge.

Spaak:
(see https://github.com/Spaak/fieldtrip/pull/1)

Spaak:
Can be reviewed and merged now, I think.

PR #1394, opened 2020-05-05
updated code to define default spmversion to be spm12, and some clean…

schoffelen:
…ups to ft_volumewrite

### DIFF
--- a/fileio/ft_write_mri.m
+++ b/fileio/ft_write_mri.m
@@ -45,7 +45,7 @@ function [V] = ft_write_mri(filename, dat, varargin)
 
 % get the options
 transform     = ft_getopt(varargin, 'transform', eye(4));
-spmversion    = ft_getopt(varargin, 'spmversion', 'SPM8');
+spmversion    = ft_getopt(varargin, 'spmversion', 'spm12');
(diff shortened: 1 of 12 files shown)


schoffelen:
Irritant! For some reason Atom has decided to replace all original indented lines with empty lines, probably this has something to do with the automatic tab-functionality. Besides from looking ugly in the file-comparison mode, it breaks the cell-mode behavior for matlab I guess. Although the latter is not so problematic, I will look into how this can be addressed in future commits.

schoffelen:
It looks as if the 'whitespace' package in Atom is causing the issue. It is suggested to disable the package altogether. Alternatively, it seems that there's an option to skip lines that are all empty (apart from the trailing spaces/tabs). I'll try the latter thing first.

schoffelen:
Plan of action:
1) merge this with the master branch (+the website changes in its own PR)
2) wait and see how the dashboard behaves
3) try and fix some remaining (open) issues on the dashboard, perhaps requiring input from @robertoostenveld because some failing test functions seem to come and go, and are not likely to be code dependent (e.g. test_ft_regressconfound), to ensure a new release-version
4) communicate the change in default behaviour in an e-mail to the discussion list.

PR #1396, opened 2020-05-07
fix some control statements if cfg.events is a matrix, trl variable i…

KristijanArmeni:
…s not declared, use cfg.events instead. Fix feedback typo in copy_ctf_files.

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1415,7 +1415,7 @@ if need_events_tsv
     else
       events_tsv = event2table([], cfg.events);
     end
-  elseif ismatrix(cfg.events) && ~isempty(cfg.events) && numel(fieldnames(cfg.events))>0
+  elseif ismatrix(cfg.events) && ~isempty(cfg.events)
     % it is a "trl" matrix formatted as numeric array, convert it to an events table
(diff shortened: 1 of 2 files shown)


data2bids.m
```diff
@@ -1415,15 +1415,15 @@
...
+  elseif ismatrix(cfg.events) && ~isempty(cfg.events)
     % it is a "trl" matrix formatted as numeric array, convert it to an events table
     begsample = cfg.events(:,1);
     endsample = cfg.events(:,2);
     offset    = cfg.events(:,3); % this is not used for the events.tsv
     if any(offset~=0)
       ft_warning('the offset in the trl matrix is ignored');
     end
-    if size(trl,2)>3
+    if size(cfg.events, 2)>3
```
schoffelen, line 1426: why is the variable name changed here?

KristijanArmeni: `trl` variable is not declared upstream, so the code breaks here. To my checking, `cfg.events` field is used throughout and there are control statments checking for whether `cfg.events` contains the trl matrix or whether event information in there is represented as a struct or as a table instead.

PR #1397, opened 2020-05-08
use isnumeric() to test if cfg.events is a trl matrix

KristijanArmeni:
Relates to #1396 and #1398. 

If `cfg.events` is a struct with no fields, `ismatrix(cfg.events) && ~isempty(cfg.events)` in `l.1418` evaluates to true and leads the script into the chunk which assumes `cfg.events` to be a matrix.

The proposal here is to use `isnumeric()` to test for matrix, as for some reason `ismatrix(struct())` evaluates to true. ("everything is a matrix" in a matrix laboratory I guess).

Note, the test runs OK past the problematic line for me, but `test_data2bids` doesn't go through all the way due to write permissions.

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1415,7 +1415,7 @@ if need_events_tsv
     else
       events_tsv = event2table([], cfg.events);
     end
-  elseif ismatrix(cfg.events) && ~isempty(cfg.events)
+  elseif isnumeric(cfg.events) && ~isempty(cfg.events)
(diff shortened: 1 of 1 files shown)


robertoostenveld:
looks good to me.

I guess that `ismatrix` returns true whenever the input is an array (like a numeric-array, struct-array, and cell-array). That is indeed not what was meant in the code.

Issue #1387, opened 2020-04-24
ft_read_event breaks when thresholding events on multiple channels


iandol:
I am loading an EDF file (Test_4.edf attached) that has two EEG channels (1:2), and 8 analog continuous trigger channels (3:8) -- the 8 event triggers are 8 bits that I want to decode to strobed words so I can define my trials. I use a threshold set to `4*nanmedian`.

[Test_4.zip](https://github.com/fieldtrip/fieldtrip/files/4528504/Test_4.zip)

I thought I could do this all in one go with ft_read_event:

```matlab
fileName = 'Test_4.edf';
triggerchannels = 3:8;
cfg				= [];
cfg.dataset 	= fileName;
cfg.header		= ft_read_header(cfg.dataset);
labels			= cfg.header.label(3:end);
data_events		= ft_read_event(cfg.dataset,'header',cfg.header,...
	'detectflank','up','chanindx', triggerchannels, 'threshold', '4*nanmedian');
```

The result is incorrect (see the markers I draw for each event channel don't align with channels, in fact markers from one channel are found on another and some events are unmarked):
![test](https://user-images.githubusercontent.com/153905/80207649-1c1f8980-8661-11ea-97b6-3c575ffd4261.png)
full code to reproduce: https://gist.github.com/iandol/118a3aeb6820ab8d7edd48fe2791774e

Step debugging through I found that the problem is on line 181 of read_trigger.m:

```matlab
if ~isempty(threshold)
  % the trigger channels contain an analog (and hence noisy) TTL signal and should be thresholded
  if ischar(threshold) % evaluate string (e.g., threshold = 'nanmedian')
    threshold = eval([threshold '(dat)']);
  end
  % discretize the signal
  dat(dat<threshold) = 0;
  dat(dat>=threshold) = 1;
end
```

With a single channel there is no problem with this eval: dat is 1xN then `eval(['4*nanmean(dat)'])` gives the correct single threshold value, but when dat is more than one channel (M channels x N) then this breaks (wrong dimension used); simple test code shows this:

```
>> dat = [0 0 0 0 10 0]

dat =

     0     0     0     0    10     0

>> eval(['4*nanmedian' '(dat)'])

ans =

     0 


>> dat = [0 0 0 0 10 0;0 0 10 0 0 0]

dat =

     0     0     0     0    10     0
     0     0    10     0     0     0

>> eval(['4*nanmedian' '(dat)'])

ans =

     0     0    10     0    10     0 
```

[...]

PR #1403, opened 2020-05-18
Threshold channel fix for #1387

iandol:
This is the pull request for #1387 

I am not exactly sure how I should format the test, I read https://github.com/fieldtrip/dashboard, is it enough to basically throw an `error` using an `assert`?

Where should I put the data ( data was shared here: https://github.com/fieldtrip/fieldtrip/files/4528504/Test_4.zip ), I'll rewrite the path to match, and I also don't know what to add for walltime / mem etc.

### DIFF
--- a/test/test_issue1387.m
+++ b/test/test_issue1387.m
@@ -1,7 +1,7 @@
 function test_issue1387
 
 % WALLTIME 00:00:10
-% MEM 40mb
+% MEM 1gb
 % DEPENDENCY ft_read_event ft_read_header ft_preprocessing
 
 %%
--- a/test/test_issue1387.m
+++ b/test/test_issue1387.m
@@ -16,9 +16,9 @@ data           = ft_preprocessing(cfg);
 
 idx            = cellfun(@(x)strcmpi(x,labels{1}),{data_events.type});
 event.label	   = labels{1};
-event.idx		= find(idx==1);
-event.evnt		= data_events(event.idx);
-event.samples	= [event.evnt.sample];
+event.idx      = find(idx==1);
+event.evnt     = data_events(event.idx);
+event.samples  = [event.evnt.sample];
 event.times	   = data.time{1}(event.samples);
 
 tm             = data.time{1};
--- a/test/test_issue1387.m
+++ b/test/test_issue1387.m
@@ -1,7 +1,12 @@
 function test_issue1387
 
-%datadir = dccnpath('/home/common/matlab/fieldtrip/data/');
-fileName       = 'Test_4.edf';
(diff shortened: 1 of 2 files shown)


iandol:
My test plots out the channel and overlays the events, however this is not necessary to actually trigger this bug, it just provides visual confirmation, so should I remove the plotting or leave it in?

robertoostenveld:
> is it enough to basically throw an error using an assert?

Yes, that would be fine. I have just added some explanation here https://github.com/fieldtrip/dashboard#how-to-add-a-test

robertoostenveld:
I just re-read #1387. It should be fine to check it in the `test_xxx.m` script. Besides checking on the correct number, perhaps you also want to check on that `unique({event.type})` or `unique({event.value})` has the expected length? Without reading the actual code in detail, I would namely expect either one to correspond to the names of the channels.

test/test_issue1387.m
```diff
@@ -0,0 +1,40 @@
+function test_issue1387
+
```
robertoostenveld, line 2: Please add MEM, WALLTIME and DEPENDENCY here

test/test_issue1387.m
```diff
@@ -0,0 +1,40 @@
...
+%datadir = dccnpath('/home/common/matlab/fieldtrip/data/');
```
robertoostenveld, line 3: I have downloaded the test file and copied it to the compute cluster

```
roboos@mentat002> mkdir /home/common/matlab/fieldtrip/data/test/issue1387
roboos@mentat002> mv Test_4.edf /home/common/matlab/fieldtrip/data/test/issue1387
roboos@mentat002> ls /home/common/matlab/fieldtrip/data/test/issue1387
Test_4.edf
```

if you use
```
fileName = dccnpath('/home/common/matlab/fieldtrip/data/test/issue1387/Test_4.edf');
```
then you should see on your local computer that `dccnpath` returns the local file (assuming it is in your present working directory. So you can run the test script locally.

On the DCCN compute environment `dccnpath` either returns the original name, or the windows version of the name with `h:` instead of `/home`, and with backward slashes.

test/test_issue1387.m
iandol: OK added. I used `tic;toc` to measure `WALLTIME` and the value returned from ft_preprocessing for `MEM`, doubling both values.

test/test_issue1387.m
iandol: OK, path added.

robertoostenveld:
Thanks for your contribution!

robertoostenveld:
this fixes #1387

PR #1407, opened 2020-05-20
copied some templates with minor changes over from playground repo

schoffelen:


schoffelen:
I take the thumbs up above as an implicit permission to merge. I'll merge out of curiosity how this pans out on the repo (I have just edited the contributing.md document, and added proper credits to the mrtrix folks from which I shamelessly copied their document as a template).

Issue #1333, opened 2020-03-11
ft_prepare_sourcemodel: improve documentation/functionality of inwardshift/moveinward


schoffelen:
Reported by @britta-wstnr by e-mail:

Documenting and taking it from here

1) we just came across some unexpected behavior (helping a student today) with ft_prepare_sourcemodel(). If we warp an MNI template, inwardshift has no effect - is that because that would shift grid points to the wrong positions after warping?

2) We often see problems with not inwardshifted sources when using the openmeeg pipeline for leadfield computation (either things breaking during leadfield computation or voxels that draw a lot of activity (several orders of magnitude higher than the rest) after source reconstruction). Fixing the template (i.e. changing what is inside/outside there) only helps in some cases. Actually, we saw yesterday that also setting cfg.moveinward but having that value being smaller than cfg.inwardshift introduces an inwardshift in the warping cases, however,  if cfg.moveinward > cfg.inwardshift it errors out.
That behavior seems a bit erratic to me, do you agree? I can check that this is true on another data set as well (so far it's only the data of the phd student, so maybe confirming with e.g. the FieldTrip example data set would be a good idea) unless you tell me this is expected behavior :)

britta-wstnr:
I can have a look with the FieldTrip example data and the FieldTrip template later and report back.

britta-wstnr:
I had a look with the FieldTrip `Subject 01` data and I see a similar behaviour pattern to what we saw with @jordannicolasalves:

Having specified `cfg.warpmni = 'yes'` (and supplying the template etc.) makes `cfg.inwardshift` have no effect (which might be wanted as it would destroy the warped MNI structure on an individual basis?).
If also `cfg.moveinward` is specified the grid points get definitely moved - unless `cfg.headmodel` is _not_ supplied, then the computation throws an error:
```
Undefined function or variable 'pos'.

Error in headsurface (line 208)
  ori = normals(pos, tri, 'vertex');

Error in ft_prepare_sourcemodel (line 683)
  [pos1, tri1] = headsurface(headmodel, [], 'inwardshift', cfg.moveinward, 'surface', 'brain');

Error in check_sourcem_inwardshift (line 77)
grid    = ft_prepare_sourcemodel(cfg);
```

If we agree that neither of the two options should be allowed with `cfg.warpmni`, I feel like this should be mentioned in the documentation and maybe a `warning` would be a good idea to give the user a heads-up.

[...]

PR #1409, opened 2020-05-28
[WIP] Adding warnings regarding inwardshift and moveinward functionality

britta-wstnr:
As discussed in https://github.com/fieldtrip/fieldtrip/issues/1333, some improvement on both the documentation and functionality of `ft_prepare_sourcemodel` might be needed.

To do list for this:
- [x] add warning for `cfg.inwardshift` if a template is already determining `model.inside`
- [x] add warning if `cfg.moveinward` is used in combination with 3D grids
- [x] make the error that is received if headmodel is missing more informative
- [x] enhance documentation of `cfg.inwardshift` and `cfg.moveinward` in docstring
- [ ] enhance documentation on website (needs extra PR)


ft_prepare_sourcemodel.m
```diff
@@ -668,6 +668,9 @@
 end
 
 if ~isempty(cfg.moveinward)
+  if ~any([strcmp(cfg.method, 'basedonshape'), strcmp(cfg.method, 'basedoncortex'), strcmp(cfg.method, 'basedonvol'), strcmp(cfg.method, 'basedonfile')])
+      ft_warning('cfg.moveinward is designed to work with surface sourcemodels, not 3D grids.')
```
britta-wstnr, line 672: I am not sure how to handle `'basedonfile'` here, I assume it can result in any other kind of method down the road, depending on what is loaded?

private/headsurface.m
```diff
@@ -195,6 +195,8 @@
     otherwise
       ft_error('other surfaces cannot be constructed this way');
   end
+else
+    ft_error('Headmodel or headshape required.')
```
britta-wstnr, line 199: @robertoostenveld This might be too low level? I tested the cases I could think of and also checked where `headsurface.m` is used throughout the code base, but cannot ensure there is not a case where a resulting empty `pos` might be wanted.

The background is that not providing a headmodel but specifying `cfg.moveinward` results in the error
``` Undefined function or variable 'pos'.```
which is maybe not so informative for novice users, what went wrong. Today I saw that an empty `cfg.headmodel` throws errors at other positions for surface models as well, thus the lower level "catch" of the problem.

File: private/headsurface.m [hunk 3]
```diff
@@ -195,6 +195,10 @@
     otherwise
       ft_error('other surfaces cannot be constructed this way');
   end
+
+  %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+else
+    ft_error('headmodel or headshape required')
```
robertoostenveld, line 201: it is a good idea to explicitly check that any of the if-elseif-elseif clauses was used.

ft_prepare_sourcemodel.m
robertoostenveld: I think that it is ok (like you do) not give a warning when `basedonfile`

private/headsurface.m
robertoostenveld: it is indeed appropriate to check that one of the if-elseif-elseif clauses is actually being used

robertoostenveld:
I have made some updates to this branch and will merge it. The only point remaining is "enhance documentation on website (needs extra PR)". I won't address that now and here: the help of the function has already been improved. If you feel that more explanation is needed, please send a PR to the website.

PR #1412, opened 2020-05-29
implemented pan tompkin algorithm for ECG beat detection

helenacockx:


robertoostenveld:
I propose that we shortly present this in the NIRS hackathon this afternoon. I will make a test/demo script with both an ECG signal and a PPG signal.

PR #1413, opened 2020-05-31
implemented neighbcorr method for summary mode in rejectvisual

robertoostenveld:
this computes for every channel in every trial the squared correlation (R2) with the neighbouring channels. If the squared correlation is low, the channel has a timecourse that is very different from its neighbours, potentially indicating an artifact.

You can test this with

```
%%

cfg = [];
cfg.dataset = 'Subject01.ds';
cfg.channel = 'all';
% cfg.channel = 'MEG';
cfg.demean = 'yes';
cfg.baselinewindow = [-inf inf];
data = ft_preprocessing(cfg);

%%

cfg = [];
cfg.method = 'summary';
cfg.neighbours = 'ctf151_neighb.mat';
data_clean = ft_rejectvisual(cfg, data);

```
and subsequently clicking the neighbcorr option.

image

In the test MEG dataset above there are reference channels (the first ~35) and a trigger channel (the last) that don't have neighbours and hence a value of 0. All MEG channels have a squared correlation of >0.95 with their neighbours.

### DIFF
--- a/ft_interpolatenan.m
+++ b/ft_interpolatenan.m
@@ -18,8 +18,7 @@ function [dataout] = ft_interpolatenan(cfg, datain)
 % interpolate using spatial information, e.g. using neighbouring channels, you should
 % use FT_CHANNELREPAIR.
 %
-% To facilitate data-handling and distributed computing with the peer-to-peer
-% module, this function has the following options:
+% To facilitate data-handling and distributed computing, you can use
 %   cfg.inputfile   =  ...
 %   cfg.outputfile  =  ...
 % If you specify one of these (or both) the input data will be read from a *.mat
--- a/ft_prepare_neighbours.m
+++ b/ft_prepare_neighbours.m
@@ -254,22 +254,32 @@ switch cfg.method
 end
 
 % only select those channels that are in the data
-neighb_chans = {neighbours(:).label};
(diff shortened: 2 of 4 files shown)


schoffelen:
Tja, je moet toch iets in je vrije weekend. Go for it!

robertoostenveld:
I noticed that the axes of the plots are not correct, i.e. it projects the maxima, whereas it should project the minima (since small R^2 values are interesting). I will try to fix this.

robertoostenveld:
I fixed the display of the smallest/negative versus the largest/positive values. Here is the same data once more, now you can see that the single-trial correlation with the neighbours is not so high after all.

Screenshot 2020-05-31 at 15 09 46

StolkArjen:
Doet Robert nou data analyse?

robertoostenveld:
> Doet Robert nou data analyse?

ja, en dan kom ik er wel achter dat ik wel wat stroef geworden ben 😁

StolkArjen:
> > Doet Robert nou data analyse?
> 
> ja, en dan kom ik er wel achter dat ik wel wat stroef geworden ben 😁

Haha ja, was 't nou ft_frequencyanalysis, ft_freqanalyze, ...

Issue #1400, opened 2020-05-08
ft_selectdata averages fields the shouldn't be averaged: sampleinfo, trialinfo and time


NirOfir:
Hi,

I came across this bug when using ft_rejectvisual in summary mode, with a data structure that has a table in trialinfo. When selecting a trial to plot in the GUI, ft_selectdata is called (rejectvisual_summary lines 634-638):
```
cfg_sd = [];
cfg_sd.trials = trls(n);
cfg_sd.avgoverrpt = 'yes';
cfg_sd.keeprpt = 'no';
tmpdata = ft_selectdata(cfg_sd, info.data);
```
line 311 in ft_selectdata is the problematic one:
`if fieldhasrpt,     varargin{i} = makeselection(varargin{i}, datfield{j}, dimtok, rptdim{i},                                selrpt{i},     avgoverrpt,      'intersect', average); end`
If trialinfo is a table, this will result in an error, because cellmatmean which is called be makeselection expects cell or matrix datfields. trialinfo, time and sampleinfo are all defined as datfields and not removed by lines 160-172 that clean the datfield. sampleinfo is used later (lines 364-374) while I think trialinfo is not used. time is used.

Solving the averaging of sampleinfo and trialinfo is possible by moving lines 402-406 to after line 250 and apply them on datfield rather than keepfield (if they don't make sense in the output, I guess they shouldn't be kept in the input too?):
```
if avgoverrpt
  datfield = setdiff(datfield, {'cumsumcnt' 'cumtapcnt' 'trialinfo' 'sampleinfo'});
end
```

This solved my specific problem for ft_rejectvisual, but I don't know if it's a good idea, and it doesn't change the fact that ft_selectdata will still average the time field.

Any suggestions for a better fix?

schoffelen:
Although in itself this might be a valid issue, I don't understand why rejectvisual_summary has functionality that averages across observations. Does this have anything to do with the (on-the-fly) visualization of the artifactual trials? If so, why would it make sense to average across trials ti inspec the morphology of an artifact?

Besides this, could you please create a test script that, in combination with some problematic data demonstrates the issue. I would say that a call to ft_selectdata (+data) would be sufficient to highlight the issue.

Then, you can submit the test code and the fix as a PR to the fieldtrip github repo, and send your data (please make it as small as possible), so that we can document it, and put the data as test data on our filesystem.

NirOfir:
So there are actually 2 issues here:
[...]

PR #1416, opened 2020-06-03
Streamlining trial selection workaround of PR #442

NirOfir:
The workaround works also without asking ft_selectdata to average over the single selected files, so it can be made shorter. See discussion in issue https://github.com/fieldtrip/fieldtrip/issues/1400.


robertoostenveld:
thanks!

Issue #1386, opened 2020-04-23
mff_v3 implementation is incomplete


schoffelen:
right now, ft_read_data returns the contents of the complete datafile, and not the specified subselections. (at least not for chanindx, but probably also not for beg/end indices).

This was identified during the toolkit. The participants kindly offered to provide a small example file to test this. (although we already might have data around to use as test data)

schoffelen:
It's still incomplete, but PR #1417 at least allows for channel selection and time point selection in continuous data.

PR #1417, opened 2020-06-04
ENH - fix selection of channels and samples for mff_v3. works for con…

schoffelen:
…tinuous recordings, not sure about the rest


schoffelen:
This addresses a small part of #1386

schoffelen:
The low level code from mffmatlabio in its current use (I did not check the implementation in detail) does not allow for channel/sample/epoch selection. For now it's fixed for continuous files, I leave it up to somebody who has direct need for a robust v3 implementation to fix the rest.

Issue #721, opened 2018-06-13
implement support for the NWB file format


robertoostenveld:
See https://github.com/NeurodataWithoutBorders/matnwb for existing code. This would involve ft_filetype, ft_read_header, ft_read_data, ft_read_event, ft_read_spike. 

For testing I propose to use the public datasets at https://www.nwb.org/allen-cell-types-database/

@dorahermes and @choldgraf, could you perhaps look at this in relation to http://bit.ly/bids_ieeg?

choldgraf:
This sounds great to me - would make me feel much better about officially supporting NWB in the iEEG spec. FieldTrip has far more users than `matnwb` so I think having export functionality in fieldtrip would go a long way.

robertoostenveld:
I think the first goal would be to implement reading. If that works, writing could be considered.

schoffelen:
Is this going to be followed up?

robertoostenveld:
Since we don't have a direct benefit for this at the Donders, I don't see any of the local people implementing this. I also don't have time for this myself. Unless someone else picks this up, I propose that we close this issue.

schoffelen:
OK, closing for now.

rivkahrivkah:
Wow, my timing is great--I was just now looking to see if Fieldtrip supports NWB!

schoffelen:
Uhm, Rebecca I just closed the issue. Do you mean that you want to reopen and self-assign it to push it forward?

rivkahrivkah:
Sorry, my tone was off...I was just amused at the timing of my finding this issue, minutes after it was closed. I am new to both NWB and Fieldtrip so not really in the position to contribute, but maybe at a later date if we end going in that direction. Thanks for your efforts with Fieldtrip!

schoffelen:
No worries, I was just probing whether I could enthuse you to take this forward... Or perhaps you know somebody in your close working environment who could take this on.

dorahermes:
Just tagging Ben Dichter, who is in the NWB development team: @bendichter do you know anyone who may want to take this on?

bendichter:
Thanks @dorahermes! I think there would be a lot of benefit for us to support FieldTrip. I've always been impressed with the level of documentation for this toolbox, and I think integration would provide value for our user base.

I'm looking through these functions and trying to think about how well the assumptions of this API maps onto the structure of NWB files, and I have some questions:

[...]

PR #1419, opened 2020-06-05
Nwb integration

sbuergers:
Jens Klinzing (@jensklinzing), Tara van Viegen (@TaravanViegen) and I wrote code to implement loading of NWB data (both LFP and spikes) with FieldTrip.

https://github.com/jens-k/fieldtrip/tree/nwb-integration

The code is tested with matNWB release 2.2.5:
https://github.com/NeurodataWithoutBorders/matnwb/releases/tag/v2.2.5.0

It is assumed that the user is familiar with the basics of MatNWB (Github MatNWB) and the user's system is in a state that would also allow loading the data using MatNWB (MatNWB must be in the path, generateCore must have been run, the correct schema must be installed, see NWB schemas).

We want to include a test_nwb.m script (draft currently in folder test, but should be added as markdown here: https://github.com/fieldtrip/website/pull/249). To adapt this to be a proper test-script we have a few questions:

1
We believe some NWB datasets should be transferred to the test data folder on the DCCN server. Our test datasets are quite large (~10GB). Is this an issue? Should we create smaller ones?

2
How to deal with the MatNWB toolbox that has to be loaded? Will there be one stable version on the server or should it automatically be pulled from https://github.com/NeurodataWithoutBorders/matnwb?

3
Also, the MatNWB will create folders when being initialized and when loading a file. Can we use the systems 'tempdir' to create these files? Are they then automatically deleted after?

The code was tested with the following example datasets that are freely available online:
From https://gui.dandiarchive.org/#/file-browser/folder/5e6eb2b776569eb93f451f8d:
sub-YutaMouse20_ses-YutaMouse20-140324_behavior+ecephys.nwb
sub-YutaMouse41_ses-YutaMouse41-150831_behavior+ecephys.nwb

From https://osf.io/hv7ja/ OR https://gui.dandiarchive.org/#/dandiset/5e6ff4140fe0ec33e6b3d5d2:
P9HMH_NOID5.nwb

See discussion of NWB-fieldtrip integration here: https://github.com/fieldtrip/fieldtrip/issues/721#issuecomment-631471642

Cheers,
Steffen.


robertoostenveld:
Regarding your questions:

1. Smaller test files would be appreciated. The test will run faster, and it takes less space.  

2. I will have a look at the `matnwb` dependency. It is not pulled automatically. For testing I can add a static copy elsewhere, where it will be automatically loaded to the path. I could add a static copy to fieldtrip/external as well, but given that it is frequently updated, I don't think that is wise at this moment. *(automatic installation of toolboxes from github would be a nice feature to have)*

3. using the systems 'tempdir' is fine, after the test script completes it will automatically be cleared so no need for you to delete the files. 

Please give me some time to update the test script and try it, first on my local MacBook, then on the Linux compute cluster. I cannot test on Windows.

PR #1424, opened 2020-06-08
Enhancement: added import of .nex5 files

NeuroExplorer:
Added import of  .nex5 files. .nex5 is an extension of .nex data format for neurophysiological data that has been in use since 2015.

### DIFF
--- a/fileio/private/ft_senstype.m
+++ b/fileio/private/ft_senstype.m
@@ -247,6 +247,10 @@ elseif isfield(input, 'nChans') && input.nChans==1 && isfield(input, 'label') &&
 elseif issubfield(input, 'orig.FileHeader') &&  issubfield(input, 'orig.VarHeader')
   % this is a complete header that was read from a Plexon *.nex file using read_plexon_nex
   type = 'plexon';
+
+elseif issubfield(input, 'orig.FileHeader') &&  issubfield(input, 'orig.VarHeaders')
+  % this is a complete header that was read from a Plexon *.nex file using read_plexon_nex
+  type = 'nex5';
   
 elseif issubfield(input, 'orig.stname')
   % this is a complete header that was read from an ITAB dataset
--- a/fileio/private/ft_senstype.m
+++ b/fileio/private/ft_senstype.m
@@ -246,7 +246,7 @@ elseif isfield(input, 'nChans') && input.nChans==1 && isfield(input, 'label') &&
   type = 'neuralynx';
   
 elseif issubfield(input, 'orig.FileHeader') && issubfield(input, 'orig.VarHeader')
-  fh = input.orig.FileHeader
+  fh = input.orig.FileHeader;
   if issubfield(fh, 'NexFileHeader') && ischar(fh.NexFileHeader) && strcmp(fh.NexFileHeader, 'NEX5')
     % this is a complete header that was read from a Nex Technologies *.nex5 file using read_nex5
     type = 'nex5';
--- a/fileio/private/ft_senstype.m
+++ b/fileio/private/ft_senstype.m
@@ -57,6 +57,7 @@ function [type] = ft_senstype(input, desired)
 %   'eeg1010'
 %   'eeg1005'
 %   'ext1020'             in case it is a small subset of eeg1020, eeg1010 or eeg1005
+%   'nex5'
 %
 % The optional input argument for the desired type can be any of the above, or any of
 % the following generic classes of acquisition systems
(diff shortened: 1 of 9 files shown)


fileio/private/ft_senstype.m
```diff
@@ -247,6 +247,10 @@
 elseif issubfield(input, 'orig.FileHeader') &&  issubfield(input, 'orig.VarHeader')
   % this is a complete header that was read from a Plexon *.nex file using read_plexon_nex
   type = 'plexon';
+
+elseif issubfield(input, 'orig.FileHeader') &&  issubfield(input, 'orig.VarHeaders')
```
schoffelen, line 251: I notice that the difference between the 'plexon' and 'nex5' senstype is just one single character in one of the fieldnames of the original header structure. This is as intended?

fileio/private/read_nex5.m
```diff
@@ -0,0 +1,167 @@
...
+%   header      structure with header information
+%   feedback    0 or 1
+%   tsonly      0 or 1, read only the timestamps and not the waveforms
+%   channel     number, or list of numbers (that will result in multiple outputs)
+%   begsample   number (for continuous only)
+%   endsample   number (for continuous only)
+%
+% See also READ_NEX5_HEADER
+%
+% Copyright (C) 2020 Robert Oostenveld
```
schoffelen, line 26: Did Robert write this? If not, feel free to add your own name here. Likewise for the other newly contributed code. Or, if the new code is based on some older code that Robert wrote, you can add both your names

schoffelen:
Would it be possible to also provide some documentation on the website regarding the nex5 format, and motivate why to what extent this differs from the 'older' nex?

http://www.fieldtriptoolbox.org/getting_started/plexon/

Also, it would be good to provide us with a small test dataset.

fileio/private/ft_senstype.m
NeuroExplorer: I did not notice ft_senstype.m code at first and simply like VarHeaders as field name more than VarHeader. And then it was handy in ft_senstype. I can check 
orig.FileHeader.NexFileHeader == 'NEX5'
instead, for example.

fileio/private/read_nex5.m
NeuroExplorer: I wrote the code myself but is is based on Robert's read_nex_* code. I will use both names.

fileio/private/ft_senstype.m
schoffelen: Ah, I see. Well I don't know who came up with the original VarHeader (so without the 's') but if this is based on a convention that is also used in other software tools, I wonder whether it's clever to adopt a slightly different fieldname for something that probably has the same content. Therefore, I would recommend sticking to the 'without-s' naming convention (even if you don't like it :) ) and do a more principled check for the 'NEX5' string.

NeuroExplorer:
Sure, I will add information on nex5 file format.  I will also provide a couple of test nex5 data files. 

By the way, I noticed that existing read_nex_* code does not handle some .nex files properly. I will provide you with the test .nex file and fix the code in the next branch.

fileio/private/read_nex5.m
schoffelen: Top! Nice to meet you online, Alex! Good to know who's the named person behind NeuroExplorer

NeuroExplorer:
Hi Jan-Mathijs,

I made changes in my branch and provided links to .nex and .nex5 test files
in Github. Should I wait for Robert to review my code?

Best Regards,

Alex Kirillov <alex@neuroexplorer.com>
Nex Technologies
www.neuroexplorer.com

schoffelen:
Thanks Alex, I reviewed your code and will merge this soon. I had to wait for our file server to be upgraded, because I want to deal with the testdata upon merging. This will happen soon! Thanks for your contributions so far, and we look forward to a fix for the old-fashioned *.nex files :)

schoffelen:
Just to raise awareness: could the reported issue with the provided *.nex file also be related to things that are in this thread #34?

schoffelen:
I am not inhibited by any knowledge here, because I didn't realize that the file extensions seem to be different, although the common denominator seems to be 'plexon'.

schoffelen:
OK, merging.... Sorry for the delay

schoffelen:
OK, it should now be in the master branch, if the overnight test batch runs through fine (this is conditional on all the accumulated new stuff in the master branch), it will percolate to the release branch as well.

NeuroExplorer:
I found that I did not add reading .nex5 to ft_read_spike.m, so my
implementation of .nex5 import is incomplete. I will work on
ft_read_spike.m in the next few days.

jens-k:
> Just to raise awareness: could the reported issue with the provided *.nex file also be related to things that are in this thread #34?

I saw this mentioning of the pl2/v2 thread. It's not really related right?

PR #1430, opened 2020-06-10
ENH - replace occurrences of colormap with ft_colormaps, allows for d…

schoffelen:
…irect brewermap functionality

The first PR of this day implements an overloaded colormap function (ft_colormap), that allows the user to specify at the highest level (i.e. cfg.colormap) things like 'RdBu' or '*RdBu', i.e. directly using brewermaps, without the need of externally calling brewermap, and then inputting the Nx3 matrix into the cfg.

### DIFF
--- a/ft_multiplotTFR.m
+++ b/ft_multiplotTFR.m
@@ -230,7 +230,7 @@ end
 % set colormap
 if isfield(cfg,'colormap')
   if ~isnumeric(cfg.colormap)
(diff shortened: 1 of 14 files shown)


schoffelen:
Note that this allows for flexibly expanding to yet other fancy color schemes.

schoffelen:
well, there's also the 'cortex' etc color specifications, but I guess that this passes through unimpeded

PR #1432, opened 2020-06-10
Improvements to nwb integration

jens-k:


robertoostenveld:
I think that even with this change it will remain problematic with the core.mat file. One possible improverment could be to use the full path of the nwb filename (`fileparts`) to determine the possible location of the corresponding core.mat file. 

What I did myself (in an earlier attempty to get it to work) is that I took the different releases from https://github.com/NeurodataWithoutBorders/nwb-schema/releases and used `generateCore` on each of them. I renamed the core.mat file for each of them and now have 
```
mac036-lan> ls nwb-schema/
nwb-schema-2.0.1-core.mat	nwb-schema-2.2.3-core.mat
nwb-schema-2.1.0-core.mat	nwb-schema-2.2.4-core.mat
nwb-schema-2.2.2-core.mat	nwb-schema-2.2.5-core.mat
```

If that is a viable way, those core files could be added to fieldtrip/external (or matnwb) and loaded (by looking at the `nwb_fileversion`) in case no custom core file is present.

PR #1434, opened 2020-06-11
improve shift of labels in ft_databrowser

helenacockx:
Labels were only shifted vertically when the events occurred at exactly the same time point, hindering the readability of the event labels.
With the suggested adaptation, the event labels are shifted when occurring close (= within 1/10 of the horizontal time axis)  to another event.

### DIFF
--- a/ft_databrowser.m
+++ b/ft_databrowser.m
@@ -1776,7 +1776,7 @@ if strcmp(cfg.plotevents, 'yes')
       fprintf(eventlabellegend);
     end
     
-    % save stuff to able to shift event labels downwards when they occur at the same time-point
+    % save stuff to able to shift event labels downwards when they occur at a similar time-point
     eventcol = cell(1,numel(event));
     eventstr = cell(1,numel(event));
     eventtim = NaN(1,numel(event));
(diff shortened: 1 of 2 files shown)


robertoostenveld:
I tested it with this code 
```
function inspect_pull434

%%

fsample = 1000;
nsamples = 10*fsample;
nchans = 10;

data = [];
for i=1:nchans
  data.label{i} = num2str(i);
end
data.trial{1} = randn(nchans, nsamples);
data.time{1} = (1:nsamples)/fsample;


nevent = 300;
event = [];
for i=1:nevent
  event(i).type = 'Trigger';
  event(i).value = i;
  event(i).sample = (i-1) * round(nsamples/nevent) + 1;
end

%%


cfg = [];
cfg.event = event;
cfg.viewmode = 'vertical';
cfg.ylim = [-5 5];
ft_databrowser(cfg, data);
```
where I changed the number of events to get them close to each other.

robertoostenveld:
original code

Screenshot 2020-06-11 at 14 39 49

updated code
Screenshot 2020-06-11 at 14 40 29

robertoostenveld:
Looking at the code, I don't directly understand why it stops shifting them down after the 4th one. I suspect it is due to the text being limited to the (here invisible) box of the first channel in which the events are plotted. I suspect that `ft_plot_text` ensures that the text remains within the box (which is the box of channel 1). Although, when I do viewmode=butterfly (which should have a much larger box) it remains the same. 

Oh, it might be that the code in lines 1737-1740 needs to be improved to ensure that it encompasses all channels. 

The situation I created might be more extreme than the usual case. Also, the code changes are already an improvement over the previous. @helenacockx if you feel like it, you might want to improve it a bit further. If not, let me know and I'll merge it as is. In any case, could you add the "inspect" script to the pull request?

helenacockx:
it was not a problem of the plotting box, but rather that the shift was limited to the number of close events (in the example above there were always 4 close events, so shift was max 3).m

helenacockx:
I improved the code by plotting the label either on en empty shift location, or on the max shift location +1

robertoostenveld:
Screenshot 2020-06-11 at 17 39 20


nice, thanks!

Issue #1438, opened 2020-06-12
Reading raw events from EDF files without annotations


iandol:
**Describe the bug**
I want to extract events from analog TTL recordings from 8 channels of an EDF file. For that, `ft_read_event.m` should work, and it does if there is also a separate annotations field in the header. But if the field is missing, then manually extracted events are discarded after they are processed.

Around line `662` of `ft_read_event.m`:

```matlab
    if ~isempty(detectflank) % parse the trigger channel (indicated by chanindx) for events
      event = read_trigger(filename, 'header', hdr, 'dataformat', dataformat, 'begsample', flt_minsample, 'endsample', flt_maxsample, 'chanindx', chanindx, 'detectflank', detectflank, 'trigshift', trigshift, 'threshold', threshold);
    else
      event = [];
    end
    
    if issubfield(hdr, 'orig.annotation') && ~isempty(hdr.orig.annotation) % EDF itself does not contain events, but EDF+ does define an annotation channel
    ...
    else
        event = [];
    end
```

if we set `detectflank` then we read our events using `read_trigger` from raw channels (this works fine), however in the next `if` block - `if issubfield(hdr, 'orig.annotation') && ~isempty(hdr.orig.annotation)` unless we have annotations then the event field is reset to be empty `else event = [];`. This code therefore throws away the events we read using `read_trigger`, and I don't understand the logic why it should do this?

**To Reproduce**

Attached EDF file:

[test-saline-run3.edf.zip](https://github.com/fieldtrip/fieldtrip/files/4769261/test-saline-run3.edf.zip)

```matlab
cfg.dataset = 'test-saline-run3.edf';
hdr       = ft_read_header(cfg.dataset);
event   = ft_read_event(cfg.dataset, 'header', hdr,...
		'detectflank', 'up', 'chanindx', [6:13],...
		'threshold', '5*nanmedian');
```

event should not be empty, but should contain 587 entries extracted from analog channels 6:13. But line `706` sets event to be empty and the emptied event structure gets returned.

**Expected behavior**
The ability to manually extract trigger events from an EDF file even if it has no 'annotations' field. For me the logic may make more sense if events is set to [] first, then annotations are added, then any manual events next. 

**Screenshots**

**Environment and versions (please complete the following information):**
 - Operating System: Ubuntu 20.04
 - MATLAB version: 2020a
 - FieldTrip version: most recent master branch from github

**Additional context**

robertoostenveld:
[...]

PR #1439, opened 2020-06-12
Fix for #1438

iandol:
This is a fix and test case for #1438


robertoostenveld:
confirmed that it works

PR #1443, opened 2020-06-15
Improve implementation of SAM beamformer

schoffelen:


schoffelen:
This PR replaces Adonai Nunes PR #1221, basically because JM messed up that PR. 

Background, with big apologies to Adonai, my intention was to push into his branch some suggestions to his changes. I created my branch off a more recent version of FT, and this caused the whole history to be messed up. Rather than trying to fix it, I suggest to close that PR and continue working from here. 
Note that the core changes suggested by Adonai are still there, and that my suggestions mainly pertain to how the data are handled at the higher lever code (i.e. ft_sourceanalysis), and how some of the arguments are called.
I suggest that Adonai reviews the current PR, and upon approval, this one will be merged.

Main differences w.r.t. Adonai's suggestion (@AdoNunes, please check the boxes if you approve):
- [ ] don't use a separate baseline argument in ft_sourceanalysis to provide the covariance, but use an explicit cfg.sam.noisecov argument. -> note that this in principle can be tricky, since it assumes that the channels (and their order) match the channels+order in the data covariance. Yet, note that in its original implementation there was no bookkeeping check on the baseline argument either, so in that sense there is nothing lost with respect to the previous implementation.
- [ ] use 'noisecov' as key-value argument, rather than 'BL_...'
- [ ] allow for cfg.sam.toi (as indices), and for cfg.sam.latency_ori (expressed in s), compute toi indices prior to passing it to beamformer_sam, where it is used as the time range for the optimal orientation computation as per 'moiseev' (as far as I understood the code).

What needs to be thought through:
- [ ] support for multiple trials in the input, i.e. as per the confusion similar to the singletrial/rawtrial etc. of the other beamformers. Is this something that is allowed/supported?
- [ ] enforce the leadfieldopt to have backprojection switched off in case of spinning and on-the-fly computation of leadfields. 
- [ ] use the same principled handling of leadfieldopt and code logic (in case of pre-computed filters) as in the other beamformer functions.

schoffelen:
OK, what's now to be fixed is the fact that with precomputed leadfields, if they are of rank 2 and non-backprojected, the code fails, since ft_sourceanalysis does not pass on the headmodel/sens anymore in case of precomputed leadfields. The most foolproof way of fixing this would be to forward project the lf on their rank 2 basis prior to the rest of the computations. This would allow beamformer_sam to stay ignorant of the headmodel/sens, and takes away the need of doing a cfg.method dependent check at the higher level of ft_sourceanalysis

schoffelen:
@AdoNunes, feel free to assign yourself here, I was not able to do so for some reason

AdoNunes:
I can't assign myself, the option is not available.

inverse/beamformer_sam.m
```diff
@@ -123,136 +156,164 @@
...
+    % convert the U matrix into the optimal orientation vector
+    switch fixedori
+      case 'spinning'
+        % do nothing, optimum orientation is already computed above
+      otherwise
+        
+        % The optimum orientation is the eigenvector that corresponds to the
+        % biggest eigenvalue (biggest value is more logical, as it relates to SNR).
+        
+        if isfield(headmodel,'singlesphere') && isempty(noisecov) && ~strcmp(fixedori,'moiseev')
```
AdoNunes, line 273: && isempty(noisecov) conditional should be removed, it will always be false at it is created above

schoffelen: OK, I removed it.

schoffelen:
If you have other suggestions for improvement, I think that it should be possible to create a PR for this branch (which mean that you create a PR for a PR :) ). The branch then is the sam_cleaning branch on github.com/schoffelen/fieldtrip

schoffelen:
@AdoNunes what is needed to finalize this?

AdoNunes:
I wanted to add the pseudoZ ratio. I will do it soon, and make a PR on your PR :)

schoffelen:
OK thanks for the notification and good luck with the git wizardy

AdoNunes:
@schoffelen I tried to make a PR on your PR but I still get:
remote: Permission to schoffelen/fieldtrip.git denied to AdoNunes.
fatal: unable to access 'https://github.com/schoffelen/fieldtrip.git/': The requested URL returned error: 403
I will make a comment with the new lines to add.

inverse/beamformer_sam.m
```diff
@@ -123,136 +156,164 @@
...
-  gain        = lf * opt_vox_or;
-  trgain_invC = gain' * inv_cov;
-  SAMweights  = trgain_invC / (trgain_invC * gain);
-
+    dipout.ori{diplop}    = opt_vox_or;
+       
+    % compute the spatial filter for the optimal source orientation
+    gain        = lf * opt_vox_or;
+    trgain_invC = gain' * inv_cov;
+    SAMweights  = trgain_invC / (trgain_invC * gain);
```
AdoNunes, line 306: if strcmp(fixedori,'moiseev')
        % get pseudoZ
        Ng                      = gain' * Nproj * gain;
        Sg                      = gain' * Sproj * gain; 
        dipout.pseudoZ{diplop}  = Sg / Ng;
    end

schoffelen: Hi Adonay, it's strange that you cannot PR on this PR, but I don't mind adding this to the code. Just for my understanding, this is only applicable to the 'moiseev' method? Why then does it need to be handled at this location separately? I understand it given the current formulation, but isn't Ng./Sg equal to Nproj./Sproj. Both Ng, Nproj etc. seem to be scalars, and the 'gain' variable is a vector (thus gain'*gain is a scalar, and that drops out of the division). Would it be OK for you if I compute the pseudoZ higher up in the code, by a direct division between Nproj and Sproj?

AdoNunes: Nope. Ng and Sg are the noise and power projected by the most salient LF direction, they are scalars. Nproj and Sproj are n_sensor x n_sensors, expressing the noise (by inv cov matrix projection) and the signal (in case of avg BF the average signal by inv cov matrix projection or the signal inv covariance). By matrix projection I mean A*B*A.
And for each LF there will be a different pseudoZ, e.g. rest and finger tapping, the pseudoZ map will show activation in the motor area as a ratio finger/rest pseudoZ.
Yes, it's only applicable to 'moiseev' method, if the noise covariance is diagonal (as in the other methods) then pseudoZ will just show the pseudo magnitude of the source.

schoffelen: My bad, of course you're totally right. Sorry for my dimwittedness and thanks for your patience.

AdoNunes:
Haha I will add dimwittedness as a new word in my vocabulary ;)

The PR looks good to me, unless you want to change the variables names.

schoffelen:
No, all fine. I added the pseudoZ stuff, and ensured it to be a matrix in dipout, rather than cell-array.

Will merge now.

PR #1444, opened 2020-06-15
calculate heart period and ectopic beat corrections in ft_heartrate

helenacockx:
Adaptations to ft_heartrate:
- heart period (= RR interval) is also calculated by ft_heartrate and added as extra channel to the data output structure
- I added the option cfg.ectopicbeat_corr to correct for single ectopic beats.


robertoostenveld:
This has been superseded by #1479

PR #1445, opened 2020-06-15
Enhancement: added import of spikes from .nex5 files

NeuroExplorer:
In ft_read_spikes.m, implemented import of timestamps and spike waveforms from .nex5 files.


fileio/ft_read_event.m
```diff
@@ -2141,6 +2141,9 @@
     
   case 'plexon_nex'
     event = read_nex_event(filename);
+
+  case 'plexon_nex5'
```
schoffelen, line 2145: This now duplicates the next 3 lines. Please remove

schoffelen:
Thanks, Alex. Quick question: does the test data you recently sent also contain spikes?

fileio/ft_read_event.m
NeuroExplorer: Done. Was messed up by auto merge?

NeuroExplorer: I generated two new test files with more variations of data formats. The files are:
https://www.neuroexplorer.com/downloads/FieldTripTest1.nex
https://www.neuroexplorer.com/downloads/FieldTripTest1.nex5

Each file has 2 wave variables.

schoffelen: My experience is that these types of things are typically user errors :). Thanks for the files, I'll upload them to the fileserver. I guess for now the expected behavior is that they can be read in without issues, right?

NeuroExplorer: .nex5 file can be read without issues, but not .nex file. I deliberately made .nex file with some data that would break existing FieldTrip .nex import code. I will work on fixing .nex import in the next pull request.

schoffelen:
Ah, I forgot, same as previously. Thanks.

PR #1446, opened 2020-06-16
Pass the option denoise from ft_read_event through to read_trigger

iandol:
The private function `read_trigger` has an option `denoise`, however as it is a private we would normally use it via `ft_read_event` -- and this function does not accept any `denoise` option so we cannot toggle it when we want to read an event channel. This pull adds the option to `ft_read_event` and passes it whenever `read_trigger` is called, with the default remaining true so it shouldn't affect existing code...

### DIFF
--- a/fileio/ft_read_event.m
+++ b/fileio/ft_read_event.m
@@ -151,7 +151,7 @@ filename = fetch_url(filename);
 % get the options
 hdr              = ft_getopt(varargin, 'header');
 detectflank      = ft_getopt(varargin, 'detectflank', 'up', true);   % note that emptymeaningful=true
-denoise          = ft_getopt(varargin, 'denoise',      true );
+denoise          = ft_getopt(varargin, 'denoise', []);
 trigshift        = ft_getopt(varargin, 'trigshift');                 % default is assigned in subfunction
 headerformat     = ft_getopt(varargin, 'headerformat');
 dataformat       = ft_getopt(varargin, 'dataformat');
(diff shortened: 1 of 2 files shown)


schoffelen:
Interesting catch, thanks Ian! I think it is good that the option should be passed to the lower level function, but this exercise raises the following questions, that we may want to think about before merging (inviting @robertoostenveld to think along):
- [] should the default value be specified irrespective of the acquisition system? In other words, should the default be defined to be true at the top, or be system specific (i.e. specified in the switch box)? I realise that in the current version (i.e. pre Ian's change the default is true in read_trigger, but this may be revisited)
- [] given the long list of key-value pairs as input to read_trigger, with only small differences based on acquisition system peculiarities, we may consider creating a fixed set of opts{:}, which can be passed into read_trigger, as per the recently implemented leadfieldopt in the inverse modeling module.

robertoostenveld:
We recently (I think in commit 7feec296f1 or thereabouts, it was related to `tmsi_poly5`) did something similar for detectflank.

Re point 1: best would be to set the default only at a single place, which (for this) must be at the lowest level in `read_trigger`. That is where the algorithmic choice must be made, whereas in `ft_read_event` in general you don't care (yet). Also at the lowest level, there is more domain specific knowledge.

Re point 2: that would be nice, but not required. I worry a bit that - when implementing it nicely - we'll discover that there are some obscure exceptions that subsequently take a lot of time to fix.

schoffelen:
I agree with both points. With respect to point 1, this means that read_trigger should include an 'if istrue(denoise)', rather than 'if denoise', and ft_read_event should specify the default denoise as []. This would maintain original functionality (as would Ian's suggested fix). For now, we won't bother cleaning up the readability of the code, nor with rethinking whether all calls to read_trigger require denoise to be strictly true (as is the case now).

@iandol could you make the requested changes in this PR? Thanks.

iandol:
OK I converted the true to [] in `ft_read_event.m` and use `istrue()` in `read_trigger.m`...

schoffelen:
Thanks, Ian!

PR #1450, opened 2020-06-16
added checks to ft_volumenormalise for custom templates

mcvinding:
Added consistency check for cfg.template and cfg.templatecoordsys and a warning when using spmmethod='new' as it will use the tissue probability maps and not the template (if I read the code correctly?).

Added possibility to reach the code for the "mars" method. It seems like the if loop starting at line 261 were not be able to read the blick where the mars params are defined.

### DIFF
--- a/ft_volumenormalise.m
+++ b/ft_volumenormalise.m
@@ -13,9 +13,7 @@ function [normalised] = ft_volumenormalise(cfg, mri)
 %   cfg.opts        = structure with configurable normalisation options,
 %                       see spm documentation for details.
 %   cfg.template    = string, filename of the template anatomical MRI (default = 'T1.mnc'
-%                     for spm2 or 'T1.nii' for spm8/spm12). If you use your
-%                     own templete you must specify the template coordsys
-%                     in cfg.templatecoordsys.
+%                     for spm2 or 'T1.nii' for spm8/spm12).
 %   cfg.parameter   = cell-array with the functional data to be normalised (default = 'all')
 %   cfg.downsample  = integer number (default = 1, i.e. no downsampling)
 %   cfg.name        = string for output filename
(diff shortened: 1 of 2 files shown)


ft_volumenormalise.m
```diff
@@ -97,6 +101,15 @@
 % check if the input data is valid for this function
 mri = ft_checkdata(mri, 'datatype', 'volume', 'feedback', 'yes', 'hasunit', 'yes', 'hascoordsys', 'yes');
 
+% Check coordsys consistency
```
schoffelen, line 104: I think that it is more logical to move this chunk (or at least a chunk of code that has an equivalent outcome) to after the definition of the cfg-options. Right now, it can happen that a user specifies cfg.template to be 'T1.nii' by hand, in which case an error is thrown, while this shouldn't be necessary (since a default 'spm' templatecoordsys would then suffice)

ft_volumenormalise.m
```diff
@@ -145,6 +151,19 @@
   end
 end
 
+% ensure that the input MRI has interpretable units and that the input MRI is expressed in
+% a coordinate system which is in approximate agreement with the template
+disp('Doing initial alignment...')
```
robertoostenveld, line 156: please use `ft_info` or `ft_notify` instead of `disp`

ft_volumenormalise.m
```diff
@@ -255,7 +274,7 @@
       cfg.tpm = fullfile(spm('dir'),'tpm','TPM.nii');
     end
     
-    fprintf('warping the individual anatomy to the template anatomy, using the new-style segmentation\n');
+    fprintf('warping the individual anatomy to the template anatomy, using the %s-style segmentation\n', cfg.spmmethod);
```
robertoostenveld, line 277: also this (and other) `fprintf` should be changed to `ft_info`

schoffelen:
Hi @mcvinding, do you need additional input, or would you know how to proceed here?

mcvinding:
Sorry, I got distracted. I'll send make a new pull request tonight.

schoffelen:
No worries, just putting it back on your radar :)

robertoostenveld:
FYI, I have started working on this

Issue #848, opened 2018-10-18
dipole fitting fails using openmeeg forward model


robertoostenveld:
This is due to changes implemented this summer, where openmeeg lead fields are computed on pre-defined positions in ft_prepare_leadfield rather than in ft_compute_leadfield. It is now not possible any more to compute lead fields on arbitrary positions.

Solution: the actual computation should be moved back to ft_compute_leadfield.

PR #1452, opened 2020-06-18
fix leadfield computation for openmeeg

schoffelen:
…nmeeg leadfield out of ft_compute_leadfield


schoffelen:
This PR now reverts a change to the code that was probably contributed by @sarangnemo , where the computation of 'openmeeg' leadfield was taken to the higher level ft_prepare_leadfield. The consequence of this change was that functions that rely on ft_compute_leadfield (e.g. ft_dipolefitting) did not work anymore with openmeeg bem volume conduction models. I changed this back, and right now the test functions that flagged this (and which were renamed into failed_*, as per #848) work again.
I need to think through and ensure that the earlier changes (with keepdsm etc.) still work.

sarangnemo:
Hi @schoffelen, there was indeed reasoning behind that change that I documented in the notes for #747. I wonder if rather the ft_dipolefitting code should be revised to fit the more consistent framework that I tried to implement in #747?

schoffelen:
ft_dipolefitting needs to be able to compute leadfields on-the-fly. For this, it directly taps into ft_compute_leadfield, and doesn't need the administrative overhead that is provided in ft_prepare_leadfield. 

In the current PR I just moved your code down to ft_compute_leadfield. This change shouldn't change functional behavior (for the moment). Yet, what needs to be discussed, is the possibility of passing on a precomputed dsm (or outputting the dsm). If this is desired, then I am not sure whether passing it in the input cfg is the way to go.
Namely, this assumes that the columns of the dsm, Number-of-dipole (x,y,z)-coordinates, match the positions in the sourcemodel.pos, and that the rows in the dsm match the rows/columns of the volume conduction model's headmodel.mat. In other, words the precomputed dsm is only meaningful in combination with a specific sourcemodel AND headmodel.

To ensure a bit more consistency, I would vote for representing the dsm in the sourcemodel, in a way that is consistent in which leadfields are represented, i.e. non-empty cells for inside dipole positions. In this way, at least it can be ensured that the pre-computed dsm matches the positions of the sourcemodel. The only thing that then would be needed, would be a check to see whether the rows match the dimensionality of headmodel.mat.

robertoostenveld:
It is not only ft_dipolefitting, but also cortico-cortical coherence in ft_sourceanalysis with dics (cfg.refdip), or cfg.supdip, but also ft_dipolesimulation, ft_megplanar, and ft_megrealign are directly calling ft_compute_leadfield.  

The design and API for forward computations is documented on <http://www.fieldtriptoolbox.org/development/module/forward/>. It is nice that with the changes of @schoffelen the OpenMEEG implementation again adheres to this.

PR #1456, opened 2020-06-20
clean up ft_sourceanalysis support for freq domain data

schoffelen:
to achieve the correct match between user expectations and what can be achieved by the code


schoffelen:
OK with the current status of this PR, the test_pull1456 returns a sourceFreq as I would assume that Arno would have expected it, i.e. a source.avg.pow which has a dimension of ndipxnfreq

schoffelen:
Note:
- these changes might have broken 'harmony', 'rv' 'music', 'mne' with frequency domain data, but I doubt whether these methods were functional at all before
- I commented out a section in prepare_freq_matrices (which called ft_selectdata with avgoverfreq and avgovertime) since this seems a bit 'dubbel op', because time and latency selection is already taking place at the level of ft_sourceanalysis (now only if needed for pcc or dics).

schoffelen:
in the course of working on this, I will also check the other inverse method that claim to support frequency data, also see bugs 2824 and 2850 on bugzilla.

http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2850
http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2824

schoffelen:
OK, I guess that this PR will benefit from a quick review @robertoostenveld.
Also, I'd want to run a subset of the test batch before merging.

Note that currently, this branch is already a bit shaky due to recent change in ft_datatype_segmentation, so I am not sure whether any failing test functions may be caused by that. Let's see

schoffelen:
Note to @arnodelorme with this change in place you can compute frequency resolved power using eloreta. If you want to compute connectivity, you'd need to be a bit more creative

schoffelen:
Shoot! conflicts due to @robertoostenveld working on supchan/refchan functionality in ft_sourceanalysis. I think I have resolved them, but wouldn't mind an extra pair of eyes.

ft_sourceanalysis.m
```diff
@@ -322,10 +327,15 @@
   % copy the descriptive fields to the output
   source = copyfields(data, source, {'time', 'freq', 'cumtapcnt'});
   
-  % HACK the remainder of the code expects a single number for the frequency and/or latency
-  cfg.frequency = mean(cfg.frequency);
-  if isfield(data, 'time')
-    cfg.latency = mean(cfg.latency);
+  if ismember(cfg.method, {'pcc' 'dics'})
+    % HACK the remainder of the code expects a single number -> JM QUESTION:
```
robertoostenveld, line 331: good point, please change accordingly

private/prepare_freq_matrices.m
```diff
@@ -95,9 +95,9 @@
 % update the cfg
 cfg.channel     = freq.label(chanindx);
 if any(strncmp(tok, 'rpt', 3))
```
robertoostenveld, line 97: rather than the `any(strncmp(...))` I think this would better be explicitly tested as `startsWith(dimord, 'rpt')`

private/prepare_freq_matrices.m
```diff
@@ -106,11 +106,11 @@
     ft_error('the requested reference channel is not found in the data');
   end
   if any(strncmp(tok, 'rpt', 3))
```
robertoostenveld, line 108: idem, see above

robertoostenveld:
or regarding the testing with a realistic dataset: perhaps we should just take subjectK and make an example script for the website (and also as test script) that demonstrates the different methods that do work.

schoffelen:
@arnodelorme supplied me with some data, it's referenced in test_pull1456 and seems to provide expected results, both for eloreta and mne

arnodelorme:
You mean that I do not need to loop through frequencies (which was not functional anyway) when computing eLoreta on spectrum. I will test once it is in the base branch.

schoffelen:
indeed.

private/prepare_freq_matrices.m
schoffelen: done for this function. I suspect that for a lot of older code in the repo this could be done (i.e. not tokenizing de dimord anymore, but use contains and startsWith

PR #1462, opened 2020-06-24
print the correct number of components that is retained

### LINKED ISSUE #1418: ft_rejectcomponent displays incorrect number of components
**Describe the bug**
When running ft_rejectcomponent the number of components displayed in the command window, e.g.
"removing 1 components
keeping 33 components"
does not necessarily display the true number of components, for example if ICA was run on reduced rank data

**To Reproduce**
1. Run ICA with reduced rank (e.g. cfg.runica.pca = numChannels - 1;)
2. Reject components
3. ft_rejectcomponent prints to the command window the number of components kept and removed, as follows (lines 128-134, fieldtrip version 010916):
fprintf('removing %d components\n', length(reject));
if ~hasdata,
  fprintf('keeping %d components\n',  ncomps-length(reject));
else
  fprintf('keeping %d components\n',  nchans-length(reject));
end

where (lines 83- 84)
hascomp = exist('comp', 'var');
hasdata = exist('data', 'var');

**Expected behavior**
I expected to see the number of components I actually did identify, but ft_rejectcomponent printed the number it would expect based on my number of channels (larger than the true value because I ran on reduced rank data).

**Suggestion**
I think the number of components should have been preferentially taken straight from ncomps, rather than the first choice being taken from nchans, so the problem would be solved by just changing the
% set the rejected component amplitudes to zero
fprintf('removing %d components\n', length(reject));
**_if ~hasdata,_** -------> **_if hascomp_**
  fprintf('keeping %d components\n',  ncomps-length(reject));
else
  fprintf('keeping %d components\n',  nchans-length(reject));
end

because the script actually already checks if there's information about the number of components, but it prioritises calculation based on the number of channels now, so it's just a question of re-prioritising.

**Environment and versions (please complete the following information):**
 - Operating System: Windows 10
 - MATLAB version R2014b
 - FieldTrip version 010916

robertoostenveld:
This fixes #1418

### DIFF
--- a/ft_rejectcomponent.m
+++ b/ft_rejectcomponent.m
@@ -127,11 +127,7 @@ end
 
 % set the rejected component amplitudes to zero
 ft_info('removing %d components\n', length(reject));
-if ~hasdata
-  ft_info('keeping %d components\n',  ncomps-length(reject));
-else
-  ft_info('keeping %d components\n',  nchans-length(reject));
-end
+ft_info('keeping %d components\n',  ncomps-length(reject));
 
 % create a projection matrix by subtracting the subspace spanned by the
 % topographies of the to-be-removed components from identity
@@ -169,7 +165,7 @@ else
   
   keepunused = 'no'; % don't need to keep the original rejected components
   bname = 'invcomp';
-
+  
   % create the initial data structure, remove all component details
   data = keepfields(comp, {'trial', 'time', 'label', 'fsample', 'grad', 'elec', 'opto', 'trialinfo', 'sampleinfo'});
 end % if hasdata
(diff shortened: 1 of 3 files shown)


ft_rejectcomponent.m
```diff
@@ -127,11 +127,7 @@
 
 % set the rejected component amplitudes to zero
 ft_info('removing %d components\n', length(reject));
-if ~hasdata
-  ft_info('keeping %d components\n',  ncomps-length(reject));
-else
-  ft_info('keeping %d components\n',  nchans-length(reject));
-end
+ft_info('keeping %d components\n',  ncomps-length(reject));
```
robertoostenveld, line 130: this is the actual fix, the rest are just some other improvements

PR #1476, opened 2020-07-01
fix test_prepare_freq_matrices

schoffelen:
prepare_freq_matrices behavior change caused regression error in test function


schoffelen:
The error only became apparent yesterday, since the test function had a hardcoded path to /home/common, which is where the release branch is located.

schoffelen:
I think that this is a clean PR. The broken test function runs through now, as does test_pull1456

Issue #1465, opened 2020-06-26
implement method to place source model points at the centroids of FEM meshes


robertoostenveld:
I discussed this in person with @mcpiastra. 

It should be implemented as `cfg.method='basedoncentroid'` and be independent from the movement of dipoles (with cfg.moveinward and cfg.spherify) and it should be independent from the inside/outside detection. As such, this does not care about the tissue types that may be assigned to each volume element.

It should work both for hexahedral and tetrahedral meshes.

There is not a resolution that the user can specify, it will simply follow the volume elements of the mesh. For specified resolutions we will come with another suggestion.

robertoostenveld:
just a note to myself: the help of ft_prepare_sourcemodel is not very clear w.r.t. the difference (and similarities) between `basedongrid` and `basedonresolution`. At a later point in time, I should update the help.

robertoostenveld:
this has been closed with #1480

PR #1480, opened 2020-07-02
add method 'basedoncentroids'

mcpiastra:
related to issue #1465

### DIFF
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -257,6 +257,7 @@ switch cfg.method
   case 'basedoncentroids'
     fprintf('creating sourcemodel based on volumetric mesh centroids\n');
     cfg.tight       = ft_getopt(cfg.sourcemodel, 'tight',       'no');
+    cfg.inwardshift = ft_getopt(cfg, 'inwardshift', 0); % in this case for inside detection
 end
 
 if (isfield(cfg, 'smooth') && ~strcmp(cfg.smooth, 'no')) || strcmp(cfg.method, 'basedonmni')
@@ -681,7 +682,6 @@ switch cfg.method
     sourcemodel.tissue = headmodel.tissue;
     sourcemodel.tissuelabel = headmodel.tissuelabel;
     sourcemodel.unit    = headmodel.unit;
-    sourcemodel.inside  = true(size(sourcemodel.pos,1),1);
 end
 
 if isfield(sourcemodel, 'unit')
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -665,22 +665,22 @@ switch cfg.method
     end
 
     % determine the size of the relevant elements
-    numhex = length(headmodel.hex);
+    numhex = size(headmodel.hex,1);
     numelem = length(headmodel.hex(1,:));
 
     sourcemodel.pos = zeros(numhex,3);
 
-    for i=1:length(headmodel.hex)
+    for i=1:numhex
         matr_help = zeros(numelem,3);
         for j=1:numelem
             matr_help(j,:) = headmodel.pos(headmodel.hex(i,j),:);
         end
-        sourcemodel.pos(i,:) = mean(matr_help);
+        sourcemodel.pos(i,:) = mean(matr_help,1);
     end
 
     sourcemodel.tissue = headmodel.tissue;
     sourcemodel.tissuelabel = headmodel.tissuelabel;
-    sourcemodel.unit    = cfg.unit;
+    sourcemodel.unit    = headmodel.unit;
     sourcemodel.inside  = true(size(sourcemodel.pos,1),1);
 end
 
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -19,6 +19,7 @@ function [sourcemodel, cfg] = ft_prepare_sourcemodel(cfg)
 %                'basedonresolution'  regular 3D grid with specification of the resolution
 %                'basedonvol'         surface mesh based on inward shifted brain surface from volume conductor
 %                'basedonfile'        the sourcemodel should be read from file
+%                'basedoncentroids'   irregular 3D grid based on volumetric mesh
 % The default for cfg.method is to determine the approach automatically, based on
 % the configuration options that you specify.
 %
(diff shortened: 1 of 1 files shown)


ft_prepare_sourcemodel.m
```diff
@@ -642,6 +646,37 @@
...
+
+    for i=1:length(headmodel.hex)
+        matr_help = zeros(numelem,3);
+        for j=1:numelem
+            matr_help(j,:) = headmodel.pos(headmodel.hex(i,j),:);
+        end
+        sourcemodel.pos(i,:) = mean(matr_help);
+    end
+
+    sourcemodel.tissue = headmodel.tissue;
```
robertoostenveld, line 681: what if the headmodel does not have the tissue field? Or is that guaranteed to be there?

ft_prepare_sourcemodel.m
```diff
@@ -642,6 +646,37 @@
...
+    sourcemodel.tissuelabel = headmodel.tissuelabel;
+    sourcemodel.unit    = cfg.unit;
+    sourcemodel.inside  = true(size(sourcemodel.pos,1),1);
```
robertoostenveld, line 684: Once you add the inside field, it will (later) be assumed to be correct and hence not revised any more. If you don't know what should be inside and outside yet, better don't add it: the code later will deal with it (and potentially set all positions to inside)

ft_prepare_sourcemodel.m
```diff
@@ -642,6 +646,37 @@
...
+        matr_help = zeros(numelem,3);
+        for j=1:numelem
+            matr_help(j,:) = headmodel.pos(headmodel.hex(i,j),:);
+        end
+        sourcemodel.pos(i,:) = mean(matr_help);
+    end
+
+    sourcemodel.tissue = headmodel.tissue;
+    sourcemodel.tissuelabel = headmodel.tissuelabel;
+    sourcemodel.unit    = cfg.unit;
```
robertoostenveld, line 683: units should not come from the configuration but from the headmodel.

ft_prepare_sourcemodel.m
```diff
@@ -642,6 +646,37 @@
...
+      headmodel = rmfield(headmodel, 'tet');
+    end
+
+    % determine the size of the relevant elements
+    numhex = length(headmodel.hex);
+    numelem = length(headmodel.hex(1,:));
+
+    sourcemodel.pos = zeros(numhex,3);
+
+    for i=1:length(headmodel.hex)
```
robertoostenveld, line 673: rather than length, better use size(..., 1).

If there is only one hexaheder, the length would return 8, which is not what you expect.

ft_prepare_sourcemodel.m
```diff
@@ -642,6 +646,37 @@
...
+        matr_help = zeros(numelem,3);
+        for j=1:numelem
+            matr_help(j,:) = headmodel.pos(headmodel.hex(i,j),:);
+        end
+        sourcemodel.pos(i,:) = mean(matr_help);
```
robertoostenveld, line 678: for the call to mean, I would explicitly add the dimension over which the mean is taken. The default might be ok, but explicit code is always better.

ft_prepare_sourcemodel.m
mcpiastra: i'm not sure if every mesh will have a tissue or tissuelabel field. i added two ifs, not sure it is the best solution.

ft_prepare_sourcemodel.m
mcpiastra: I modified it. 
with a hexa mesh with 5 compartments (scalp, skull, csf, brain, lesion) the inside sources are in brain+csf.

mcpiastra:
one comment.
from my understanding, the method 'basedonvol' (or better, the function `headsurface`) does not cover the case of 
`cfg.headmodel = vol mesh` with `cfg.headmodel.type = 'simbio'`.
so when the method is omitted, because of these lines:
```
elseif ~isempty(cfg.headmodel)
    cfg.method = 'basedonvol';
```
the function `headsurface` does not return anything and therefore there is an error.
it seems to me that 'basedoncentroids' can be seen as the 'basedonvol' for fem meshes. even though, for fem meshes, the output would not be a superficial dipole layer but a volumetric source space. still the name 'basedonvol' suggests that the headmodel is going to play the crucial role, as it does in 'basedoncentroids'.
I see many ways to deal with this. (we could also discuss in person):
1.  include basedoncentroids in basedonvol (and specify that the output will be a volume if in input there is a volumetric mesh)
2. keep basedoncentroids and basedonvol separate (and avoid the option to give a volumetric mesh as an input for basedonvol)

robertoostenveld:
'basedonvol' could be implemented for FEM consistent with the other methods by determining the outer surface and working from that. That would be similar to how the 'surfaceonly' option in `ft_plot_mesh` works. 
...   hmm, looking in the code that is actually what should be supported already. On line 72 in `headsurface it creates the surface mesh from the volume mesh.  But on line 568 in `ft_prepare_sourcemodel` it is called with the geometry in the 1st `headmodel` input, not with the tet/hex mesh in the optional `headshape`  argument. That is why it does not work. 

To be honest, `headsurface` is a very old function, and especially in the first few years of FieldTrip it was abused to do all sorts of magic in the background. The way that the function is called and works is therefore a mess. If I now were to fix this,  I would change the function interface and make the different methods that are implemented much more explicit ("make it explicit" is a common theme we learned the last 15 years). 

Also, the default for `ft_prepare_sourcemodel` to be 'basedonvol' is actually stupid: AFAIK it is only relevant for `ft_megrealign` and `ft_megplanar`, and it would also there be much better to make it explicit.

robertoostenveld:
perhaps we should just not automatically end up with `cfg.method = 'basedonvol'` on line 194 in ft_prepare_sourcemodel and require that specific option to be explicitly specified in the two functions that rely on it. 

I sometimes see that FT test scripts (but probably also users) result in a 642 dipole grid, and then *I* realize it is a 3-fold refined icosahedron that is shifted inward from the volume conductor surface. That is hardly ever what one would want.

robertoostenveld:
I have simplified the code a bit. @mcpiastra can you have a look?

mcpiastra:
looks good to me!
I could change lines ~122-133 in ft_inside_heamodel in the same way. (not urgent maybe)

mcpiastra:
> perhaps we should just not automatically end up with `cfg.method = 'basedonvol'` on line 194 in ft_prepare_sourcemodel and require that specific option to be explicitly specified in the two functions that rely on it.

what about eliminating lines 193-194? 

> I sometimes see that FT test scripts (but probably also users) result in a 642 dipole grid, and then _I_ realize it is a 3-fold refined icosahedron that is shifted inward from the volume conductor surface. That is hardly ever what one would want.

robertoostenveld:
but we would have to ensure that the corresponding fieldtrip functions then do specify `cfg.method = 'basedonvol'` (which they currently are probably not doing)

robertoostenveld:
since this PR is now ready, I will merge. Further ideas have to be discussed and implemented elsewhere.

Issue #1466, opened 2020-06-26
implement cfg.movetocentroid for FEM meshes in ft_prepare_sourcemodel


robertoostenveld:
as discussed with @mcpiastra and to complement #1465 

This will allow users to use `cfg.method='basedongrid'` or `cfg.method='basedoncortex'` or any of the other methods. What should happen is that the source model is in 1st instance created, and that - similar to `cfg.moveinward` and `cfg.spherify` the positions of the sources are shifted toward the nearest centroid of the hexahedral or tetrahedral head model mesh.  

This should only work for head models that consist of FEM meshes.

This should still be independent of the inside/outside detection of the points of the source model, which is actually the 3rd step that is done in `ft_prepare_sourcemodel`.

robertoostenveld:
now that #1480 is implemented, the next step is to implement the cfg.movetocentroid option. @mcpiastra please do pull the master from fieldtrip/fieldtrip to ensure that your development branch is up to date.

robertoostenveld:
this has been closed with #1484

PR #1484, opened 2020-07-07
implement "cfg.movetocentroids" in ft_prepare_sourcemodel  

mcpiastra:
this is linked to #1466.
tested with:
```
cfg = [];
cfg.resolution = 1; 
cfg.headmodel = mesh;
cfg.headmodel.type = 'simbio';
cfg.movetocentroid = 'yes';
reg_grid = ft_prepare_sourcemodel(cfg);
```
the regular 1cm source grid is shifted to the closest centroids of the volumetric mesh

### DIFF
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -719,7 +719,11 @@ if ~isempty(cfg.movetocentroid) && strcmp(cfg.movetocentroid, 'yes')
     % eliminate duplicates (if, e.g., cfg.resolution is smaller than the
     % mesh resolution)
     sourcemodel.pos = unique(grid_shifted,'rows','stable');
-    cfg.tight       = ft_getopt(cfg.sourcemodel, 'tight',       'no');
+    if isfield(sourcemodel,'dim')
+        sourcemodel = rmfield(sourcemodel,'dim');
+    end
+
+%     cfg.tight       = ft_getopt(cfg.sourcemodel, 'tight',       'no');
 
 end
 
--- a/ft_prepare_sourcemodel.m
+++ b/ft_prepare_sourcemodel.m
@@ -706,25 +706,21 @@ if ~isempty(cfg.moveinward)
   end
 end
 
-if ~isempty(cfg.movetocentroid) && strcmp(cfg.movetocentroid, 'yes')
-    % 1. compute centroids
+if strcmp(cfg.movetocentroid, 'yes')
+    % compute centroids of the tetrahedral or hexahedral mesh
     centroids = compute_centroids(headmodel);
 
-    % 2. move the sourcemodel to the closest centroids
+    % move the dipole positions in the sourcemodel to the closest centroid
     grid_shifted = zeros(size(sourcemodel.pos));
     for i = 1:length(sourcemodel.pos)
-        [~, amin] = min(sum((sourcemodel.pos(i,:) - centroids.pos).^2,2));
+        [dum, amin] = min(sum((sourcemodel.pos(i,:) - centroids.pos).^2,2));
         grid_shifted(i,:) = centroids.pos(amin,:);
     end
-    % eliminate duplicates (if, e.g., cfg.resolution is smaller than the
-    % mesh resolution)
-    sourcemodel.pos = unique(grid_shifted,'rows','stable');
-    if isfield(sourcemodel,'dim')
-        sourcemodel = rmfield(sourcemodel,'dim');
-    end
-
-%     cfg.tight       = ft_getopt(cfg.sourcemodel, 'tight',       'no');
-
+    % eliminate duplicates, this applies for example if cfg.resolution is smaller than the mesh resolution
+    sourcemodel.pos = unique(grid_shifted, 'rows', 'stable');
+    
+    % the positions are not on a regular 3D grid any more, hence dim does not apply
+    sourcemodel = removefields(sourcemodel, {'dim'});
 end
 
 % determine the dipole locations that are inside the source compartment of the
@@ -734,6 +730,9 @@ if ~isfield(sourcemodel, 'inside')
 end
 
 if strcmp(cfg.tight, 'yes')
+  if ~isfield(sourcemodel, 'dim')
+    ft_error('tightgrid only works for positions on a regular 3D grid');
+  end
   fprintf('%d dipoles inside, %d dipoles outside brain\n', sum(sourcemodel.inside), sum(~sourcemodel.inside));
   fprintf('making tight grid\n');
   boolvol = reshape(sourcemodel.inside, sourcemodel.dim);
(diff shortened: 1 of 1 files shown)


ft_prepare_sourcemodel.m
```diff
@@ -725,6 +725,44 @@
...
+    % 2. move the sourcemodel to the closest centroids
+    grid_shifted = zeros(size(sourcemodel.pos));
+    for i = 1:length(sourcemodel.pos)
+        [~, amin] = min(sum((sourcemodel.pos(i,:) - centroids.pos).^2,2));
+        grid_shifted(i,:) = centroids.pos(amin,:);
+    end
+    % eliminate duplicates (if, e.g., cfg.resolution is smaller than the
+    % mesh resolution)
+    sourcemodel.pos = unique(grid_shifted,'rows','stable');
+    cfg.tight       = ft_getopt(cfg.sourcemodel, 'tight',       'no');
```
robertoostenveld, line 762: why do you want cfg.tight=no and not leave that to the user or the normal default settings?

robertoostenveld: if you start with 
```
cfg.method='basedonresolution'
cfg.movetocentroids='yes'
```
then trying to set the default here again has no effect anyway, since it happens already at line 214

ft_prepare_sourcemodel.m
```diff
@@ -725,6 +725,44 @@
   end
 end
 
+if strcmp(cfg.movetocentroid, 'yes')
```
robertoostenveld, line 728: right now the cfg` does not have the `movetocentroids` field. You should add that somewhere around line 150

ft_prepare_sourcemodel.m
```diff
@@ -725,6 +725,44 @@
...
+
+    % 2. move the sourcemodel to the closest centroids
+    grid_shifted = zeros(size(sourcemodel.pos));
+    for i = 1:length(sourcemodel.pos)
+        [~, amin] = min(sum((sourcemodel.pos(i,:) - centroids.pos).^2,2));
+        grid_shifted(i,:) = centroids.pos(amin,:);
+    end
+    % eliminate duplicates (if, e.g., cfg.resolution is smaller than the
+    % mesh resolution)
+    sourcemodel.pos = unique(grid_shifted,'rows','stable');
```
robertoostenveld, line 761: removing the duplicates might cause some applications of this to break: imagine starting with basedongrid or basedonresolution; in both cases `sourcemodel.dim` is used to determine how the data can be reshaped in a 3D array. After removing positions, the estimated data cannot be reshaped any more and `size(pos,1)` is not equal to `prod(dim)`. 

You either remove positions and remove the dim field, or you keep positions and dim. In the latter case, `ft_sourceplot` might still be able to plot it as if it were a regular 3D grid.

mcpiastra: I was about to write a comment on this (that is connected to setting `cfg.tight       = ft_getopt(cfg.sourcemodel, 'tight',       'no');` ).

I think `ft_sourceplot` works also with a sourcemodel that does not have the field "dim" (I need to check again). I would then rather eliminate the dim field in sourcemodel. 

If cfg.tight is set to yes,  then the dim is needed, but if the user wants a sourcemodel moved to the centroids, it does not seem necessary to have a smaller bounding box (i.e., to use cfg.tight), since it is already based on the volumetric mesh in input.

Maybe I missed something.

robertoostenveld: The dim is present whenever the positions form a regular 3D grid, as in the example at the top of https://github.com/Donders-Institute/meg-hackathon/tree/master/2020-07-06-earth-moving-statfun. The tight option only applies when there is a dim. Please leave it as-is. 

The only thing you should worry about is whether you want to allow the slightly shifted positions to be still interpretable on their original regular positions (with a dim) or not (without a dim). Probably the best is not to keep the dim, since the positions are not exactly on a grid any more.

mcpiastra: It is not clear to me.
I removed the dim from the sourcemodel and removed the cfg.tight = no (inside the cfg.movetocentroids).
If I run:
```
cfg                 = [];
cfg.resolution      = 1; 
cfg.method          ='basedonresolution';
cfg.movetocentroids ='yes';
cfg.headmodel       = mesh;
cfg.headmodel.type  = 'simbio';
cfg.movetocentroid  = 'yes';
reg_grid = ft_prepare_sourcemodel(cfg);
```
it will enter cfg.tight (because is set to yes when choosing `cfg.method          ='basedonresolution';`) and will give an error because the field dim is not there.

I do not think you meant what I did.

robertoostenveld: you should not specify both cfg.movetocentroid=yes and cfg.tight=yes. That is a user error, i.e. not something the code has to fix.

robertoostenveld: it is the same as getting an error with cfg.movetocentroid=yes and giving it a spherical head model.

robertoostenveld:
thanks!

ft_prepare_sourcemodel.m
mcpiastra: I think still there is something to fix.
If the user specify `cfg.method          ='basedonresolution';`, then cfg.tight is automatically set to yes (without explicitly specifying it in the cfg). See line 221.
so, when I run:
```
cfg                 = [];
cfg.resolution      = 1; 
cfg.method          ='basedonresolution';
cfg.movetocentroids ='yes';
cfg.headmodel       = mesh;
cfg.headmodel.type  = 'simbio';
cfg.movetocentroid  = 'yes';
reg_grid = ft_prepare_sourcemodel(cfg);
```
is giving me the error:
Error using ft_prepare_sourcemodel (line 735) tightgrid only works for positions on a regular 3D grid

sorry if I am still missing something crucial.

mcpiastra: shall we remove line 221?

ft_prepare_sourcemodel.m
```diff
@@ -706,25 +706,21 @@
   end
 end
 
-if ~isempty(cfg.movetocentroid) && strcmp(cfg.movetocentroid, 'yes')
-    % 1. compute centroids
+if strcmp(cfg.movetocentroid, 'yes')
```
mcpiastra, line 710: if I now run:
```
cfg                 = [];
cfg.resolution      = 1; 
cfg.method          ='basedonresolution';
cfg.headmodel       = mesh;
cfg.headmodel.type  = 'simbio';
reg_grid = ft_prepare_sourcemodel(cfg);
```
then I get the error:
Reference to non-existent field 'movetocentroid'.

robertoostenveld: the is an "s" missing on line 710 where strcmp(..., 'yes') is checked. I will fix it. I will also add the option to the documentation....

See commit 266e2aec5cfadfc102d9873125f7ee5960a644f9

PR #1489, opened 2020-07-08
ft_version does not return version string for compiled fieldtrip-lite

arnodelorme:


fileio/ft_read_atlas.m
```diff
@@ -138,7 +138,7 @@
     % important to keep the original numbers or to make the list with
     % labels compact. This could be made optional.
     compact = true;
-    if compact
+    if compact && ~isempty(atlas.tissuelabel)
```
robertoostenveld, line 141: I don't think this is harmful, but it should only be needed in case there are no tissue labels at all, right? Which means that tissue==0 everywhere.

robertoostenveld: this might be *a solution* to *a problem*, but perhaps there is a better solution for the specific problem.

arnodelorme: Hum, I only wanted to commit the change to ft_version.m but somehow, previous pull requests were added. I should have reseted my local Fieldtrip repo to the master in a hard way (although I am not sure how to do that). I guess it can simply be deleted and forked again.

arnodelorme: About the code above "atlas.tissuelabel". I remember adding it to read some specific file format which was not supported.

robertoostenveld: don't worry, the extra change is fine. Next time you better make a feature branch and make the changes there, rather than making them directly on your master. It is best practice to keep the master clean from any changes.

arnodelorme: Yes, that makes sense.

PR #1427, opened 2020-06-09
tetra mesh as input for simbio 

mcpiastra:
improve prepare_mesh_tetrahedral.m so that the tetra mesh in output can be read by ft_headmodel_simbio.m

PR #1490, opened 2020-07-08
tetra mesh as input for simbio

robertoostenveld:
this closes #1427 with some small code updates.


robertoostenveld:
oh, it also includes some other small code cleanups of the last few days that are unrelated to simbio meshes.

Issue #1500, opened 2020-07-16
ft_plot_vector returns the handle of the text rather than line object, if a label is provided


NirOfir:
**Describe the bug**
ft_plot_vector returns the handle of the text rather than line object, if a label is provided. This happens because all handles in the code are given the same variable name "h". I'll make a PR to correct that.

**To Reproduce**
Run the following code:
```
t = linspace(0,4*pi);
y = sin(t);

% This returns a text object
h_line = ft_plot_vector(t, y, 'label', 'sine wave');
class(h_line)

% This returns a line object
h_line = ft_plot_vector(t, y);
class(h_line)
```

**Expected behavior**
ft_plot_vector should return the handle to the line object that was created, not its label.

**Additional context**
This error was found during work on #1499.

NirOfir:
Just noticed that the parent is set according to the 'Parent' name-value pair input many times in many different ifs, and then once again at the end of the code for all conditions. I removed all 
```
if ~isempty(parent)
        set(h, 'Parent', parent);
end
```
Except for the last one.

PR #1501, opened 2020-07-16
Fixed erronous returned handle when label is provided

NirOfir:
see #1500.


plotting/ft_plot_vector.m
```diff
@@ -337,19 +337,12 @@
         % plot each line with its own color
         h = plot(hdat, vdat(i,:), style, 'LineWidth', linewidth, 'Color', color(i), 'markersize', markersize, 'markerfacecolor', markerfacecolor);
       end
-      if ~isempty(parent)
```
schoffelen, line 340: if the variable h is recycled for each line to be plotted, I'd say that indeed it doesn't make sense to add the parent property to the handle, if it's to be overwritten in the next iteration. Yet, what I don't understand in general from this part of the code, is why there is a loop across i, but i is only used in line 338 and not in the other situations. Does this mean that with matricial input the matrix is plotted i times on top of itself?

schoffelen: Ah, it seems that this is incorrect only here. further down in the code the handles are indeed correctly updated in the loop, i.e. h(i) rather than h. I would assume that this also needs to be done here.

In that case (with a vector (or even matrix) of handles) I'd assume that the parent needs to be added to each of them. This can be done once at the bottom of the function indeed, but may require a for-loop there.

schoffelen: OK, overall it seems that the section that deals with the saturation case is messy.
in case multiple handles are needed (i.e. with multiple colors and/or with multiple saturated segments) the handle currently overwrites itself, which is probably not intended. Also the for loop across i is misplaced.

While you're at it, could I ask you to address this inconsistency?

NirOfir:
Sure, I can do that, but I don't completely understand what goes on there. What handles are supposed to be returned, and for what purpose? For rejectvisual there's only a single graphical object that interests us (the line itself, to make things efficient and quick by accessing properties instead of redrawing). However, when highlighting is used, there's more than one relevant object (the full data line and the highlight box or line), and I don't know which function is then using it and for what? I searched across the repository for "= ft_plot_vector" and "=ft_plot_vector" (to catch anywhere the function is called with output), and the only results are the modified rejectvisual_channel and ft_databrowser. In both cases the function is called without highlighting. I'd be happy to get more details about what the desired behavior is.

schoffelen:
> I'd be happy to get more details about what the desired behavior is.

That makes two of us :)

OK, I guess that your changes don't result in unexpected behaviour elsewhere, so perhaps we could safely proceed

NirOfir:
I checked through all subfolders. Only ft_databrowser and the new rejectvisual_channel use it with an output.

Well, since we're the ones setting the behavior here, I suggest that we set it to something reasonable :)
The output will have 3 optional varargout:
{1} The full data line object
{2} The highlight object
{3} The label text object

2 and 3 will be attached as children to 1.

I think there are major possible improvements to the way color is handled:
1. We can sort everything out in the top of the function instead of all the ifs when plotting happens. This should save us some for loops.
2. The function is skipping plotting altogether if color = 'none'. This is inconsistent with the way a matlab user would expect plot() to behave, and would crash the upstream code if it tries to plot in a loop and keep the handles. This snippet
```
t = linspace(0,4*pi);
y = sin(t);
color = {'k', 'g', 'none'};

for i = 1:3
    h_line(i) = ft_plot_vector(t, y*i, 'label', 'sine wave', 'color', color{i});
end
```
returns the following error: `Unable to perform assignment because the left and right sides have a different number of elements.`

I'll fix that and edit the PR.

NirOfir:
Ok, never mind. I still think that this would be the right approach, but it's too much work to try and make everything consistent right now. Ideally I would make a 1Xnline cell array with the colors at the beginning, then have one plot call for the full lines, one plot call for the highlighting and set the the colors with set(h, {'Color'}, color_cell_array). This will handle all colors, in any format or combination or formats you want. For example: color_cell_array = {'k', [0, 0, 1], 'none', '#fcba03'} will work just fine (if your matlab is new enough to handle hex).

Would it be ok to accept this PR so I can push the fix to rejectvisual_channel which is ready?

schoffelen:
also, was it intended to have the changes of rejectvisual_channel in this PR?

NirOfir:
> can you consistently make the number of spaces per tab 2?
> Sorry, this is really nitty gritty.

Sure, I corrected it in my local fork, and will update the PR.

> also, was it intended to have the changes of rejectvisual_channel in this PR?

Oops, no :( I'm still learning the ropes here. Is it ok that I'll delete the file and make a separate PR with it? That was my initial idea to keep things clear.

schoffelen:
Yes, sure. I think it will be cleanest to have the changes to rejectvisual_channel in a separate branch (and PR). To be safe, I'd copy over the changed version to some other location, then you can update the file with the version in the repo for this branch. Then, you can create a new branch and copy the function back add/commit/push/PR.

Indeed git can be a beast, but once you get the hang of it (a bit) it can be very useful. As long as you don't want to do anything too fancy.

NirOfir:
Sorry, I completely messed this up. May I close this pull request and open new clean ones?

schoffelen:
sure, no problem

NirOfir:
Thanks for your patience!

Issue #1499, opened 2020-07-15
Speeding up rejectvisual_channel


NirOfir:
Hi,

I find the rejectvisual_channel function very useful, but it's quite slow. I think this comes from the fact that after each interaction with the GUI, the entire graphics is plotted again (by the redraw subfunction). This means that each time the user marks a trial to exclude by clicking on it, or each time the user moves to a new channel, all graphics are recreated, and the GUI takes several seconds to become responsive again (obviously this becomes worse with more trials). The solution I thought about is this:

Create the entire graphical layout once at the beginning. In the trial plotting loop, get the handles of individual lines from ft_plot_vector, and assign them as children to the figure. When an object is interacted with, change the properties of only that object when those are interacted with. When the user moves to a different channel, loop and change only the YData property of all trials (the layout and excluded trials remain the same, so this should be simple enough).

I wanted to check with you whether it makes sense before submitting a PR. If it does make sense, I think the same logic applies to rejectvisual_trial.

schoffelen:
Go for it! It's been a while that I used this functionality myself, but if I recall well, not all trials/channels might fit into one figure, right? I guess then that the redraw stil needs to recreate handles etc. when this happens

NirOfir:
Great, I'll start working on the code and hopefully make a PR soon. As to your question, right now only one figure is created, regardless of the number of trials. The way the code is structured now, with uiwait, I don't think can be extended to more than one figure. I don't see any reason that this code won't work without uiwait, but that will take more work to do. Anyway, theoretically, what I would do if we want multiple figures is to create them in a loop and make sure that the assignments happen within the loop. The rest should "take care of itself" as long as the handles are in the right place.

schoffelen:
sorry for the confusion, i did not suggest that more than one figure should be created, I just (wrongly) thought that this was the case.

NirOfir:
Oh ok, no problem. I've starting fixing the code, and this is what I've fixed so far:

1. There's a redundant condition in the while loop that maintains focus on the GUI ('interactive' is set to 1 before the loop but nothing else ever touches it, so it's a condition that can't fail).
2. During redraw, tmin (min timepoint for all trials), tmax (max timepoint for all trials) and amax (maximum amplitude, later ymin and ymax are calculated and used) are calculated but not used.
3. All subfunctions had varargout that weren't set.
[...]

PR #1504, opened 2020-07-21
Speedup for rejectvisual_trial

NirOfir:
The same logic as in rejectvisual_channel: Use plot() at the beginning once and save the handles, then modify objects' properties as needed (see #1499).
I also made a small change in rejectvisual_channel to get rid of a redundancy I introduced in the previous fix there.


schoffelen:
Thanks.

PR #1508, opened 2020-07-27
ENH - clean up prepare_mesh_tetrahedral to fix regression errors

schoffelen:


schoffelen:
OK, cleaning up prepare_mesh_tetrahedral fixes the failing test functions.

TO DO: there's currently a hard coded assumption that the segmentation volume of the input mri is called 'seg', in case cfg.tissue is a numeric vector. I suspect that this is a remnant of old code. Also, I don't know whether the functionality of tetrahedral should be similar to hexahedral. If so, these functions need to be critically compared, and if needed, streamlined.

PR #1511, opened 2020-07-30
Fix default component selection in ft_dipolefitting

NirOfir:
The documentation states that if cfg.component is defined as empty, then all components are selected. However, a default selection wasn't implemented, so instead it would just choose no components, creating an empty Vdata which returned an error on line 307.


NirOfir:
Sorry, this PR is not ready yet. If cfg.component is defined as empty, cfg.model is set to regional because of lines 177-180, and only a single dipole is fit. On the other hand, I can't change the default value for cfg.component in the ft_getopt call, because that will affect cfg.model even if the data is not a component structure. I'm still working on it.

NirOfir:
I found a solution. Data type was determined later in the code explicitly anyway, so I just moved this bit up, and relied on its results for cfg.model defaults.

PR #1512, opened 2020-07-31
improvement of trf computations

schoffelen:


schoffelen:
As inspired by @sslaats observation that the computations were very slow, and after some profiling, some important temporal bottlenecks were removed. Also, an explicit protection of the user against herself was implemented, originally causing a potential massive duplication of redundant reflags, if inappropriately specified.

schoffelen:
Another note: some of the unavoidable administrative overhead (calls to ft_senstype etc) can be significantly sped up, if the information about the grad/elec is removed from the data, prior to calling the function. FieldTrip's inherent efficiency measures (i.e. working with persistent variables) don't work if 1) both a grad and elec are present in the data, and 2) if subsequent calls to these functions are with different input arguments. 1) can be avoided in this case, 2) cannot.

PR #1514, opened 2020-07-31
simple fix to print in ft_checkdata getting stuck for atlas

mcvinding:
`ft_checkdata` would not read labels from an atlas (MATLAB R2016a). `ft_info` in `line 1837` would throw an error when the name of the label was the maximum label length because `pad(tissuelabel{m}, width)` would return empty for reasons. Adding +1 to width fixed the issue.

### DIFF
--- a/compat/matlablt2016b/pad.m
+++ b/compat/matlablt2016b/pad.m
@@ -82,9 +82,7 @@ else
   assert(numel(c)==1);
   assert(ischar(c));
 
-  if length(s)>=n
-    return
-  else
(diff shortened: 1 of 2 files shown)


schoffelen:
Hi Mikkel, thanks for catching this. May I ask what the reason is when pad(tissuelabel{m}, width) returns empty?

mcvinding:
It turned out that `varargout ` never gets assigned in pad when the label name is the same length as `width` which instead stops the function at `line 86`. The new commit should fix this.

PR #1516, opened 2020-08-03
second try to fix break of simbio/tetraedral mesh code

schoffelen:


schoffelen:
OK, I kind of assumed that the intended functionality for the tetrahedral mesh creation is that it results in a tetrahedral mesh, in which each of the tets has an indexed tissue assigned to it. Until now, this did not work, because 1) the volume that entered the lower level code of iso2mesh was boolean (so no different tissues to be identified), and 2) (with 1) fixed, or perhaps also without 1) fixed) the iso2mesh functions failed. 

I fixed this by calling a different 'submethod' from iso2mesh, which seems to at least result in a tetrahedral mesh. But now, the problem is that the orientation of the tetrahedra, as checked by sb_test_ori, apparently is not good enough. I can imagine that the order of the cornerpoints of the tetrahedron should be according to some convention (right/left handed, similar as the order of the corners of a triangle, because that determines the direction of the normal). If this imagination of mine holds, I don't see why the sb_test_ori should return a boolean (which is an all pass/fail), and why this could not be changed into something like sb_fix_ori, which, for the affected elements, juggles around the order of the points such that the tetrahedra are oriented correctly.

In other words, and this is a question to @robertoostenveld, who are the main stakeholders of external/simbio, and can we go in and add some functionality there?

schoffelen:
OK, I think my latest commit fixed the orientation issue for the tetrahedral mesh, without tampering with external/simbio

schoffelen:
OK, ready for review @robertoostenveld . 

Documentation still needs to be done.

private/prepare_mesh_tetrahedral.m
```diff
@@ -86,11 +86,12 @@
 % this requires the external iso2mesh toolbox
 ft_hastoolbox('iso2mesh', 1);
 
-[node, elem, face] = vol2mesh(seg, 1:mri.dim(1), 1:mri.dim(2), 1:mri.dim(3), 2, 2, 0, 'cgalsurf');
+[node, elem, face] = vol2mesh(uint8(seg), 1:mri.dim(1), 1:mri.dim(2), 1:mri.dim(3), 1, 1, 0, 'cgalmesh');
+elem(:,1:4)        = meshreorient(node(:,1:3), elem(:,1:4));
 
 mesh = keepfields(mri, {'coordsys', 'unit'});
-mesh.pos = ft_warp_apply(mri.transform, node);
-mesh.tet = elem(:,[1 2 4 3]); % re-order elements
```
robertoostenveld, line 93: Hmm, I hope @schoffelen carefully considered that the reordering of elements was not needed any more. By just looking at the code, it is not possible to see whether this is a proper change.

schoffelen: Yes, I did. The reordering is now done in meshorient, which gets the orientation consistent. sb_test_ori now passes (which it didn't before, even with the magical [1 2 4 3] swap, because with the new mesh generation the orientation of the tetraeders was totally random (mix of left and right handed elements)). sb_test_ori kind of assumes all elements to be either left or right handed.

robertoostenveld:
I made some additional changes: mainly comments and help and extended the test script a bit. The `test_pull1427` script now looks as good as it gets, i.e. all meshes are aligned with the input segmentations. 

However, when comparing the refined mesh with the original segmentation, there are still shifts. This suggests that `ft_volumereslice` introduces some small shifts in the coordinates when interpolating the volumetric data on another volumetric grid. That is not to be solved with this PR.

```
git push upstream master
Enumerating objects: 130, done.
Counting objects: 100% (118/118), done.
Delta compression using up to 8 threads
Compressing objects: 100% (92/92), done.
Writing objects: 100% (92/92), 19.77 KiB | 3.29 MiB/s, done.
Total 92 (delta 70), reused 0 (delta 0)
remote: Resolving deltas: 100% (70/70), completed with 21 local objects.
   dd2ca8c01..36df25c38  master -> master 
```

robertoostenveld:
Oh, and relevant to mention is that I fixed the `surfaceonly` option in ft_plot_mesh

robertoostenveld:
I tried replicating the displacement error but failed. 
```
%%

seg3 = [];
seg3.tissue = zeros(3,3,3);
seg3.tissue(2,2,2) = 1;
seg3.tissuelabel = {'brain'};
seg3.dim = size(seg3.tissue);
seg3.unit = 'cm';
seg3.coordsys = 'ctf';
seg3.transform = eye(4);
seg3.transform(1,4) = -2;
seg3.transform(2,4) = -2;
seg3.transform(3,4) = -2;

% determine the range of the bounding box
source3 = ft_checkdata(seg3, 'datatype', 'source');

% this should be [0 0 0]
centerofmass = mean(source3.pos .* [source3.tissue source3.tissue source3.tissue])

cfg = [];
cfg.funparameter = 'tissue';
cfg.method = 'ortho';
cfg.location = [0 0 0]; % this is the center of the volume, in this plot the crosshair will go through the nearest voxel
ft_sourceplot(cfg, seg3);

%%

seg6 = [];
seg6.tissue = zeros(6,6,6);
seg6.tissue(3:4,3:4,3:4) = 1;
seg6.tissuelabel = {'brain'};
seg6.dim = size(seg6.tissue);
seg6.unit = 'cm';
seg6.coordsys = 'ctf';
seg6.transform = eye(4);
seg6.transform(1,1) = 0.5;
seg6.transform(2,2) = 0.5;
seg6.transform(3,3) = 0.5;
seg6.transform(1,4) = -3.5/2;
seg6.transform(2,4) = -3.5/2;
seg6.transform(3,4) = -3.5/2;

% determine the range of the bounding box
source6 = ft_checkdata(seg6, 'datatype', 'source');
% this should be [0 0 0]
centerofmass = mean(source6.pos .* [source6.tissue source6.tissue source6.tissue])

cfg = [];
cfg.funparameter = 'tissue';
cfg.method = 'ortho';
cfg.location = [0 0 0]; % this is the center of the volume, in this plot the crosshair will go through the nearest voxel
ft_sourceplot(cfg, seg6);

%%

cfg = [];
cfg.resolution = 0.5;
cfg.xrange = [min(voxpos3(:,1))-0.5 max(voxpos3(:,1))+0.5];
cfg.yrange = [min(voxpos3(:,2))-0.5 max(voxpos3(:,2))+0.5];
cfg.zrange = [min(voxpos3(:,3))-0.5 max(voxpos3(:,3))+0.5];
seg3to6 = ft_volumereslice(cfg, seg3);


% determine the range of the bounding box
source3to6 = ft_checkdata(seg3to6, 'datatype', 'source');
source3to6.tissue(isnan(source3to6.tissue)) = 0;
% this should be [0 0 0]
centerofmass = mean(source3to6.pos .* [source3to6.tissue source3to6.tissue source3to6.tissue])

%%

cfg = [];
cfg.resolution = 0.5;
cfg.xrange = [min(voxpos6(:,1))-0.25 max(voxpos6(:,1))+0.25];
cfg.yrange = [min(voxpos6(:,2))-0.25 max(voxpos6(:,2))+0.25];
cfg.zrange = [min(voxpos6(:,3))-0.25 max(voxpos6(:,3))+0.25];
seg6to3 = ft_volumereslice(cfg, seg6);

% determine the range of the bounding box
source6to3 = ft_checkdata(seg6to3, 'datatype', 'source');
source6to3.tissue(isnan(source6to3.tissue)) = 0;
% this should be [0 0 0]
centerofmass = mean(source6to3.pos .* [source6to3.tissue source6to3.tissue source6to3.tissue])
```

for all carefully crafted options the `centerofmass` stays in [0 0 0]

I wonder whether it might be a problem with `ft_plot_ortho` instead.

schoffelen:
spuit 11: what happens with the center-of-mass if you use cfg.method = 'nearest' for ft_volumereslice (which is what you'd need to keep the integer indexing of the segmentation)?

robertoostenveld:
I will check it out. I do indeed suspect that ft_volumereslice is inaccurate with grid positions in some cases.

PR #1523, opened 2020-08-14
improve organization of ft_volumenormalise to allow for custom templates

robertoostenveld:
This takes #1450 as the starting point but takes it further in reorganizing the code (especially the input data annd cfg checking), simplifying the complex `if-elseif-else` statements, and in general refactoring the code. 

Overall, there should not really be any functional differences for normal use cases, although some small bugs have been fixed. I hope that the behaviour of these bugs is not being relied upon by users' scripts.


fileio/ft_write_mri.m
```diff
@@ -47,6 +48,19 @@
 transform     = ft_getopt(varargin, 'transform', eye(4));
 spmversion    = ft_getopt(varargin, 'spmversion', 'spm12');
 dataformat    = ft_getopt(varargin, 'dataformat'); % FIXME this is inconsistent with ft_read_mri, which uses 'format'
+unit          = ft_getopt(varargin, 'unit');
+
+% convert the input to the desired units
+if ~isempty(unit)
```
robertoostenveld, line 54: this section makes the API consistent with ft_write_headshape

ft_prepare_sourcemodel.m
```diff
@@ -46,7 +46,6 @@
 % inversely warped from MNI coordinates to the individual subjects MRI.
 % It uses the following configuration options:
 %   cfg.mri             = structure with anatomical MRI model or filename, see FT_READ_MRI
-%   cfg.warpmni         = 'yes'
```
robertoostenveld, line 49: the `cfg.warpmni` option is now deprecated, people should use cfg.method instead

utilities/ft_checkopt.m
```diff
@@ -20,6 +20,7 @@
 %   'struct'
 %   'function_handle'
 % Furthermore, the following custom types can be specified
+%   'empty'
```
robertoostenveld, line 23: the allowed type `empty` was never documented, but is implemented

robertoostenveld:
I worked on this for more than two days, and as far as I can tell everything works now. The core functionality should not have changed, but the code is refactored and should now in principle allow for custom templates. However, that new functionality is something that I have not tested yet. 

A new test script is needed to ensure that inverse normalization (from MNI to single-subject) works. Note also that for spm12 and new/mars  it requires a template TPM which makes the inverse normalisation more challenging.

robertoostenveld:
Note that in the test script around https://github.com/fieldtrip/fieldtrip/blob/67f66b1829715e70427a3ad7438f9f06709f375b/test/test_ft_volumenormalise.m#L152 there is now a section that step-by-step converts from individual CTF, to approximate ACPC, to MNI and back. That might be valuable for some online documentation as a FAQ.

PR #1526, opened 2020-08-19
Adapt ft_timelockgrandaverage for data containing nan's .

helenacockx:


robertoostenveld:
I extended the test script, now with a 1-trial, 2-trial and a 3-trial data structure.

The checks for `method='across'` appear to be OK to me, but the within method is not. It even returns negative variance. Note that I am not sure that this is due to your changes, but now that we have detected it, we should deal with it. Can you have a look?

robertoostenveld:
Hi @bmaess, just to let you know that we are working on ft_timelockgrandaverage in relation to it supporting nans (that indicate artifacts) in the input averages that it gets.

PR #1529, opened 2020-08-27
fixed cfg.dip problem for custom signal properties

Richter2552:
This is a fix for the cfg.dip problem described in issue1505


schoffelen:
Thanks, Craig!

Richter2552:
My pleasure!!

PR #1531, opened 2020-08-27
Fix qsub slurm support

octomike:
We migrated to Slurm (16.x) and this PR adds three basic changes to make things work (at all)

1. I believe using `srun` or worse `nohup srun` is conceptually broken, since `qsubfeval` should be non-blocking, right?
2. Use a proper check for a completed slurm job
3. actually make use of the `timreq` and `memreq` options

### DIFF
--- a/qsub/qsubfeval.m
+++ b/qsub/qsubfeval.m
@@ -378,11 +378,11 @@ switch backend
     end
 
     if ~isempty(timreq) && ~isnan(timreq) && ~isinf(timreq)
-      submitoptions = [submitoptions sprintf('--time=%d ', round((timreq+timoverhead)/60))];
+      submitoptions = [submitoptions sprintf(' --time=%d ', round((timreq+timoverhead)/60))];
     end
 
     if ~isempty(memreq) && ~isnan(memreq) && ~isinf(memreq)
-      submitoptions = [submitoptions sprintf('--mem=%.0f ', round((memreq+memoverhead)./1024^2))];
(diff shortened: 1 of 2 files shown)


robertoostenveld:
`qsubfeval` should indeed be non-blocking. The slurm support was contributed quite some time ago (through email, not as PR) by some people at the https://www.esi-frankfurt.de, especially Kai Rönnburg. He seems not to be on github, but I will reach out to him by email to ask whether they still use it. 

If not, I am fine adopting your changes. I am not capable of testing myself, therefore I rely on active users like you.

robertoostenveld:
I have sent an email to Kai.

octomike:
> I am not capable of testing myself, therefore I rely on active users like you.

I'm happy to test this and other Slurm-related things.

robertoostenveld:
I checked with Kay, he wrote

> at some point the changes we made to fieldtrip were no longer committed. Additionally the way of handling jobs
changed in qsubfeval and the way slurm does the tracking of jobs did not really fit into the scheme used with fieldtrip.
> 
> From my understanding the slurm/fieldtrip implementation in the repo is quite old and resembles more a proof of
concept. We used srun to better handle the job tracking and than moved to batch processing using sbatch.
>  
> I could check for the last version of code we used on the cluster, but some time ago we rewrote the whole
submission engine into slurm bypassing qsub totally and streamlined the code.

so the `fieldtrip/qsub` code is not used any more at the ESI. That also means that you are free to make the changes required to getit to run on newer and more standard SLURM installations.

robertoostenveld (approved):
this looks good to me

PR #1533, opened 2020-08-31
improve plotting, especially for NIRS data

robertoostenveld:


robertoostenveld:
this also makes the `cfg.figure`option more consistently supported by all high-level FT functions. The default is to create a new figure, but you can also specify to clean and reuse the current figure, or to add it to the current figure without cleaning, or to add it to a specific figure handle

PR #1540, opened 2020-09-07
improve support for Homer nirs format and for Snirf format   

robertoostenveld:
This PR adds support for writing both formats and includes some changes to the reading (e.g. reading from Snirf now also returns the opto structure). 

Furthermore, it implements homer2fieldtrip and fieldtrip2homer to convert the nirs structure in-memory (just like how it is done for EEGLAB).


robertoostenveld:
@helenacockx I have asked you to review this pull request. I don't expect you to read though all code changes (although you are of course welcome to do so), but could you read (online, in the github web interface) the `test_homer` and the `test_snirf` functions to check whether they make (logically) sense to you?

My idea in implementing them is that repeatedly reading, writing, reading, ... should not change the interpretation of the data content. The only differences that happen is in channel labels (either Tx/Rx or S/D) and in the representation of the stimuli (as a continuous channel in Homer, as an event table in Snirf). 

The reading and writing of events (in the FT_READ_EVENT representation) still requires some attention.

robertoostenveld:
Good idea to update the test script such that the matching channels are compared.  Also good idea to use S/D rather than Tx/Rx for Homer and SNIRF channel names. I will implement these.

The question on multiple SNIRF data blocks is better addressed elsewhere.

robertoostenveld:
when reading from Artinis, writing to snirf, and reading that again, there are still some differences in the name of channels and in the type/value of events. All numeric comparisons now work fine.

Issue #1541, opened 2020-09-10
label rejection during ft_preprocessing


Nichalas:
[dat.zip](https://github.com/fieldtrip/fieldtrip/files/5200605/dat.zip)
Hi, 
It seems that ft_preprocessing function rejects labels from the data structure (dat; attached)
**To Reproduce**
Code for reproducing: 
        
cfg=[];      
cfg.length=2;      
data2=ft_redefinetrial(cfg,dat);
        
cfg=[];     
cfg.demean='yes';             
data3=ft_preprocessing(cfg,data2);

**Expected behavior**
data3 should contain the same amount of labels as starting dat.

**Environment and versions (please complete the following information):**
 - Operating System: Ubuntu 18.04.4 LTS
 - MATLAB version R2019b
 - FieldTrip version most recent master branch from github"

schoffelen:
dear @Nichalas

This is probably not the most appropriate place to ask this question. Have your read our contributing guidelines? 

https://github.com/fieldtrip/fieldtrip/blob/master/.github/CONTRIBUTING.md

I think you should ask this on the discussion list (if this hasn't been reported before), and you should be a little bit more informative about the problem that you perceive. 'it seems that ft_preprocessing function rejects labels from the data structure' is a bit too vague. At least, it doesn't inspire me to download the zip-file and look into it.

robertoostenveld:
I did download the data and had a look at it. I can confirm that `data2` still has 12 channels, but that `data3` only has 6. That is caused by the original `dat` having duplicate channel names:

```
>> dat.label'
ans =
  12×1 cell array
    {'Hippocampus_L_1'     }
    {'Hippocampus_L_2'     }
    {'Hippocampus_L_3'     }
    {'Hippocampus_L_1'     }
    {'Hippocampus_L_2'     }
    {'Hippocampus_L_3'     }
    {'Cerebellum_Crus1_L_1'}
    {'Cerebellum_Crus1_L_2'}
    {'Cerebellum_Crus1_L_3'}
    {'Cerebellum_Crus1_R_1'}
    {'Cerebellum_Crus1_R_2'}
    {'Cerebellum_Crus1_R_3'}
```

schoffelen:
So, in other words, this is a user error, right?

robertoostenveld:
It is difficult to say whether this qualifies as a bug in the code, or a bug in the data. I would say that the data is incorrect. 

[...]

PR #1542, opened 2020-09-10
ENH - improve error handling in case of inconsistent input data

robertoostenveld:
This deals with the problem reported in #1541


robertoostenveld:
I also had a look at ft_datatype_segmentation and ft_datatype_parcellation, they both include a call to https://github.com/fieldtrip/fieldtrip/blob/master/utilities/private/fixsegmentation.m which I believe (*) deals with this if it were an anatomical or functional atlas.

(*) although I did not check in detail.

Issue #1473, opened 2020-06-30
make the prewhitening consistent across the codebase


robertoostenveld:
This applies to ft_dipolefitting, ft_denoise_prewhiten and ft_inverse_mne

schoffelen:
goed idee buur!

Issue #1307, opened 2020-02-15
Inconsistency in minimumnormestimate usage of lambda


rbruna:
Hello,

I have been looking at the code for the calculation of the minimum norm and there is an inconsistency, and some issues.

First, the Cholesky factorization approach:
<pre>
      Rc      = chol(R, 'lower');
      [U,S,V] = svd(A * Rc, 'econ');
      s  = diag(S);
      ss = s ./ (s.^2 + lambda);
      w  = Rc * V * diag(ss) * U';
</pre>
 is only used with pre-whitening. Why is this? This approach should be more stable numerically in any case, is it not?

Either way, even with the current implementation, which is valid, the usage of lambda is inconsistent between the pre-whitening approach (above) and the non-pre-whitening one:
<pre>
      denom = (A*R*A'+(lambda^2)*C);
      if cond(denom)<1e12
        w = R * A' / denom;
      else
        fprintf('taking pseudo-inverse due to large condition number\n');
        w = R * A' * pinv(denom);
      end
</pre>

In the later case the lambda is squared (as should be, according with Liu 2002, Lin 2004 and even the MNE website: https://mne.tools/0.17/manual/source_localization/inverse.html#computation-of-the-solution), but in the case using Cholesky factorization the square is missing.

This is, when using pre-whitening, the real value of lambda is the square root of the desired value. And it creates an inconsistency between the two different approaches. The value of lambda should be squared there, too.

schoffelen:
Hi Ricardo,

Thanks for working on this. The minimumnormestimate code for historical reason has the original Lin implementation, as well as the numerically poorly behaving pinv() version.
That being said, all implementations have been written, carefully looking at the MNE documentation, and also at the equivalent MATLAB implementation in brainstorm (who also looked closely at the documentation, and listened carefully to John Mosher and Matti Hamalainen, it seems). 
Of course, this doesn't guarantee that a small error has slipped in (e.g. omitting a square).

Yet, I am not entirely sure, whether your perceived inconsistency is actually inconsistent at all. The chol-and-svd version, essentially computes the inverse by operating on the svd of A, whereas the Lin implementation explicitly inverts (A*R*A'). In other words, the singular values in the first case, relate to the (not explicitly, yet implicitly used) singular values in the second case, as a square root. 
I think that for this reason the regularization matrix in the second case weighs in with the square of lambda in the second case, in comparison to the first case.

I ran a quick numerical simulation, which seems to confirm this:

[...]

PR #1544, opened 2020-09-11
mne prewhiten

schoffelen:
This is a PR that eventually should address #1473.


schoffelen:
The first step is to inventorize for ft_inverse_mne, whether the 'prewhiten' option behaves the same as a call to ft_denoise_prewhiten, followed by ft_sourceanalysis

schoffelen:
Currently, the test function checks whether ft_sourceanalysis (with 'mne'+'prewhiten') results in the same output as a sequential call to ft_denoise_prewhiten/ft_sourceanalysis (with 'mne' + 'prewhiten'/'no-prewhiten').
With prewhitened data in the input (i.e. an eye(nchan) noise cov matrix), AND cfg.mne.prewhiten = 'yes', the results are the same.

With prewhitened data in the input AND cfg.mne.prewhiten = 'no', the results will only be the same IF cfg.mne.lambda is square rooted :(. this is the consequence of the fact that the estimation of the inverse operator in ft_inverse_mne applies a different formula, and is incosistent in its treatment of lambda (squared vs. non-squared). See #1307

PR #1547, opened 2020-09-17
added blink start and end Eyelink messages in read_eyelink_asc.m

kloosterman:
read_eyelink_asc.m does not read the start and end times of blink messages (sblink and eblink) from the asc files. I added this so that these messages can be used in subsequent analysis steps, such as blink interpolation.


schoffelen (approved):
looks clean and good to me. Thanks!

PR #1549, opened 2020-09-18
incorporate some of jm's changes in external/dss

schoffelen:
motivation and justification:

-the external party that wrote the toolbox has not been actively maintaining it, so no newer version to be expected.
-jm's changes are needed for a more memory efficient implementation, and incorporates a new denoising function that is aware of trial boundaries (as opposed as the original implementation from which it is derived).

-the update is needed, because I'd like to finalize some documentation that is using this functionality, to efficiently unmix multichannel data, in order to identify the heartbeat (e.g. to be used for data cleaning, or to extract the heartrate)

### DIFF
--- a/external/dss/VERSION
+++ b/external/dss/VERSION
@@ -1,1 +1,10 @@
-1-0
+1-0, with some modifications made to the following functions, in order to support cell-data as input:
+
+dss_core_defl.m
+dss_core_pca.m
+dss_core_symm.m
+dss_preprocess.m
+dss_sphere.m
+pre_sphere.m
+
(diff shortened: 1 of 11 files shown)


schoffelen:
This PR will also upgrade one of Johanna's still open projects on bugzilla, which was to write an example script about how to efficiently remove the BCG from EEG-fMRI data. (I will rewrite it slightly, in order to use ArtifactRemoval.ds) The principle is the same

schoffelen:
The example script page is online now, let's merge, and see whether the test function runs through on the cluster (it works on my end).

PR #1552, opened 2020-09-29
ENH - bypass expensive calls to ft_channelselection if not needed

schoffelen:


schoffelen:
the graphical diff looks a bit strange, but the intention is hopefully clear. I cannot think of any unwanted side effects, but would appreciate a quick check by @robertoostenveld

Issue #1319, opened 2020-02-23
Atlas VTPM not correct


yingjoeyzhou:
`load ([ftpath,'/template/atlas/vtpm/vtpm.mat'], 'vtpm');`

Note here ftpath is the path in DCCN cluster, which i would assume to have the most updated version of fieldtrip.

While `numel( vtpm.tissuelabel ) = 50`, `max( vtpm.tissue(:) ) = 25` and `min( vtpm.tissue(:) ) = 0`.

Needs to be fixed.

schoffelen:
Please go ahead and fix it. I don't know about this atlas, perhaps you could first check what the expected number of parcels is, and whether the tissuelabel is consistent with this number, or whether the values in tissue are consistent with this label. Then, you can try and find out why one of the two is inconsistent. Next, we can try and find a way to make the atlas consistent (and/or update the reading code so that all is aligned again.)

yingjoeyzhou:
ok, I will try to fix it. I will label it as work in progress.

schoffelen:
One obvious cause of a mismatch could be that the tissuelabel distinguishes the left and right hemispheres, while the tissue-field does not.

schoffelen:
Note: if that is true, it is straightforward to fix, but don't forget to update the description on the website

schoffelen:
Please confirm that the vtpm.mat consists of a blend of the original maxprob volumes lh/rh, as present in the Probatlas_v4/subj_vol_all. To me this means that the volumetric data is consistent with the distributed maxprob atlas, only the indices are not consistent with the labels, and as of yet the labels do not distinguish left from right it seems.

schoffelen:
Also, the tissuelabel is a row vector as of yet, should be column vector.

yingjoeyzhou:
I am currently working on it ~ I will send you the updated atlas and the
code to make it by the end of today.

schoffelen:
I suggest at minimum to update the template/atlas/vtpm/vtpm.mat file to be internally consistent.
Probably in terms of file size it might be good to consider to keep  the tissue field in int8.

Starting from the original volumetric maxprob files:

x = ft_read_mri('maxprob_vol_lh.nii.gz');
y = ft_read_mri('maxprob_vol_rh.nii.gz');

z = rmfield(z,'anatomy');
y.anatomy(y.anatomy~=0) = y.anatomy(y.anatomy~=0)+25;
z.tissue = x.anatomy+y.anatomy;

[...]

PR #1556, opened 2020-10-02
ENH - improved handling of probabilistic fsl derived atlases

schoffelen:
The probabilistic atlases in fsl are in int8, and expressed in percent values (it seems that is: 100 = max). This is not recognized by FT as a probabilistic representation.

### DIFF
--- a/fileio/ft_read_atlas.m
+++ b/fileio/ft_read_atlas.m
@@ -1993,6 +1993,15 @@ switch fileformat
           fn = strrep(fn, ')', '');
           
           atlas.(fn) = tmp.anatomy(:,:,:,m);
+          if ~isa(atlas.(fn), 'double') && ~isa(atlas.(fn), 'single')
+            % ensure that the probabilistic values are either double or
+            % single precision, do single precision to save memory
+            atlas.(fn) = single(atlas.(fn));
+          end
+          if any(atlas.(fn)(:)>1)
+            % convert to probability values, assuming 100 to be max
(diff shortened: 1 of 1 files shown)


schoffelen:
I solved it here for the fsl based atlases, but this reminds me of a previous issue #1319, related to the vtpm.mat atlas (which is maxprob based, but for which also probabilistic maps exist (which are not part of fieldtrip).

Consider whether this needs to be addressed for the vtpm atlas as well.

fileio/ft_read_atlas.m
```diff
@@ -1993,6 +1993,15 @@
           fn = strrep(fn, ')', '');
           
           atlas.(fn) = tmp.anatomy(:,:,:,m);
+          if ~isa(atlas.(fn), 'double') && ~isa(atlas.(fn), 'single')
+            % ensure that the probabilistic values are either double or
+            % single precision, do single precision to save memory
+            atlas.(fn) = single(atlas.(fn));
+          end
+          if any(atlas.(fn)(:)>1)
```
robertoostenveld, line 2001: would it not be better to also explicitly test whether values are below 100? If >1 and >100 I would be confused, and dividing by 100 would not be the best option.

schoffelen: good point. I thought about that, but decided for the 'binnenbocht' right now.

but adding a && all(atlas.(fn)(:)<=100) to the if clause would do it, right?

PR #1560, opened 2020-10-08
implement a ft_virtualchannel function

schoffelen:


robertoostenveld:
the development looks good!

private/univariate2bivariate.m
```diff
@@ -31,7 +31,7 @@
 %    GNU General Public License for more details.
 %
 %    You should have received a copy of the GNU General Public License
-%    along with FieldTrip. If not, see <http://www.gnu.org/licenses/>.
+%    along with FielddTrip. If not, see <http://www.gnu.org/licenses/>.
```
Spaak, line 34:
```suggestion
%    along with FieldTrip. If not, see <http://www.gnu.org/licenses/>.
```

ft_virtualchannel.m
```diff
@@ -0,0 +1,382 @@
...
+        cnt = cnt+1;
+        montage.labelnew{cnt} = sprintf('%s_dipole%03d_orientation%03d', cfg.parcel{i}, k, kk);
+      end
+    end
+    bname = matlab.lang.makeValidName(cfg.parcel{i}); % FIXME not sure how universal this is.
+    
+  end
+  
+  % apply the montage to the numeric data, but make an exception for the
+  % svd method: this one is much more efficient, in that its implementattion
```
Spaak, line 233:
```suggestion
  % svd method: this one is much more efficient, in that its implementation
```

ft_virtualchannel.m
```diff
@@ -0,0 +1,382 @@
...
+  % operates on the covariance matrix directly, bypasses ft_componentanalysis, 
+  % and does not require the intermediate (possibly memory greedy, and slow)
+  % data representation.
+  if ~isequal(cfg.method, 'svd')
```
Spaak, line 237: Might be clearer with `if isequal(cfg.method, 'svd')` first as the special case, and with the default case last in the `else` block.

ft_virtualchannel.m
```diff
@@ -0,0 +1,382 @@
...
+    % sure that the to-be-computed covariance matches the labels, not sure
+    % whether this is needed, this could probably also be achieved by doing
+    % a check on the input data's labels, related to the source.filter
+    % labels
+    assert(isequal(i1(:)', 1:numel(data.label)));
+    
+    unmixing{1,i}(:,i1) = montage.tra(:,i2);
+    topolabel{1,i}      = data.label(i1);
+    
+    if ~exist('tlck', 'var')
```
Spaak, line 298:
```suggestion
    if ~exist('tlck', 'cov')
```
It's possible that `tlck` does have a `var` but no `cov`.

ft_virtualchannel.m
```diff
@@ -0,0 +1,382 @@
...
+        unmixing{i} = tmpdata{i}.unmixing * unmixing{i};
+        
+        for k = 1:numel(tmpdata{i}.label)
+          if usepos
+            str = sprintf('virtualchannel%03d', i);
+          else
+            str = cfg.channel{i};
+          end
+        end
+        tmpdata{i}.label{k} = sprintf('%s_%s%03d', str, cfg.method, k);
```
Spaak, line 272: I suspect the most common use case here will be `ncomp == 1`, right? In that case, people might not care for the `runicaXXX` suffix, and just want the parcel (/pos) label. So I propose:

```suggestion
        if cfg.numcomponent == 1
          tmpdata{i}.label{k} = str;
        else
          tmpdata{i}.label{k} = sprintf('%s_%s%03d', str, cfg.method, k);
        end
```

ft_virtualchannel.m
```diff
@@ -0,0 +1,382 @@
...
+    unmixing{1, i} = u(:,1:ncomp)' * unmixing{1, i};
+    label{1, i}    = cell(ncomp,1);
+    if usepos
+      str = sprintf('virtualchannel%03d', i);
+    else
+      str = cfg.parcel{i};
+    end
+    for k = 1:ncomp
+      label{1, i}{k} = sprintf('%s_svd%03d', str, k);
+    end
```
Spaak, line 334: Same as above, special case for `numcomponent == 1`.

```suggestion
    if ncomp == 1
      label{1, i}{1} = str;
    else
      for k = 1:ncomp
        label{1, i}{k} = sprintf('%s_svd%03d', str, k);
      end
    end
```

Spaak (commented):
Great addition. I made some code suggestions inline.

ft_virtualchannel.m
schoffelen: well, I guess that my intention would be that it might still be relevant to reflect in the label the heuristic used for the orientation estimation, in which case 'pca'/'dss' etc still are relevant. I agree that the k can be dropped if ncomp=1

ft_virtualchannel.m
schoffelen: indeed, but the variable needs to exist, so a separate check needs to be done using isfield.

ft_virtualchannel.m
schoffelen: same comment to this as above :)

ft_virtualchannel.m
schoffelen: good suggestion, thanks

ft_virtualchannel.m
Spaak: Nvm, I was not reading properly here.

ft_virtualchannel.m
Spaak: I'm inclined to say the orientation estimation method does not need to be explicitly reflected in the virtual channel label. (We also don't have a `freq.powspctrm_hanning` or `freq.powspctrm_wavelet`.) In most cases I expect people to be happy/happier with `data.label = {'Occipital_R_4', ...}` than `data.label = {'Occipital_R_4_runica', ...}`. But, having said that, I don't have any strong objections either ;)

Note, by the way, I think you can accept my suggestions directly in GitHub (and then pull them locally), so you don't have to write them again and push (like you did with the typo).

schoffelen: That's a fair argument. And I can imagine that a person who cares about this, would probably know how to extract the information (assuming that all goes well with the cfg tracking...)

schoffelen: and yes, I of course could just have accepted the typo fixes, didn't think of that. Thanks.

schoffelen:
The basic functionality is in place, and works as expected according to the added test function.
Can be merged, as far as I am concerned, unless it requires conceptual discussion prior to that

Issue #1554, opened 2020-10-01
Add .amplitude field to ft_spike_maketrials for segmentation into trials, and overwrite .samples field in new trial dimension


stephenwhitmarsh:
I'd like ft_spike_maketrials to also segment an additional amplitude field - when present - into trials.

Currently there are some try-catch statements that add a .waveform, .unit or .fourierspctrm field to the spike structure by ft_spike_maketrials. It would be easy to add an .amplitude field here.

Additionally, these trial-based fields currently overwrite existing 'raw' fields in the spike structure. However, .samples field stays in the original dimension, i.e. does not get segmented. 

The most consistent to me would be to create a trial-based .samples field, and overwrite the existing one. Then also segment into trials any other fields that have the same dimensions as the input raw structure. This would make the dimensions of the data consistent within a single data structure. 

Currently fields to be segmented are hard-coded. I am unsure / have no opinion about what would be a better. Hardcoded is easy (already done on my side). But I can make a PR of either.

stephenwhitmarsh:
To clarify further - an unusual aspect of the current output is that the total number of indices in the trial segmented fields have a different number of elements than the remaining .samples field (i.e. when not all spikes are selected, or trials overlap), which then becomes orphaned, i.e. cannot be indexed in any meaning way anymore.

robertoostenveld:
can you explain how `amplitude` relates to the existing fields as documented/defined in https://github.com/fieldtrip/fieldtrip/blob/master/utilities/ft_datatype_spike.m

I think (recall) that the spike data structures predate the more explicit definition of the dimord. Although the dimord is in general not present explicitly, we have the function https://github.com/fieldtrip/fieldtrip/blob/master/private/getdimord.m. It would be good to have clear what the dimord is of each of the fields in the `spike` data structure.

stephenwhitmarsh:
Amplitude would describe the amplitude of a spike, either real (in uV) or e.g. normalized to a template, and will have the same dimensions as e.g. the `time`, `trial `or `timestamp`. These would all have a `dimord `of "{chan}_spike". Of the remaining fields, we have `trialtime `and `sampleinfo`, both "rpt", and `waveform`, which is already described in ft_datatype_spike as "{chan}_lead_time_spike"

robertoostenveld:
[...]

PR #1561, opened 2020-10-08
test for issue #1554 concerning parsing of extra fields in ft_spike_m…

stephenwhitmarsh:
…aketrials


schoffelen (changes requested):
I am right in concluding that this PR contains the exact same test function as the other PR that is related to this #1562. If so, could you close #1561, and then we can take if from there

stephenwhitmarsh:
yep - okidoki

PR #1563, opened 2020-10-13
plot events as boxes in ft_databrowser when duration is > 0

helenacockx:
Events were previously plotted as lines, regardless of their duration. With the proposed adaptation, durations of events will also be visible in ft_databrowser. When duration of the event =0, the event will still be plotted as a line.


data2bids.m
```diff
@@ -920,7 +960,7 @@
 
 need_events_tsv       = need_events_tsv       || need_meg_json || need_eeg_json || need_ieeg_json || need_emg_json || need_exg_json || need_nirs_json || need_eyetracker_json || need_motion_json || (contains(cfg.outputfile, 'task') || ~isempty(cfg.TaskName) || ~isempty(cfg.task)) || ~isempty(cfg.events);
 need_channels_tsv     = need_channels_tsv     || need_meg_json || need_eeg_json || need_ieeg_json || need_emg_json || need_exg_json || need_nirs_json;
-need_coordsystem_json = need_coordsystem_json || need_meg_json || need_electrodes_tsv;
+need_coordsystem_json = need_coordsystem_json || need_meg_json || need_electrodes_tsv || need_nirs_json;
```
robertoostenveld, line 963: need_nirs_json is already on line 962

ft_databrowser.m
```diff
@@ -1806,9 +1806,9 @@ function redraw_cb(h, eventdata)
       
       % compute the time of the event
       eventtim(ievent) = (event(ievent).sample-begsample)/opt.fsample + opt.hlim(1);
-      
-      lh = ft_plot_line([eventtim(ievent) eventtim(ievent)], [-1 1], 'tag', 'event', 'color', eventcol{ievent}, 'hpos', opt.hpos, 'vpos', opt.vpos, 'width', opt.width, 'height', opt.height, 'hlim', opt.hlim, 'vlim', [-1 1]);
-      
+      % plot events as box or line (when duration = 0)
```
robertoostenveld, line 1809: I am afraid that in some cases event.duration would be [] (i.e. unknown) instead of 0. To make the code robust, I propose that you add something like

```
if isempty(event(ievent).duration) || event(ievent).duration==0
  plot line
else
  plot box  
end
```

robertoostenveld: also, lines will be faster than boxes (which are 4 lines) in the cases when the duration is 0 or []

robertoostenveld (changes requested):
the suggested changes include NIRS-related changes to `data2bids`. I know you were working on them, and I would not mind merging them as well if you feel that you are done with those changes (mainly metadata fields). 

Please see the specific comments that I made

ft_databrowser.m
helenacockx: Okay, I'll change that in a new commit!

data2bids.m
helenacockx: woops, that was indeed not the purpose that both branches got mixed up... I am not really done on this function, but will try to work on it later this week.

helenacockx: I don't really understand what you mean... line 962 is channels.tsv while line 963 is about need_coordsystem.json, right?

PR #1565, opened 2020-10-15
plot events as boxes in ft_databrowser when duration is > 0

helenacockx:
Events were previously plotted as lines, regardless of their duration. With the proposed adaptation, when an event has a duration >0, event durations will be plotted as a box. When duration of the event =0, the event will still be plotted as a line.


robertoostenveld:
this contains the good parts of #1563

PR #1566, opened 2020-10-15
append_common should append chans in data with dimord chan_freq_time

dieloz:
This pull request fixes append_common to append channels in freq data with dimord 'chan_freq_time'. Run this to check that powspctrm is not in the output while freq.label is correct:

freq1=[];
freq1.label={'chan1'};
freq1.dimord='chan_freq_time';
freq1.freq=[1:20];
freq1.time=[0:0.05:0.5];
freq1.powspctrm = rand(1,size(freq1.freq,2),size(freq1.time,2));

freq2=freq1;
freq2.label={'chan2'};

cfg=[];
cfg.appenddim = 'chan';
cfg.parameter = 'powspctrm';
freq = ft_appendfreq(cfg,freq1,freq2)

freq = 

  struct with fields:

     label: {2×1 cell}
      freq: [1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20]
      time: [0 0.0500 0.1000 0.1500 0.2000 0.2500 0.3000 0.3500 0.4000 0.4500 0.5000]
    dimord: 'chan_freq_time'
       cfg: [1×1 struct]

### DIFF
--- a/private/append_common.m
+++ b/private/append_common.m
@@ -131,19 +131,12 @@ switch cfg.appenddim
             chansel = match_str(varargin{j}.label, oldlabel{j});
             data.(cfg.parameter{i})(:,chansel,chansel) = varargin{j}.(cfg.parameter{i})(:,chansel,chansel);
           end
-         
-        case {'chan_freq_time'}
+           
+        case {'chan' 'chan_time' 'chan_freq' 'chan_freq_time'}
           data.(cfg.parameter{i}) = nan(dimsiz);
           for j=1:numel(varargin)
             chansel = match_str(varargin{j}.label, oldlabel{j});
             data.(cfg.parameter{i})(chansel,:,:) = varargin{j}.(cfg.parameter{i})(chansel,:,:);
-          end
-          
-        case {'chan' 'chan_time' 'chan_freq'}
-          data.(cfg.parameter{i}) = nan(dimsiz);
-          for j=1:numel(varargin)
-            chansel = match_str(varargin{j}.label, oldlabel{j});
-            data.(cfg.parameter{i})(chansel,:) = varargin{j}.(cfg.parameter{i})(chansel,:);
           end
           
         case {'rpt_chan_time' 'subj_chan_time' 'rpt_chan_freq' 'rpttap_chan_freq' 'subj_chan_freq'}
--- a/test/test_pull1566.m
+++ b/test/test_pull1566.m
@@ -22,7 +22,12 @@ freq2_s = ft_selectdata(cfg,freq2);
 cfg=[];
 cfg.appenddim = 'chan';
 cfg.parameter = 'powspctrm';
-freqA = ft_appendfreq(cfg,freq1_s,freq2_s)
+freqA = ft_appendfreq(cfg,freq1_s,freq2_s);
+
+% The below assertions are 'incorrect', i.e. they should be throwing an
+% error if this issue is solved
+assert(isequal(freqA.freq, [0;0]));
+assert(~isfield(freqA, 'powspctrm'));
 
 % another situation
 freq1=[];
@@ -45,12 +50,16 @@ cfg.appenddim = 'chan';
 cfg.parameter = 'powspctrm';
 freqB = ft_appendfreq(cfg,freq1_s,freq2_s);
 
+% This scenario yields an OK structure
+assert(isfield(freqB, 'powspctrm'));
+assert(isequal(freqB.freq, 0));
+
 % another situation
 freq1=[];
 freq1.label={'chan1'};
 freq1.dimord='chan_freq';
 freq1.freq=[1:2];
-freq1.powspctrm = rand(2,size(freq1.freq,2));
+freq1.powspctrm = rand(1,size(freq1.freq,2));
 
 freq2=freq1;
 freq2.label={'chan2'};
@@ -58,4 +67,10 @@ freq2.label={'chan2'};
 cfg=[];
 cfg.appenddim = 'chan';
 cfg.parameter = 'powspctrm';
-freqC = ft_appendfreq(cfg,freq1,freq2)
+freqC = ft_appendfreq(cfg,freq1,freq2);
+
+% This scenario seems to correctly append the powerspectra, but the
+% freq-field is incorrect, the second assertion should throw an error once
+% this has been solved.
+assert(isequal(freqC.powspctrm, [freq1.powspctrm;freq2.powspctrm]));
+assert(isequal(freqC.freq, [1 2;1 2]));
--- a/test/test_pull1566.m
+++ b/test/test_pull1566.m
@@ -74,3 +74,11 @@ freqC = ft_appendfreq(cfg,freq1,freq2);
 % this has been solved.
 assert(isequal(freqC.powspctrm, [freq1.powspctrm;freq2.powspctrm]));
 assert(isequal(freqC.freq, [1 2;1 2]));
+
+% the thing that seems to go wrong is the following:
+% line 94 in append_common calls ft_selectdata with multiple inputs (and
+% outputs). Once there is just a single channel in the input data, it seems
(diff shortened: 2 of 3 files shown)


private/append_common.m
```diff
@@ -131,6 +131,13 @@
             chansel = match_str(varargin{j}.label, oldlabel{j});
             data.(cfg.parameter{i})(:,chansel,chansel) = varargin{j}.(cfg.parameter{i})(:,chansel,chansel);
           end
+         
+        case {'chan_freq_time'}
+          data.(cfg.parameter{i}) = nan(dimsiz);
+          for j=1:numel(varargin)
+            chansel = match_str(varargin{j}.label, oldlabel{j});
+            data.(cfg.parameter{i})(chansel,:,:) = varargin{j}.(cfg.parameter{i})(chansel,:,:);
```
schoffelen, line 139: I wonder whether this works generically, i.e. I think this should work also, when the dimord = 'chan', 'chan_freq', or 'chan_time'.

private/append_common.m
```diff
@@ -131,6 +131,13 @@
             chansel = match_str(varargin{j}.label, oldlabel{j});
             data.(cfg.parameter{i})(:,chansel,chansel) = varargin{j}.(cfg.parameter{i})(:,chansel,chansel);
           end
+         
+        case {'chan_freq_time'}
```
schoffelen, line 135: could you try and combine the two cases? i.e. rather than having

case {'chan_freq_time'}
  ...some-stuff
case {'chan' 'chan_time' 'chan_freq'}
  ...some-almost-identical-stuff

do something like

case {'chan' 'chan_freq' 'chan_time' 'chan_freq_time'}

...do the improved stuff by you (adding the semi-colons for the possibly present third dimension.

schoffelen (changes requested):
Thanks Diego, well caught.

Could you try out my suggestion (provided I was clear enough)?
And quickly test it, i.e. try and append a 'chan_freq' with the updated code and assert that it works as expected? If not, I stand corrected, and we can revert to your original PR

dieloz:
Hi Jan-Mathis
Thanks for your input. Short question: why ft_apppend**freq** should accept dimord data without **freq** field like 'chan', or 'chan_time'?

schoffelen:
> Hi Jan-Mathis
> Thanks for your input. Short question: why ft_apppend**freq** should accept dimord data without **freq** field like 'chan', or 'chan_time'?

Hm, good question. I am not sure when a data structure with a dimord of 'chan' can occur, nor whether it is supposed to be accepted by ft_appendfreq. Probably there will be a higher order check, that checks whether the input to the function is a freq structure. If that is defined to contain a 'freq'-field, AND assuming that then the dimord is consistently 'chan_freq', I think that ft_appendfreq will throw an error. however, append_common is also called by ft_appendtimelock, so for a 'chan_time' dimord it should be supported for sure, and it could be that one can end up in append_common with a 'chan' dimord as well

robertoostenveld:
A data structure with a dimord with only 'chan' can result from e.g. ft_selectdata with avgoverfreq=yes and keepfreqdim=no. We don't have them often, I think only the FT stats functions (with avgoverxx) are somewhat likely to make them. The difference whether a structure has only chan, or chan_something with a singleton "something" is minute. 

Since  we don't have an explicit datatype for those data structures with only chan, the regular functions (for timelock and freq) have to be able to deal with them.

dieloz:
Sorry to bother you again but could you please which way do you prefer to deal with the ft_checkdata and {'chan','chan_chan','chan_time'} cases?
Thanks beforehand

schoffelen:
> Sorry to bother you again but could you please which way do you prefer to deal with the ft_checkdata and {'chan','chan_chan','chan_time'} cases?
> Thanks beforehand

Sorry Diego to keep you waiting, append_common can also be called by ft_appendtimelock, so it is necessary to keep instances like 'chan_time' (does not exist for freq either), and 'chan'. Does this pose a problem?

dieloz:
No worries!
Yes, it's a problem. If you run the following test (with or without the current PR fix) freq data with 'chan' dimord will not output powspctrm:

freq1=[];
freq1.label={'chan1'};
freq1.dimord='chan_freq';
freq1.freq=[1:20];
freq1.powspctrm = rand(1,size(freq1.freq,2));

freq2=freq1;
freq2.label={'chan2'};

cfg=[];
cfg.avgoverfreq='yes';
cfg.keepfreqdim='no';
freq1_s = ft_selectdata(cfg,freq1);
freq2_s = ft_selectdata(cfg,freq2);

cfg=[];
cfg.appenddim = 'chan';
cfg.parameter = 'powspctrm';
freq = ft_appendfreq(cfg,freq1_s,freq2_s)

freq = 

  struct with fields:

     label: {2×1 cell}
      freq: [2×1 double]
    dimord: 'chan_freq'
       cfg: [1×1 struct]

This is because the internal ft_checkdata call in ft_appendfreq transforms the 'chan' dimord to 'chan_freq' as follows:
    label: {'chan1'}
powspctrm: 0.4633
   dimord: 'chan_freq'
      cfg: [1×1 struct]
     freq: 0

My fieldtrip knowledge is too rusty to decide whether the problem is at the level of ft_checkdata or at the append_common? If you help me with this I can fix it

schoffelen:
gotcha, I'll have a look soon.

schoffelen:
OK, at first sight something goes wrong with the call to ft_selectdata in line 94 of append_common
The data structures that come out are unexpected, in that the freq fields are 2x1 .

I now wonder whether this is an unforeseen special case, caused by the single data point in the structure.

schoffelen:
When I try to input 2 data structures with 2 channels each (again no freq and dimord = 'chan') it seems to work fine. I guess that ft_selectdata needs to be fixed. I'll look into this.

schoffelen:
freq1=[];
freq1.label={'chan1'};
freq1.dimord='chan_freq';
freq1.freq=[1:20];
freq1.powspctrm = rand(1,size(freq1.freq,2));

freq2=freq1;
freq2.label={'chan2'};

cfg=[];
cfg.avgoverfreq='yes';
cfg.keepfreqdim='no';
freq1_s = ft_selectdata(cfg,freq1);
freq2_s = ft_selectdata(cfg,freq2);

cfg=[];
cfg.appenddim = 'chan';
cfg.parameter = 'powspctrm';
freqA = ft_appendfreq(cfg,freq1_s,freq2_s)

% another situation
freq1=[];
freq1.label={'chan1';'chan2'};
freq1.dimord='chan_freq';
freq1.freq=[1:20];
freq1.powspctrm = rand(2,size(freq1.freq,2));

freq2=freq1;
freq2.label={'chan3';'chan4'};

cfg=[];
cfg.avgoverfreq='yes';
cfg.keepfreqdim='no';
freq1_s = ft_selectdata(cfg,freq1);
freq2_s = ft_selectdata(cfg,freq2);

cfg=[];
cfg.appenddim = 'chan';
cfg.parameter = 'powspctrm';
freqB = ft_appendfreq(cfg,freq1_s,freq2_s)

% another situation
freq1=[];
freq1.label={'chan1'};
freq1.dimord='chan_freq';
freq1.freq=[1:2];
freq1.powspctrm = rand(2,size(freq1.freq,2));

freq2=freq1;
freq2.label={'chan2'};

cfg=[];
cfg.appenddim = 'chan';
cfg.parameter = 'powspctrm';
freqC = ft_appendfreq(cfg,freq1,freq2)

schoffelen:
The plot thickens: also when there are 2 frequency bins (but only 1 channel per data structure), ft_appendfreq misbehaves. That is leading up to freqC in the above

schoffelen:
hi @dieloz I think the issue in append_common / ft_selectdata that you ran into is now fixed. Can you check and if confirmed, finalize this? Thanks

dieloz:
Works in my data: Thanks a lot Jan-Mathijs!

PR #1569, opened 2020-10-20
DEBUG: eLORETA on 3D data

tiborauer:
This PR aims to resolve the issue when using eLORETA on 3D (e.g. chan_freq_time) data.

### DIFF
--- a/inverse/ft_inverse_eloreta.m
+++ b/inverse/ft_inverse_eloreta.m
@@ -198,8 +198,8 @@ end
 estimate.inside  = originside;
 estimate.pos     = origpos;
 if isfield(sourcemodel, 'pow') % here pow is cell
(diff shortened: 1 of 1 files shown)


schoffelen (approved):

schoffelen:
This PR seems to solve an issue that is unrelated to the issue that was raised on the discussion list, i.e. the one with empty leadfields for dipoles that are labelled as 'inside'. Diego told me that Tibor and he will be working on this. I suggest to create an issue about this, to document the diagnostic process and its resolution.

Regarding this PR, reading the code, it looks as if the projection of the sensor data through the spatial filters is computationally extremely inefficient. I think that this can be improved, but that's for another time.

 I will merge, typically we'd request some snippets of test code to go along with some code change, but I think for this current change it is not necessary.

schoffelen:
Thanks!

PR #1570, opened 2020-10-23
first proposal of BEP030 (NIRS data) into data2bids

helenacockx:
These commits are a proposal to integrate BEP30 for nirs data into data2bids. BEP30 is still subject to change, so data2bids should get updated accordingly.


robertoostenveld:
looks good to me, thanks!

PR #1571, opened 2020-10-23
fix - correcting distance calculation of Rx-Tx combinations

Horschig:
Computation of optode distances could have been wrong in some cases. The error was that indexing of optode positions was based on optode index in a local array instead of global optode number.


Horschig:
will create a new one, need to clean up code a bit

PR #1572, opened 2020-10-23
fix - correcting distance calculation of Rx-Tx combinations

Horschig:
Computation of optode distances could have been wrong in some cases. The error was that indexing of optode positions was based on optode index in a local array instead of global optode number.


schoffelen (approved):
looks good to me! permission to merge?

Horschig:
you have my permission - does that count?

Horschig:
oh, too late... well done, JM! 👍

schoffelen:
Yes, it counts. Because I did not know whether you'd want me to wait for it.

PR #1574, opened 2020-10-27
added trialfun to fix Neuromag STI016 bug

natmegsweden:
There is a known bug where trigger channel STI016 in MEG data recorded on Neuromag devices (prevalent for Elekta Neuromag Triux systems) that it is stored in the datafiles with negative trigger value -32768. I have not seen this issue mentioned on the FieldTrip webpage or found a solution. It is a known bug in the Neuomag software.

The trial function I have added solves the issue when used in combination with ft_definetrial. It is perhaps not the best solution as it does not solve the underlying issue and it would have been better if I could solve the issue directlæy in one of the low-level functions that read the raw data. The current solution is a quick fix I made for data collected by a PhD student.

I have even written a small FAQ for the users of our lab and can push it to FieldTRip if you want?


trialfun/ft_trialfun_neuromagSTI016fix.m
```diff
@@ -0,0 +1,90 @@
...
+
+% Get only specific event type
+cfg.trialdef.eventtype = ft_getopt(cfg.trialdef, 'eventtype', 'STI101');
+cfg.trialdef.eventvalue = ft_getopt(cfg.trialdef, 'eventvalue', []);
+
+% read the header information and the events from the data
+hdr   = ft_read_header(cfg.dataset);
+
+% Read trigger channels
+chanindx = find(not(cellfun('isempty', strfind(hdr.label,'STI0'))));
```
robertoostenveld, line 31: could you do 
```
chanindx = find(startsWith(hdr.label,'STI0'));
```
here?

trialfun/ft_trialfun_neuromagSTI016fix.m
```diff
@@ -0,0 +1,90 @@
+function [trl, event] = ft_trialfun_neuromagSTI016fix(cfg)
```
robertoostenveld, line 1: I am not so happy with the name of the function; I wonder whether people will find it this way. I think the purpose is not to fix something, but to work as a general trial function for all neuromag data (regardless of whether STI016 is broken or not). Let me try some others ...

- ft_trialfun_neuromag
- ft_trialfun_triux
- ft_trialfun_STI016

trialfun/ft_trialfun_neuromagSTI016fix.m
```diff
@@ -0,0 +1,90 @@
...
+
+% FT_TRIALFUN_NAME is supposed to fix the error with STI016 in Neuromag
```
robertoostenveld, line 3: the function name in the help should be updated

robertoostenveld (commented):
I have a few suggestions to improve the code, see my review comments.

robertoostenveld:
some documentation on the issue and the proposed fix in this function would be very welcome on the FT website. I propose to place it here https://www.fieldtriptoolbox.org/getting_started/neuromag/#special-issues or to make a dedicated section on that page.

schoffelen:
I don't know whether this is related, but read_trigger has instances of 'fix<something>' where some known quirks w.r.t. some acquisition systems are dealt with. Is the current issue something that can be solved there in principle as well? (i.e. thus not requiring a dedicated trialfun)

robertoostenveld:
@schoffelen the options you refer to are the ones in https://github.com/fieldtrip/fieldtrip/blob/master/fileio/private/read_trigger.m right? We could indeed also add a `fixSTI016` option there. The question then is: how does this option become accessible to the end-user?

The user calls it like this

analysisscript -> ft_definetrial -> ft_trialfun_general -> ft_read_event -> read_trigger 

The option could be added to https://github.com/fieldtrip/fieldtrip/blob/54b46195349788225ffd625af69b6705a42f2051/fileio/ft_read_event.m#L1648, and then to https://github.com/fieldtrip/fieldtrip/blob/master/trialfun/ft_trialfun_general.m (see around line 18) and https://github.com/fieldtrip/fieldtrip/blob/54b46195349788225ffd625af69b6705a42f2051/trialfun/ft_trialfun_general.m#L105 and then it would be available in `cfg.trialdef` which is available in  `ft_definetrial` and the user's script.

robertoostenveld:
Having slept over it for a night, I actually think that we should include this trialfun anyway. The extra option to `read_trigger` would also be nice to have, but could be additional to this trialfun. 

What I like about the proposed solution is that it is a very clean and exemplary trialfun that Neuromag users can read and understand in a glance. It might not be as flexible and generic as a `fixSTI016` option to `read_trigger`, but it is more transparent. And in general I think it is preferred if people write their own trialfuns rather than relying on situation-specific intelligence that is implemented 4 layers deep.

PR #1576, opened 2020-10-28
first proposal of BEP029 (motion data) into data2bids

helenacockx:
These commits are a proposal to integrate BEP029 for nirs data into data2bids. BEP29 is still subject to change, so data2bids should get updated accordingly.

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -610,8 +610,8 @@ cfg.channels.status_description = ft_getopt(cfg.channels, 'status_description' ,
 cfg.channels.source             = ft_getopt(cfg.channels, 'source'             , nan);  % REQUIRED. Name of the source as specified in the *_optodes.tsv file. n/a for channels that do not contain NIRS signals (acceleration).
 cfg.channels.detector           = ft_getopt(cfg.channels, 'detector'           , nan);  % REQUIRED. Name of the detector as specified in the *_optodes.tsv file. n/a for channels that do not contain NIRS signals (acceleration).
 cfg.channels.wavelength         = ft_getopt(cfg.channels, 'wavelength'         , nan);  % REQUIRED. Wavelength of light in nm. n/a for channels that do not contain raw NIRS signals (acceleration).
-% FIXME: optional nirs channel fields (e.g. frequencies for frequency
-% domain NIRS) should still be added here.
+% FIXME: optional nirs channel fields (e.g. frequencies for frequency domain NIRS) should still be added here.
+
 % motion specific fields
 cfg.channels.source             = ft_getopt(cfg.channels, 'source'             , nan);  % REQUIRED. Label of the object that is being tracked, for example, label of a tracker or a marker.
 cfg.channels.component          = ft_getopt(cfg.channels, 'component'          , nan);  % REQUIRED. Component of the representational system that the channel contains.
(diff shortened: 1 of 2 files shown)


fileio/private/xsens_mvnx.m
```diff
@@ -71,6 +71,10 @@
   hdr = [];
   hdr.label = {}; 
   hdr.chanunit={};
+  hdr.chantype={};
+  hdr.chansource={};
```
robertoostenveld, line 75: the header that is returned by `ft_read_header` adheres to a strict [API](https://en.wikipedia.org/wiki/API) to ensure that all FieldTrip functions, but also external toolboxes that use the fileio functions (like SPM amd EEGLAB) know which fields they can expect. Only fields that are common to all file formats should be in the header itself. Low-level details that apply only to a single file format can be in `hdr.orig`.

robertoostenveld: chansource, chancomponent and chanrefspace cannot be represented like this.

data2bids.m
```diff
@@ -2030,7 +2085,17 @@
...
   end
-  tab = table(name, type, units, sampling_frequency, source, detector, wavelength);
+  % add motion specific fields (empty columns will be removed in a later
+  % step)
+  component=cell(length(name), 1); reference_space=cell(length(name),1);
+  if isfield(hdr, 'chansource');
+    for i=1:length(name)
+      source{i}=hdr.chansource(i);
+      component{i}=hdr.chancomponent(i);
+      reference_space{i}=hdr.chanrefspace(i);
```
robertoostenveld, line 2095: how would this work with a dataset that is converted from C3D to BIDS? Or a dataset that was already read in MATLAB memory, e.g. using `ft_preprocessing`?

robertoostenveld (changes requested):

robertoostenveld:
Can you move the chansource, chancomponent and chanrefspace fields to hdr.orig? And can you change the data2bids such that it works with a `data` input argument where that data is from ft_preprocessing? I realize that this complicates passsing the metadata about the Xsens motion sensors from the file to the BIDS representation. We want the FT code to be generic, not specific to data from one system that is represented in one file format. 

Can you add a test script `test_pull1576` with a small datafile in mvnx format, the same data in c3d format, and then demonstrate all 4 ways of converting that data to BIDS, i.e.

1) directly from mvnx file to bids
2) directly from c3d file to bids
3) ft_preprocessing of the mvnx file, then `data2bids(cfg, data)`
4) ft_preprocessing of the c3d file, then `data2bids(cfg, data)`

fileio/private/xsens_mvnx.m
helenacockx: Hmmm, but when I put the chansource, chancomponent and chanrefspace in hdr.orig, it is kicked out again by `ft_preprocessing` in line 631:
```
    % don't keep hdr.orig in the output data if it is too large
    % hdr.orig can be large when caching data from specific file formats, such as bci2000_dat and mega_neurone
    if isfield(hdr, 'orig')
      s = hdr.orig;
      s = whos('s');
      if s.bytes>10240
        hdr = rmfield(hdr, 'orig');
      end
    end
```

robertoostenveld: Have a look at https://www.fieldtriptoolbox.org/faq/how_are_the_various_data_structures_defined/ which explains two FT principles:

1. we have a small number of structures (raw, timelock, freq, etc) 
2. we try to keep the structures themselves small

With the 2nd point we more resemble SPM (which always writes data at any level of processing to a nifti file on disk) than EEGLAB (which maintains an elaborate data structure) or BrainStorm (which maintains a whole database on disk).  

If you consider my step 2 and 4 from the proposed example, you'll see that those are also not solved by a mvnx specific fieds in the header. Hence, the `hdr.orig` strategy does not result in a generic solution.

helenacockx: So would you think then that it might be a better solution if the researcher would always specify the motion channels with `cfg.channels` for `data2bids`?

schoffelen:
Hi Helena, what's the status of this?

robertoostenveld:
I reviewed this, but realize that it is a PR to a rather old version of data2bids, and that most updates have already been implemented independently from this. Consequently, the remaining changes mostly cause merge conflict and don't add relevant functionality. 

Hence I will close this PR without merging. If needed, we can revisit it. A better strategy for a follow-up attempt would be to start with a concrete case of converting motion data (from shared data) to bids, e.g. like https://www.fieldtriptoolbox.org/example/bids_motion/

PR #1577, opened 2020-10-29
enhancement - added plotting options component view ft_databrowser

jannystapel:
I have added plotting options to ft_databrowser, which can be used in the component view mode. It concerns plotting options that are also used in for instance ft_topoplotER and ft_topoplotTFR.


robertoostenveld:
LGTM

ft_databrowser.m
```diff
@@ -2060,8 +2070,22 @@ function redraw_cb(h, eventdata)
       
       % laychan is the actual topo layout, in pixel units for .mat files
       % laytopo is a vertical layout determining where to plot each topo, with one entry per component
+     
+     opt = {'interpmethod', cfg.interpolation, ...
```
robertoostenveld, line 2074: this introduces a bug in the code, since `opt` was already used as a variable name (and was a struct rather than a cell-array)

schoffelen: poor variables called 'opt'...

PR #1582, opened 2020-11-04
BUG: empty array of artifacts

ruiliu-psy:
The latest version returned error when the artifacts array was empty. The error report is appended below.

Index in position 2 exceeds array bounds.

Error in ft_rejectartifact (line 309)
    begsample = artifact{i}(:,1);


schoffelen:
Do you have some test code and small example data that demonstrates the original issue (and which we can include in a test fucntion)?

PR #1586, opened 2020-11-08
Update ft_multiplotTFR.m

raintwoto:
Need hold on, otherwise, the plot will be replaced by new ones in Matlab 2020b

### DIFF
--- a/ft_multiplotTFR.m
+++ b/ft_multiplotTFR.m
@@ -459,7 +459,7 @@ chanHeight = cfg.layout.height(sellay);
 % open a new figure, or add it to the existing one
 % note that in general adding a TFR to an existing one does not make sense, since they will overlap
 open_figure(keepfields(cfg, {'figure', 'clearfigure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
-
+hold on
 % Get physical z-axis range (color axis):
 if strcmp(cfg.zlim, 'maxmin')
   zmin = nanmin(datamatrix(:));
--- a/ft_multiplotTFR.m
+++ b/ft_multiplotTFR.m
@@ -460,6 +460,7 @@ chanHeight = cfg.layout.height(sellay);
 % note that in general adding a TFR to an existing one does not make sense, since they will overlap
 open_figure(keepfields(cfg, {'figure', 'clearfigure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
 hold on
+
 % Get physical z-axis range (color axis):
 if strcmp(cfg.zlim, 'maxmin')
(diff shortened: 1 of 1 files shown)


schoffelen:
Thanks! I guess this also needs to be changed in ft_multiplotER.m. This issue has been also recently raised by a  user on the dicussion list. It was probably introduce by a recent clean up operation, that sychronized the opening of new figures across the code base (using the new function open_figure). @robertoostenveld can you confirm that open_figure does not 'think' about holding operations (which might be context dependent anyhow), and that the hold on (and possibly later hold off) need to be done in the caller function?

robertoostenveld:
is the problem you identified specific to MATLAB 2020b?

robertoostenveld:
>  can you confirm that open_figure does not 'think' about holding operations (which might be context dependent anyhow), and that the hold on (and possibly later hold off) need to be done in the caller function?

correct, the `hold on` and `hold off` remains the responsibility from the calling code, i.e. the high-level FT function, or for the low-level `ft_plot_xxx` function if that loops over elements

schoffelen:
disconfirmed. It also occurs in MATLAB2018b on both MacOS and Linux

robertoostenveld:
I can reproduce the problem with 

```
ntrial = 5;

data = [];
data.label = {'1', '2', '3'};
for i=1:ntrial
  t = (1:1000)/1000;
  s = sin(i*2*pi*t); % make a sine with 1, 2, 3, ... Hz
  data.time{i} = t;
  data.trial{i} = 0.1*randn(3,1000);
  data.trial{i}(1,:) = s;  % insert it in the first channel
end

%%

cfg = [];
timelock = ft_timelockanalysis(cfg, data);

%%

cfg = [];
cfg.layout = 'vertical';
figure
ft_multiplotER(cfg, timelock); % it should show a mix

%%

cfg = [];
cfg.method = 'wavelet';
cfg.toi = t;
freq = ft_freqanalysis(cfg, data);

%%

cfg = [];
cfg.layout = 'vertical';
figure
ft_multiplotTFR(cfg, freq); % it should show a mix
```

robertoostenveld (approved):

schoffelen:
Should we make a similar change in ft_multiplotER as well? I can do that, no problem

robertoostenveld:
my test code did not reveal any problem in `ft_multiplotER`. I see in ft_plot_vector that it is doing it

```
% everything is added to the current figure
holdflag = ishold;
if ~holdflag
  hold on
end
``` 

it might actually be good to make ft_plot_vector and ft_plot_matrix more consistent, and remove it again from ft_multiplotTFR

Issue #1568, opened 2020-10-16
Event display per channel for ft_databrowser


Frederik-D-Weber:
The problem is that events/or artifacts that relate to only one or mulitple channels (but not others) cannot be visualized (also with different colors) in the **ft_databrowser**. This is a typical usecase when there are events that do not relate to all channels (e.g. channel-specific artifacts.)

There should be an option to load cfg.event and on recognition of the optional field cfg.event.channel to specify where those events were stemming from/were detected. 
There should be a default or other option to give each unique event type a color. 
Maybe a funtion for transfer of artifacts to events would also be helpful e.g. **_ft_artifact2event_**

Example:
```
...
event.type = 'artifact_1';
event.sample = 3000;
event.channel = {'C3', 'Cz', 'C4'}
%event.seconds = 30.0;
%event.fsample = 100;
cfg.event = {event};
...
cfg = ft_databrowser(cfg, data);
...
```

There is already an undocumented cfg.event but it does not fit the mulit-channel highlighting and is used for in-databrowser marking. Also the use over cfg.artifact.xxx is not specific to one channel.
I already tried it on two event types in a nonsystematic manner to plot in the SpiSOP browser, so plotting speed is OK with consideration of many events different for all channels. However a consistent implementation in FieldTrip would be nice.
Also event do not consider the original sampling rate, maybe it would be a solution to add an optional field like event.seconds or event.fsample to check if consistency with data or recreate sample from data.fsample before execution.

Modifications would need to be added to **ft_read_events** and **ft_databrowser** and maybe also **ft_filter_event**.
This was discussed between @gdamaskos86 and @Frederik-D-Weber, but we would like to get some feedback from others if this would be the way to go or if this was already talked about.

robertoostenveld:
Hi @Frederik-D-Weber 

Thanks for reporting the issue. This is actually closely related to #1564 on which I started working earlier this week, but also to much earlier work from @IngridNieuwenhuis on Monkey sleep ECoG and ft_databrowser. And in a way it also relates to the recently merged #1565 from @helenacockx . 

[...]

PR #1588, opened 2020-11-09
made plotting of events and artifacts more consistent in ft_databrowser 

robertoostenveld:
this PR is mainly a cleanup of the ft_databrowser code, with only a few changes to the defaults and functionality. The big update is the refactoring of the events are artifacts plotting code, related to #1568


robertoostenveld (commented):
I tested various things interactively and think that it still works as it should

PR #1594, opened 2020-11-13
improve nan-handling in preproc module, if possible, and broadcast this to higher level preproc

schoffelen:


schoffelen:
this relates to bug3238 on bugzilla, and generically to the handling of nans.
I came across this issue, because I want to be able to do a Hilbert transform on (long) data segments, with nans in there.
Note that the current version of preproc seems to bypass all ft_preproc_ functions, so the possible nan awareness (and capability to deal with it appropriately) is not available from the level of ft_preprocessing yet.

schoffelen:
I upgraded the failed_bug3238 to test_bug3238 again, and commented out for now the parts that are failing. Specifically, what fails at the moment is the computation of covariance in ft_timelockanalysis, when there are channel specific nans in the data.

schoffelen:
http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=3238

robertoostenveld (approved):
looks good to me. 

There will still be cases where nans in the data will cause preprocessed results that do not to match with the users' expectations. Perhaps we should in the help of each ft_preproc_xxx function explicitly specify what happens if the input data contains nans.

PR #1597, opened 2020-11-18
data2bids onset duration added to events.tsv

DidiLamers:
Some forms of cfg.events input are of form "onset" "duration" whereas others are of form "begsample", "endsample", "offset". In the later case, an error was given when the table gets sorted so that onset and duration are the first columns of the table (old lines 1590 to 1603). It should now be solved by the addition of columns "onset" and ""duration" to tables of the latter form.

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1536,8 +1536,16 @@ if need_events_tsv
     % use the events table as it is
     events_tsv = cfg.events;
   elseif istable(cfg.events) && all(ismember({'begsample', 'endsample', 'offset'}, fieldnames(cfg.events)))
-    % it is a "trl" matrix formatted as table, use it as it is
+    % it is a "trl" matrix formatted as table, use it as it is, but add
+    % onset and duration
     events_tsv = cfg.events;
+    begsample                   = table2array(events_tsv(:,{'begsample'}));
+    endsample                   = table2array(events_tsv(:,{'endsample'}));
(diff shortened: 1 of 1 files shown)


data2bids.m
```diff
@@ -1536,8 +1536,16 @@
     % use the events table as it is
     events_tsv = cfg.events;
   elseif istable(cfg.events) && all(ismember({'begsample', 'endsample', 'offset'}, fieldnames(cfg.events)))
-    % it is a "trl" matrix formatted as table, use it as it is
+    % it is a "trl" matrix formatted as table, use it as it is, but add
+    % onset and duration
     events_tsv = cfg.events;
+    begsample                   = table2array(events_tsv(:,{'begsample'}));
+    endsample                   = table2array(events_tsv(:,{'endsample'}));
+    onset                       = begsample./hdr.Fs;
```
robertoostenveld, line 1544: the first sample of the recording is 1, but the first time in the recording is 0. So `onset = (begsample-1)/hdr.Fs`

robertoostenveld (changes requested):

data2bids.m
DidiLamers: Yes, my mistake!

robertoostenveld: but as far as I can see, you have not changed the code yet

robertoostenveld (changes requested):

PR #1599, opened 2020-11-25
critical fix for dics method, accidental channel reordering

schoffelen:


robertoostenveld:
I suspect this problem to be persistent, i.e. it might have happened in the past and might happen again in the future.  

Searching through the code, I found some other occurrences of this problem. E.g. on https://github.com/fieldtrip/fieldtrip/blob/2a67cb59746eb81fe05bc118c96d5f257a63f51c/ft_channelrepair.m#L181 the problem is addressed differently. But on https://github.com/fieldtrip/fieldtrip/blob/2a67cb59746eb81fe05bc118c96d5f257a63f51c/ft_channelrepair.m#L162 the problem is not addressed. 

Should we not by default do `setdiff(..., 'stable')` everywhere to make it more explicit and to increase the chances of other code contributors becoming aware of this?

robertoostenveld:
in `prepare_freq_matrices` I have added
_The order of the channels in the output data is according to the input cfg.channel, which therefore must be specified as a cell-array with actual labels, not as an input like 'all' that still needs to be interpreted by FT_CHANNELSELECTION.`

in `ft_prepare_vol_sens` I see 
_The channel order returned by this function corresponds to the order in the 'channel' option, or if not specified, to the order in the input sensor array._

However, when testing the order returned by `ft_prepare_leadfield`, I see that it does **not** honor cfg.channel. For the high-level functions that return data with full metadata (i.e. labels as a cell array in the same structure) it is also less of a problem. For low-level functions that only return numeric arrays it is important that the channel order is stable and well defined.

PR #1600, opened 2020-11-25
Introduction of Laplacian re-referencing

pietroamerio:
Dear FieldTrip maintainers,

I had the pleasure to work with your toolbox on the analysis of human intracranial EEG (iEEG) data. 
Trying to optimize my preprocessing pipeline to achieve the best possible S/N ratio in my dataset, I decided to adopt a re-referencing method different from those currently implemented in FieldTrip. It is called "Laplacian re-referencing" and consists of the re-referencing of each channel within a stereotactic electrode against the average of its two closest neighboring channels (the top and bottom channels are simply re-referenced against their only closest neighbor). This method, at least for iEEG, has been shown to minimize inter-channel correlations in the evoked responses and to increase the correlation of both high- and low-band activity with the task (while reducing the number of task-related channels), leading to an overall higher S/N ratio compared to other re-referencing methods, including bipolar re-referencing (for further details, here is the relative study, on which this pull request is based: https://doi.org/10.1016/j.neuroimage.2018.08.020).

The main modifications are to be found in the ft_prepare_montage.m function. The 'laplace' feature is designed to work in a fashion similar to that of the 'bipolar' feature, so to minimize the need for changes in other functions (in order to make sure that the code would not crash in the rare case of only one channel being given in input to ft_prepare_montage.m, an admittedly artificial mechanism was adopted, which is meant to simply behave in the same way as the 'bipolar' feature would in the case of a single input channel).
The modifications in the other functions (ft_preprocessing.m and ft_preproc) are simply meant to allow the user to set the re-referencing in "Laplacian mode" from the higher-level function ft_preprocessing, by setting cfg.refmethod = 'laplace'.

It is my personal opinion that this method could prove valuable also for EEG data analysis. Therefore, given the broader diffusion of FieldTrip for EEG data analysis than for iEEG data analysis, the pipeline currently treats all channels given in input to ft_prepare_montage.m.

I hope you will also find this feature interesting and worth of being introduced in the toolbox! Please feel free to contact me for further questions or if you think that I neglected to modify/adapt something and should therefore take care of it.

Kind regards,

Pietro Amerio
Decision and Awareness Group, German Primate Center (DPZ), Göttingen


pietroamerio:
Sorry I skipped a part of my explanation. Let me reconnect to the previous message:

It is my personal opinion that this method could prove valuable also for EEG data analysis. Therefore, given the broader diffusion of FieldTrip for EEG data analysis than for iEEG data analysis, the pipeline currently re-references all channels given in input to ft_prepare_montage.m together, without making distinctions among them. The users who are interested in iEEG data analysis should make sure that the re-referencing is conducted separately for each group of channels belonging to the same electrode; given the smaller pool of people interested in iEEG and the fact that electrodes' names might be arbitrarily assigned, I have not introduced the possibility of sorting channels prior to re-referencing.
I would love to hear your feedback on whether the automatic channel sorting should be implemented and I am willing to work on its implementation myself, if necessary.

Best,
Pietro Amerio

schoffelen:
Thanks Pietro,

I agree that using more locally re-referenced data is useful, both for iEEG and EEG. Yet, I would be a bit worried about the currently proposed changes, if there is no explicit check on the montages being valid. That is, and admittedly this also holds for the already implemented 'bipolar' method, there is no guarantee whatsoever that the channels are ordered in a particular way. The changes you suggest assume the channels to be ordered as a 1D array, with one channel next to the next one.

I think we should consider to improve ft_prepare_montage such, that it does not make explicit assumptions about the channel order, but that it requires an explicit definition of which channels are each others' neighbours. This is a heuristic that we already use in ft_scalpcurrentdensity (for laplace-like transforms of EEG data) and ft_megplanar (for spatial derivatives of meg data, cf. megplanar_sincos in fieldtrip/private). Alternatively, we could consider to use the already existing ft_scalpcurrentdensity, and make this more generic to also work on iEEG data.

@robertoostenveld what do you think?

pietroamerio:
Thank you very much for your reply and your attention!

Yes that's correct! For my data, I make sure that the channels are sorted and divided into the correct groups before the call to ft_preprocessing, which has to be looped for every group.
How do you wish to coordinate the implementation of the changes you suggest among us?

robertoostenveld:
Hi Pietro

For scalp EEG similar approaches have been proposed and discussed, see e.g. https://www.frontiersin.org/articles/10.3389/fnins.2019.00941/full for a recent overview. Specifically the Hjorth (1975) reference scheme shares resemblance, which have later been refined (inspired by denser sampling of the potential distribution) under the name "scalp current density" (SCD, which is confusingly similar to CSD cross spectral density) and "surface laplacians".  We have (already since 2004) https://github.com/fieldtrip/fieldtrip/blob/master/ft_scalpcurrentdensity.m for that. 

I thnk it would be more clear to end-users if the implementation of rereferencing for intracranial EEG with linear probes (along 1 dimension) remains conceptually separate from that of re-referencing 2D or 2.5D (curved scalp surface) arrangements.

robertoostenveld:
Looking at the suggested code changes, I don't see anything wrong with it. But I do realize that it is actually non-trivial to check the implementation by simply reading the code because of lack of documentation. That is not only for the new Laplace implementation, but also for the bipolar implementation. 

I think that all would benefit from some online documentation (on the FT website) that explains and demonstrates the different reference schemes and assumptions that they make (or fail to make). E.g. explain that if the data has bipolar channels, you should exclude them from the avgref. or the way that channels have to be named in sEEG to recognize that they share a shaft.  Perhaps this can be formatted as a FAQ.

robertoostenveld:
Regarding coordinate the implementation of the changes. I would like @pietroamerio to take the lead. If we consider an example page rather than a faq, we could also have some example data (and of course we can reuse existing data).  How about a page `/example/rereference.md` on the website with subsections for each reference method 

- avg
- median
- rest
- bipolar
- laplace

You don't have tu fill all sections yourself, but could you provide the bipolar and laplace section? For that you could take some data from https://www.fieldtriptoolbox.org/tutorial/human_ecog/ (which can be a small piece of the original preprocessed data - saved to a new mat file) and use that for the last 2. 

For the first three we can take an EEG example from one of the tutorials.I would probably also add the explanation of a linked-mastoids version, and possibly also some text on when bipolar recording channels are present, or when EOG was also recorded with a reference common to the EEG electrodes.

pietroamerio:
Thank you very much for the detailed reply!

Yes I can take care of ft_prepare_montage. I will also take a look at the functions you suggested to get inspiration.
I guess I will try to introduce the possibility of having the channels to be sorted based on their names (maybe asking the user to specify them as a letter code followed by a number, where every letter code corresponds to an electrode) and to be separately re-referenced for each group. This could be an option that the user has to specifically ask for, otherwise everything will be re-referenced in the order it is given in input (a warning message could be sent if electrodes with different letter codes are detected to be contained in the input). 
Unfortunately, I am not sure if will be able to do it in the next few days, I hope this is not a problem!

Regarding the example page, I am surely willing to help with the bipolar and laplacian parts! Will you create the page and define to outline of it or should I do it?

P.S. I am not an expert programmer, nor an expert EEG/iEEG data analyst, if you feel that I am not getting some points of your directions, please excuse me and feel welcome to give any additional input!

robertoostenveld:
I have added an example page at http://fieldtriptoolbox.org/example/rereference that explains the different options. It also has placeholders for rest (which I guess I will have to add) and for bipolar and laplace methods. @pietroamerio can you have a look and suggest some text for the bipolar and laplace sections (perhaps first only bipolar, then discuss, then continue with laplace)? 

Contributing to the website goes the same as to the fieldtrip code itself, i.e. you make a fork on github, then a local clone, set up the `origin` and `upstream` remotes, then make a branch, change, commit, push, and make a pull request on [fieldtrip/website](https://github.com/fieldtrip/website).

pietroamerio:
Thank you very much for the detailed instructions!

I am almost done with the modifications to ft_prepare_montage and will upload them soon. 
In the next days I will work on the example page.

pietroamerio:
Dear @robertoostenveld and @schoffelen 

I am extremely sorry for such a long delay!
I finally added the possibility of grouping channels automatically and re-reference them within groups (this is only compatible with bipolar and Laplacian re-referencing). The grouping is done on the basis of channels' names, which should be designed to have an alphabetic portion followed by a numeric one (where letters represent the group the channel belongs to and numbers represent the position of the channel within the group, e.g. 'LTR1'). 

To do so, I had to resort to an external function for the correct sorting of channels' names.
Here is the function's page on the MatLab FileExchange: 
https://de.mathworks.com/matlabcentral/fileexchange/47434-natural-order-filename-sort
I was not sure where I should place the .m file, so, momentarily, I placed it in the "external" folder.

I will try to add the information to the example page in the following days!
Sorry again for the delay!

Best regards,
Pietro Amerio

pietroamerio:
Dear @schoffelen and @robertoostenveld,

Sorry for double-messaging, I was just wondering what you think about the last modifications and about the text in the example page. Would you suggest further changes before merging with the main repository?

Thank you in advance for your attention!

Best regards,
Pietro Amerio

robertoostenveld:
@pietroamerio can you review the changes that are now merged in the master branch? There is now a test script `test_pull1600`, and you may want to look at #1765 to see the code changes.

pietroamerio:
Dear @robertoostenveld,

I just found out that the pull request was released! Thank you so much!
I am sorry for being absent in the past few days, I was finishing up and submitting my Master's Thesis.
May I still help or is it unnecessary now that the changed were released?

Thanks again!
Pietro Amerio

robertoostenveld:
all is fine, thanks again for your contribution!

PR #1604, opened 2020-11-30
ft_inside_headmodel is slow due to dsearchn

schoffelen:


robertoostenveld:
assuming that you have tested it properly, the suggested change looks fine to me

PR #1606, opened 2020-12-01
data2bids events json update

DidiLamers:
- changed CapModelName to CapManufacturersModelName for the eeg events json, as that is the standard according to the BIDS spec

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1542,7 +1542,7 @@ if need_events_tsv
     begsample                   = table2array(events_tsv(:,{'begsample'}));
     endsample                   = table2array(events_tsv(:,{'endsample'}));
     onset                       = (begsample-1)./hdr.Fs;
-    duration                    = (endsample-begsample)./hdr.Fs; 
+    duration                    = (endsample-begsample+1)./hdr.Fs; 
     table_onset_duration        = table(onset, duration);
     events_tsv                  = [table_onset_duration events_tsv];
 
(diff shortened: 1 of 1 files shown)


data2bids.m
```diff
@@ -1542,7 +1542,7 @@
     begsample                   = table2array(events_tsv(:,{'begsample'}));
     endsample                   = table2array(events_tsv(:,{'endsample'}));
     onset                       = (begsample-1)./hdr.Fs;
-    duration                    = (endsample-begsample+1)./hdr.Fs; 
+    duration                    = (endsample-begsample)./hdr.Fs;
```
robertoostenveld, line 1545: Can you please revert this? If an event lasts from sample N to sample N+1, I would say that its duration is 2 samples. So if it lasts from N to N, its duration is 1 sample.

data2bids.m
```diff
@@ -1563,7 +1563,7 @@
     end
     % convert to the required fields
     onset     = (begsample-1)/hdr.Fs;
-    duration  = (endsample-begsample+1)/hdr.Fs;
+    duration  = (endsample-begsample)/hdr.Fs;
```
robertoostenveld, line 1566: Can you please revert this?

robertoostenveld (changes requested):

robertoostenveld:
dank je!

data2bids.m
DidiLamers: Okay, done

PR #1608, opened 2020-12-02
FIX - missing function argument to checkpos when appenddim = auto

schoffelen:


schoffelen:
I did not try to reproduce the error, it seemed clear from reading the code. Surprising that no one has encountered this before.
Note that in the current version of the code the relevant line number is 100, and not 101 as originally reported

PR #1610, opened 2020-12-04
test for issue #1554 concerning parsing of extra fields in ft_spike_m…

stephenwhitmarsh:
…aketrials

Hi @robertoostenveld here is the testscript


schoffelen:
It looks as if the newer PR #1614 also has this test function incorporated, so I assume that this particular one can be closed.

PR #1614, opened 2020-12-07
Issue1554

stephenwhitmarsh:
This completes @robertoostenveld 's fix


schoffelen:
can this be merged? @stephenwhitmarsh @robertoostenveld

schoffelen:
I checked with @stephenwhitmarsh , and it seems that the problem that this PR aims to solve, is not a problem in the current master branch (anymore), which suggests that this PR does need to be merged. Nonetheless, I would be reluctant to close this without merging. @robertoostenveld, shall we merge this either way, because no doubt some thinking has gone into this, and it could even be that the suggested changes reflect a general improvement of the codebase, irrespective of the original issue.

robertoostenveld:
agreed. I had a look at the code, and it all looks like good legit improvements. It also includes a new test function, which is always welcome.

PR #1616, opened 2020-12-10
improve reading of iEEG header for BIDS dataset 

### LINKED ISSUE #1601: error in electrodes_tsv.name when running ft_read_data > ft_read_header
In my BIDS dataset, I have both an electrodes.json and an electrodes.tsv. The electrodes.tsv contains the electrode's names, position etc. The electrodes.json contains the description of the fieldnames in electrodes.tsv. 

When running ft_read_data, it runs ft_read_header and within ft_read_header, bids_sidecar.m loads an electrodes-file of which electrodes_tsv.name is used later on in ft_read_header. The bids_sidecar.m searches for the accompanying electrodes.json OR electrodes.tsv. It then loads the electrodes.json based on alphabethical order instead of the electrodes.tsv. This results in an error later in ft_read_header, because name is not a field in the electrodes_tsv variable when the electrodes.json is loaded.

I propose to remove '.json' in line 127 of bids_sidecar.m to resolve this error. 

https://github.com/fieldtrip/fieldtrip/blob/9a6153cf7dd568efc532cf224e05286b1c3551a9/fileio/private/bids_sidecar.m#L127

robertoostenveld:
this fixes #1601

### DIFF
--- a/fileio/private/bids_sidecar.m
+++ b/fileio/private/bids_sidecar.m
@@ -8,12 +8,12 @@ function sidecar = bids_sidecar(filename, suffix)
 %   sidecar = bids_sidecar(filename, sidecar, retval)
 % where filename refers to a BIDS data file and suffix is a string that refers to the
 % specific sidecar file. To read the json sidecar corresponding to the data itself,
-% you can keep the suffix empty. In that case the suffix (e.g. meg, eeg or ieeg) will
+% you can keep the suffix empty. In that case the suffix (e.g., meg or eeg) will
 % be determined from the filename.
 %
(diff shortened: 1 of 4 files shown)


fileio/private/bids_sidecar.m
```diff
@@ -119,12 +121,16 @@
 % start with an empty return value
 sidecar = [];
 
+% if there is both a tsv and a json file for the desired sidecar, we want to return the tsv
```
robertoostenveld, line 124: this is the most significant improvement for issue #1601 as it ensures that the tsv is read instead of the json

PR #200, opened 2016-07-27
Visualization with BESA Plot in FieldTrip

todoriordanov:
New functions facilitating data visualization with BESA Plot in FieldTrip.
BESA Plot is a free tool created by Patrick Berg (BESA GmbH) for visualization of MEEG data. Initially it was meant to be complementary visualization tool to BESA Research, however, it could be used in more general context. For more information please visit [Visualization with BESA Plot in FieldTrip](http://wiki.besa.de/index.php?title=Visualization_with_BESA_Plot_in_FieldTrip).

PR #1619, opened 2020-12-17
option scaling for ints is now not discareded in ft_volumewrite.m

### LINKED ISSUE #1618: ft_volumewrite option cfg.scaling='no' overwritten for ints
In ft_volumewrite, the output is scaled no matter if cfg.scaling='no'  for ints because the value is overwritten, in line 116:
[https://github.com/fieldtrip/fieldtrip/blob/b626fbc5f7ffda59e8fb4c9d60f23ee140c7eb09/ft_volumewrite.m#L115](https://github.com/fieldtrip/fieldtrip/blob/b626fbc5f7ffda59e8fb4c9d60f23ee140c7eb09/ft_volumewrite.m#L115)

DanielMiklody:
closes #1618

### DIFF
--- a/ft_volumewrite.m
+++ b/ft_volumewrite.m
@@ -17,23 +17,26 @@ function ft_volumewrite(cfg, volume)
 %   cfg.parameter     = string, describing the functional data to be processed,
 %                         e.g. 'pow', 'coh', 'nai' or 'anatomy'
 %   cfg.filename      = filename without the extension
-%   cfg.filetype      = 'analyze', 'nifti', 'nifti_img', 'analyze_spm', 'mgz',
-%                         'vmp' or 'vmr'
+%   cfg.filetype      = 'analyze_old', 'nifti', 'nifti_img', 'analyze_spm',
+%                       'mgz', 'mgh', 'vmp' or 'vmr'
 %   cfg.vmpversion    = 1 or 2 (default) version of the vmp-format to use
 %
 % The default filetype is 'nifti', which means that a single *.nii file
-% will be written using the freesurfer toolbox. The 'nifti_img' filetype uses SPM for
-% a dual file (*.img/*.hdr) nifti-format file.
+% will be written using code from the freesurfer toolbox. The 'nifti_img' filetype
+% uses SPM for a dual file (*.img/*.hdr) nifti-format file.
 % The analyze, analyze_spm, nifti, nifti_img and mgz filetypes support a homogeneous
 % transformation matrix, the other filetypes do not support a homogeneous transformation
 % matrix and hence will be written in their native coordinate system.
 %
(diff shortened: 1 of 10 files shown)


ft_volumewrite.m
```diff
@@ -110,13 +110,15 @@ function ft_volumewrite(cfg, volume)
 cfg.markorigin   = ft_getopt(cfg, 'markorigin',   'no');
 cfg.markfiducial = ft_getopt(cfg, 'markfiducial', 'no');
 cfg.markcorner   = ft_getopt(cfg, 'markcorner',   'no');
-cfg.scaling      = ft_getopt(cfg, 'scaling',      'no');
+cfg.scaling      = ft_getopt(cfg, 'scaling');
 
-if any(strmatch(cfg.datatype, {'int8', 'int16', 'int32'}))
-  cfg.scaling = 'yes';
+if any(strmatch(cfg.datatype, {'uint8','int8', 'int16', 'int32'}))&&isempty(cfg.scaling)
```
schoffelen, line 115: should we be exhaustive here, and also include uint16/uint32 (u)int64?

DanielMiklody: then you also need to include them around line #200. The rescaling is not yet implemented for these datatypes and they are also not mentioned in the help. If you think this could be of use it is not a big thing implementing them.

schoffelen:
I think that this is the good time to fix this in a principled manner, and to revisit the code. Currently, it is quite ugly. ft_volumewrite depends for most filetypes on ft_write_mri, and it is not clear (historic reasons) why some of the filetypes are supported from ft_volumewrite only, and not for others. Also, the typecasting of the data needs to be made consistent, i.e. do all filetypes support all datatypes? Also, currently - for the nifti files - not all permitted datatypes work because of restrictions imposed by the lower-level freesurfer code.

schoffelen:
OK, this is as far as I am going to push it. I have made some changes that allows for a correct writing (and reading) of different data types, using the freesurfer code. In short, the scaling and intercept parameters are saved along in the file, and applied upon reading in the file. I didn't get this to work for SPM though, unfortunately. This is documented in the help of ft_volumewrite. Moreover, I did a code overhaul, and moved relevant lower-level code into ft_write_mri.

schoffelen:
OK, this is as far as I am going to push it. I have made some changes that allows for a correct writing (and reading) of different data types, using the freesurfer code. In short, the scaling and intercept parameters are saved along in the file, and applied upon reading in the file. I didn't get this to work for SPM though, unfortunately. This is documented in the help of ft_volumewrite. Moreover, I did a code overhaul, and moved relevant lower-level code into ft_write_mri.

schoffelen:
To do still: file a PR with the freesurfer folks to get the MRIwrite/MRIread/load_nifti/save_nifti in sync on their end.

PR #1625, opened 2021-01-05
Add BrainVision timestamps for markers

nbeliy:
A small modification of `ft_read_event.m` that copy the BrainVision timestamps (if present) into `events.timestamp`. 
Timestamps are converted into `datetime` structure.

In BrainVision format, the acquisition time can be only derived from markers timestamps, but they are ignored during conversion.
This pull corrects that by copying BV timestamp (stored as last field in marker file) into events array.

The changing is not invasive and affects only BrainVision files, the timestamp column, for my knowledge,  is not standardised in fieldtrip.

### DIFF
--- a/fileio/ft_read_event.m
+++ b/fileio/ft_read_event.m
@@ -514,9 +514,8 @@ switch eventformat
               event(end  ).value    = tok{2};
               event(end  ).sample   = str2num(tok{3});
               event(end  ).duration = str2num(tok{4});
-              if size(tok, 2) >= 6
-                event(end).timestamp = datetime(tok{6}, 'InputFormat', ...
-                'yyyyMMddHHmmssSSSSSS');
+              if size(tok, 2) >= 6 && ft_platform_supports('datetime')
+                event(end).timestamp = datetime(tok{6}, 'InputFormat', 'yyyyMMddHHmmssSSSSSS');
               end
             end
           end
--- a/fileio/ft_read_event.m
+++ b/fileio/ft_read_event.m
@@ -514,6 +514,10 @@ switch eventformat
               event(end  ).value    = tok{2};
               event(end  ).sample   = str2num(tok{3});
               event(end  ).duration = str2num(tok{4});
+              if size(tok, 2) >= 6
+                event(end).timestamp = datetime(tok{6}, 'InputFormat', ...
+                'yyyyMMddHHmmssSSSSSS');
+              end
             end
           end
         end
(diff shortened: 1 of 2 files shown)


robertoostenveld:
thanks for the suggestion.

The MATLAB [datetime](https://nl.mathworks.com/help/matlab/ref/datetime.html) function was introduced in 2014b, which is just old/established enough to consider. See https://www.fieldtriptoolbox.org/development/guideline/code/#ensure-that-it-runs-on-older-matlab-versions. But nevertheless, I suggest that we accommodate this robustly. Let me propose some changes in they code itself ...

robertoostenveld:
... oh, direct edits by me on your branch are not allowed.

fileio/ft_read_event.m
```diff
@@ -514,6 +514,10 @@
               event(end  ).value    = tok{2};
               event(end  ).sample   = str2num(tok{3});
               event(end  ).duration = str2num(tok{4});
+              if size(tok, 2) >= 6
+                event(end).timestamp = datetime(tok{6}, 'InputFormat', ...
```
robertoostenveld, line 518: please put this on a single line, without the `...`

fileio/ft_read_event.m
```diff
@@ -514,6 +514,10 @@
               event(end  ).value    = tok{2};
               event(end  ).sample   = str2num(tok{3});
               event(end  ).duration = str2num(tok{4});
+              if size(tok, 2) >= 6
```
robertoostenveld, line 517: please change this into `if size(tok, 2) >= 6 && ft_platform_supports('datetime')` to make it conditional on the MATLAB version

robertoostenveld (changes requested):

fileio/ft_read_event.m
robertoostenveld: and can you please add the check for datetime to `fieldtrip/utilities/ft_platform_supports.m`?

nbeliy:
Hi Robert, just added the write access

robertoostenveld:
since you are at it anyway, could you do the small edits yourself? thanks

nbeliy:
For sure.

robertoostenveld:
Oh, I just got the write access request for your whole repo. That is not needed. At the moment of making the PR, you can indicate somewhere that maintainers (of the repo at which the PR is aimed) have write access to the specific branch (from your repo) that should be pulled.

fileio/ft_read_event.m
nbeliy: Done, just for my culture , why the `...` is depricated?

fileio/ft_read_event.m
nbeliy: Done

nbeliy: Check for datetime also done

robertoostenveld:
thanks!

PR #1628, opened 2021-01-10
ENH - cross-channel CFC

tiborauer:


ft_crossfrequencyanalysis.m
```diff
@@ -82,40 +93,46 @@
...
 freqhigh = ft_checkdata(freqhigh, 'datatype', 'freq', 'feedback', 'yes');
 
-% prior to 19 Jan 2017 this function had input options cfg.chanlow and cfg.chanhigh,
-% but nevertheless did not support between-channel CFC computations
-cfg = ft_checkconfig(cfg, 'forbidden', {'chanlow', 'chanhigh'});
+if isfield(cfg, 'chanlow') && isfield(cfg, 'chanhigh')
+    docrosschan = true;
+    cfg.chanlow  = ft_channelselection(cfg.chanlow, freqlow.label);
+    cfg.chanhigh  = ft_channelselection(cfg.chanhigh, freqhigh.label);
+    labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label));
```
schoffelen, line 100: This is not robust. It assumes that chanlow and chanhigh are cell-arrays with the same dimensionality (number of elements). This is not in general guaranteed. It seems that the desired behavior is to have an all-by-all pairwise low-to-high CFC, i.e. an NxM matrix of CFC estimates. I don't see a need to kick out the auto combinations a priori, nor to represent the data in a linear array of 'labelcmb'

ft_crossfrequencyanalysis.m
```diff
@@ -82,40 +93,46 @@
...
+    labelcmb(arrayfun(@(x) strcmp(labelcmb{x,:}), 1:size(labelcmb,1)),:) = []; % remove auto
```
schoffelen, line 101: why?

ft_crossfrequencyanalysis.m
```diff
@@ -257,12 +274,18 @@
...
 end % switch method for actual computation
 
-crossfreq.label      = cfg.channel;
 crossfreq.crsspctrm  = crsspctrm;
 crossfreq.dimord     = dimord;
 crossfreq.freqlow    = LF;
 crossfreq.freqhigh   = HF;
 
+if docrosschan
+    crossfreq.labelcmb = labelcmb;
```
schoffelen, line 283: In this representation it is ambiguous as to which channel is the phase providing signal, and which one is the amplitude highfreq signal.

schoffelen (changes requested):
Please come up with another strategy to create the labelcmb variable, because the ft_channelcombination call will not work in general.

Also we would need a test function to go along with this PR (according to the test functions in fieldtrip/test) which demonstrates how the function can be used, and which checks some common user scenarios (also in particular it should test the old behavior, and demonstrate that the current version of the code does not break old functionality)

ft_crossfrequencyanalysis.m
tiborauer: I do not have any such assumption. The pairwise low-to-high CFC would be a (N*M) x 2 matrix, where each row describes a channel pair from where the CFC is computed. In this way, the data has the same dimensionality for both within- and cross-channel analysis. with dimord `chan_freqlow_freqhigh` (for within) and `chancmb_freqlow_freqhigh` (for cross).

ft_crossfrequencyanalysis.m
tiborauer: Contrary to what you suggest, I found the `ft_channelcombination` quite useful, however, it seems to add the auto-combination even if I ask not to. That is why I explicitly remove them here.

ft_crossfrequencyanalysis.m
tiborauer: I try following the existing logic of the code, i.e. `freqlow` -> `freqhigh`, so that the first item of the pairs (see my comment on `labelcmb` above) corresponds to `freqlow`, while the second item of the pairs corresponds to `freqhigh`.

(I assume `freqlow` corresponds to phase  and `freqhigh` corresponds to `amplitude`.)

schoffelen: That's I think a valid assumption ( although I don't wrote the code myself). It's OK to follow the code's convention, but what I meant is that perhaps it is useful to make this implicit convention explicit to the user (either in the documentation, or in the labelname). E.g. rather than having labelcmb = {'chan01' 'chan02' ; ...}
have labelcmb = {'chan01phase' 'chan02amp'} or so.

ft_crossfrequencyanalysis.m
schoffelen: OK, I stand corrected. Apparently ft_channelcombination is smarter than I remember.

ft_crossfrequencyanalysis.m
tiborauer: I have added an explicit description of the output. Can you, please, take a look? I have kept the original convention ("low" and "high").

schoffelen: you mean in the documentation section? I checked, and rephrased it a bit.

schoffelen:
OK, pending a test function (and pending acceptable behavior of the function as per the test function) this can be merged...

tiborauer:
I have extended the test function now including cross-channel analysis. I also had to change the function a bit because it failed with having only a single channel or channel combination, such as the test.

ft_crossfrequencyanalysis.m
tiborauer: yes. thanks

schoffelen:
OK, I pulled this PR locally, and the test function runs through without error. 
I am fine with merging it as is, but for the next time would recommend:
- to start with a clean revision history -> right now there are two commit messages that pertain to changes that are irrelevant for this PR (because no files seem to have changes)
- file the PR from a branch that is not your local master branch (it's easier to keep track of the different branches).

Issue #1632, opened 2021-01-13
Omission in ft_compute_leadfield for MEG and OpenMEEG headmodel


bknipscheer:
**Describe the bug**
There seems to be an omission in the ft_compute_leadfield script causing some problems down the line. 
In the most recent version of ft_compute_leadfield, when calculating the leadfields for MEG using a headmodel based on OpenMEEG, this is done for the MEG coils and not the sensors.
There seem to be a few lines missing. These lines are:
  _if isfield(sens, 'tra')
    % construct the channels from a linear combination of all magnetometer coils
    lf = sens.tra * lf;
  end_

**To Reproduce**
1. Go to  fieldtrip/forward/ft_compute_leadfield.m .
2. Scroll down to  _case 'openmeeg'_ line 270.
3. See that the following lines are missing, while present in other headmodel types for example 'singleshell' on lines 265-268.
_if isfield(sens, 'tra')
        % compute the leadfield for each gradiometer (linear combination of coils)
     lf = sens.tra * lf;
end_

**Expected behavior**
It is expected that the leadfield is computed for each gradiometer for 'openmeeg'  similar to other headmodel types.

**Environment and versions (please complete the following information):**
 - Operating System: MATE 1.20.4
 - MATLAB version: 9.8.0.1323502 (R2020a)
 - FieldTrip version: revision 84a0dd028.

schoffelen:
Confirmed indeed. I have whipped up a small test function that demonstrates the issue as well. I will wait until you decide how to take this further until incorporating this test function in the further procedure.

schoffelen:
The test script looks good overall, so it will be usable in our test suite. The bottom line is, that currently the script should fail (since the number of rows in the leadfield is not according to the expected number of sensors), but in the future the script should not fail (i.e. after ft_compute_leadfield is fixed)

TODO (later) is to expand it, so that the leadfields are computed within the script, so that it can be ensured that future code changes do not disable the fix later on.

schoffelen:
PS: I'd recommend against using a 'column' normalization for the leadfields

schoffelen:
this has been fixed with #1633

PR #1633, opened 2021-01-13
Fix for issue #1632 by adding lf=sens.tra*lf for case 'openmeeg'

bknipscheer:
Fix for issue #1632 Omission in ft_compute_leadfield for MEG and OpenMEEG headmodel.
The following lines have been added in ft_compute_leadfield (lines 281-285):
_if isfield(sens, 'tra')
% compute the leadfield for each gradiometer (linear combination of coils)
lf = sens.tra * lf;
end_


schoffelen (approved):
Good stuff, thanks Bram!

PR #1635, opened 2021-01-14
Update ft_connectivityanalysis.m

fabiobaselice:
Added support to the ft_connectivity_plm function

### DIFF
--- a/ft_connectivityanalysis.m
+++ b/ft_connectivityanalysis.m
@@ -42,6 +42,7 @@ function [stat] = ft_connectivityanalysis(cfg, data)
 %     'wppc'       weighted pairwise phase consistency
 %     'corr'       Pearson correlation, support for timelock or raw data
 %     'laggedcoherence', lagged coherence estimate
+%     'plm'        phase linearity measurement
 %
 % Additional configuration options are
 %   cfg.channel    = Nx1 cell-array containing a list of channels which are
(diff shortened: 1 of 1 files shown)


schoffelen:
please also add the ft_connectivity_plm and a test function (or add some snippet of code to an existing testfunction).

also, for consideration: if I recall well, the ft_connectivity_plm that you wrote at some point is performing the Hilbert transform within the function. You could consider to make this conditional on complex-valued input, and let ft_preprocessing take care of both the bandpassfiltering and the Hilbert transform. The former is already implicitly assumed, i.e. the ft_connectivity_plm expects bandpassfiltered data in the input.

ft_connectivityanalysis.m
```diff
@@ -66,6 +67,8 @@
 %                     method 'powcorr' and 'amplcorr'.
 %   cfg.bandwidth   = scalar, needed for 'psi', half-bandwidth of the integration
 %                     across frequencies (in Hz, default is the Rayleigh frequency)
+%                     needed for 'plm', half-bandwidth of the integration window (in Hz)
+%   cfg.fsample     = scalar, needed for 'plm', sampling frequency of the data in [Hz]
```
schoffelen, line 71: If the input data structure is of type 'raw', the fsample is typically present, or can be easily deduced from the time-axis. I believe that ft_checkdata has an option that does this (and can thus be included in line 399). In other words, there should be no need to define the fsample separately in the cfg.

schoffelen (changes requested):

ft_connectivityanalysis.m
schoffelen: I see now that fsample is not necessarily present in 'raw' data.

PR #1637, opened 2021-01-14
ENH - docu fix and added nearest to private to keep preproc standalone

schoffelen:


schoffelen:
This PR addresses two issues:
- a typo in the documentation, causing confusion with respect to the correct naming of the input key-value arguments, which might have led to unexpected behavior if this documentation was literally followed, as per the original contribution of @sabineleske @sarangnemo 
- added nearest to the private folder of the preproc module, to avoid issues with overloaded matlab function if users do not have fieldtrip/utilities on their path

PR #1641, opened 2021-01-16
New mvpa interface

treder:


treder:
This version is now backwards compatible, it runs with the FT/MVPA tutorial and I've also tested it with your example code and seems to run fine. 

I've radically stripped down the code in _mv_statistics_mvpa_. The previous version contained a lot of bespoke code which I didn't fully understand. I hope that we can derive a shorter version of the function with the new interface, but likely some of the old code has to come back. 

If you think there's something missing let me know and I can try to put it in, in a way that exploits the MVPA-Light well. There's still a few things I don't fully understand:

- What's the purpose of cfg.mvpa.timwin? Isn't cfg.latency meant to select a time window from the data?
- What's the purpose of cfg.connectivity, can't we use cfg.mvpa.neighbours instead?

Lastly, on a related note, I'd like to extend the tutorial a bit with the following items, what do you think:
- Searchlight across both time and channels
- Time-frequency classification with different variants (searching across time, or freq, or both, or time/freq/channel) - not sure if this should be a separate tutorial on TF classification though

ft_statistics_mvpa.m
```diff
@@ -171,257 +173,132 @@
...
-if istrue(cfg.timextime) && istrue(cfg.searchlight)
-  ft_error('you should not set timextime = ''yes'' and searchlight = ''yes'' simultaneously')
+if isfield(cfg,'timextime') && strcmp(cfg.timextime, 'yes')
+    cfg.generalize = 'time';
+end
+if isfield(cfg,'searchlight') && strcmp(cfg.searchlight, 'yes')
+    cfg.features = 3;
+end
+if isfield(cfg,'normalise') && ~isfield(cfg.mvpa,'preprocess')
+    add_to_preprocess(cfg.normalise);
```
schoffelen, line 219: our convention is to have the subfunctions have explicit output arguments. I am not familiar with this syntax, but I assume that it somehow ends up in the cfg of the caller function anyhow? 

to avoid confusion, I suggest to change this (i.e. have add_to_preprocess output cfg.preprocess explicitly.)

ft_statistics_mvpa.m
```diff
@@ -171,257 +173,132 @@
...
+end
+if isfield(cfg,'balance') && ~isempty(cfg.balance)
+    add_to_preprocess(cfg.balance);
```
schoffelen, line 222: would it be a big deal to consistently have the tab indentation to be at 2 spaces?

schoffelen (changes requested):

ft_statistics_mvpa.m
treder: It's an internal function so it has access to everything in the caller. I don't mind changing the syntax though for consistency.

ft_statistics_mvpa.m
treder: nope - i switched from 4 to 2 spaces throughout the file

treder:
both changes are now in the file

gdemarchi:
@treder Since you are restructuring the ft_ to mvpa interface, any chance to be able to access the cross decoding from the higher level ft_ functions directly, instead of relying to "lower level" mv_? Didn't want to hijack the thread, just to mention a nice feature that would be nice to be able to "tap" from ft_ ...

ft_statistics_mvpa.m
schoffelen: thanks

schoffelen:
> If you think there's something missing let me know and I can try to put it in, in a way that exploits the MVPA-Light well. There's still a few things I don't fully understand:
> 
> * What's the purpose of cfg.mvpa.timwin? Isn't cfg.latency meant to select a time window from the data?

This scalar serves to define the width of the temporal searchlight (in samples), it is used to create a 'T' matrix that defines neighbours across time.

> * What's the purpose of cfg.connectivity, can't we use cfg.mvpa.neighbours instead?

Uhm, dunno. connectivity is a boolean matrix, neighbours (at least in FT) is a struct-array, which still needs to be interpreted, i.e. converted into a matrix. In general the high level code knows which channels are present in the data, so it can correctly create a matrix from a neighbours structure array, the low level code does not necessarily know this, but assumes that the order of channels/parcels/spatial points in the data are the same as in the connectivity matrix. FT supports both a connectivity matrix, and a neighbours struct array, where the connectivity matrix is used at the user's own risk. As a side note, with a triangulated mesh the connectivity is defined as per the triangles, and not via a struct array.


 
> Lastly, on a related note, I'd like to extend the tutorial a bit with the following items, what do you think:
> 
> * Searchlight across both time and channels

Yes please! I am totally in favour, and this is the strength of EEG/MEG to begin with.

> * Time-frequency classification with different variants (searching across time, or freq, or both, or time/freq/channel) - not sure if this should be a separate tutorial on TF classification though

No, this could well go into here. Note that we then need to check whether ft_freqstatistics smoothly interfaces to ft_statistics_mvpa

schoffelen:
> @treder Since you are restructuring the ft_ to mvpa interface, any chance to be able to access the cross decoding from the higher level ft_ functions directly, instead of relying to "lower level" mv_? Didn't want to hijack the thread, just to mention a nice feature that would be nice to be able to "tap" from ft_ ...

Hi GP, no worries (as far as I am concerned). If you already have some example code handy, please paste it in this thread, or file a PR onto this PR :)

gdemarchi:
Well ... no, no code yet :-)

We often use (or always, for one reason or another we always train and test on different things/conditions/modalities): 
[acc, res] = mv_classify_timextime(cfg, data_condA, labelsA, data_condB, labelsB);

When the first ft_mvpa_ tutorial came out, I remember discussing of this issue with Matthias, but if I'm not mistaken he mentioned that a major rework of the ft_statfun was needed ... I was just wondering whether this would be the right time, but the ft_statfun internals are still mysterious to me, that's why we were sticking to the (relatively) low level mv_*, but it would be a nice feature doing straight from ft_

treder:
> > * What's the purpose of cfg.mvpa.timwin? Isn't cfg.latency meant to select a time window from the data?
> 
> This scalar serves to define the width of the temporal searchlight (in samples), it is used to create a 'T' matrix that defines neighbours across time.

ah OK, I can include this and update the PR.


> > * What's the purpose of cfg.connectivity, can't we use cfg.mvpa.neighbours instead?
> 
> Uhm, dunno. connectivity is a boolean matrix, neighbours (at least in FT) is a struct-array, which still needs to be interpreted, i.e. converted into a matrix. In general the high level code knows which channels are present in the data, so it can correctly create a matrix from a neighbours structure array, the low level code does not necessarily know this, but assumes that the order of channels/parcels/spatial points in the data are the same as in the connectivity matrix. FT supports both a connectivity matrix, and a neighbours struct array, where the connectivity matrix is used at the user's own risk. As a side note, with a triangulated mesh the connectivity is defined as per the triangles, and not via a struct array.

Ah I see. That's a clash of terminologies then, in MVPA-Light neighbours are always represented as a Boolean matrix, so cfg.mvpa.neighbours should be this Boolean matrix. What I'm doing right now is set cfg.mvpa.neighbours = cfg.connectivity if the former doesn't exist but the latter does. Furthermore, any struct arrays are converted to binary matrices using channelconnectivity.
So setting either cfg.connectivity or cfg.mvpa.neighbours should work, and both boolean matrix and struct arrray should work too.


> > * Searchlight across both time and channels
> 
> Yes please! I am totally in favour, and this is the strength of EEG/MEG to begin with.
> 
> > * Time-frequency classification with different variants (searching across time, or freq, or both, or time/freq/channel) - not sure if this should be a separate tutorial on TF classification though
> 
> No, this could well go into here. Note that we then need to check whether ft_freqstatistics smoothly interfaces to ft_statistics_mvpa

OK great.

treder:
@GP thanks for the reminder, this is a thing that's on the todo list. So far only mv_classify_timextime supports cross-decoding, mv_classify doesn't. I was planning to develop a generic cross-decoding solution in mv_classify first which works for any number/order of dimensions, and then put this into FT. 

An alternative would be to implement a solution that only works for the time dimension for 3D data for the time being, but this would still need thinking on how to submit the extra dataset to FT.

schoffelen:
> @gp thanks for the reminder, this is a thing that's on the todo list. So far only mv_classify_timextime supports cross-decoding, mv_classify doesn't. I was planning to develop a generic cross-decoding solution in mv_classify first which works for any number/order of dimensions, and then put this into FT.
> 
> An alternative would be to implement a solution that only works for the time dimension for 3D data for the time being, but this would still need thinking on how to submit the extra dataset to FT.

As an alternative - rather than adding a data argument to the function - you could consider supporting a non-CV classification/regression using a pre-computed model, i.e. passing the model parameters directly in the input, and apply it to the data. That would do it, right?
This would require the model to be present in the output of a (first pass) of ft_timelockstatistics.

treder:
> > * What's the purpose of cfg.connectivity, can't we use cfg.mvpa.neighbours instead?
> 
> Uhm, dunno. connectivity is a boolean matrix, neighbours (at least in FT) is a struct-array, which still needs to be interpreted, i.e. converted into a matrix. In general the high level code knows which channels are present in the data, so it can correctly create a matrix from a neighbours structure array, the low level code does not necessarily know this, but assumes that the order of channels/parcels/spatial points in the data are the same as in the connectivity matrix. FT supports both a connectivity matrix, and a neighbours struct array, where the connectivity matrix is used at the user's own risk. As a side note, with a triangulated mesh the connectivity is defined as per the triangles, and not via a struct array.

I think the question of specifying boolean matrices for searchlights needs more clarification. In ML, cfg.mvpa.neighbours should be a cell array if there's multiple search dimensions. For instance, if you search across three dimensions channels/time/freq, it should be a cell array with 3 boolean matrices.

I'm now wondering whether it's better if a FT user does not use cfg.mvpa.neighbours directly but rather simply sets cfg.connectivity (or alternatively cfg.neighbours), cfg.timwin (for time), and cfg.freqwin (same idea for frequency). Then, _mv_statistics_mvpa_ collects everything in cfg.mvpa.neighbours, using cfg.dimord to identify the order of the search dimensions. (we can still allow cfg.mvpa.neighbours to be set by the user if they want to specify say more unusual boolean matrices by hand)

Just to be extra clear, now cfg.neighbours is the FT neighbours struct array and cfg.mvpa.neighbours is the  boolean matrices that serve as input to ML. Would this work?

treder:
> As an alternative - rather than adding a data argument to the function - you could consider supporting a non-CV classification/regression using a pre-computed model, i.e. passing the model parameters directly in the input, and apply it to the data. That would do it, right?
> This would require the model to be present in the output of a (first pass) of ft_timelockstatistics.

Not sure how this would exactly work, since you still need to train/test multiple times. E.g. for time cross-decoding you still need to train the classifier/regression model for every training time point separately.

Another option might be something like this (?): 
```
ft_timelockstatistics(cfg, dataset_subject1, dataset_subject2)
```

Then both datasets get pasted together by FT. The only thing we need to do is add an extra row to cfg.design which is [1 1 1 1 1....... 2 2 2 2] i.e. specifies which trials belongs to which subject. The only problem I see here is if the two dataset do not have the same number of time points, because the data can't be concatenated by FT?

schoffelen:
> > As an alternative - rather than adding a data argument to the function - you could consider supporting a non-CV classification/regression using a pre-computed model, i.e. passing the model parameters directly in the input, and apply it to the data. That would do it, right?
> > This would require the model to be present in the output of a (first pass) of ft_timelockstatistics.
> 
> Not sure how this would exactly work, since you still need to train/test multiple times. E.g. for time cross-decoding you still need to train the classifier/regression model for every training time point separately.
> 
> Another option might be something like this (?):
> 
> ```
> ft_timelockstatistics(cfg, dataset_subject1, dataset_subject2)
> ```
> 
> Then both datasets get pasted together by FT. The only thing we need to do is add an extra row to cfg.design which is [1 1 1 1 1....... 2 2 2 2] i.e. specifies which trials belongs to which subject. The only problem I see here is if the two dataset do not have the same number of time points, because the data can't be concatenated by FT?

As long as the data are sampled on the same time axis, they can be concatenated with ft_appenddata/ft_appendtimelock (filling missing points with NaNs).

schoffelen:
> > > As an alternative - rather than adding a data argument to the function - you could consider supporting a non-CV classification/regression using a pre-computed model, i.e. passing the model parameters directly in the input, and apply it to the data. That would do it, right?
> > > This would require the model to be present in the output of a (first pass) of ft_timelockstatistics.
> > 
> > 
> > Not sure how this would exactly work, since you still need to train/test multiple times. E.g. for time cross-decoding you still need to train the classifier/regression model for every training time point separately.
> > Another option might be something like this (?):
> > ```
> > ft_timelockstatistics(cfg, dataset_subject1, dataset_subject2)
> > ```
> > 
> > 
> > Then both datasets get pasted together by FT. The only thing we need to do is add an extra row to cfg.design which is [1 1 1 1 1....... 2 2 2 2] i.e. specifies which trials belongs to which subject. The only problem I see here is if the two dataset do not have the same number of time points, because the data can't be concatenated by FT?
> 
> As long as the data are sampled on the same time axis, they can be concatenated with ft_appenddata/ft_appendtimelock (filling missing points with NaNs).

but thinking it over, the concatenation dimension is the epochs, right? The assumption is that the features (at least the channels) are comparable across data structures, or not?

gdemarchi:
A lot of ideas going on ...

> On 18.01.2021, at 17:24, Matthias Treder <notifications@github.com> wrote:
> 
> 
> As an alternative - rather than adding a data argument to the function - you could consider supporting a non-CV classification/regression using a pre-computed model, i.e. passing the model parameters directly in the input, and apply it to the data. That would do it, right?
> This would require the model to be present in the output of a (first pass) of ft_timelockstatistics.
> 
> Not sure how this would exactly work, since you still need to train/test multiple times. E.g. for time cross-decoding you still need to train the classifier/regression model for every training time point separately.
> 
> 

As I said, from ft_stat*, “ich verstehe nur Bahnhof ..."
> Another option might be something like this (?):
> 
> ft_timelockstatistics(cfg, dataset_subject1, dataset_subject2)
> Then both datasets get pasted together by FT. The only thing we need to do is add an extra row to cfg.design which is [1 1 1 1 1....... 2 2 2 2] i.e. specifies which trials belongs to which subject. The only problem I see here is if the two dataset do not have the same number of time points, because the data can't be concatenated by FT?
> 
… that indeed happens all the time, that I train on (many) trials of condition A, and test on (less many) trials of condition B. 
The “top down” solution from the ft_ side would be obviously more elegant, on the other hand I indeed use *only* mv_classify_timextime, if I need the train/test on the same time point I just take the diagonal of the output of the previous function … I know it’s a waste of computing power, but your toolbox Matthias is so fast anyway that I don’t bother :-)

gdemarchi:
> 
> but thinking it over, the concatenation dimension is the epochs, right? The assumption is that the features (at least the channels) are comparable across data structures, or not?
> 
> 

Yes, from the user perspective we train/test on different epochs/trials of different conditions, which would require some bookkeeping here … I cannot imagine now something different …. more general … dunno … that one trains on left temporal channels/voxels and tests on the right side ?!

schoffelen:
when I try to run test_tutorial_mvpa_light, which is a markdown2matlab conversion of the tutorial page (once upon a time, so I am not sure whether there's a 1-to-1 match with the current version), I get an error already early on (in line 81). The test function is in fieldtrip/test, and should run throuth without errors, before we can merge this PR. This would either mean adjusting the tutorial code, or fixing it in ft_statistics_mvpa.

treder:
OK working on this. The error was due to small changes I hadn't committed. I'm simultaneously working on extending the tutorial to make sure the code will run for channel x time searchlights etc.

One issue I'm facing is how to specify 'neighbours' for different dimensions (e.g. channels, time points, frequencies ...) simultaneously in a consistent and extendible way. What we have so far:

- cfg.mvpa.neighbours: struct array or [channel x channel] boolean matrix
- cfg.mvpa.timwin: can be used to specify the width of a window of neighbouring time points
( - cfg.mvpa.freqwin: not implemented but freq could be specified in the same way as time)


This isn't great because (1) syntax and semantics for channels vs time points are very different and (2) there's a name clash because in MVPA-Light the neighbours field is used to specify any type of neighbours (chan / time / freq / ...) so eventually all the chan/time/freq neighbours need to go into the cfg.mvpa.neighbours field. Alternative proposal:

- cfg.mvpa.chan_neighbours: struct array or channel x channel boolean matrix (same as former cfg.mvpa.neighbours). If cfg.neighbours or cfg.connectivity are given cfg.mvpa.chan_neighbours is set to that value.
- cfg.mvpa.time_neighbours: either scalar (which makes it act like cfg.mvpa.timwin) or time x time boolean matrix
- cfg.mvpa.freq_neighbours: either scalar (which makes it act like cfg.mvpa.freqwin) or freq x freq boolean matrix

_ft_statistics_mvpa_ would then compile these together into the cfg.mvpa.neighbours field which goes into MVPA-Light. 

An alternative would be to keep cfg.mvpa.timwin as is and make it also accept a [time x time] matrix as input (same for cfg.mvpa.freqwin). 

Any preference, also in light of usage of neighbours/connectivity elsewhere in FT?

treder:
@gdemarchi  unfortunately, I won't manage to implement a proper solution for cross-decoding just now, but it's first item now on my todo list, I would like to get it right & generic so will need to take some time for it later

gdemarchi:
Thanks, no worries … as said it would be a “nice to have” but as far as I’m concerned I can keep using mvpa light at the “low level” ... it would be just cooler doing in field trip from A to Z, though it will “hide” you nice toolbox :-) !

treder:
Hi Jan-Mathijs, do you have any comments per above? The main question is how to solve the 'neighbours' issue. In MVPA-Light, all dimensions are equal in principle, i.e. you can define neighbours for time points, channels, frequencies, and any other dimension. 

What's the best way to integrate this into the FT interface, in a way that's both flexible and convenient?

schoffelen:
Hi @treder, sorry that this slipped off my radar. I will ponder a bit about this and get back to you soon!

schoffelen:
Hi Matthias, these are my thoughts:

The to-be-converged upon fields cfg.mvpa structure will need to be according to the syntax in the mv_* functions. As such we should take care that cfg.mvpa.neighbours will be consistent with what the lower-level is used to working with. 
The wrapper ft_statistics_mvpa can take care of the necessary reformatting and packaging of the options, by translating FT-style options cfg.neighbours (struct-array, as per its use in ft_megplanar/ft_timelockstatistics etc.) into a boolean cfg.mvpa.neighbours or so. Likewise it can translate a scalar timwin/freqwin into an appropriate boolean matrix. Note that for permutation based statistics (depending on ft_statisitcs_montecarlo as lower-level function) the user specifies cfg.neighbours (or cfg.connectivity) and not cfg.montecarlo.neighbours. Essentially, unless I am mistaken, this is what you already sketched in your last comment, with the only difference that I think of the neighbourhood-defining options to be specified at the top-level of the cfg, and not in cfg.mvpa. Does this make sense?

PS we could think of building in a sanity check to avoid users defining neighbours at different levels of the cfg, e.g. have stuff that is defined in cfg.mvpa take precedence. This is typically achieved with code chunks like:

cfg.mvpa.neighbours = ft_getop(cfg.mvpa, 'neighbours', []);
if isempty(neighbours)
  % convert cfg.neighbours, cfg.timwin, cfg.freqwin into the appropriate boolean matrices and package them appropriately into the cfg.mvpa neighbours field
end

treder:
OK I performed an update per your points: if cfg.mvpa.neighbours is not set, it is being built from cfg.neighbours/cfg.connectivity/cfg.timwin/cfg.freqwin. 
Note that I also submitted a PR with a new tutorial which you can use to check the code if you like. Let me know if any more fixed are required.

schoffelen:
hmmm. I have tested this branch with some data (and cfg) that I use with a collaborator on her project. Apart from the need to define cfg.features explicitly as empty, the code runs. Yet, the results are vastly different:

Screenshot 2021-03-09 at 17 15 26

on the left there's a plot of the results obtained with the new code, on the right the results based on the old code.

I think it is not desired to have such a large difference in functional behavior with virtually the same input into the functions. Let me see whether I can create a lean data-structure that I can upload somewhere so that @treder can also think along.

schoffelen:
oh, and what needs to be addressed still (but which was also incorrect before) is that the output label does not match the channel dimension in the data.

treder:
Thanks for the quick check. If you can share any cfg struct and data dimensions I could try to replicate the issue.

just to be extra clear: cfg.searchlight is deprecated and ideally shouldn't be used anymore, you can define searchlight more flexibly now using cfg.features. Let's assume your data is [samples x chan x time]. Then cfg.features = 'chan' (or equivalently cfg.features = 2) performs a classification for every time point (=search across time) using channels as features. If cfg.features = 'time' (or equivalently cfg.features = 3) performs a classification for every channel using all time points as features. If cfg.features = [] the search is performed across both chan and time, yielding a 2D matrix of classification results. This approach generalizes to any dimensionality of the data (e.g. same for time-frequency) including 5D, 6D etc data so it's future compatible ;)

Would it help to make the default value for cfg.features context dependent? Something like this:
- if cfg.connectivity or cfg.neighbours is not empty, cfg.features defaults to 3 (3rd dimension is usually time)
- optional: if cfg.timwin is not empty (but cfg.connectivity/neighbours is) cfg.features default to 'chan' 
- optional: similar approach for cfg.freqwin
- otherwise cfg.features = []

schoffelen:
Yes, I think it would be useful to have some educated guessing by the code to specify the default.
Perhaps, also, we may consider building in some explicit checks w.r.t. deprecated options. This is usually done with ft_checkconfig:   ft_checkconfig(cfg, 'deprecated', <option-names that are deprecated>).

I will try and send you some data + cfg

schoffelen:
Ouch! well spotted with respect to the diagonal being zero. I guess a straightforward fix would be that the diagonal is added after the call to channelconnectivity, right?

schoffelen:
OK, I noticed 2 things with the latest version:
- cfg.features still needs to be explicitly defined, in order for the searchlight to kick in.
- cfg.timwin has moved up one level in the structure (originally I had cfg.mvpa.timwin). 
The latter in principle I can live with, it would be nice to have a solution for the first one, but it's not a big deal either, so can be gracefully ignored

treder:
>Note that I made a small fix (directly into this PR) to the way that the new labels are going to be defined, in case the neighbour matrix is non-square

Thanks for the fix.

> Ouch! well spotted with respect to the diagonal being zero. I guess a straightforward fix would be that the diagonal is added after the call to channelconnectivity, right?

Yes indeed. I implemented this (l.301). 
Obviously if the matrix is non-square it can't be fixed, so if cfg.connectivity is given directly (like in the example you sent) the source channel has to be in the matrix.

> cfg.timwin has moved up one level in the structure (originally I had cfg.mvpa.timwin).

I thought we figured in the end that FieldTrip/mv_statitstics_mvpa specific features should be in cfg and only things that go into MVPA-Light should be in cfg.mvpa? timwin does not exist in the toolbox, but I'm happy to move it back to cfg.mvpa if that's better.

> cfg.features still needs to be explicitly defined, in order for the searchlight to kick in.

Ah yes the default assumed you want to use time as features because I didn't verify whether timwin/freqwin are empty, fixed it now. The defaults don't do anything smart (yet) when only cfg.connectivity/neighbours and cfg.freqwin are set.

```
if ~isfield(cfg, 'features')
  % no features provided, we have to guess
  if ~isempty(cfg.neighbours) || ~isempty(cfg.connectivity) 
    if isempty(cfg.timwin)
      cfg.features = 3; % dimension 3 = time (usually)
    end
  elseif ~isempty(cfg.timwin) || ~isempty(cfg.freqwin)
    cfg.features = 2; % dimension 2 = chan (usually)
  end
end
cfg.features = ft_getopt(cfg, 'features', []);
```

schoffelen:
Great, thanks for the updates to the code. It runs fine I think (at least on my limited use case). Keeping the timwin/freqwin etc. at the level of cfg indeed is good (and as discussed). If you're happy with the PR, I'll merge, just give me a thumbs up.

treder:
OK great.

We're almost there, I just found another problem when running through the tutorial again: Due to changing the channel labels in the tutorial "Search across channels ('where')", the channel labels in `stat` are not recognised and it crashes at this part:

```
  cfg              = [];
  cfg.parameter    = 'accuracy';
  cfg.layout       = 'CTF151_helmet.mat';            
  cfg.xlim         = [0, 0];
  cfg.colorbar     = 'yes';
  ft_topoplotER(cfg, stat);
```

The obvious fix would be to set the channel labels back to only the source channel name. But this wouldn't work in the data you sent me when cfg.connecttivity is provided because it's not clear which channel is the source channel. Any ideas?

schoffelen:
I would say that the labels need to be updated only, if there's a spatial neighbourhood matrix that deviates from identity, in which case it would be the responsibility of the user to replace the labels to something that the plotting function can recognize. In a 'simple' user scenario, I'd not create fancy labels to begin with (i.e. bypass the creation of 'combined('channellist')' as new labels

treder:
OK done. If the neighbours matrix is square and has a diagonal the original labels are used. I've double checked the tutorial as well, it all runs, so the website should be ready for an update, too. Ready to launch 🚀

schoffelen:
Awesome. Thanks for your hard work.

PR #1642, opened 2021-01-17
updated line 558 and 559

sagihaider:
We believe these two lines need to be swapped after assignment operator (i.e. line number 558 and 559). When we are validating the BIDS data it was throwing an error and after carefully looking at it we believed they need to be swapped.
Many thanks!


schoffelen (approved):

schoffelen:
Thanks!

schoffelen:
Sorry, wrong button!

Issue #1644, opened 2021-01-18
reading BrainVision timestamps gives error


robertoostenveld:
**Describe the bug**

Reading BrainVision timestamps gives an error due to the changes introduced by @nbeliy in #1625

**To Reproduce**

This was detected in the `fieldtrip/test/test_bug2482.m` script, with the file `/home/common/matlab/fieldtrip/data/test/bug2482/001 P11 M1 004_Seg_fs6.vhdr`

**Expected behavior**

The existing test scripts should never start failing, since that would also mean that existing scripts on the computer of FT users might fail. 

**Screenshots**

```
Error using datetime.datetime (line 635)
Unable to convert '00000000000106774000' to datetime using the format 'yyyyMMddHHmmssSSSSSS'.
Error in ft_read_event (line 528)
Error in ft_trialfun_brainvision_segmented (line 48)
Error in ft_definetrial (line 181)
Error in test_bug2482 (line 17)
```

nbeliy:
the timestamp in the log `00000000000106774000` looks weird (anonymised?)

BV gives it's timestamps [explicetly as](https://www.brainproducts.com/files/public/products/more/BrainVisionCoreDataFormat_1-0.pdf) 
```
Date Example19990311140312003012means 11 March 1999, 14:03:12.003012
```

robertoostenveld:
yes, but the one you give is explicitly an example and might be the ideal case, whereas in reality they can be more variable. The file I have is one from an actual experiment and seems to be generated by BrainVision Analyzer. So files like these are out and about.

This is the start of the vhdr:
```
 Brain Vision Data Exchange Header File Version 2.0
; Data created from history path: 001 P11 M1 004/Raw Data/InitChans_electrodepos_256/new_positions/Filters_0.1-80Hz/without_channels_cheek/ICA/Inverse ICA/Raw Data Inspection/interpol_39,81,90132,144/new_Ref_linked_Mastoid/Filters_1-20Hz/Seg_fs6

[Common Infos]
Codepage=UTF-8
DataFile=001 P11 M1 004_Seg_fs6.eeg
MarkerFile=001 P11 M1 004_Seg_fs6.vmrk
DataFormat=BINARY
; Data orientation: VECTORIZED=ch1,pt1, ch1,pt2..., MULTIPLEXED=ch1,pt1, ch2,pt1 ...
DataOrientation=MULTIPLEXED
DataType=TIMEDOMAIN
NumberOfChannels=204
[...]

PR #1646, opened 2021-01-20
Fixing warnngs as discussed in #1644

nbeliy:
Added `readTime` boolean variable, which is switched off when parcing timestamps fails


robertoostenveld:
👌

PR #1647, opened 2021-01-20
Fix visualisation (as done for rejectvisual_trial)

benjybarnett:
Visualisation was fixed for the trial based view of ft_rejectvisual a few days ago, but the issue remained for the channel based view. I've fixed this now.


schoffelen (approved):
Thanks!

PR #1649, opened 2021-01-21
speed up the checking of the grad and elec structures

### LINKED ISSUE #1551: Appreciable slowdown of code performance with both grad and elec field in datastructure
I am wondering about the following thing:

Currently, it seems that potentially a lot of 'computation time' is spent doing low-level data-structure bookkeeping operations (ft_senstype related), when a MEG raw data structure contains both a grad and elec field. This is mostly due to the fact that the persistent variable trick in the lower-level utility functions does not work, because the slightly higher-level low level functions alternate between using grad and elec as input. 
If I encounter this slowdown (and get annoyed by it), I usually ditch the elec field, and then I am happy again.

Would it be worthwile to consider making it a standard feature of the code, to ditch elec/grad, conditional on the data.label?

robertoostenveld:
This hopefully fixes #1551.

### DIFF
--- a/fileio/ft_chanunit.m
+++ b/fileio/ft_chanunit.m
@@ -67,7 +67,7 @@ if isheader
 end
 
 current_argin = {input, desired};
-if isequal(current_argin, previous_argin)
+if false % isequal(current_argin, previous_argin)
   % don't do the type detection again, but return the previous output from cache
   chanunit = previous_argout{1};
   return
--- a/fileio/ft_chanunit.m
+++ b/fileio/ft_chanunit.m
@@ -67,7 +67,7 @@ if isheader
 end
 
 current_argin = {input, desired};
-if false % isequal(current_argin, previous_argin)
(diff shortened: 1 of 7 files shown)


fileio/ft_chanunit.m
```diff
@@ -67,7 +67,7 @@
 end
 
 current_argin = {input, desired};
-if isequal(current_argin, previous_argin)
+if false % isequal(current_argin, previous_argin)
```
robertoostenveld, line 70: oops, this needs to be fixed

utilities/ft_datatype_sens.m
```diff
@@ -121,7 +121,7 @@
 persistent previous_argin previous_argout
 
 current_argin = [{sens} varargin];
-if isequal(current_argin, previous_argin)
+if false %% isequal(current_argin, previous_argin)
```
robertoostenveld, line 124: oops, also needs to be fixed

PR #1651, opened 2021-01-21
Fix data formatting in ft_connectivityanalysis.m for 'powcorr_orth' method

cascam07:
When using the powcorr_orth method in ft_connectivityanalysis, I found that all of my results looked like random noise. After contacting a former coleague, Sean Tanabe, who had used this method, I found that he had previously discussed a fix for this code with Jan-Mathis, who requested that the code be changed and a pull request made. Reference to correspondance: https://mailman.science.ru.nl/pipermail/fieldtrip/2018-November/038411.html

This evidently never occurred, so I am making the pull request now.


schoffelen:
thanks

PR #1652, opened 2021-01-22
eLoreta + precision

tiborauer:


ft_statistics_montecarlo.m
```diff
@@ -77,6 +77,7 @@
 %   cfg.voxelstatistic   deprecated
 %   cfg.voxelthreshold   deprecated
 %   cfg.precondition     before|after|[], for the statfun
+%   cfg.precision        single|double (default)
```
schoffelen, line 80: I understand why you would want this, but are you sure that this is going to work, i.e. will the typecasting be done correctly if cfg.precision is 'single' and the data is 'double'. Also, why would it only be used as an argument in this function, and not in other functions with a potentially large memory footprint. 
I think in general, also, it would be good not to put commits with two different purposes into a single PR for transparency.

inverse/ft_inverse_eloreta.m
```diff
@@ -133,37 +133,37 @@
...
-Ndip          = numel(sourcemodel.leadfield);
-leadfield     = permute(reshape(cat(2,sourcemodel.leadfield{:}),Nchan,Nori,Ndip),[1 3 2]);
-
 % use existing filters, or compute them
 if ~hasfilter
-  filt = mkfilt_eloreta(leadfield, lambda);
-  for i=1:size(sourcemodel.pos,1)
-    sourcemodel.filter{i,1} = squeeze(filt(:,i,:))';
-  end
+    % deal with reduced rank
```
schoffelen: It's unclear what the change in the code is here, the largest diff is due to indentation changing from the (conventional) 2, to 4.

schoffelen (commented):
It's still not clear what the changes are to the eloreta code

ft_statistics_montecarlo.m
tiborauer: I agree with you, that it is not ideal to pack different things into one PR, but I did not want to submit another PR just for this minor change. 
I agree with you also in that other functions may also benefit from setting precision. but this was the only function I had an issue with even on our cluster with a generous amount of RAM. Considering temporal data combined with a large number of permutations, this function in no doubt consumes the most memory. Besides, going through the whole codebase is beyond my paycheck. :)
Regarding functionality, yes, it works as expected at least for my datasets.

inverse/ft_inverse_eloreta.m
tiborauer: Indentation changing is due to refractoring the code. The main issue was that the function was looking for the leadfield even when a precomputed filter was provided. This commit puts the leadfield processing part under the condition check `~hasFilter`.

schoffelen: Well, I changed some of the indentation, because the tab length was also changed.

schoffelen:
I am fine with the changes in ft_inverse_eloreta. Please make a new PR with only those changes.

If you want the precision argument in ft_statistics_montecarlo (or something similar) please file this in a separate PR. I think for the latter one, a suggestion that does not rely on the introduction of new cfg options, and which is equally defendible (and which would even fit your  - and mine, for that matter - paygrade) would be the preservation of the input data type (i.e. if the numeric data in the input is single precision, statrand will be single precision).

tiborauer:
I thought introducing the cfg option (what I have seen in other ft functions, too) would be a more transparent solution., but if you prefer type-preservation, then I am hppy to implement that, as well.

schoffelen:
If I grep for precision, I get: 

(base) jm-laptop:fieldtrip jansch$ grep precision *.m
fieldtrip2spss.m:dlmwrite(filename, data, '-append', 'delimiter', '\t', 'precision', 4);
ft_databrowser.m:cfg.precision           = ft_getopt(cfg, 'precision', 'double');
ft_databrowser.m:  % convert the data to another numeric precision, i.e. double, single or int32
ft_databrowser.m:  if ~isempty(cfg.precision)
ft_databrowser.m:    dat = cast(dat, cfg.precision);
ft_databrowser.m:% convert the data to another numeric precision, i.e. double, single or int32
ft_databrowser.m:if ~isempty(cfg.precision)
ft_databrowser.m:  dat = cast(dat, cfg.precision);
ft_databrowser.m:  tim = [tim linspace(tim(end),tim(end)+nsamplepad*mean(diff(tim)),nsamplepad)];  % possible machine precision error here
ft_freqanalysis.m:cfg.precision = ft_getopt(cfg, 'precision', 'double');
ft_freqanalysis.m:      if powflg, powspctrm     = zeros(nchan,nfoi,ntoi,cfg.precision);             end
ft_freqanalysis.m:      if csdflg, crsspctrm     = complex(zeros(nchancmb,nfoi,ntoi,cfg.precision)); end
ft_freqanalysis.m:      if fftflg, fourierspctrm = complex(zeros(nchan,nfoi,ntoi,cfg.precision));    end
ft_freqanalysis.m:      if powflg, powspctrm     = nan(ntrials,nchan,nfoi,ntoi,cfg.precision);                                                                 end
ft_freqanalysis.m:      if csdflg, crsspctrm     = complex(nan(ntrials,nchancmb,nfoi,ntoi,cfg.precision),nan(ntrials,nchancmb,nfoi,ntoi,cfg.precision)); end
ft_freqanalysis.m:      if fftflg, fourierspctrm = complex(nan(ntrials,nchan,nfoi,ntoi,cfg.precision),nan(ntrials,nchan,nfoi,ntoi,cfg.precision));       end
ft_freqanalysis.m:      if powflg, powspctrm     = zeros(ntaptrl,nchan,nfoi,ntoi,cfg.precision);        end %
ft_freqanalysis.m:      if csdflg, crsspctrm     = complex(zeros(ntaptrl,nchancmb,nfoi,ntoi,cfg.precision)); end
ft_freqanalysis.m:      if fftflg, fourierspctrm = complex(zeros(ntaptrl,nchan,nfoi,ntoi,cfg.precision));    end
ft_mvaranalysis.m:  % ensure that the input arguments are double precision
ft_mvaranalysis.m:      % ensure that the input arguments are double precision
ft_mvaranalysis.m:% precision this prevents an instability (bug) in the computation of the
ft_preprocessing.m:%   cfg.precision     = 'single' or 'double' (default = 'double')
ft_preprocessing.m:cfg.precision      = ft_getopt(cfg, 'precision', 'double');
ft_preprocessing.m:      % convert the data to another numeric precision, i.e. double, single or int32
ft_preprocessing.m:      if ~isempty(cfg.precision)
ft_preprocessing.m:        dat = cast(dat, cfg.precision);
ft_resampledata.m:        % temporary convert this trial to double precision
ft_resampledata.m:        % temporary convert this trial to double precision
ft_resampledata.m:        % temporary convert this trial to double precision
ft_resampledata.m:        % temporary convert this trial to double precision
ft_resampledata.m:        % temporary convert this trial to double precision
ft_sourceinterpolate.m:% ensure the functional data to be in double precision
ft_sourceinterpolate.m:          fv    = double_ifnot(dat_array{i}(:,:,:,k,m)); % ensure double precision to allow sparse multiplication
ft_sourceinterpolate.m:% SUBFUNCTION to cast array to double precision, only if needed
ft_sourceplot.m:    ft_warning('converting functional data to double precision');
ft_sourcewrite.m:%   cfg.precision = string, can be 'single', 'double', etc.
ft_sourcewrite.m:cfg.precision      = ft_getopt(cfg, 'precision');       % is used for cifti
ft_sourcewrite.m:    ft_write_cifti(cfg.filename, source, 'parameter', cfg.parameter, 'brainstructure', cfg.brainstructure, 'parcellation', cfg.parcellation, 'precision', cfg.precision);
ft_statistics_mvpa.m:%                         accuracy auc tval dval confusion precision

Indeed, for ft_preprocessing and ft_freqanalysis precision is a cfg-option. Yet, the behavior of this option is that it determines the datatype of the output data. your suggested changes only affects the intermediate data. That should be properly documented, or the data should be also outputted in the specified precision (which would have my preference, to stay consistent). Note that ft_statistics_montecarlo is not typically called by the user directly, so in order for this change to be availabel higher up, ft_timelockstatistics etc, should also be adjusted (+their documentation sections)

tiborauer:
Makes sense. Thank you for the clarification.

PR #1655, opened 2021-01-24
ENH - implemented reading header and data from matlab file

robertoostenveld:
I noticed that it was already supported to read **events** from a mat file, but not the **data** or **header**. That meant that `ft_definetrial` still would not work. I now made all three ft_read_xxx functions consistent. The mat file should either contain hdr, dat, event, or it should contain a FT raw data structure.

### DIFF
--- a/fileio/ft_read_event.m
+++ b/fileio/ft_read_event.m
@@ -1341,9 +1341,6 @@ switch eventformat
     end
     
   case {'itab_raw' 'itab_mhd'}
-    if isempty(hdr)
-      hdr = ft_read_header(filename);
-    end
     for i=1:hdr.orig.nsmpl
       event(end+1).type = 'trigger';
       event(end  ).value    = hdr.orig.smpl(i).type;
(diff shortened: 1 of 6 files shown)


robertoostenveld:
code-wise it is only a small change, but has potentially large impact in how people (including me) can use (or abuse) this in their analysis scripts. 

@schoffelen do you think this is in general a good idea?

schoffelen:
I think it is a good idea. In particular for the ft_read_data I am not sure what the possible user scenarios are, and whether this will be widely useful, but I don't see why this would be problematic.
Also, if it's not advertised too widely, hardly anyone will notice this functionality, right?

robertoostenveld:
There might be a small chance of it somehow conceptually interfering with `cfg.inputfile`. And of people thinking that FieldTrip is also a file format. 

The new functionality allows ft_definetrial and standard trialfuns to be used, similar to how EEGLAB datasets can be processed (which are also stored in a renamed mat ".set" file). Other than that, I don't see a reason why one would use this. Also ft_preprocessing will just work fine on the continuous data in memory.

robertoostenveld:
we might want to review the strategies outlined on https://www.fieldtriptoolbox.org/faq/how_can_i_import_my_own_dataformat/ and see how this fits in.

fileio/ft_read_event.m
```diff
@@ -1362,9 +1362,15 @@
...
-    tmp   = load(filename, 'event');
-    event = tmp.event;
+    % read the event structure from a MATLAB file
+    % it should either contain an "event" structure, or a FieldTrip data structure according to FT_DATATYPE_RAW
+    w = whos(matfile(filename));
+    if any(strcmp({w.name}, 'event'))
+      event = loadvar(filename, 'event');
+    elseif any(strcmp({w.name}, 'data')) || length(w)==1
+      data = loadvar(filename, 'data');
+      event = ft_fetch_event(data);
```
robertoostenveld, line 1372: this can be further improved for the case that the data structure does not have events yet (in cfg.previous.previous ...). In that case ft_read_event could also do the onset/offset detection on trigger channels using the low-level read_event function.

i.e. ft_read_event calls private/read_trigger, which calls ft_read_header+ft_read_data to get the trigger channel, threshold, onset-detect, and construct the event structure on the fly.

PR #1658, opened 2021-01-26
ENH - precision of ft_statistics_montecarlo according to the input

tiborauer:


schoffelen (approved):
OK, if you are fine doing it like this, I suggest to do so

PR #1663, opened 2021-02-07
integration of DUNEuro into FieldTrip

mcpiastra:
Integration of the DUNEuro-MATLAB into FieldTrip, whose main contributor is @s-schrader.

I created a test script that computes and compares two MEG leadfields using DUNEuro: one created from a hexahedral mesh of a 4-layered cube and one created from a tetrahedral mesh of the same cube. The relative error should stay below 10%.

License info is printed out when calling DUNEuro.

Tested on the Donders cluster with MATLAB 2018b (the mex file also created in the same machine).

I have some doubts about how I check the units of the computed leadfield. 

Many thanks,
Maria Carla


mcpiastra:
@vitpia: this is a good step forward for our project together :)

s-schrader:
I think some of the remaining issues involve:
- physical units: it is probably easiest to convert input units to SI units (i.e., _m_ for mesh/sensor and dipole locations, _S/m_ for conductivity, _Am_ for dipole moments) in ft before passing it to DUNEuro / computing B_primary /multiplying with magnetic constant 
- subtraction: either disable for MEG or add u_infinity contribution
- discuss how to best compile/provide mex file

s-schrader:
Should we also include a test script for an EEG leadfield? So far, @mcpiastra, you modified the MEG part of my test scripts to use fieldtrip input, right?

forward/ft_compute_leadfield.m
```diff
@@ -336,14 +336,24 @@
...
+      
       %TODO: involve unit checking
 
       % compute secondary leadfield numerically
-      lf = leadfield_duneuro(dippos, headmodel);
+      lf = leadfield_duneuro(dippos, headmodel, 'meg');
 
       % compute primary B-field analytically
-      mu = 4*pi*1e-4; %unit: Tmm/A
+      % check unit of permeability constant mu
```
robertoostenveld, line 349: I don't think that unit conversions in the low-level code are appropriate. This code is called very often, e.g. when dipole fitting. Better ensure (once) at t he higher level that the units meet the requirements of being expressed according to SI. Also: if people want to mix units and work with arbitrary units, we don't have to stop them.

forward/ft_compute_leadfield.m
```diff
@@ -356,6 +366,13 @@
         % construct the channels from a linear combination of all magnetometer coils
         lf = sens.tra * lf;
       end
+      
+      % convert lf into siunits: Tm/A
```
robertoostenveld, line 370: see previous comments about unit conversions in low-level code

forward/ft_headmodel_duneuro.m
```diff
@@ -0,0 +1,244 @@
...
+% as output a volume conduction model which can be used to compute EEG/MEG
+% leadfields.
+%
+% Use as
+%   headmodel = ft_headmodel_duneuro(mesh,'conductivity', conductivities, ...)
+%   headmodel = ft_headmodel_duneuro(mesh,'grid_filename', grid_filename, 'tensors_filename', tensors_filename, ...)
+%
+% Required input arguments should be specified in key-value pairs and have
+% to include either
+%   grid_filename   = string, filename for grid
```
robertoostenveld, line 17: it is unclear what these files are, how to create them, and which which file format these are to be specified.

forward/ft_prepare_vol_sens.m
```diff
@@ -533,6 +536,22 @@
         headmodel.elec = sens;
       end
 
+    case 'duneuro'
+      ft_hastoolbox('duneuro', 1);
+      duneuro_license; % show the license (only once)
```
robertoostenveld, line 541: this already happens in ft_hastoolbox, right?

forward/ft_prepare_vol_sens.m
```diff
@@ -302,7 +302,10 @@
       % don't do anything, h2em or h2mm generated later in ft_prepare_leadfield
 
     case 'duneuro'
-
+      
+      % show the license (only once)
+      duneuro_license;
```
robertoostenveld, line 307: this already happens in ft_hastoolbox, right?

forward/private/leadfield_duneuro.m
```diff
@@ -0,0 +1,51 @@
+function [lf] = leadfield_duneuro(pos, vol, method)
```
robertoostenveld, line 1: is this function not beter located in `external/duneuro`?

test/test_pull1663.m
```diff
@@ -0,0 +1,185 @@
...
+for i=1:size(coils,1)
+  meg_labels(i) = cellstr(strcat('meg',num2str(i)));
+end
+
+sens = [];
+sens.coilpos = coils;
+sens.coilori = projections;
+sens.chanpos = coils;
+sens.chanori = projections;
+sens.label = meg_labels;
```
robertoostenveld, line 108: please add `sens.tra=eye(5)` to make it explicit that there mapping from coils to channels is one-to-one (i.e. magnetometers).

robertoostenveld (changes requested):
It mostly looks OK, but please have a look at the comments.

forward/private/leadfield_duneuro.m
mcpiastra: I guess that leadfield_duneuro.m was put in `forward/private` because all other functions leadfield_{fns, interpolate, simbio}.m are there. Shall I move it anyways?

test/test_pull1663.m
mcpiastra: added in commit 4375db7

forward/ft_compute_leadfield.m
mcpiastra: solved in commit c9080cd. Now the permeability constant mu is in si units.

forward/ft_compute_leadfield.m
mcpiastra: deleted unit conversion at this level here: c9080cd

forward/ft_prepare_vol_sens.m
mcpiastra: yes, right, removed here 05bfc64

forward/ft_prepare_vol_sens.m
mcpiastra: yes, removed here 05bfc64 (where I deleted also in other locations, i.e., ft_headmodel_duneuro.m and ft_comute_leadfield.m )

forward/ft_headmodel_duneuro.m
mcpiastra: addressed here: 2b40b95

mcpiastra:
> I think some of the remaining issues involve:
> 
> * physical units: it is probably easiest to convert input units to SI units (i.e., _m_ for mesh/sensor and dipole locations, _S/m_ for conductivity, _Am_ for dipole moments) in ft before passing it to DUNEuro / computing B_primary /multiplying with magnetic constant

at the moment we assume the user to have used si units and the constant mu is now in si units. The next step might be to help the user to convert the input into si units, for example, in prepare_headmodel.m

> * subtraction: either disable for MEG or add u_infinity contribution

the default source model is "venant" so we could open a separate ENH pull request about this point..?

> * discuss how to best compile/provide mex file

yes, let's discuss. I think this should not stop the review of this pull request though.

mcpiastra:
> Should we also include a test script for an EEG leadfield? So far, @mcpiastra, you modified the MEG part of my test scripts to use fieldtrip input, right?

It is correct, the test script is computing MEG leadfields

forward/ft_headmodel_duneuro.m
```diff
@@ -0,0 +1,243 @@
...
+  headmodel.pos = mesh.pos;
+else
+  error('Vertex field is required!')
+end
+
+if isfield(mesh,'tet')
+  headmodel.tet = mesh.tet;
+elseif isfield(mesh,'hex')
+  headmodel.hex = mesh.hex;
+  %%assume fieldtrip ordering, reorder to conform to DUNE standard
```
robertoostenveld, line 149: should there be an explicit check on the ordering? I don't think that there is any specific "fieldtrip" ordering, or if there is, it is not explicitly defined.

forward/private/leadfield_duneuro.m
robertoostenveld: no, you can leave it in private

robertoostenveld:
It looks good to me and I suggest to merge. There are further improvements possible, but those can wait. Better work on some specific use cases (like that of @vitpia) and use that to get feedback on the most important remaining issues.  

agreed?

mcpiastra:
note: the mex file was created when compiling DUNEuro on the Donders cluster ("CentOS Linux 7 (Core)"), the test script was tested in the same machine with MATLAB 2018b.

mcpiastra:
> It looks good to me and I suggest to merge. There are further improvements possible, but those can wait. Better work on some specific use cases (like that of @vitpia) and use that to get feedback on the most important remaining issues.
> 
> agreed?

I agree! thanks a lot!

PR #1664, opened 2021-02-08
ENH - created test function to check the functionality of the spatial…

schoffelen:
… whitening.

This is inspired by some recent e-mails back-and-forth concerning the technote.


schoffelen:
Pinging @britta-wstnr and @vlitvak and @robertoostenveld 

Hi Vladimir! My e-mail that I sent last week was inspired by an 'if memory serves me well'. Just to be sure, I whipped up a test function (which is the new function in this PR) that runs an lcmv analysis on subject 15 of the Wakeman&Henson data. (starting off from a tutorial that was created for the meguk SPM/FT workshop in 2015) It seems that - for me at least - the spatial whitening does not prohibit ft_prepare_leadfield / ft_sourceanalysis to run through fine. I can confirm that the data.chanunit after whitening is 'snr'. However, it seems that the data.grad.chanunit still contains the physical units of the signal.

schoffelen:
@vlitvak thanks for picking this up. For convenience, I paste here the e-mail you wrote to the technote-team.

++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
There is no bug in FT, it's just the way those functions are called in SPM that was problematic (for instance it was with 'amplitude', 'T' option for ft_datatype_sens). I now made some fixes and could run the pipeline with whitening on Caroline's visual gamma data and it looks fine. 

I'm now trying to make the example pipeline in DAiSS consistent with Jan-Mathijs's example script https://github.com/fieldtrip/fieldtrip/pull/1664/commits/737e50eb0e34db31775bf591502c9579f0c03dce

So it looks that the script:

1) Separates mags and grads before prewhitening which effectively zeroes the cross terms.
2) Prewhitens with the default options which are 'tichonov' with lambda of 0 and empty kappa 71.
3) Then merges the data and computes covariance, which I understand creates non-zero cross-terms.
4) For beamforming uses the same kappa.

I tried that and also zeroing the cross terms. If you don't zero there are two cliffs in the eigenspectrum. If you do, there is one and it all looks cleaner. As the two cliffs are after the cutoff it doesn't make a difference in practice, but just wanted to make sure that we are all doing the same thing also consistent with MNE. If that's the case, I'll update the text tomorrow.
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

The test function is not intended to be the gold standard. In particular: I did the manual separation of mags and grads prior to the whitening to ensure the correct kappa-based inversion of the baseline 'noise'. In principle ft_denoise_prewhiten supports a cfg.split argument which could achieve the sensor-type specific whitening. I assume that in this case, with the same kappa for both the mags and grads, this would have worked well.

w.r.t. your point 3): indeed, in the covariance the cross terms are not zeroed out. To be honest, I never realized this explicitly myself, I always thought of the zero'ing out of cross-terms in terms of the prewhitening itself, not in terms of the beamforming. Do you happen to know how MNE-python 'thinks' of this? Or else @britta-wstnr ?

w.r.t. your point 4): indeed: if the prewhitening imposes (either implicitly or explicitly) a lower rank on the data, it would make sense to use the truncated svd inversion heuristic for the beamformer for optimally predictable behavior.

vlitvak:
Alex mentioned in the comments that they only zero cross-terms between EEG
and MEG when beamforming.

Your last sentence suggests that the cross-terms should be zeroed in both
cases since only in then the rank of the covariance matrix in beamforming
will be the same as set by pre-whitening.

Vladimir

britta-wstnr:
Regarding MNE-Python: I double-checked with the code and had a quick discussion with @larsoner as well.
@agramfort was right, only the cross terms between EEG and MEG get zero'ed out, but not between different MEG channel types. This happens during the whitening procedure. In MNE-Python, we always invert only the none-zero singular values though.

schoffelen:
There's no apparent reason to keep this unmerged it seems.

PR #1665, opened 2021-02-09
ENH - CF phase-amplitude coupling

tiborauer:
This PR allows  ft_crossfrequencyanalysis.m to output the PAC estimates.


schoffelen:
Hi Tibor, Thanks for this. I assume that it is all well tested and validated? Would it make sense to add a little bit of detail w.r.t. the help documentation to explain the difference between mi and pac? Or should the user just read the paper?

tiborauer:
I only provide access to data at an earlier stage, i.e. the output of data2pac. Therefore, I have not felt the need for further explanation or validation, because I do not introduce any new measure.

schoffelen:
OK, thanks.

PR #1666, opened 2021-02-10
FIX - update read_eeglabheader according to EEGLAB 2021.0 defaut

tiborauer:
EEGLAB 2021.0 saves the content of the header as default

### DIFF
--- a/fileio/private/read_eeglabheader.m
+++ b/fileio/private/read_eeglabheader.m
@@ -38,6 +38,9 @@ end
 
 if ~isstruct(filename)
   load('-mat', filename, 'EEG');
+  if strcmp(lastwarn,'Variable ''EEG'' not found.') % EEGLAB > 2021.0 saves content of EEG as default
+      EEG = load('-mat', filename);
+  end
 else
   EEG = filename;
 end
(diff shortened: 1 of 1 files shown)


fileio/private/read_eeglabheader.m
```diff
@@ -38,6 +38,9 @@
 
 if ~isstruct(filename)
   load('-mat', filename, 'EEG');
+  if strcmp(lastwarn,'Variable ''EEG'' not found.') % EEGLAB > 2021.0 saves content of EEG as default
```
schoffelen, line 41: Perhaps it would be better here not to rely on the warning, e.g. it could be that the user has non-standard settings for the warning behavior, in which case this might fail. Would it also work if you in advance check the contents of the file, using the output of a call to which.m, and then proceed depending on whether the variable 'EEG' exists or not?

schoffelen (changes requested):

fileio/private/read_eeglabheader.m
tiborauer: Do you have any specific solution in your mind?

tiborauer:
Thanks for the more robust solution!

schoffelen:
no problem. I haven't been able to test it, so please keep your eyes open...

PR #1670, opened 2021-02-18
ENH - implemented ft_badchannel and ft_badsegment, modelled after the HCP megconnectome pipelines

robertoostenveld:
This is a first implementation, containing two straightforward algorithms and two from the [HCP pipelines](https://github.com/Washington-University/megconnectome) and specifically 

- https://github.com/Washington-University/megconnectome/blob/master/analysis_functions/hcp_qc_neighcorrel.m
- https://github.com/Washington-University/megconnectome/blob/master/analysis_functions/hcp_qc_neighstdratio.m

What still needs to be done is to figure out how to ensure that one bad channel not tainting all its (good) neighbors) and how to decide on the multitude of neighbours.


ft_badchannel.m
```diff
@@ -0,0 +1,204 @@
...
+    ft_info('min(range)    = %g\n', min(tmp));
+    ft_info('max(range)    = %g\n', max(tmp));
+    ft_info('mean(range)   = %g\n', mean(tmp));
+    ft_info('median(range) = %g\n', median(tmp));
+    ft_info('mad(range)    = %g\n', mad(tmp));
+    figure; histogram(tmp, round(length(tmp)/10));
+    figure; imagesc(channeighcorrel);
+    
+    % find channels with a median neighbour value BELOW the threshold
+    bad = nanmedian(channeighcorrel,2)<cfg.threshold;
```
robertoostenveld, line 189: each channel has a number of neighbours with which the correlation can be low. 

Imagine a single bad channel surrounded by good chanels: all good channels would have a low correlation with the bad channel. To prevent the good ones to be marked as bad, the good ones need more neighbors with which they would have a high correlation. So the good ones correlate only poorly with a (single) bad one, and correlate well with other neighbours.

The bad one correlates poorly with _all_ its neighbors.

ft_badchannel.m
```diff
@@ -0,0 +1,204 @@
...
+    ft_info('min(range)    = %g\n', min(tmp));
+    ft_info('max(range)    = %g\n', max(tmp));
+    ft_info('mean(range)   = %g\n', mean(tmp));
+    ft_info('median(range) = %g\n', median(tmp));
+    ft_info('mad(range)    = %g\n', mad(tmp));
+    figure; histogram(tmp, round(length(tmp)/10));
+    figure; imagesc(chanstdratio);
+    
+    % find channels with a median neighbour ABOVE the threshold
+    bad = nanmedian(chanstdratio,2)>cfg.threshold;
```
robertoostenveld, line 161: the same rationale as for `neighcorrel` also applies here, see below.

robertoostenveld:
I have kept any potential updates to ft_artifact_threshold out of this PR. However, that function might also need some updating to make it more consistent with ft_rejectvisual. 

@schoffelen could you have a look at my suggested contribution?

ft_badchannel.m
```diff
@@ -0,0 +1,239 @@
...
+
+% ensure that the preproc specific options are located in the cfg.preproc substructure
+cfg = ft_checkconfig(cfg, 'createsubcfg',  {'preproc'});
+
+% set the defaults
+cfg.nbdetect      = ft_getopt(cfg, 'nbdetect', 'all');
+cfg.feedback      = ft_getopt(cfg, 'feedback', 'no');
+cfg.thresholdside = ft_getopt(cfg, 'thresholdside', []); % the default depends on cfg.method, see below
+
+if isempty(cfg.thresholdside)
```
schoffelen, line 105: You probably thought about this, but are there use cases where cfg.thresholdside's value is meaningfully non-default, as a function of cfg.method? In other words, does this argument need to be configureable by the user?

test/test_ft_badchannel.m
```diff
@@ -0,0 +1,95 @@
...
+% each method has its own optimal threshold
+% FT_REJECTVISUAL can be used to determine the appropriate value
+
+
+
+cfg = [];
+cfg.feedback = 'yes';
+cfg.neighbours = neighbours;
+
+% cfg.method = 'range';
```
schoffelen, line 48: in principle, this can all be commented in, right (+ followed by a call to ft_badchannel), so that it just cycles through those possibilities. On the other hand perhaps there is not much added value in doing so.

schoffelen (approved):
Looks good to me. Two small comments for consideration.

ft_badchannel.m
robertoostenveld: you are right. I initially made it configurable to postpone thinking about it, but later I indeed identified that the default is obvious. There is only one obscure case left where it is ambiguous. I will remove it.

test/test_ft_badchannel.m
robertoostenveld: the test script is indeed not optimal yet, because it does not go through all options. It is more for manual testing at the moment. This will be updated later!

robertoostenveld:
@helenacockx this is the one I mentioned

robertoostenveld:
This got a bit bigger than expected, but I think this is a good improvement of the existing code base and a few good improvements to the functionality.

PR #1675, opened 2021-02-22
pass through non-MEG channels in ft_megplanar

### LINKED ISSUE #1671: ft_megplanar should transparently pass through non-MEG channels
If we have an axial gradient MEG data structure with some added (e.g. reference or EMG) channel, `ft_megplanar` now throws away those channels in the output (i.e. it *only* returns the computed planar gradient channels). We could consider passing through any extra non-MEG channels unchanged.

If you agree, I could file a PR for this.

Spaak:
This fixes #1671 by simply changing the default `cfg.channel` to `'all'`. (The issue was not in `ft_apply_montage` but in `ft_megplanar` itself.)

As requested by @robertoostenveld, I also had a look at `ft_megrealign`. While I didn't run any explicit tests for `ft_megrealign`, the code handles channel selection differently, explicitly taking care of non-MEG channels:
https://github.com/fieldtrip/fieldtrip/blob/d8e913aa20a7478946d90af66e41a5a8d8afb76d/ft_megrealign.m#L163-L168

so I don't expect that the same issue affects the code there.

### DIFF
--- a/ft_megplanar.m
+++ b/ft_megplanar.m
@@ -116,7 +116,7 @@ cfg = ft_checkconfig(cfg, 'renamed', {'grid',    'sourcemodel'});
 cfg = ft_checkconfig(cfg, 'renamed', {'pruneratio', 'tolerance'});
 
 % set the default configuration
-cfg.channel      = ft_getopt(cfg, 'channel',      'MEG');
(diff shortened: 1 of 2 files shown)


robertoostenveld:
this is indeed a better default.

PR #1677, opened 2021-02-23
Adding eeglab keyword

arnodelorme:


robertoostenveld:
the suggested change is fine, but I think there are other places where we want to make this consistent. For example in the help of the functions, e.g. here

https://github.com/fieldtrip/fieldtrip/blob/738b88479a2855786bd0b9eba52d23d943376d3c/ft_meshrealign.m#L23

and in the corresponding code.

PR #1678, opened 2021-02-23
Allow for missing *.eeg file in neuroscope data

wuffi:
ft_read_header would handle missing *.eeg files in neuroscope data, by checking whether the variable lfpfile is empty. However, ft_read_header currently throws an error at this point, because the lfpfile variable would not be generated, if there is no *.eeg file. This commit fixes the problem by generating an empty lfpfile array in case there is no *.eeg file.


robertoostenveld:
👍🏻

PR #1680, opened 2021-02-24
FIX - removed incorrect file, created neighbour template for combined planar NM122

schoffelen:
OK, I removed the incorrect file from the repo, and created a template neighbours structure for the combined gradiometers. I don't think that this is trivial for the odd user, so I suggest to include this in the repo.


schoffelen:
The template was created using the label list from the combined 122 channel layout file, and the positions of the sensors as per sens.chanpos(1:2:end,:). Subsequently, I used ft_prepare_neighbours with method 'triangulation', and with cfg.compress = 'yes'. A quick inspection in ft_neighbourplot looked good to me.

robertoostenveld:
👍

PR #1685, opened 2021-02-28
HOTFIX - fix for issue uncovered with PR1675

schoffelen:


schoffelen:
Eelke's recent PR #1675 uncovered 2 issues.
The PR changed the default cfg.channel from 'MEG' to 'all'. 
As a consequence:
- megplanar_orig started failing due to improper handling of the reference coils. Upon investigation, the megplanar_orig was slightly inconsistent with megplanar_sincos which works correctly. This PR makes the 2 functions consistent.
-  cfg.method = 'sourceproject' started failing for grad arrays with  reference sensors (so the first one encountered was bti248), due to the fact that the lower-level code did not produce valid leadfields. Ensuring tha the leadfields are only produced for the sensors close to the head solves this issue.

I haven't inspected the output of the function, i.e. whether the data contains the non-planar gradient transformed reference channels if these were present in the input. This would be the natural consequence and expected behavior with the changed default cfg.channel = 'all'.

Spaak:
Thanks JM! Happy Sunday ;)

PR #1686, opened 2021-02-28
Cluster-based statistics for 2D and 4D data

tiborauer:


schoffelen:
Hi Tibor, I assume that this PR is related to your attempts to do statistical inference on CFC data. In order for us to really judge your changes, you'd need to provide a test function that implements the intended use-case, and which demonstrates that - without the suggested changes to the code - your use case doesn't work. Also, the test function should demonstrate that backward compatibility is ensured, i.e. that the more 'traditional' use cases are not broken.

Specific concerns at the moment (which I cannot judge due to missing code+data): what is special about CFC data? You mention in the title of this PR 2D data, and 4D data, but isn't a channel versus time matrix (i.e. an EEG ERP matrix) also a 2D matrix? What is meant with a 4D matrix? Do you mean space x space x freq(low) x freq(high) or space x freq(low) x freq(high) x time, where space can be either channel-level, or source-level? The latter might be reshapeable to 3D volumetric, but could also be defined on a cortical surface, which would require - in analogy to channel level data - a spatial neighbourhood definition. 
If more than one of your intended dimensions in the 4D data is spatial in nature, then your use case is not really specific to CFC, but more akin to channel/source level connectivity data in general, for which we have so far not been able to solve the spatial neighbourhood issue (although attempts have been made already a long time ago).

robertoostenveld:
To follow up on what @schoffelen wrote, I also wonder whether this is for source-level all-to-all connectivity, or channel level? Going up in dimensions, I would enumerate the following, where chan=2d, sheet=2d, grid=3d

3D
- grid
- sheet-time or sheet-freq (i.e. on a cortical sheet, where neighbors are still irregular, like on the channel level) 

4D
- chan-time-freq
- grid-time or grid-freq (i.e. on a regular 3D grid)
- sheet-sheet (e.g. all-to-all connectivity on the cortical sheet)
- sheet-time-freq

5D  
- chan-chan-time-freq
- sheet-sheet-time or sheet-sheet-freq

6D 
- grid-grid (dense all-to-all connectome at a single frequency and/or time)
- sheet-sheet-time-freq

7D
- grid-grid-time or grid-grid-freq

8D
- grid-grid-time-freq 🤯

Not sure whether this is complete, but you will get the  idea.

The ones with regular neighbours should (I think) be easy to implement with MATLAB bwlabeln (not with the SPM version though), and the irregular ones (with chan and sheet) require an extra layer of FT code to cluster over the explicit neighbours.

schoffelen:
btw, more recent versions of MATLAB have bwconncomp as a replacement for bwlabeln which is supposed to be more memory efficient...

tiborauer:
The first part of the PR is to fix the issue (?) with 2D data. The current code parametrizes `spm_bwlabel` differently for 2D (n=4) and 3D (n = 6) https://github.com/fieldtrip/fieldtrip/blob/7b4b47593120d6373c490de38ef91105dcb059b1/private/clusterstat.m#L191 while `spm_bwlabel` accepts only n=6, 18, and 26 https://github.com/fieldtrip/fieldtrip/blob/7b4b47593120d6373c490de38ef91105dcb059b1/external/spm12/spm_bwlabel.m#L6 and throws an error for n=4.

tiborauer:
The second part of the PR was, indeed, motivated by my CF use case, which - when using `method=pac`, it generates 4D or 5D data 
https://github.com/fieldtrip/fieldtrip/blob/198bc9a47589b5ccab75d45de0c9d4d8c8be65fa/ft_crossfrequencyanalysis.m#L294-L303.

tiborauer:
I do not consider sensor vs source differently. When I create virtual channels I store their position (e.g. the COG of the brain areas), so that I can define the spatial neighbourhood See https://github.com/tiborauer/automaticanalysis/blob/a05ff67d9c60230b76be9244fae1ad515967fd1e/aa_modules/aamod_meeg_sourceatlasing.m, L63-L69

private/clusterstat.m
```diff
@@ -326,7 +326,7 @@
       
       numdims = length(cfg.dim);
       if numdims == 2 || numdims == 3 % if 2D or 3D data
-        [negclusrnd, negrndnum] = spm_bwlabel(tmp, 2*numdims); % use spm_bwlabel for 2D/3D to avoid usage of image toolbox
+        [negclusrnd, negrndnum] = spm_bwlabel(tmp, 6); % use spm_bwlabel for 2D/3D to avoid usage of image toolbox
```
schoffelen, line 329: Agreed. Changes are clear and make sense. Apparently - before - nobody ended up at this point with numdims being 2, and thus never got an error. Indeed it is clear that spm_bwlabel only supports 6 (and not 4) as an input value

private/findcluster.m
```diff
@@ -82,26 +83,33 @@
...
+labelmat = reshape(labelmat, spatdimlength, prod(ndatadim));
 
 % combine clusters that are connected in neighbouring channel(s)
 % (combinations). Convert inputs to uint32 as that is required by the mex
 % file (and the values will be positive integers anyway).
 cluster = combineClusters(uint32(labelmat), logical(spatdimneighbstructmat), uint32(total));
 
 % reshape the output to the original format of the data
-cluster = reshape(cluster, spatdimlength, nfreq, ntime);
+cluster = reshape(cluster, spatdimlength, celldims{:});
```
schoffelen, line 112: OK, the changes to this function look clean enough. We look forward to some code added to an existing test function, or a new test function which highlights your use case.

schoffelen (changes requested):
some test code is still needed before this can be merged

tiborauer:
I have added a test function with a little nice touch ;)

schoffelen:
Thanks.

schoffelen:
I have locally pulled this PR, and ran the test functions that I identified with a grep clusterstat *.m
On the way, I noticed a failed_ft_statistics_montecarlo function, which I will investigate a bit further (see whether it can be promoted back into  a test function)

tiborauer:
I also take another look.

schoffelen:
Sorry, Tibor, I have not been clear. This failing test function is not the consequence of your PR: there's a bunch of functions in the test directory whose names start with 'failed_...', rather than with 'test_'. Those functions were operational at some point in time, but somehow started failing. Thus, I just flagged this as a placeholder, and will look into it.

schoffelen:
This PR caused test_ft_clusterplot to fail (and also another test function, which I have not investigated yet).

The cause is that ft_clusterplot now started choking on data, complaining that there was nothing 'significant' to plot. With ft_statistics_montecarlo (and clusterstat etc.) running through fine, this was surprisining initially, but further inspection showed that the latest PR caused clusters to be bogus, in case the cfg.dim contains only 2 elements (i.e. the input data is space versus time (or freq)). The cause is the intended more generic data dimensionality handling in findcluster.

When doing something like:

siz = size(onoff);
siz = siz(2:end); % remove the spatial dim

siz now can be a scalar if the input of onoff was originally a 2D matrix.

This caused problems down the line, since downstream the siz variable (i.e. the ndatadim created) is 
1) explicitly checked for numel(ndatadim)>1
2) used for reshaping of a matrix

regarding 1); spm_bwlabel should also be called if there is just a single additional dimension beyond the spatial, only if the additional dimension is singleton, this step can be skipped, thus if numel(ndatadim) should be replaced by something else. Solution: change this into if ~(numel(ndatadim)==1 && ndatadim(1)==1)

regarding 2): the variable celldim is created based on ndatadim, and is incomplete (because it contains just a single element in case of 2D (space x something else) input data. Solution: add a singleton to ndatadim -> this causes the reshape to be always correct (trailing singletons don't matter.

I will make these changes in a new PR.

PR #1687, opened 2021-03-02
ENH - pass parcel labels from brainordinate field to lower level

schoffelen:


schoffelen:
This PR results in the labels of the parcels (if specified in the input's brainordinate field) to be passed on to the figures containing the location-specific time courses

PR #1688, opened 2021-03-02
Data2bids for motion data. 

sjeung:
Changes : 

- added metadata entries for motion in data2bids.m 
- xdf2bids can select streams by range of effective sampling rates or by keywords in stream name


data2bids.m
```diff
@@ -1323,10 +1334,11 @@
...
 if need_motion_json
-  motion_json.SamplingFrequency = hdr.Fs;
-  motion_json.StartTime = nan;
-  motion_json.Columns = hdr.label;
-  motion_json.RecordingDuration = (hdr.nSamples*hdr.nTrials)/hdr.Fs;
+    
+  motion_json.SamplingFrequency     = hdr.Fs;
+  motion_json.StartTime             = nan;
+  motion_json.MotionChannelCount    = hdr.nChans; 
+  motion_json.RecordingDuration     = hdr.nSamples/hdr.Fs;
```
robertoostenveld, line 1341: it should in general be `hdr.nSamples*hdr.nTrials` rather than `hdr.nSamples` to deal with data that is represented in segments. That might not happen for common motion formats, but is e.g. the case for EDF (which stores data in 1-second chunks) or with epoch FIF files.

data2bids.m
```diff
@@ -545,13 +545,21 @@
...
 
 %% motion is not part of the official BIDS specification
-% this follows https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continuous-recordings.html
-cfg.motion.Columns                        = ft_getopt(cfg.motion, 'Columns'              );
-cfg.motion.StartTime                      = ft_getopt(cfg.motion, 'StartTime'            );
-cfg.motion.SamplingFrequency              = ft_getopt(cfg.motion, 'SamplingFrequency'    );
-cfg.motion.RecordingDuration              = ft_getopt(cfg.motion, 'RecordingDuration'    );
-
-%% information for the coordsystem.json file for MEG, EEG and iEEG
+% this follows (Link to BEP)
```
robertoostenveld, line 548: can you add the link https://bids.neuroimaging.io/bep029 here?

xdf2fieldtrip.m
```diff
@@ -42,18 +44,50 @@
 % $Id$
 
 % process the options
-streamindx = ft_getopt(varargin, 'streamindx');
+streamindx      = ft_getopt(varargin, 'streamindx');
```
robertoostenveld, line 47: it seems to me that these ways of indexing the XDF file are mutually exclusive, right? That is fine with me, but you may want to add a check  to ensure that the user knows that as well. Like

````
% not more than one cal be selected
assert(sum(cellfun(@isempty, {streamindx, streamkeywords, sraterange}))<=1);
````

robertoostenveld (changes requested):

robertoostenveld:
could you also add a test script similar to `test_issue789` that demonstrates the new functionality in the XDF reading and send me (not via the repository) a test file so that I can see how it works? See https://www.fieldtriptoolbox.org/faq/how_should_i_send_example_data_to_the_developers/

robertoostenveld:
Since it still seems work in progress, I have converted this PR into a draft. @sjeung  once you think it is ready for final review and merging, please convert it back again from draft into a normal PR.

sjeung:
> Since it still seems work in progress, I have converted this PR into a draft. @sjeung once you think it is ready for final review and merging, please convert it back again from draft into a normal PR.

Hi Robert, thanks for the comments. I will do once the test script & data are ready.

xdf2fieldtrip.m
sjeung: Fixed - I can imagine people using two or more methods in conjunction but for now it is restricted to be exclusive for simplicity.

sjeung:
Hi Robert,

could you review the pull request 1688 with the added test script?
I was not sure if I am supposed to add data viewer in the test script, but
it is there at the moment.

Here is the file that is used in the test script.
https://drive.google.com/file/d/1PZmbKUNZr5Arq5kM1BQYNHiCg-z0skLL/view?usp=sharing

Please let me know if there is any issue in the process of PR submission.

Many thanks,
Sein

Am Di., 2. März 2021 um 13:43 Uhr schrieb Robert Oostenveld <
***@***.***>:

robertoostenveld (approved):
Sorry for this taking some time. I reviewed your suggestions, they look fine to me. Also the text script works as expected.

robertoostenveld:
is it an idea to add an XDF example to this page? https://www.fieldtriptoolbox.org/example/bids_motion/

JuliusWelzel:
We are about to finalize the BEP029 for BIDS and it might make sense to add an XDF example after this

schoffelen:
Nice! Please feel free to add an example.

Issue #1689, opened 2021-03-05
Include mex.h in geometry.c


gllmflndn:
`geometry.c` calls `mxGetNaN()` so it should include `mex.h` (it used to include `matrix.h`, which was removed for Octave compatibility)
https://github.com/fieldtrip/fieldtrip/blob/e2abaa05085b35f80450ae146ef63634136d86b6/src/geometry.c#L19-L20

PR #1690, opened 2021-03-05
FIX - include mex.h

gllmflndn:
Prevent warning: implicit declaration of function ‘mxGetNaN’. See #1689


robertoostenveld:
thanks!

PR #1694, opened 2021-03-08
ENH - within- & cross-channel CFC

tiborauer:
This PR allows specifying within- & cross-channel CFC in the same execution.

### DIFF
--- a/ft_crossfrequencyanalysis.m
+++ b/ft_crossfrequencyanalysis.m
@@ -112,9 +112,6 @@ if isfield(cfg, 'chanlow') && isfield(cfg, 'chanhigh')
   cfg.chanlow   = ft_channelselection(cfg.chanlow, freqlow.label);
   cfg.chanhigh  = ft_channelselection(cfg.chanhigh, freqhigh.label);
   labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label));
-  % - remove duplicates
-  labelcmb = cellfun(@(cmb) strsplit(cmb,'|'),unique(arrayfun(@(x) strjoin(labelcmb(x,:),'|'), 1:size(labelcmb,1), 'UniformOutput', false)),'UniformOutput',false);
-  labelcmb = vertcat(labelcmb{:});
 elseif ~isfield(cfg, 'chanlow') && ~isfield(cfg, 'chanhigh')  % within-channel analysis (default)
   docrosschan = false;
   % ensure that we are working on the intersection of the channels
--- a/ft_crossfrequencyanalysis.m
+++ b/ft_crossfrequencyanalysis.m
@@ -112,7 +112,9 @@ if isfield(cfg, 'chanlow') && isfield(cfg, 'chanhigh')
   cfg.chanlow   = ft_channelselection(cfg.chanlow, freqlow.label);
   cfg.chanhigh  = ft_channelselection(cfg.chanhigh, freqhigh.label);
   labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label));
-  labelcmb(arrayfun(@(x) strcmp(labelcmb{x,:}), 1:size(labelcmb,1)),:) = []; % remove auto
+  % - remove duplicates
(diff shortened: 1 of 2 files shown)


ft_crossfrequencyanalysis.m
```diff
@@ -112,7 +112,9 @@
   cfg.chanlow   = ft_channelselection(cfg.chanlow, freqlow.label);
   cfg.chanhigh  = ft_channelselection(cfg.chanhigh, freqhigh.label);
   labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label));
-  labelcmb(arrayfun(@(x) strcmp(labelcmb{x,:}), 1:size(labelcmb,1)),:) = []; % remove auto
+  % - remove duplicates
+  labelcmb = cellfun(@(cmb) strsplit(cmb,'|'),unique(arrayfun(@(x) strjoin(labelcmb(x,:),'|'), 1:size(labelcmb,1), 'UniformOutput', false)),'UniformOutput',false);
```
schoffelen, line 116: Are there situations in which ft_channelcombination returns duplicates? I would expect that ft_channelcombination warrants that this is not the case. If - for some reason - ft_channelcombination is not robust enough to ensure this, in case it is called as in line 114, I think that the issue of duplicates should be solved within ft_channelcombination.

schoffelen (changes requested):

schoffelen:
The use case of calling ft_channelcombination with two cell-arrays of labels is handled in lines 76-92 in ft_channelcombination. This part of the code indeed is not smart enough to exclude double occurrences.

schoffelen:
Note, also, that you probably want behavior of ft_channelcombination according to as if dirflag==2, i.e. a combination {'a' 'b'} and {'b' 'a'} should both be included, which by default should not be the case (but which happen to occur in this particular use case).

tiborauer:
You can try the following code: `ft_channelcombination({{'a' 'b'} {'a' 'b'}},{'a' 'b'})`, which is my use case for specifying both wthin and across-channe combinations (all-to-all). The `dirflag=0` (default) means that the order is preserved, while the `includeauto=0` (default) means that no (extra) autocombination should be added. Yet, this is the output:

```
ans =

  8×2 cell array

    {'a'}    {'a'}
    {'a'}    {'b'}
    {'b'}    {'a'}
    {'b'}    {'b'}
    {'a'}    {'a'}
    {'b'}    {'b'}
    {'a'}    {'a'}
    {'b'}    {'b'}
```

schoffelen:
I know. I tried it myself also.

schoffelen:
The general behavior (i.e. outside your use case) of ft_channelcombination is to translate a shorthand recipe for requested combinations, and to return the lower triangular non auto combinations.

Specifically, when you do something like ft_channelcombination({'all' 'all'}, {'a';'b'}), you get:

{'b' 'a'}

I would assume - if I were a naive user - that calling ft_channelcombination({ {'a';'b'}, {'a';'b'} }, {'a';'b'}) would return the same

And that if you want ft_channelcombination({ {'a';'b'}, {'a';'b'} }, {'a';'b'}) to return {'a' 'a';'b' 'a';'a' 'b';'b' 'b'}, you'd call it like this:

ft_channelcombination({ {'a';'b'}, {'a';'b'} }, {'a';'b'},1,2)

tiborauer:
Thank you for the clarification. It makes sense. 
I will amend this PR to change the ft_channelcombination instead so that it behaves the way you specified.

schoffelen:
Top. Thanks!

tiborauer:
Done!

schoffelen:
Thanks. Ideally, you would have also updated the test_ft_channelcombination function, but I'll let that pass for now.

tiborauer:
Thanks.

schoffelen:
This is still not correct. If I now do

ft_channelcombination({{'a';'b'},{'c';'d'}},{'a';'b';'c';'d'}, opt1, opt2); I would expect that you'd want in the output 'ac','ad','bc', and 'bd'. This now only happens when opt1 and opt2 are both 0.

on the other hand, if the input is ft_channelcombination({{'a';'b'},{'b' 'c';'d'}},{'a';'b';'c';'d'}, opt1, opt2); I would expect that you'd want in the output 'ab','ac','ad','bb','bc','bd'. with opt1 and opt2 both zero, 'bb' is not present, with opt1 1 and opt2 0 'aa','cc',and 'dd' are present as well.

schoffelen:
In other words, the user's expectations will be violated most of the time, and will be only sometimes correct, depending on the specific lists in the 2 cells, and the values for opt1 and opt2. For sure, I think that opt2 should never be 2

Issue #1696, opened 2021-03-09
ft_channelcombination behavior is unexpected


schoffelen:
**Describe the bug**
PR #1664 provided an update to ft_channelcombination, with the intention to make its behavior according to expectations, with respect to an input to the function consisting of {{cell1} {cell2}}, initiated by @tiborauer 
Now, it has become more explicit that the function's behavior might not work as expected

**To Reproduce**
the expected output of  ft_channelcombination({{'a';'b'},{'c';'d'}}, {'a';'b';'c';'d'}, opt1, opt2) should be thought through, for all combinations of opt1 and opt2, also if {'a';'b'} and {'c';'d'} are partially overlapping. Right now it's messy.

**Expected behavior**
an output that is consistent with what's documented in hte function's help.

**Environment and versions (please complete the following information):**
 - Operating System: [e.g. Windows 10]: linux
 - MATLAB version [e.g. R2019b, type `ver`]: matlab2018b 
 - FieldTrip version [e.g. 2020312, or "most recent master branch from github", type `ft_version`]: master branch

**Additional context**
use cases + expectations should be explicitly written in test_ft_channelcombination, and investigated thoroughly

schoffelen:
================================
grep ft_channelcombination *.m
ft_connectivityanalysis.m:    tmpcmb = ft_channelcombination(cfg.channelcmb, data.label);
ft_connectivityanalysis.m:    cfg.channelcmb = ft_channelcombination(cfg.channelcmb(:, 1:2), tmpchan, 1);
ft_connectivityanalysis.m:    cfg.channelcmb = ft_channelcombination(cfg.channelcmb, data.label, 1);
ft_crossfrequencyanalysis.m:  labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label),1,2);
ft_freqanalysis.m:  cfg.channelcmb = ft_channelcombination(cfg.channelcmb, data.label);
ft_mvaranalysis.m:  cfg.channelcmb = ft_channelcombination(cfg.channelcmb, data.label);

+++++++++++++++++++++++++++++++++
grep ft_channelcombination */*.m

connectivity/ft_connectivity_csd2transfer.m:  channelcmb = ft_channelcombination(channelcmb(:,1:2), freq.label);
private/fourier2crsspctrm.m:chncmb  = ft_channelcombination(cfg.channelcmb, freq.label);
[...]

PR #1697, opened 2021-03-09
investigate and ensure correct behavior of ft_channelcombination

schoffelen:
This is related to an issue that was sparked by @tiborauer

### DIFF
--- a/ft_crossfrequencyanalysis.m
+++ b/ft_crossfrequencyanalysis.m
@@ -111,7 +111,8 @@ if isfield(cfg, 'chanlow') && isfield(cfg, 'chanhigh')
   docrosschan   = true;
   cfg.chanlow   = ft_channelselection(cfg.chanlow, freqlow.label);
   cfg.chanhigh  = ft_channelselection(cfg.chanhigh, freqhigh.label);
-  labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label),1,2);
+  labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label),1,0);
+  labelcmb(ismember(labelcmb(:,1),cfg.chanhigh)&strcmp(labelcmb(:,1),labelcmb(:,2)),:) = [];
 elseif ~isfield(cfg, 'chanlow') && ~isfield(cfg, 'chanhigh')  % within-channel analysis (default)
   docrosschan = false;
   % ensure that we are working on the intersection of the channels
@@ -131,7 +132,7 @@ cfg.keeptrials = ft_getopt(cfg, 'keeptrials');
 
 % make selection of frequencies and channels
 tmpcfg = [];
-tmpcfg.channel   = cfg.chanlow;
+tmpcfg.channel   = unique(labelcmb(:,1));
 tmpcfg.frequency = cfg.freqlow;
 freqlow = ft_selectdata(tmpcfg, freqlow);
 [tmpcfg, freqlow] = rollback_provenance(cfg, freqlow);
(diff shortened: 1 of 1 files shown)


schoffelen:
The function test_ft_channelcombination now enumerates all possible combinations of calling ft_channelcombination with two cell-arrays in the input + the different values for includeauto and dirflag. The test script should be completed with building in all expectations correclty, and then it can be used to address the failing assertions. The ones which I for now have identified to be failing for sure are commented out, and appended with a % this one fails. Other assertions are placeholders (and copied over from somewhere else, and should be adjusted and checked.

Also, once this is in place we should address the functions that I grepped in the corresponding issue #1696 to ensure that there are no unwanted side effects of strictly imposing consistent behavior, because it could be that old code is relying on old yet inconsistent behavior.

One point to add is that it could be that different applications (use cases) have different expectations in mind with respect to the includeauto flag, in particular when the input cell-arrays are (partly) disjoint: in the case of crossfrequencycoupling (new use case) I can imagine that one only wants those occurences of channels which are present in both the left and right lists. Yet, on the other hand, in the context of connectivity, one could expect all occurences of all unique channels.

schoffelen:
with a fix to ft_channelcombination all seems to be in order now. Next step is to investigate the calling behavior in the function calls to ft_channelcombination to check whether unwanted side effects might occur.

ft_crossfrequencyanalysis.m
```diff
@@ -111,7 +111,8 @@
   docrosschan   = true;
   cfg.chanlow   = ft_channelselection(cfg.chanlow, freqlow.label);
   cfg.chanhigh  = ft_channelselection(cfg.chanhigh, freqhigh.label);
-  labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label),1,2);
+  labelcmb = ft_channelcombination({cfg.chanlow,cfg.chanhigh},union(freqlow.label, freqhigh.label),1,0);
+  labelcmb(ismember(labelcmb(:,1),cfg.chanhigh)&strcmp(labelcmb(:,1),labelcmb(:,2)),:) = [];
```
tiborauer, line 115: Hi @schoffelen, is there a reason why you changed the code here? It is not what I intended. Can you change it (line 114-115) back?
In case when `isequal(cfg.chanhigh,cfg.chanlow)`, labelcmb should list all combinations with all directions and including auto.

schoffelen:
But what is the intended behavior if cfg.chanhigh ~= cfg.chanlow?

ft_crossfrequencyanalysis.m
schoffelen: But what is the intended behavior if cfg.chanhigh ~= cfg.chanlow?

tiborauer: If you have a specific hypothesis about connectivity and do not what to test the whole space but rather a subset of the locations only.

tiborauer: E.g.: fronto-parieta PAC only

tiborauer: In general, the intended behaviour is that `labalcmb` contains pairs of all items of chanlow to all items of chanhigh regardless, whether there is an overlap between the two.

schoffelen: I understand that, but if you call ft_channelcombination with ({low high},labellist,1,2), (with low and high disjoint) then you will end up with n^2 pairs, which means that what was initially intended to be channels with the high frequency envelope also end up as being the candidate channels to deliver the 'low' phase signal. To me it seems that this is not intended, and might fail. I am happy to change it  back, no problem.

schoffelen:
I will submit the change in a new PR

ft_crossfrequencyanalysis.m
tiborauer: Yes. But I would leave that decision to the user, whether (s)he specifies pairs with both directions (i.e. overlap between chanlow and chanhigh)..

tiborauer:
Thank you (I accidentally clicked on the "Start review". I hope it did not cause an issue.).

PR #1700, opened 2021-03-12
this reimplements Tibor's PR #1695

schoffelen:
The error Tibor got was a low-level file's too small error thrown by SPM(12) when trying to create a deformation volume on-the-fly, based on the spatial normalisation parameters. test_ft_volumenormalise (which also calls ft_warp_apply with sn2indivdual as a method) has not complained before, and Tibor's use case differed with respect to the ones in the existing test function in that cfg.nonlinear was 'no' in his call to ft_volumennormalise (in the other use case it was set to 'yes').

I couldn't find why this caused the error, but I checked that with the new version of the code both warps work, i.e. with parameters obtained from ft_volumenormalise (spm12) with nonlinear = 'yes' , and 'no'

### DIFF
--- a/test/test_issue1695.m
+++ b/test/test_issue1695.m
@@ -18,44 +18,15 @@ cfg     = [];
 cfg.dim = mri.dim;
 mri     = ft_volumereslice(cfg,mri);
 
+% this seemed to initially crash
 cfg = [];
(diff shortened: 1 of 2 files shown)


schoffelen:
This PR is a replacement PR for #1695, the latter one will be closed without merging

PR #1706, opened 2021-03-15
New mvpa interface: fix chan labels

treder:
@schoffelen  this should hopefully fix the channel label issue.


schoffelen:
Thanks!

PR #1708, opened 2021-03-15
Fix failed ft statistics montecarlo

schoffelen:
This PR fixes a failing test function, which was due to inconsistent data bookkeeping in ft_datatype_source, and obvious typo in the test function and an issue with inside dipole position allocation


schoffelen:
This addresses an issue that I noticed while working on a PR from @tiborauer (yet the issue was unrelated to that actual PR #1686).

PR #1709, opened 2021-03-16
clean up clustering code and implement TFCE

schoffelen:


schoffelen:
Reviewing @tiborauer latest PR and some errors introduced I noticed that the code that deals with clustering for statistics is rather messy. In addition, it has been on my list for a long time to also include tfce as a 'cluster' statistic. I checked with the implementation in LIMO, and it seems easier to just write it from scratch. I wrote a tfce function in the past (while in Glasgow) but cannot find it anymore, will just do it from scratch. I want to tackle these things in one go.

1) clean up the code in ft_statistics_montecarlo/findcluster/clusterstat
2) add tfce

robertoostenveld (approved):
looks good. I will update the documentation and help of the functions a bit, and will cleanup the code here and there.

PR #1712, opened 2021-03-17
make ft_source2full and ft_source2sparse compatible with current source structures

robertoostenveld:


robertoostenveld:
the reimplementation of source2full with fewer assumptions on the input data structure was a bit of a challenge (or beter: a fun puzzle). It now does not require an input dim+transform, but reconstructs those from the regular positions. However, it does need a position approximately in the center of the head that must have has regular neighbours.

robertoostenveld:
@schoffelen can you review and merge?

schoffelen (approved):
Looks good to me! Seems well thought through, and looks like a clean implementation. Good work.

PR #1717, opened 2021-03-19
add README file for each external toolbox 

robertoostenveld:
@jannystapel mentioned that it is not so clear who is in charge of the code in the `fieldtrip/external` directories. There is some documentation on https://www.fieldtriptoolbox.org/faq/external/, but I agree that it is not as clear as it should.

Quite some of the external toolboxes have a readme file or so. With this PR I propose to also add a readme file to those external toolboxes that don't have one at the moment.


robertoostenveld (commented):
I think this is enough for now. There are some that still have an empty README, these take more time to figure out where the code came from. I don't want to spend that time on it now, so will leave those files empty.

Issue #1701, opened 2021-03-12
ft_info and other notification functions slow down many FieldTrip functions


robertoostenveld:
**Describe the bug**

When looking at `ft_databrowser` with the MATLAB profiler, I noticed that quite some time was spent on `ft_info`.  We have replaced many occurrences of `fprintf` with that, and also use the same underlying code for `ft_warning` and `ft_error`. However, I do worry that the sophistication that we rely on for e.g. "warning once" slows down FT in general.

**To Reproduce**

```
n = 1000;

stopwatch = tic;
for i=1:n
  fprintf('hello\n');
end
t1 = toc(stopwatch);

stopwatch = tic;
for i=1:n
  ft_info hello
end
t2 = toc(stopwatch);

t1
t2
```

gives me this

```
t1 =
    0.0107
t2 =
    1.3086
```

So ft_info is more than 100 times slower.

**Expected behavior**

I don't see a fundamental reason why it would be 100x so slow.

**Environment and versions (please complete the following information):**
 - Operating System: macOS
 - MATLAB version: R2020b
 - FieldTrip version: latest master

schoffelen:
hehe, if I run it, it takes > 4 seconds. Half of the time seems to be spent on the line (in ft_notification) p = fileparts(which('ft_defaults')), which is I guess needed to get ft_path

robertoostenveld:
I see it varying quite a bit over repeated calls, sometimes it is a factor 100x, sometimes 250x.

robertoostenveld:
replacing that line with 
```
[v, p] = ft_version;
```
already speeds it up quite a bit.

schoffelen:
It's actually an interesting profiling exercise. I notice that fileparts in itself is quite 'slow' when it evaluates ispc: I wonder if it's worthwhile to overload it, and use a persistent variable (or input flag to denote pc'ness)

Might be something to consider in general for own FT-code that calls ispc often

schoffelen:
Not sure, but the repeated evaluation ismember(<something>, {s.identifier}) might be speeded up, if the {s.identifier} is created once as a temporary variable, which removes the need of doing the { } expansion multiple times per call to ft_notification

robertoostenveld:
I made a `test_issue1701.m` script and made `ft_version` a bit faster by beter use of the persistent variables, resulting in `fileparts` only being called once. I also made some other changes in the commit 13a0700 that I just merged with master. 

I suspect there is further room for improvement.

PR #1719, opened 2021-03-22
ENH - some substantial speedups based profiling it on a Mac with Matl…

schoffelen:
…ab 2018b

### DIFF
--- a/utilities/ft_version.m
+++ b/utilities/ft_version.m
@@ -53,6 +53,11 @@ persistent issvn
 persistent isgit
 persistent ftver
 persistent ftpath
+persistent is_pc
+
+if isempty(is_pc)
+  is_pc = ispc;
+end
 
 if isempty(ftpath)
   ftpath = fileparts(mfilename('fullpath'));
@@ -63,7 +68,7 @@ end
 clean = 'no';
 branch = 'unknown';
 
-if ispc
+if is_pc
   % this requires a file extension
   ext = '.exe';
 else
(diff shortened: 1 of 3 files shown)


schoffelen:
This is following up on @robertoostenveld 's #1701, based on which already some changes were made. As a profiling exercise, I also looked into the largest bottlenecks, and noticed that:
- expansion of a bunch of strings in a struct array, as per {s.identifier} can be sped up if it's done only once within a function, and thus when creating it as a local variable in ft_notification it does not need to be done in each of the if/elseif evaluations in ft_notification
- using the local variable ident = {s.identifier} as additional input arguments to the subfunctions getstate/setstate etc, removes the need to re-expand the {s.identifier} within those subfunctions
- ismember(string, cellarray of strings) is slower than any(strcmp(cellarray of strings, string)) 
- defaultId can take as input the pruned stack, so that it does not need to be re-created within defaultId again.
- ispc is relatively slow and can be created as a persistent variable in ft_version

schoffelen:
The changes in this PR reduce (on my Mac, matlab2018b) the profiling time of the function pasted below from 6 seconds to 3.8:

%%%%%%%%%%%%%%%
% start of main function
function [t1,t2] = info_test

n = 1500;

t1  = run_subfun1(n);
t2  = run_subfun2(n);

% subfunctions
function t1 = run_subfun1(n)

stopwatch = tic;
for i=1:n
  fprintf('hello%s\n','_blup');
end
t1 = toc(stopwatch);

function t2 = run_subfun2(n)

stopwatch = tic;
for i=1:n
  ft_info('hello%s','_blup');
end
t2 = toc(stopwatch);

% end
%%%%%%%%%%%%%%%%%%%%%%%%%

robertoostenveld (approved):
Looks good, you made some smart improvements!

PR #1724, opened 2021-03-25
FIX - new attempt for speed up, before there was an infinite recursio…

schoffelen:
…n, causing all test functions to fail

The cause of this was the fact that I failed to realize that the 'ident' variable (that was computed from the {s.identifier} at the top of the function, should be updated in some cases of the setstate subfunction call (since this could lead to an expansion of the s struct array, which required the ident to be expanded in parallel). This caused an infinite recursion in getstate (calling itself) in case the s struct array input was inconsistent with its corresponding ident variable.

This error was undetected by my clunky test script, because it called ft_notification directly rather than from underlying fieldtrip-code. Hence, setstate was not called and tested appropriately. 

Anyhow, given the time spent on this already, and the still non negligible speed up of the code, we should consider merging this new PR again, but I will first run some more tests... (note that I made the fix on my Mac, and want to run the tests on the cluster)

### DIFF
--- a/utilities/private/defaultId.m
+++ b/utilities/private/defaultId.m
@@ -39,8 +39,8 @@ else
   
   % remove the non-FieldTrip functions and scripts, these should not be part of the message identifier
   [v, p] = ft_version;
-  keep = ~startsWith({stack.file}, p);
-  stack = stack(keep);
+  keep   = startsWith({stack.file}, p);
+  stack  = stack(keep);
 end
 
 switch numel(stack)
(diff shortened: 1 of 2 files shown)


schoffelen:
I think it's now in order.

The failure of test_warning_once was due to the fact that the stack was incorrectly pruned, i.e. the ft-functions were pruned off, rather than the non-ft-functions

schoffelen:
I ran the test batch. There are some dashboard errors, but these pertain to file permission issues and toolbox-not-on-the-path things it seems. I'll merge and god zegene de greep

Issue #1725, opened 2021-03-26
spm_sample_vol 'Bug' in ft_warp_apply


linrongqi:
Hey FT community,
I want to use the ft_warp_apply to tranform the subject's coordinate into MNI.
When i use the fuction 'ft_warp_apply' , there was a bug
##########
elec_mni = elec_acpc_fr;
elec_mni.elecpos = ft_warp_apply(fsmri_mni.params, elec_acpc_fr.elecpos, 'individual2sn');
elec_mni.chanpos = ft_warp_apply(fsmri_mni.params, elec_acpc_fr.chanpos, 'individual2sn');
elec_mni.coordsys = 'mni';
###########
##########
Wrong use spm_sample_vol
File too small.
Error in spm_preproc_write8 (line 182)
 f          = spm_sample_vol(res.image(n),x1,x2,o*x3(z),0);
Error in individual2sn (line 81)
 spm_preproc_write8(P, zeros(6,4), [0 0], [1 0], 1, 1, bb, 1);
Error in  ft_warp_apply (line 189)
warped = individual2sn(M, input);
########
I want to know how to fix this bug.
Thanks for the great toolbox.
Rongqi lin.

schoffelen:
I think that this has been addressed. Could you upgrade to the latest release version of the toolbox and try again?

linrongqi:
Thank you for your suggestion! I have downloaded the latest version frildtrip-20210325, but the error still here. My code was following.
###### 
mrit1 = ft_read_mri('T1.nii'); 
mrit1.coordsys = 'acpc';

cfg            = [];
cfg.nonlinear  = 'yes';
cfg.spmversion = 'spm12';
cfg.spmmethod  = 'new';
fsmri_mni = ft_volumenormalise(cfg, mrit1);

elec_mni = elec_acpc_fr;
elec_mni.elecpos = ft_warp_apply(fsmri_mni.params, elec_acpc_fr.elecpos, 'individual2sn');
elec_mni.chanpos = ft_warp_apply(fsmri_mni.params, elec_acpc_fr.chanpos, 'individual2sn');
elec_mni.coordsys = 'mni';
##########
#########
Wrong use spm_sample_vol
File too small.
Error in spm_preproc_write8 (line 182)
f = spm_sample_vol(res.image(n),x1,x2,o*x3(z),0);
Error in individual2sn (line 81)
[...]

PR #1726, opened 2021-03-28
Update individual2sn.m

linrongqi:
According to the suggestion from #1725, i fix the error in individual2sn.m


schoffelen:
Thanks!

Issue #1507, opened 2020-07-26
Support for variable time axis in ft_multiplotER with more than one condition?


NirOfir:
**Is your feature request related to a problem? Please describe.**
One advantage of fieldtrip is working with trials with variable duration. I would like to plot on a single layout responses to a stimulus with variable duration for clear visual comparison. Currently, ft_multiplotER forces the x axes to be the same for all plots, because it assumes that all inputs have the same time field in their structures.

**Describe the solution you'd like**
I'm thinking about creating a single x axis before the xlim calculations are done in lines 409-428 by taking the union of all time fields from all inputs. To make sure ft_plot_vector won't compress or dilate vectors, I'll probably use nan padding to make everything the same duration, which I think is the solution in ft_databrowser now.

Does that sound like a helpful contribution to fieldtrip? Ideas to things it might breaks?

schoffelen:
I think in principle this is a good feature request.

I would recommend exploiting the functionality that is already in ft_selectdata. There's an option cfg.select, that can either be 'intersect' (default), or 'union'. The latter is the thing you describe. Typically, in a function with multiple input arguments, ft_selectdata handles the selection of channels/time points, e.g. as per line 108/109 in ft_timelockstatistics. ft_multiplotER and (I think) ft_singleplotER still do this selection in the function itself, most likely due to the fact that the varargins are untouched, because these are to be passed to the next function unchanged (interactive mode). Yet, in order to avoid code duplication I think there is no reason (apart from an acceptable increased memory footprint) against passing the data through ft_selectdata for channel and latency selection, creating a temporary data-array, which is subsequently used for ylim selection and to-be-plotted datamatrix (and mask) creation.

PR #1735, opened 2021-04-16
ENH - allow for union selection for plotting multiple inputs

schoffelen:
This addresses issue #1507


NirOfir:
Thanks! Looks like a much better solution than what I had in mind. I will give it a try.

NirOfir:
It does work great. Thanks again!

PR #1739, opened 2021-04-20
ENH - added 129-channel egi geodesic net layout

schoffelen:


schoffelen:
It would make sense to do this for the 256 channel (and perhaps 64 etc) caps as well. From what I can tell, the difference between the 128/129, and 256/257 is the addition of the named Cz electrode, otherwise the electrode labels and positions seem to be the same in 3D

robertoostenveld:
it would be good to add this to  the template/layout page, and to mention it on the EGI getting started page

schoffelen:
The former has been done, the latter not yet...

schoffelen:
but I see that the figures are quite small, I will increase them a bit

PR #1741, opened 2021-04-22
added try-catch if the mex file is invalid with instructions

mcpiastra:
hi all, hi @s-schrader,
I added a try-catch if the duneuro_matlab.mex64 file is invalid. 
a warning points to the instructions on how to generate a valid mex file (DUNEuro Wiki page).
what do you think?
best,
MC

### DIFF
--- a/forward/ft_headmodel_duneuro.m
+++ b/forward/ft_headmodel_duneuro.m
@@ -221,7 +221,12 @@ else
   end
 end
 
-headmodel.driver = duneuro_meeg(cfg);
+try
+    headmodel.driver = duneuro_meeg(cfg);
+catch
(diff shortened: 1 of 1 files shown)


mcpiastra:
note: I tested it on my laptop with test_pull1663.m where the error was thrown. there should be no error thrown on mentat.

s-schrader:
Hi, I think this is a better intermediate solution than the original mex file error, however, in the long term it would be great to directly provide better tested and more flexible mex files. I contacted Andreas to update the DUNEuro homepage which is referenced in the gitlab wiki you link, so that the clone and config files there are up-to-date (version 2.6).

PR #1743, opened 2021-04-23
[BUG] Fix nmt_sourceplot for time-resolved source data

britta-wstnr:
This fixes a bug with `nmt_sourceplot`, which was not able to plot time-resolved source data (plotting `source.avg.mom` as a result of LCMV beamformer), throwing an error about a missing field `freq`.

@sarangnemo Can you have a look whether this would break other functionality of `nmt_sourceplot`?

### DIFF
--- a/contrib/nutmegtrip/nmt_sourceplot.m
+++ b/contrib/nutmegtrip/nmt_sourceplot.m
@@ -432,10 +432,10 @@ for funidx = 1:length(funparameters)
             insideindx = find(functional.inside);
             
             tmpfun = cell2mat(tmpfun);
-            switch(dimord)
-                case {'{pos}_freq_time','{pos}_ori_time','{pos}_unknown_time'} % ori is hack... FT thinks that 3 freq bands is an ori!
-                    tmpfun = reshape(tmpfun,[size(tmpfun,1)/length(insideindx) length(insideindx) size(tmpfun,2)]);
-                    tmpfun = permute(tmpfun, [2 1 3]);
+            if any(strcmp(dimord, {'{pos}_freq_time', '{pos}_unknown_time'})) | strcmp(dimord, '{pos}_ori_time') & isfield(functional, 'freq')
+                tmpfun = reshape(tmpfun,[size(tmpfun,1)/length(insideindx) length(insideindx) size(tmpfun,2)]);
+                tmpfun = permute(tmpfun, [2 1 3]);
+                dimord = '{pos}_freq_time';
             end
             
             %    tmpfun = real(tmpfun);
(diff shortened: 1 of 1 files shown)


robertoostenveld:
I don't have a test script or dataset to test this with, so won't review it in great detail. Nevertheless, either @schoffelen or I will have to berge it. Let us know when you think it is ready to be merged.

britta-wstnr:
Thanks, @robertoostenveld 
Ping @sarangnemo can you check?

britta-wstnr:
Just pinging @jordannicolasalves as well: this is what we stumbled upon a while back.

schoffelen:
The code changes themselves look clean enough to me. However, rumour has it that @sarangnemo and his team have a newer version of the code. Merging this PR may introduce git conflicts w.r.t. this newer function, but this is the whole idea behind git to begin with. As such I would be willing to take the chance, and just merge this, and have the burden of fixing any conflicts with the remote stakeholders that so far haven't updated the code regularly enough.

However:
I tried to pull @britta-wstnr 's branch, and had the ambition to quickly write a test function. Using a source structure off the shelf did not work. After some tweaking I got some results, but this would require further changes to nmt_sourceplot. In principle, the intended functionality (I think) is nice, but at the moment a lot of band aids will be needed I suspect in order to be able to exploit this functionality. 
I am mostly worried about the fact that a large chunk of the code is concerned with data handling and reformatting (i.e. based on a fieldtrip-type source structure) in order to create the data matrix that can be handed over to the lower-level nutmeg code (as represented in contrib/nutmegtrip/private). Over the past years, fieldtrip has become much better at data structure bookkeeping, while the current nmt_sourceplot code seems to reflect a very old version of ft_sourceplot, and an outdated way of data handling. It would be nice if the the data bookkeeping part of the code could be rewritten, rather than adding adding a bandaid for each dysfunctional use case encountered along the road. 
Some things I was pondering about: why does the code assume that the frequencies are represented in the field freqbands, as opposed to freq? How should the function behave in case of vector-beamformer input? What does a correct fieldtrip structure with time frequency data look like? What would be the use case for visualizing spatial filter coefficients or forward solutions? Should we add spm_orthview and spm_image to fieldtrip/external/spm8/12 so that the user does not need a 'full' installation of spm?

All in all, I think that with a well-defined scope of the function, as well as with some effort in cleaning up the code in general and the data bookkeeping part in particular there is really some merit to have this function. In its current state, I seriously doubt its usefulness.

sarangnemo:
I've been snowed under, but have planned to review the code next week (and also attempt to merge any relevant updates from my end). Indeed, I took the ft_sourceplot structure as the basis, as discussed with you guys when I visited Nijmegen ca. 2016. :-) It has indeed been difficult to keep up-to-date with ft_sourceplot as it has evolved since then.

_why does the code assume that the frequencies are represented in the field freqbands, as opposed to freq?_
Frequencies are defined as frequency bounds, to allow for unequal frequency bandwidths as is commonly done with Hilbert transformed data (e.g. 8-12 Hz, 12-20 Hz, etc.)

_How should the function behave in case of vector-beamformer input?_
I believe vector beamformer input is not currently completed, but could be added if somebody had a need for it.

_What does a correct fieldtrip structure with time frequency data look like?_
The code as designed is able to plot standard FieldTrip beamformer evoked responses or similarly formatted output from my group's Hilbert source localization pipeline, e.g. _source.mom_ with voxels containing 1 x N time points for evoked responses for evoked responses, or M frequency bands x N time points)

_ What would be the use case for visualizing spatial filter coefficients or forward solutions?_
Often as a sanity check, or to visualize how a hypothesized source would project to the sensors, etc.

_Should we add spm_orthview and spm_image to fieldtrip/external/spm8/12 so that the user does not need a 'full' installation of spm?_
I personally don't recommend that approach, since it frequently leads to conflicts when a user would like to operate both toolboxes side-by-side. "external" versions also tend to get out-of-sync and miss bug fixes or new features.

sarangnemo:
Basic functionality works for me, but appears to break masking functionality. Investigating and working on a patch...

sarangnemo:
Made a patch, trying to figure out how to update @britta-wstnr's PR with it. (My attempts today failed due to permissions errors.)  Will give it a try pushing it to her original branch tomorrow.

sarangnemo:
update: seems there's no straightforward way for me to push my revision based on @britta-wstnr's update. I'm coordinating directly with her to get it into this PR

schoffelen:
I suggest to follow up on the issues that @sarangnemo is having in a separate branch.

PR #1744, opened 2021-04-26
ENH - re-implementation of (undocumented) patchsvd and lbex options

schoffelen:
patch svd now allows atlas in input

### DIFF
--- a/test/test_lbex.m
+++ b/test/test_lbex.m
@@ -1,0 +1,228 @@
+function test_lbex(datadirs)
+
+% WALLTIME 00:30:00
+% MEM 12gb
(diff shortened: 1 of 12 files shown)


schoffelen:
This might be relevant for @sjeung and @reneterporten

schoffelen:
I need to still make a test function, so stay tuned

test/test_lbex.m
```diff
@@ -0,0 +1,228 @@
+function test_patchsvd(datadirs)
```
robertoostenveld, line 1: the function name here and in the other test script are swapped

schoffelen: oeps

PR #1745, opened 2021-04-29
improve support for processing of fMRI time series data 

robertoostenveld:
This pull request started off from a test script that I wrote following the MEG toolkit. 

The idea is that it should be possible (and not too hard) to convert a 4D fMRI representation into a source representation, and from there into a parcellated raw data representation, i.e. with each channel representing the average over a parcel. 

From there on it would be straight-forward to use the time-domain analysis methods. 

But already at the level of the source  representation it should be possible to use ft_sourcestatistics.

### DIFF
--- a/fileio/private/ft_checkdata.m
+++ b/fileio/private/ft_checkdata.m
@@ -771,13 +771,13 @@ if isequal(current, desired)
   % nothing to do
   
 elseif strcmp(current, 'fourier') && strcmp(desired, 'sparsewithpow')
-  if startsWith(dimord, 'rpttap')
+  if startsWith(data.dimord, 'rpttap')
     nrpt = size(data.cumtapcnt,1);
   else
     nrpt = 1;
   end
-  if contains(dimord, 'freq'), nfrq = length(data.freq); else nfrq = 1; end
-  if contains(dimord, 'time'), ntim = length(data.time); else ntim = 1; end
+  if contains(data.dimord, 'freq'), nfrq = length(data.freq); else nfrq = 1; end
+  if contains(data.dimord, 'time'), ntim = length(data.time); else ntim = 1; end
   
   fastflag = all(data.cumtapcnt(:)==data.cumtapcnt(1));
   flag     = nrpt==1; % needed to truncate the singleton dimension upfront
@@ -859,13 +859,13 @@ elseif strcmp(current, 'fourier') && strcmp(desired, 'sparsewithpow')
   
 elseif strcmp(current, 'fourier') && strcmp(desired, 'sparse')
   if isempty(channelcmb), ft_error('no channel combinations are specified'); end
-  if startsWith(dimord, 'rpttap')
+  if startsWith(data.dimord, 'rpttap')
     nrpt = size(data.cumtapcnt,1);
   else
     nrpt = 1;
   end
-  if contains(dimord, 'freq'), nfrq = length(data.freq); else nfrq = 1; end
-  if contains(dimord, 'time'), ntim = length(data.time); else ntim = 1; end
+  if contains(data.dimord, 'freq'), nfrq = length(data.freq); else nfrq = 1; end
+  if contains(data.dimord, 'time'), ntim = length(data.time); else ntim = 1; end
   
   flag      = nrpt==1; % flag needed to squeeze first dimension if singleton
   ncmb      = size(channelcmb,1);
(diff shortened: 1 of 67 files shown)


schoffelen:
Moet ik er even doorheen lezen?

robertoostenveld:
@schoffelen yes, that was indeed my idea ;-)

Could you first have a look at `test_pull1745.m` (just read it) to see whether it makes sense? And also whether the planned extensions of the test script (or example on the FT website) are realistic?

fileio/private/ft_checkdata.m
```diff
@@ -685,17 +695,17 @@
   data = rmfield(data, 'dof');
 end % if hasdof
 
-if ~isempty(cmbrepresentation)
+if ~isempty(cmbstyle)
```
schoffelen, line 698: Perhaps I am the only one, but I have quite a few old scripts lying around, where I call ft_checkdata directly, using 'cmbrepresentation' as key. Could we gracefully have ft_checkdata to silently rename 'cmbrepresentation' into 'cmbstyle' for backward compatibility purposes?

ft_sourceparcellate.m
```diff
@@ -94,19 +94,25 @@
...
 % ensure that the source and the parcellation are anatomically consistent
-if ~isalmostequal(source.pos, parcellation.pos, 'abstol', 1000000*eps)
-  ft_error('the source positions are not consistent with the parcellation, please use FT_SOURCEINTERPOLATE');
+if ~strcmp(source.unit, parcellation.unit)
+  ft_error('the units of the source and parcellation structure are not consistent, please use FT_SOURCEINTERPOLATE');
+end
+
+% ensure that the source and the parcellation are anatomically consistent
+tolerance = 0.1 * ft_scalingfactor('mm', source.unit);
+if ~isalmostequal(source.pos, parcellation.pos, 'abstol', tolerance)
```
schoffelen, line 114: perhaps quickly check whether this is similarly handled in ft_virtualchannel. It could be that you already did this, but I am going through the functions in alphabetical order.

fileio/private/ft_checkdata.m
robertoostenveld: That is already taken care off at the start of ft_checkdata with 
```
sel = find(strcmp(varargin(1:2:end), 'cmbrepresentation'));
if ~isempty(sel)
  ft_warning('the option ''cmbrepresentation'' is deprecated, please use ''cmbstyle'' instead');
  varargin{2*sel-1} = 'cmbstyle';
end
```

ft_sourceparcellate.m
robertoostenveld: good idea, I will update it there as well.

fileio/private/ft_checkdata.m
schoffelen: ah I missed that, sorry for bringing it up

robertoostenveld:
```
roboos@mentat002> cd /home/megmethods/roboos/fieldtrip/dashboard/logs/latest
roboos@mentat002> grep FAILED *.txt | wc -l
57
```

there are still some regression problems that require attention.

schoffelen:
some?!! :)

robertoostenveld:
```
roboos@mentat002> grep FAILED *.txt | wc -l
5
```

things are improving, but the batch is not yet complete

robertoostenveld:
it would be good if `private/parameterselection` would deal with the situation that the automatically constructed `volume.seg/seglabel` is now renamed to `volume.tissue/tissuelabel`. E.g. it could detect that `seg` is requested as parameter, that it is not present but that 'tissue` is, and then return that instead (plus give a warning). I am not sure how often  `parameterselection` is still being used, but it might help rescue a few existing users' scripts.

schoffelen (approved):
Looks good to me. In particular, the skeleton test function for this PR looks good. It remains to be seen how smoothly the parcellation step will work out, w.r.t. applying the inverse nonlinear warp to the 4D data (while the parameters of the warp are estimated from the 3D image). We'll only find out if it's tried out I guess

robertoostenveld:
Whether or not to add the inside in case of `insidestyle=[]` requires further thought. Initially I made it conditional, i.e. only call fix inside if there is an inside field, but that broke a number of rather random test scripts. So in https://github.com/fieldtrip/fieldtrip/pull/1745/commits/df1ea4c73fa11047a8d7dfe49e8a234adf567fc9 I reverted that change. The inside is now just as before. It was not the target of this pull request, and I would like to move on with getting this merged. 

The follow up parts in the  test_pull1745 script will need to be done later. For now I want to see if this is release-ready.

Issue #1746, opened 2021-04-29
add support for the polhemus 3D motion tracker file formats from kinemagic 


robertoostenveld:
Following discussion with @ruiliu-psy we identified that  the [kinemagic](https://github.com/lennartverhagen/kinemagic) toolbox from @lennartverhagen can read and process data from the Polhemus Liberty 3D tracker in the DCCN behavioral lab. It would be nice to have that functionality also in FieldTrip. 

This will allow to convert the data to BIDS using `data2bids` just like here https://www.fieldtriptoolbox.org/example/bids_motion/. 

Note that we **do not aim** to replace the existing motion processing that kinemagic can do, just the minimal preprocessing. Following data actual processing in kinemagic, we might want to implement a `kinemagic2fieldtrip` function for follow-up analysis, but that would be for a separate issue.

Specifically it pertains to this function https://github.com/lennartverhagen/kinemagic/blob/master/proc/read/km_read_polhemus.m.

robertoostenveld:
@ruiliu-psy can you send me an example data file? You can send it through `/home/common/temporary` or email.

ruiliu-psy:
> @ruiliu-psy can you send me an example data file? You can send it through `/home/common/temporary` or email.

Hi, Robert! I uploaded the data onto '/home/common/temporary/HoiRobert', you cannot miss it ;)
The README.txt file illustrates contents of each file. The Polhemus recording doesn't have a hrd file, please let me know if you need the configuration file, I can fetch it from the lab PC tomorrow (for the analysis I didn't use it).

robertoostenveld:
Thanks, found it! I have copied the data to `/home/common/matlab/fieldtrip/data/test/issue1746`. We can use that location in a test script that can be executed on the compute cluster.

robertoostenveld:
Hi @ruiliu-psy couldd you try the code in https://github.com/fieldtrip/fieldtrip/blob/master/test/test_issue1746.m on your dataset and check whether it makes sense?

ruiliu-psy:
Hi @robertoostenveld thanks for the quick action!
The headerformat ('*.csv') is not supported yet. Error occured in both ft_read_header and ft_read_data.

robertoostenveld:
did you use the latest master version from github?

ruiliu-psy:
yeah, the one your updated 17h ago. Here's the error report, but 'delimitedtext' seems correct to me.

```
Error using readtable (line 216)
[...]

PR #1749, opened 2021-05-01
fixed data reshape problem relates to issue #1746

ruiliu-psy:
Re #1746

okay, I see the rational behind, i.e. to treat position and origination of the same 'Polhemus sensor' as separate 'channel'. Then, the data.label{1}= 'sensor' and the first row of data.trial{1} = [3 3 3 3....] would be unnecessary? Not sure if you add it for any particular reason, so I leaved it in.

The dat and label was mismatched, due to data reshaping. This PR just fixed it.


robertoostenveld:
@ruiliu-psy and I reviewed the suggested changes in detail.
thanks!

PR #1751, opened 2021-05-04
implement signal quality index (SQI) algorithm for NIRS data

sofisappia:
Fieldtrip implementation of the [ Signal Quality Index (SQI) algorithm](https://github.com/Artinis-Medical-Systems-B-V/SignalQualityIndex/blob/master/SQI.m), published in [Biomedical Optics Express](https://doi.org/10.1364/BOE.409317). SQI assesses the signal quality of fNIRS signals and provides a continuous rating between 1 (very low signal quality) and 5 (very high signal quality). It can also be used as a binary rating of the signal quality by thresholding the output scores, as has been done in this implementation.

### DIFF
--- a/external/artinis/ft_nirs_signalqualityindex.m
+++ b/external/artinis/ft_nirs_signalqualityindex.m
@@ -1,0 +1,216 @@
+function data_out = ft_nirs_signalqualityindex(cfg, data_in)
+% FT_NIRS_SIGNALQUALITYINDEX processes NIRS data and returns a copy of the 
+% original data replaced with nans in the signal segements that are bellow 
+% the specified quality threshold.
+%
+% Use as
+%   [out_data] = ft_nirs_signalqualityindex(cfg, in_data)
+% where indata is raw NIRS-data (in optical densities, ODs)
+% and cfg is a configuration structure.
+%
+%
+%   cfg.threshold    = scalar, the SQI (signal quality index) value that 
+%                      has to be exceeded to be labelled as a 'good' 
+%                      channel (default = 3.5)
+%   cfg.windowlength = scalar, the length (in seconds) of the signal
+%                      segments to be analyzed (default = 10)
+%
+% To facilitate data-handling and distributed computing you can use
+%   cfg.inputfile   =  ...
+%   cfg.outputfile  =  ...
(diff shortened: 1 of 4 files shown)


robertoostenveld:
To ensure that this works now but also in the future on different versions of MATLAB, with different operating systems, and that it continues to work - also when we do code maintenance in the future  (aka https://en.wikipedia.org/wiki/Regression_testing) - I would like to have a small test script and data set to complement this. 

The data (which preferably is as small as possible, but large enough to make sense) should not be added to the repository, but see here https://www.fieldtriptoolbox.org/faq/how_should_i_send_example_data_to_the_developers/ 

The test script should be placed in `fieldtrip/test` and can follow the format of the many examples there. You don't have to bother with the `dccnpath` function, just write the script for a file in the present working directory. Once I have received the file and added it to our test directory own the compute cluster, I will update the script so that the code responsible for https://www.fieldtriptoolbox.org/development/testing/ will find it.

robertoostenveld (changes requested):
To ensure that this works now but also in the future on different versions of MATLAB, with different operating systems, and that it continues to work - also when we do code maintenance in the future (aka https://en.wikipedia.org/wiki/Regression_testing) - I would like to have a small test script and data set to complement this.

The data (which preferably is as small as possible, but large enough to make sense) should not be added to the repository, but see here https://www.fieldtriptoolbox.org/faq/how_should_i_send_example_data_to_the_developers/

The test script should be placed in fieldtrip/test and can follow the format of the many examples there. You don't have to bother with the dccnpath function, just write the script for a file in the present working directory. Once I have received the file and added it to our test directory own the compute cluster, I will update the script so that the code responsible for https://www.fieldtriptoolbox.org/development/testing/ will find it.

robertoostenveld:
`6c7ec2c/test_pull1751.txt:test_pull1751 FAILED in 13 seconds`

This** was expected. Can you send an `oxy3` format version of the same file?

sofisappia:
> `6c7ec2c/test_pull1751.txt:test_pull1751 FAILED in 13 seconds`
> 
> This** was expected. Can you send an `oxy3` format version of the same file?

I just added the oxy3 file in the same drive folder.

PR #1755, opened 2021-05-10
Update savevar.m

matsvanes:
If the pathname doesn't yet exist, create it before saving.

### DIFF
--- a/private/savevar.m
+++ b/private/savevar.m
@@ -15,6 +15,11 @@ ft_info('writing ''%s'' to file ''%s''\n', varname, filename);
 
 eval(sprintf('%s = value;', varname));
 
+% Create path if it doesn't exist
+if ~exist(fileparts(filename))
+  mkdir(fileparts(filename))
+end
+
 s = whos(varname);
 
 % if variable < ~500 MB, store it in old (uncompressed) format, which is faster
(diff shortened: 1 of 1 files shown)


private/savevar.m
```diff
@@ -15,6 +15,11 @@ function savevar(filename, varname, value, hashfile)
 
 eval(sprintf('%s = value;', varname));
 
+% Create path if it doesn't exist
+if ~exist(fileparts(filename))
+  mkdir(fileparts(filename))
```
robertoostenveld, line 20: this will not make the directory if the parent directory does not yet exist. I suggest to implement it using a sub function with recursion, e.g. like

```
function mkdir_and_parent(d)
parent = fileparts(d)
if ~isfolder(parent)
  % this will recursively make all parent directories
  mkdir_and_parent(parent);
end
mkdir(d);
```

robertoostenveld (changes requested):

private/savevar.m
matsvanes: I believe `mkdir` already does this.  " `mkdir parentFolder folderName` creates `folderName` in `parentFolder`. If `parentFolder` does not exist, MATLAB attempts to create it.
https://www.mathworks.com/help/matlab/ref/mkdir.html#:~:text=If%20parentFolder%20does%20not%20exist,error%20to%20the%20Command%20Window.

robertoostenveld: you are right
```
>> mkdir('/tmp/a/b/c/d')
>> ls -al /tmp/a/b/c/d
total 0
drwxr-xr-x  2 roboos  wheel  64 May 10 16:20 .
drwxr-xr-x  3 roboos  wheel  96 May 10 16:20 ..
```

PR #1757, opened 2021-05-11
added multitaper feature to ft_connectivity_powcorr_ortho.m

toludwig:
As of Jan-Mathijs Schoffelen's 2012 version, orthogonalized power correlation was not implemented for multiple tapers. Instead, multiple tapers were simply averaged and then orthogonalized, which discarded important phase information. Here, we propose to orthogonalize and correlate tapers individually and only then average. Our code was tested in a project in the Siegel lab. Credits to Andrea Ibarra-Chaoul.

### DIFF
--- a/connectivity/ft_connectivity_powcorr_ortho.m
+++ b/connectivity/ft_connectivity_powcorr_ortho.m
@@ -1,29 +1,27 @@
 function [c] = ft_connectivity_powcorr_ortho(mom, varargin)
 
-% FT_CONNECTIVITY_POWCORR_ORTHO computes power correlation after removing the
-% zero-lag contribution on a trial-by-trial basis. This implements the method
-% described in JF Hipp, DJ Hawellek, M Corbetta, M Siegel, AK Engel. Large-scale
-% cortical correlation structure of spontaneous oscillatory activity. Nature
-% neuroscience 15 (6), 884-890.
+% FT_CONNECTIVITY_POWCORR_ORTHO computes power correlation after removing
+% the zero-lag contribution on a trial-by-trial basis, according to Hipp's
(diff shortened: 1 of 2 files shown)


schoffelen:
Thanks! Before looking at this in detail:
1) could you also include a snippet of test code, as per the functions in fieldtrip/test? Ideally, this snippet of code should demonstrate the functionality of the code (and also that the new functionality does not break old functionality). You might get started by grepping the m-files in fieldtrip/test for powcorr_ortho.
2) at first glance it looks as if it is assumed that all(tapvec==tapvec(1)). If so, this should be checked (and an error thrown if this is not the case).
3) the higher level caller function (ft_connectivityanalysis) probably does not know about optionally including the tapvec as an input option to ft_connectivity_powcorr_ortho. I am not sure whether ft_connectivity is commonly used as an interface to access the lower level code, but ideally this should also be added.

toludwig:
Hi, I finally came to work on it...
1. there is a test_pull1757.m included now. It just simulates data, FFTs it using 7 tapers and applies the new powcorr_ortho.
2. true, this assumption is actually already checked in your original version, an error would be thrown in line 58
3. I'm not sure about these points... the tapvec is explicitly set as data.cumtapcnt in ft_connectivityanalysis.m line 796, but as you say, there doesn't seem to be a possibility to set it as an option of ft_connectivityanalysis. If this is a common usecase (e.g. dropping the last taper?), then this would make sense to include (e.g. as a non-boolean cfg.keeptap). But at the moment, I don't see an urgent need for this, and it might be a separate issue

schoffelen:
Thanks Tobias for your contribution!

PR #1761, opened 2021-05-17
Fixed error when reading brainvision data

mcvinding:
This might be a compatibility issue (Matlab R2016), but the function `split ` does not work in this context. Replaced with `strsplit`.


robertoostenveld:
I indeed suspect it is a version issue, on 2020b both functions work the same in this case (except that one returns a Nx1 and the other a 1xN cell-array). But since `strsplit` is older and more specific for the purpose (i.e. one dimensional array of characters), this is a good change

PR #1767, opened 2021-05-27
changed default cfg.viewmode to component and  cfg.compscale to local

robertoostenveld:
I also added all cfg.zlim options from ft_topoplot, including [zmin zmax]


robertoostenveld:
@schoffelen do you agree with this change in defaults?

The reason that triggered this was that I was looking with someone at EEG ICA components, where the EEG data was referenced to a mastoid and the different components were rather asymmetric

PR #1768, opened 2021-05-28
HOTFIX - fixing regression error due to enthusiastic search and replace

schoffelen:


schoffelen:
Initially, I thought that @robertoostenveld had changed the headmodel.bnd into headmodel.mesh throughout the whole code base, but I think that the regression error was caused by an all too enthusiastic search-replace from 'bnd' into 'mesh', in ft_plot_headmodel

robertoostenveld (approved):
thanks for fixing this.

PR #1773, opened 2021-06-01
create functional support for mim

schoffelen:
The original stub created by Laura Marzetti in the early HCP age was never finished


schoffelen:
test function is to be written still. Pinging @reneterporten

PR #1784, opened 2021-06-10
Update xdf2fieldtrip so that it also return events

robertoostenveld:
This follows the idea of #1209 but implements it differently. I tested it with an LSL recorder file from one of our NIRS experiments.

### DIFF
--- a/test/test_pull1209.m
+++ b/test/test_pull1209.m
@@ -1,0 +1,42 @@
+function test_pull1209
+
+% WALLTIME 00:10:00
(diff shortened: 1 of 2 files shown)


robertoostenveld:
This is the result from the test script. 

Screenshot 2021-06-10 at 11 54 55

@varghese-peter if you want you can review the code in that script. I will now merge it, but if further changes are needed, please suggest them as a new issue or pull request.

This closes #1209

PR #1785, opened 2021-06-10
implement support for epoched fif files where the events are described with labels rather than numbers

robertoostenveld:
see #769


robertoostenveld:
this includes a test script and example data, it closes #769

PR #1788, opened 2021-06-16
ft_channelrepair fix

schoffelen:


schoffelen:
This PR fixes an issue that was introduced possibly in part by commit (but not completely) https://github.com/fieldtrip/fieldtrip/commit/1bac36a0e9c8e5da685adac9aad26f9a874cb710#diff-8928759be4040f099277872586116abef33641eeac403ae7a744b1055b36ebe8

The reported issue referred to the fact that ft_channelrepair choked on trying to extract (for the use cases described at least) sensor data from the data/cfg, but  didn't manage. I have cleaned up ft_channelrepair to explicitly state (also in the documentation) when a sens-array (with chanpos is needed) -> for the methods 'weighted', 'spline', 'slap', and make sure that no attempts are made to extract it from cfg/data (or reconstruct it from a layout), if not needed. I have added some test cases to the test function. 

-I still don't fully understand all details in the issue report, particularly why it seemed to have worked with older versions of the code, i.e. using the default 'weighted' option it seems (which would require sensor distances)
-also I don't understand well why 'weighted' is the default method, while the documentation + an explicit warning mention that this might be a dispreferred option.

robertoostenveld:
should we have a look at this together and discuss in person (online)?

schoffelen:
> should we have a look at this together and discuss in person (online)?

Ja, is goed!

robertoostenveld:
please make the `cfg.layout` parameter explicitly forbidden, otherwise it is fine.

PR #1811, opened 2021-06-26
added test for each of the functions in the PREPROC module

robertoostenveld:
- removed the benchmark scripts, those were not effective to ensure code quality 
- updated the help of some of the preprocessing functions
- changed the name of some input/output variables for consistency
- fixed two other text scripts

The new test scripts all follow a similar pattern, which consists of calling the function many times with different input arguments/options. If the inputs do something, then the results should be different. The test consists of checking that all results are different.


schoffelen (approved):
Looks good to me. Right now the assertions within the test functions are rather generic, but it's a good start.

PR #1816, opened 2021-06-30
DEV - added test_ft_preproc_xxxfilter functions

LucaLaFisca:
Maybe we could use this pull request as the starting point of the meeting of Thursday. Just to be sure I properly understood your "coding philosophy".

### DIFF
--- a/test/test_ft_preproc_bandpassfilter.m
+++ b/test/test_ft_preproc_bandpassfilter.m
@@ -1,0 +1,83 @@
+function test_ft_preproc_bandpassfilter
+
+% WALLTIME 00:10:00
+% MEM 2gb
+% DEPENDENCY ft_preproc_bandpassfilter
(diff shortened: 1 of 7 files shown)


robertoostenveld:
@LucaLaFisca and I discussed this and decided to close this pull request. These scripts are too detailled and overlap with the ones in #1811. Merging them in a single test_ft_preprocessing also is not trivial, since that test script already exists. For now the goal is to implement only test scripts (with minimal tests) for those high-level functions that do not have a test script yet. 

CC to @malinINCF  and @vijayiyer05, so that you know we are working on this

vijayiyer05:
Thank you for the update @robertoostenveld. Focusing on filling in the gaps first sounds like a sensible approach. FYI, I'm now watching the repository so will see all the updates :)

Issue #1802, opened 2021-06-23
implement a minimal test_ft_xxx script for all “main” fieldtrip functions


malinINCF:
a. Make a list of functions that do and do not have a test script
b. Group the functions according to their input data and document this. 
c. For each group, implement a test script

robertoostenveld:
using the following command-line code I identified the list of functions:

```
cd fieldtrip
for file in ft_*.m ; do if [ -n test/test_$file ] ; then echo test_$file ; fi; done
```

test_ft_analysispipeline.m
test_ft_annotate.m
test_ft_anonymizedata.m
test_ft_appenddata.m
test_ft_appendfreq.m
test_ft_appendlayout.m
test_ft_appendsens.m
test_ft_appendsource.m
test_ft_appendspike.m
test_ft_appendtimelock.m
test_ft_artifact_clip.m
test_ft_artifact_ecg.m
test_ft_artifact_eog.m
test_ft_artifact_jump.m
test_ft_artifact_muscle.m
test_ft_artifact_nan.m
test_ft_artifact_threshold.m
test_ft_artifact_tms.m
test_ft_artifact_zvalue.m
test_ft_audiovideobrowser.m
test_ft_badchannel.m
test_ft_badsegment.m
test_ft_channelnormalise.m
test_ft_channelrepair.m
test_ft_clusterplot.m
test_ft_combineplanar.m
test_ft_componentanalysis.m
test_ft_conjunctionanalysis.m
test_ft_connectivityanalysis.m
test_ft_connectivityplot.m
test_ft_connectivitysimulation.m
test_ft_crossfrequencyanalysis.m
test_ft_databrowser.m
test_ft_defacemesh.m
test_ft_defacevolume.m
test_ft_defaults.m
test_ft_definetrial.m
test_ft_denoise_dssp.m
test_ft_denoise_pca.m
test_ft_denoise_prewhiten.m
test_ft_denoise_synthetic.m
test_ft_denoise_tsr.m
test_ft_detect_movement.m
test_ft_dipolefitting.m
test_ft_dipolesimulation.m
test_ft_electrodeplacement.m
test_ft_electroderealign.m
test_ft_electrodermalactivity.m
test_ft_eventtiminganalysis.m
test_ft_examplefunction.m
test_ft_freqanalysis.m
test_ft_freqanalysis_mvar.m
test_ft_freqbaseline.m
test_ft_freqdescriptives.m
test_ft_freqgrandaverage.m
[...]

PR #1821, opened 2021-07-05
Test scripts - raw data input

LucaLaFisca:
Goal #1802 : test scripts for functions that input raw data


robertoostenveld:
I looked at the scripts, this is indeed what I had in mind. I also tested all scripts, they execute without problems (as expected).

Thanks, keep 'em coming like this!

robertoostenveld:
For your information: I have renamed `data.sens` into `data.elec` in some scripts, and removed `data.sens` from those where the tested function does not need it. For an example of a good electrode structure please check test_ft_dipolefitting or test_ft_sourceanalysis.

PR #1824, opened 2021-07-06
FIX - for issue 1822

schoffelen:
This is a fix for 1822 to make the partialisation option in ft_connectivityanalysis a little bit less messy.

### DIFF
--- a/connectivity/ft_connectivity_corr.m
+++ b/connectivity/ft_connectivity_corr.m
@@ -131,7 +131,7 @@ elseif ~isempty(pchanindx)
     
     this = powindx(k,:);
     sela = powindx(:,1)==this(1)&powindx(:,2)==pchanindx;
-    if sum(sela)>0
+    if any(sela)
       F_ap = input(:,sela,:,:);
     else
       sela = powindx(:,2)==this(1)&powindx(:,1)==pchanindx;
@@ -139,7 +139,7 @@ elseif ~isempty(pchanindx)
     end
     
     selb = powindx(:,2)==this(2)&powindx(:,1)==pchanindx;
-    if sum(selb)>0
+    if any(selb)
       F_pb = input(:,selb,:,:);
     else
       selb = powindx(:,1)==this(2)&powindx(:,2)==pchanindx;
--- a/connectivity/ft_connectivity_corr.m
+++ b/connectivity/ft_connectivity_corr.m
@@ -53,7 +53,7 @@ function [c, v, outcnt] = ft_connectivity_corr(input, varargin)
 %
 % Partialisation can be performed when the input data is (chan x chan). The following
 % option needs to be specified:
-%   pchan   = index-vector to the channels that need to be partialised
+%   pchanindx   = index-vector to the channels that need to be partialised
 %
 % See also CONNECTIVITY, FT_CONNECTIVITYANALYSIS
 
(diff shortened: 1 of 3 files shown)


connectivity/ft_connectivity_corr.m
```diff
@@ -81,14 +80,13 @@
...
-cmplx       = ft_getopt(varargin, 'complex', 'abs');
+hasjack     = ft_getopt(varargin, 'hasjack',  false);
+cmplx       = ft_getopt(varargin, 'complex',  'abs');
 feedback    = ft_getopt(varargin, 'feedback', 'none');
 dimord      = ft_getopt(varargin, 'dimord');
 powindx     = ft_getopt(varargin, 'powindx');
 pownorm     = ft_getopt(varargin, 'pownorm', 0);
-pchanindx   = ft_getopt(varargin, 'pchanindx');
-allchanindx = ft_getopt(varargin, 'allchanindx');
+pchan       = ft_getopt(varargin, 'pchanindx');
```
robertoostenveld, line 89: here the variable is `pchan`, whereas the option is `pchanindx`. In the help you document the option `pchan`

robertoostenveld: I see further in the code that the two are rather interchangeable. If you did not change the API, I suggest to keep the name `pchanindx`, it is more consistent with other `chanindx` options elsewhere

ft_connectivityanalysis.m
```diff
@@ -221,22 +221,18 @@
...
   case {'coh' 'csd'}
     if ~isempty(cfg.partchannel)
-      if hasrpt && ~hasjack
+      if hasrpt && ~hasjack && ~isfield(data, 'labelcmb')
         ft_warning('partialisation on single trial observations is not supported, removing trial dimension');
         try
           data = ft_checkdata(data, 'datatype', {'freqmvar' 'freq'}, 'cmbstyle', 'fullfast');
           inparam = 'crsspctrm';
-          hasrpt = 0;
+          hasrpt  = contains(data.dimord, 'rpt');
```
robertoostenveld, line 229: can you use `getdimord` instead, or possibly the data dimord that has been determined already earlier in the ft_connectivityanalysis code?

ft_connectivityanalysis.m
```diff
@@ -658,17 +649,25 @@
   case 'coh'
     % coherence (unsquared), if cfg.complex = 'imag' imaginary part of coherency
     optarg = {'complex', cfg.complex, 'dimord', data.dimord, 'feedback', cfg.feedback, 'pownorm', normpow, 'hasjack', hasjack};
-    if ~isempty(cfg.pchanindx), optarg = cat(2, optarg, {'pchanindx', cfg.pchanindx, 'allchanindx', cfg.allchanindx}); end
-    if exist('powindx', 'var'), optarg = cat(2, optarg, {'powindx', powindx}); end
+    if ~isempty(cfg.pchanindx), optarg = cat(2, optarg, {'pchanindx', cfg.pchanindx}); end
```
robertoostenveld, line 652: here it again is `pchanindx`, which is inconsistent with the change in the low-level function

robertoostenveld (changes requested):
I think there are still a few things that need to be changed, please see the comments

connectivity/ft_connectivity_corr.m
schoffelen: this has been resolved I think

PR #1832, opened 2021-07-13
DEV - added last test functions

LucaLaFisca:
Sorry, I didn't precise it was a matter of time to add the last test functions. Here there are.
I don't add any label as this part is closed in the MSoC.

### DIFF
--- a/test/test_ft_volumewrite.m
+++ b/test/test_ft_volumewrite.m
@@ -1,0 +1,16 @@
+function test_ft_volumewrite
+
+% WALLTIME 00:10:00
+% MEM 2gb
+% DEPENDENCY ft_volumewrite
+
+mrifilename   = dccnpath('/home/common/matlab/fieldtrip/template/headmodel/standard_mri.mat');
+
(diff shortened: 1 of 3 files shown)


test/test_ft_volumewrite.m
```diff
@@ -0,0 +1,16 @@
+function test_ft_volumewrite
+
+% WALLTIME 00:10:00
+% MEM 2gb
+% DEPENDENCY ft_volumewrite
+
+mrifilename   = dccnpath('/home/common/matlab/fieldtrip/template/headmodel/standard_mri.mat');
```
robertoostenveld, line 7: could you use something like
```
[ftver, ftpath] = ft_version;
mrifilename = fullfile(ftpath, 'template', 'headmodel', 'standard_mri.mat');
```
here? That allows the script to run anywhere with fieldtrip installed, not only on our cluster.

LucaLaFisca: It's done. I think some of the previous functions have the same issue... I didn't check if you already modified them.

Issue #1841, opened 2021-08-03
Duplication of the code for EEG channel interpolation based on splines


rbruna:
I am unsure if this should be considered a bug, so I decided to present it as a request.

**Context**

I am walking through the methods for the calculation of the Laplacian and I see that the spherical splines method, based in Perrin 1989, also provide bad/missing channel interpolation. It uses the function splint.

Meanwhile, the ft_channelrepair function also allows for interpolation of bad/missing channels using spherical splines. It uses the function sphericalSplineInterpolate, which is also based in Perrin 1989.

The implementation is similar in both cases, but there are two functions doing virtually the same, and I think it should be a good idea to merge them, to avoid future problems.

**Differences between implementations**

There are two differences between the two functions:
* splint generates, in addition to the potential, an estimate of the surface Laplacian. However, it is cheap to compute, so it should not be a problem.
* sphericalSplineInterpolate calculates the Legendre polynomials up to degree 500, stopping when the change is small.

**Extra**

splint calculates the Legendre polynomials using the plgndr MEX function, and sphericalSplineInterpolate calculates it explicitly in Matlab. In both cases, the polynomials are calculated for all degrees up to a maximum value.

I have created a MEX function similar to plgndr, but which:
a) works for matrix inputs and
b) returns all the polynomias up to i-th order, instead of only the i-th order polynomial.

Using this code would improve the efficiency of the code, with no numerical error at all. I attach the C-code here.

I offer to modify the functions for the calculation of the surface Laplacian and the channel interpolation myself, if you consider it useful.

[my_plgndr.c.zip](https://github.com/fieldtrip/fieldtrip/files/6924468/my_plgndr.c.zip)

schoffelen:
Hi Ricardo,

Thanks for your detailed report. I agree that it is not optimal to have two different implementations of doing the same thing. I assume that you are referring to ft_scalpcurrentdensity?

So, in brief, ft_channelrepair uses the splines to estimate 'missing' data based on the distribution of voltages, but it does not return the laplacian transformed data. Ft_scalpcurrentdensity uses the laplacian transformed data from splint (which also outputs some mapping matrices for free)

Our philosophy is also that we to avoid compiled as much as possible, as long as this does not at too huge a cost in terms of computation time spent. The reason is that maintaining mex files on the different supported platforms + versions is quite challenging. Moreover, for robust functional support, we'd need a m-file based overloaded version anyhow.  

[...]

PR #1846, opened 2021-08-05
Issue #1841

rbruna:
Pull request to solve issue #1841:
* Creates a new function sphsplint to replace both splint and sphericalSplineInterpolate.
* Creates a test file to check that the output is similar.
* Replaces the call for the previous functions in ft_channelrepair and ft_scalpcurrentdensity.
* Adds a new function mplgndr to calculate the Legendre polynomials (and associated functions) more efficiently.

In addition, the new sphsplint function center the electrode set in the origin of the best fitting sphere, to get a more accurate spherical interpolation.

### DIFF
--- a/ft_scalpcurrentdensity.m
+++ b/ft_scalpcurrentdensity.m
@@ -188,7 +188,6 @@ goodchanpos = allchanpos(goodindx,:);     % the position of good channels
 
 % compute SCD for each trial
 if strcmp(cfg.method, 'spline')
-  
   fprintf('Checking spherical fit... ');
   [c, r] = fitsphere(chanpos);
   d = chanpos - repmat(c, numel(find(sensidx)), 1);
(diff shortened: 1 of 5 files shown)


ft_scalpcurrentdensity.m
```diff
@@ -185,24 +185,32 @@
...
-      % compute scd for all channels, also for the bad ones
-      fprintf('computing scd also at locations of bad channels');
-      [V2, L2, L1] = splint(goodchanpos, data.trial{trlop}(goodindx,:), allchanpos, cfg.order, cfg.degree, cfg.lambda);
-      scd.trial{trlop} = L2;
-    else
-      % just compute scd for input channels, specify arbitrary channel to-be-discarded for interpolation to save computation time
-      [V2, L2, L1] = splint(goodchanpos, data.trial{trlop}(goodindx,:), anychanpos, cfg.order, cfg.degree, cfg.lambda);
-      scd.trial{trlop} = L1;
-    end
+    scd.trial{trlop} = WLo * data.trial{trlop}(goodindx,:);
```
schoffelen, line 213: I think that at this point we could consider using ft_apply_montage, both to the data object, and to the electrode description. I know that it is not part of this PR per se, but for code consistency w.r.t. the other methods it would be nice to have it. I can also imagine that we keep this for later

private/mplgndr.m
```diff
@@ -0,0 +1,72 @@
...
+%    GNU General Public License for more details.
+%
+%    You should have received a copy of the GNU General Public License
+%    along with FieldTrip. If not, see <http://www.gnu.org/licenses/>.
+%
+% $Id$
+
+
+% Checks the input.
+if order < 0 || round ( order ) ~= order
```
schoffelen, line 35: Could I ask you to update this file to have 2 spaces for a tab, rather than 4. 
Robert is quite keen on having this consistent.

This comment also counts for the other new function

schoffelen (changes requested):
Looks overall very good.

One comment w.r.t. formatting which I like to see solved.
We can discuss about the other one (related to ft_apply_montage)

ft_scalpcurrentdensity.m
rbruna: I think it is a good idea, as it would also solve part of the issues I found with the Hjorth method (see https://github.com/fieldtrip/fieldtrip/issues/1841#issuecomment-893580074).

I will change it.

private/mplgndr.m
rbruna: Of course. I do prefer 2 spaces, in fact, 4 is the default Matlab.

Do you know if I can change it? It would be great...

ft_scalpcurrentdensity.m
rbruna: Done.

Should we use ft_apply_montage too in ft_channelrepair? All methods but 'nan' use a spatial filter to build the interpolated data from the original.

private/mplgndr.m
rbruna: Done

schoffelen: In the matlab preferences you can specificy the number of spaces per tab:

environment -> preferences -> editor/debugger -> tab

PR #1848, opened 2021-08-09
[BUG] Fix sLORETA

britta-wstnr:
The gram matrix in `ft_inverse_sloreta` was computed based on a leadfield with wrong dimensions. Instead of `M x 3N` (`M` = number of sensors, `N` = number of source points), the input was of dimensions `MN x 3`.

### DIFF
--- a/inverse/ft_inverse_sloreta.m
+++ b/inverse/ft_inverse_sloreta.m
@@ -218,51 +218,56 @@ elseif ~isempty(subspace)
   end
 end
 
-L = cell2mat(sourcemodel.leadfield');
-G = L*L'; % Gram matrix
-invG = inv(G + lambda * eye(size(G))); % regularized G^-1
+if ~hasfilter
+    % compute the Gram matrix and its inversion across all voxels
+    L = cell2mat(sourcemodel.leadfield');
+    G = L*L'; % Gram matrix
+    invG = inv(G + lambda * eye(size(G))); % regularized G^-1
+end
 
 ft_progress('init', feedback, 'scanning grid');
 for i=1:size(sourcemodel.pos,1)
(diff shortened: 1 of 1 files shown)


inverse/ft_inverse_sloreta.m
```diff
@@ -218,7 +218,7 @@
   end
 end
 
-L = cell2mat(sourcemodel.leadfield);
+L = cell2mat(sourcemodel.leadfield');
 G = L*L'; % Gram matrix
 invG = inv(G + lambda * eye(size(G))); % regularized G^-1
```
britta-wstnr, line 223: Just realized that there is actually yet another problem with those lines: if I pass a precomputed filter, `ft_sourceanalysis` will delete `leadfield` from `sourcemodel`, hence these lines will fail. 
What is the best way to make sure they are only run if needed? 
An if-statement like `if ~hasfilter`?

britta-wstnr: I attempted a fix for this, it works on my own data.

robertoostenveld:
looks good to me, thanks!

rakansha:
The sLORETA returns the following as error-
Requested 98064x98064 (71.6GB) array exceeds maximum array size preference. Creation of arrays greater
than this limit may take a long time and cause MATLAB to become unresponsive. See array size limit or
preference panel for more information.

Error in ft_inverse_sloreta (line 224)
    G = L*L'; % Gram matrix

Error in ft_sourceanalysis (line 1085)
      dip(i) = ft_inverse_sloreta(sourcemodel, sens, headmodel, squeeze_avg, squeeze_Cy, methodopt{:},
      leadfieldopt{:});

Error in trysLORETALegs (line 56)
    source=ft_sourceanalysis(cfg, timelock);

Issue #1844, opened 2021-08-04
data2bids output is not "kosher"  for bids-validator wrt *CoordinateSystem


gdemarchi:
Dear @robertoostenveld and @schoffelen (who are the ones most engaged in the introduction of BIDS in the MEG community),
I'm trying to finalize the conversion to BIDS of an old dataset, that I started to work with a few years ago and then data2bids worked, now I'm testing the output with bids-validator and I'm getting some troubles. Few of them I managed to sort out (like true is different than 'true' and "true", here I appreciated much the 'flexibility' of fieldtrip :-)   ), but I keep encountering a problem that I didn't manage to solve alone after one day of fighting against data2bids and bids-converter.

Note: I'm using the online bids-converter, since it's much more straightforward to see the errors, but I get the same results using the command line.

TL;DR: from the big dataset now I'm trying to convert a single .fif file, and I get 3 errors, namely 

> Error 1: [Code 55] JSON_SCHEMA_VALIDATION_ERROR
Error 2: [Code 130] CHANNELS_COLUMN_TYPE_UPPER_CASE
Error 3: [Code 131] CHANNELS_COLUMN_TYPE

The first one boils down to (at first):

> Reason:
Invalid JSON file. The file is not formatted according the schema.
Evidence:
.HeadCoilCoordinateSystem should be equal to one of the allowed values

and this is due to the default for Elekta in data2bids.m being wrong now:

`
coordsystem_json.HeadCoilCoordinates                 = []; % FIXME it might be possible to get these from the dataset header
coordsystem_json.HeadCoilCoordinateSystem            = 'Neuromag';
coordsystem_json.HeadCoilCoordinateUnits             = 'm';
coordsystem_json.HeadCoilCoordinateSystemDescription = 'Neuromag head coordinates, orientation RAS, origin between the ears';
`

In particular, according to the Appendix VIII: Coordinate systems of the BIDS v1.6.0,

`coordsystem_json.HeadCoilCoordinateSystem            = 'Neuromag';
`
should be 

`coordsystem_json.HeadCoilCoordinateSystem            = 'ElektaNeuromag';
`

Easy fix, I'll PR as soon as we have the next one fixed. 
Once this is properly fixed with 

`
cfg.coordsystem.HeadCoilCoordinateSystem            = 'ElektaNeuromag';
`
 the next problem that pops up is then 

> Reason:
Invalid JSON file. The file is not formatted according the schema.
Evidence:
[...]

PR #1850, opened 2021-08-11
Various fixes to make data2bids compatible with Neuromag fif files. #1844 

gdemarchi:
For details see #1844


robertoostenveld:
thanks!

PR #1854, opened 2021-08-13
updated loadama for file version 11

ThomOostendorp:
This is the updated version of loadama, that works with both file format version 10 (used by old dipoli in fieldtrip) and new version 11 (used by new version of dipoli, not yet in field trip).


robertoostenveld:
thanks!

PR #1856, opened 2021-08-16
directly convert FT nirs data to snirf data without first conversion to homer .nirs data

helenacockx:
Two remaining questions:
- would you like to store the extra code in another function because it has become quite long?
- I've added to subfunction 'getorthoviewpos' from ft_prepare_layout to project 3D optode positions to 2D optode positions. To avoid double functions, should we make the getorthoviewpos a standalone function?


fileio/ft_write_data.m
```diff
@@ -875,12 +875,97 @@ function ft_write_data(filename, dat, varargin)
...
+%       all_wavelengths(i); % is not yet supported by the snirf toolbox
+      if any(all_wavelengths(i)==WL1.values)
+        snirf.data.measurementList(i).wavelengthIndex = 1;
+      else
+        snirf.data.measurementList(i).wavelengthIndex = 2;
+      end
+      snirf.data.measurementList(i).dataType = 99999;
+      snirf.data.measurementList(i).dataTypeLabel = 'dOD';
+    end
+    ft_warning('FT assumes that the input data is change in optical density. Adjust manually if not correct')
```
robertoostenveld, line 925: can you do 
    ft_warning('assuming that the input data represents (change in) optical densities')

The "Adjust manually" instruction is not specific enough (i.e. I would not know what to do if that were to appear on screen), so I have left that out. If you want to keep that, you should elaborate. And in that case it wou8ld probably be better do link to https://www.fieldtriptoolbox.org/getting_started/snirf/#writing-snirf-data and extend the help there rather than in the warning message that appears in the MATLAB the command window.

fileio/ft_write_data.m
```diff
@@ -875,12 +875,97 @@ function ft_write_data(filename, dat, varargin)
...
+    detector_idx=find(contains(data.hdr.opto.optotype, {'receiver', 'detector'}));
+    tra=data.hdr.opto.tra;
+    all_wavelengths=data.hdr.opto.wavelength(tra(find(tra>0)));
+    split=median(all_wavelengths);
+    WL1.values=all_wavelengths(all_wavelengths<split);
+    WL2.values=all_wavelengths(all_wavelengths>split);
+    WL1.nominal=round(median(WL1.values),-1);
+    WL2.nominal=round(median(WL2.values),-1);
+    num_WL=2;
+    ft_warning('Fieldtrip assumes that nominal wavelengths of %d and %d nm were used. Please adjust manually when this is not correct', WL1.nominal, WL2.nominal)
```
robertoostenveld, line 897: can you change this into 

    ft_warning('assuming that the nominal wavelengths are %d and %d nm')

As with the other comment on ODs the "Please adjust" instruction is not informative enough right now. How would one need to adjust the nominal wavelengths? Or should ft_write_data have an key-value option `nominalwavelength`, and similarly an option `datatype` for OD and the other SNIRF types?

fileio/ft_write_data.m
```diff
@@ -875,12 +875,97 @@ function ft_write_data(filename, dat, varargin)
...
+    % devide data in nirs channels, stimulus channels and auxillary channels
+    seldat  = startsWith(hdr.chantype, 'nirs');
+    selstim = strcmp(hdr.chantype, 'stimulus');
+    selaux  = ~seldat & ~selstim;
+    
+    % create empty SnirfClass
+    snirf = SnirfClass();
+    
+    % collect information for creation of snirf file
+    source_idx=find(contains(data.hdr.opto.optotype, {'transmitter', 'source'}));
```
robertoostenveld, line 887: Please add spaces around the `=`, i.e. as `<space>=<space>` for readability

fileio/ft_write_data.m
```diff
@@ -917,3 +1002,61 @@ function ft_write_data(filename, dat, varargin)
       type{i} = 'Other';
   end
 end
+
+function pos = getorthoviewpos(pos, coordsys, viewpoint)
```
robertoostenveld, line 1006: it would indeed be good to make this a regular private function. A copy is needed in `fieldtrip/private` and another copy in `fieldtrip/fileio/private`, and they should both be in `bin/synchronize-private.sh`. If you make a private function, please also update the "see also" help in `private/setviewpoint.m`

robertoostenveld (changes requested):

robertoostenveld:
> would you like to store the extra code in another function because it has become quite long?

No, I think it is better for clarity to keep the code in this function as it is.

helenacockx:
I've updated the code according to your suggestions.

helenacockx:
I ran into some issues when trying out the test_snirf.m file and debugged this code accordingly.

fileio/ft_write_data.m
```diff
@@ -875,12 +875,130 @@ function ft_write_data(filename, dat, varargin)
     data.label = hdr.label;
     data.sampleinfo = [1 size(dat,2)];
     
-    % convert the raw data structure to Homer format
-    nirs = fieldtrip2homer(data, 'event', evt);
+    % devide data in nirs channels, stimulus channels and auxillary channels
+    seldat  = startsWith(hdr.chantype, 'nirs');
+    selstim = strcmp(hdr.chantype, 'stimulus');
```
robertoostenveld, line 880: this could also be trigger channels, right? I propose something like 

```
selstim = contains(hdr.chantype, 'stimulus') | contains(hdr.chantype, 'trigger');
```

to make it more general

fileio/ft_write_data.m
```diff
@@ -875,12 +875,130 @@ function ft_write_data(filename, dat, varargin)
...
+    num_WL = 2;
+    ft_warning('assuming that the nominal wavelengths are %d and %d nm', WL1.nominal, WL2.nominal)
+     
+    % metaDataTags
+    snirf.metaDataTags(1).tags.LengthUnit = data.hdr.opto.unit; 
+    snirf.metaDataTags(1).tags.TimeUnit = 's'; 
+    snirf.metaDataTags(1).tags.FrequencyUnit = 'Hz';
+    
+    % data 
+    % (FIX ME: in theory this can contain multiple data blocks)
```
robertoostenveld, line 907: I think this can be fixed like this

```
for trl=1:numel(data.trial)
  snirf.data(trl).dataTimeSeries = data.trial{trl}(seldat,:)';
  snirf.data(trl).time = data.time{trl}';
end
```

the same would be needed for stim and aux

fileio/ft_write_data.m
```diff
@@ -875,12 +875,130 @@ function ft_write_data(filename, dat, varargin)
...
+    end
+    ft_warning('assuming that the input data represents (change in) optical densities')
+    % sort the channels according to wavelengths (because this is the way
+    % that homer handles data) and change the order of the data accordingly
+    [~, idx] = sort(([snirf.data.measurementList(:).wavelengthIndex]));
+    snirf.data.measurementList = snirf.data.measurementList(idx);
+    % update the data accordingly
+    snirf.data.dataTimeSeries=snirf.data.dataTimeSeries(:, idx);
+    
+    % stim
```
robertoostenveld, line 935: so I understand that you do not want to store the stim channels that are continuously represented. But what if there is a stim channel, but no evt?

I propose something like

```
if ~isempty(evt) && ~any(selstim)
   % the current code
elseif isempty(evt) && any(selstim)
   % store the stim channels just like the data and aux
elseif ~isempty(evt) && any(selstim)
   % both are present
   % hmm, a decision is needed here -> store both? Or have an key-value option that specifies which one?
else
  % no stim channels and no events
end
```

robertoostenveld (changes requested):

fileio/ft_write_data.m
helenacockx: It is not entirely clear to me whether the data blocks are intended to use this way. 
In the [notes on Ted Huppert's talk](https://docs.google.com/document/d/1AqkIff1SkLmwEkYCXuwEnUrz7Wo-N7T2DKGLUGknpXQ) on snirf I found this:
"indexing for data{i} → e.g. store different kinds of data (also hybrid) (eg. frequency domain and time domain), or junks of data when your device cannot measure it continuously"
Also in the stim field it is not clear to me how stimuli in different data block would be represented, because there is only 1 stim structure for all the data blocks.

fileio/ft_write_data.m
helenacockx: so if there is no evt but stimulus channels, you won't convert them to events, but add them as (aux) channels? Sorry, I don't have much experience with handling stim channels.

fileio/ft_write_data.m
robertoostenveld: It is also not 100% clear to me whether epoched data can be represented in SNIRF. Furthermore, in https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsidataj there is `/data(1)` but also `/data1`. I propose we ask somewhere in the SNIRF community.

An alternative is 

`/data(1).dataTimeSeries = cat(2, data.trial{:});`
`/data(1).data(trl).time    = cat(2, data.time{:});`

which will work fine if the data is pseudo continuous (i.e. segmented in subsequent pieces), but confusing when data is time locked to the onset of a stimulus. In that case the time axis would jump back and forth and `plot(/data(1).data(trl).time` would show a sawtooth pattern.

fileio/ft_write_data.m
robertoostenveld: Sorry, I did not read https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsistimjdata properly. I now see that the `stim` section is a table (a bit like `trl`), not a representation similar to continuously sampled trigger channels. Please ignore this code suggestion.

In general in FT if we use `ft_read_event` to parse continuous trigger channels (i.e. determine the onset of the blocks), then we still represent the trigger channel continuously as well. Throwing away the stim channels means that certain information gets lost, hence I would prefer being able to keep them as aux channels.

fileio/ft_write_data.m
helenacockx: I propose to leave it this way for now to make sure that no errors will happen there. I will also ask to the SNIRF community how they see it.
It might for example be that data can be divided into multiple blocks, but that the timestamps of the events always correspond to the data.time of the different blocks. This seems the most logic actually,... However, because events are linked to the data with sample numbers in FT, a lot of assumptions on how to relate the events to the data should be made here. I therefor prefer to let this up to the user and only have the option to provide one data block with one event structure.

robertoostenveld: agreed

robertoostenveld:
Now that I am executing some code to try it out, I realize that the discussion on multiple trials does not apply. We pass `dat` and `hdr` to `ft_write_data` and subsequently construct a one-trial FT raw data representation.

PR #1860, opened 2021-08-18
enable the support for EEG electrode positions from the BESA elp format

robertoostenveld:


Charl0ttedeB93:
Some .elp files contain a line to indicate the number of channels, see for example https://osf.io/xheg6/

PR #1868, opened 2021-09-01
Adding electrodes to surfaces for use in FEMfuns

meronvermaas:
This function creates a tetrahedral two-sphere mesh with cylindrical electrodes lying on the boundary between the two spheres (i.e., intracranial electrodes)
which can be used to perform (e.g., ECoG) forward simulations, particularly in FEMfuns (Finite Element Method for useful neuroscience simulations, https://github.com/meronvermaas/FEMfuns).
It has been created in the IntoTheBrain project (@mcpiastra @ThomOostendorp) and functions will be extended to allow the full use of FEMfuns within FieldTrip.
The structure of this script is more or less:
1. Create two triangulated spheres representing the brain and skull and electrode coordinates on top of the brain surface
2. Create cylinders at each electrode coordinate.
  a. make the cylinder surface crossing through the spherical surface it is combined with
  b. refine the spherical surface where the cylinder crosses it
  c. cut off the part of the cylindrical surface sticking into the sphere
3. Combine all the surfaces into one surface with unique faces and nodes
4. Make a tetrahedral mesh


robertoostenveld:
Hi Meron, 

thanks for the code suggestion. Prior to reviewing the code in detail, I want to get some general aspects clarified. 

I see that the code changes are only in `fieldtrip/external/femfuns`. On https://www.fieldtriptoolbox.org/faq/requirements/#external-requirements it is explained that FieldTrip "makes use" (i.e. depends) on some external toolboxes. Here it seems the other way around (or perhaps not even: since femfuns does not call FT, and FT does not call femfuns).  Why do you think that adding these functions to `fieldtrip/external` is the best way of distributing to the wider world. How many FieldTrip users do you think would use these functions? Why not a http://github.com/Donders-Institute/femfuns or a http://github.com/meronvermaas/femfuns ? 

I also recommend that you add an explicit LICENCE to your code. And a README that explains what the dependencies are, e.g. where can I find the source code and documentation of the `trident.exe` executable.

meronvermaas:
Hi Robert, thanks for your quick reaction. 

After discussing with @mcpiastra and @ThomOostendorp we agree that some clarifications are indeed missing. As you know, this is part of the into the brain project, in which we intend to integrate FEMfuns into FieldTrip, so it will be, in the end, FieldTrip calling FEMfuns. In the contract with the grant organisation, StITPro, we agreed upon certain mile stones at certain time points. The acceptance of the pull request would be proof that the mile stone is reached.

The first milestone was to be able to insert electrodes in a FEM model.  FEMfuns itself is not yet included in this pull request. In the pull request there is preliminary work for FEMfuns: create a 2d and 3d mesh of the headmodel with intracranial electrodes.
we placed the code in the external folder FEMfuns for now with the idea of exploring other applications where the meshing tool might be of use within ft: visualization of ECoG and stereoEEG data/setting might be a good example?

In the end, FEMfuns will be used by researchers interested in performing forward simulations with intracranial electrodes without profound coding skills. example of users can be our partners in Utrecht (Geertjan Huiskamp and colleagues) interested in performing simulations of electric brain mapping in epileptic patients.

You are right that this particular step does not yet connect FieldTrip and FEMfuns. We did not think about this when we set up the milestones and the way they must be verified.

I can see two options to go ahead:

    1. we go ahead with this pull request, in the confidence that the next step (a complete working example including FEMfuns, i.e., our second milestone) will make sense from the point of view of fieldtrip
    2. We show in the progress report that we have reached the first milestone, even though this isn’t proven by an accepted pull request. In that case a written statement from you that we did reach the milestone, and explaining why this cannot yet be accepted in FieldTrip, would be helpful.

As you understand, we would prefer option 1.

robertoostenveld:
I will in general not merge code with the fieldtrip release that does not meet the [requirements](https://www.fieldtriptoolbox.org/development/contribute/#contribute-code). 

Lacking here are the design document or a description of the development plan, a draft of the  end-user documentation, and a draft of the code architecture.

Furthermore, as per previous comment: the license is not specified on the MATLAB code, and the binary `trident.exe` has an unclear background. It might contain a virus, or might be copyrighted by someone else and not redistributable software.

I propose that you make a page on https://www.fieldtriptoolbox.org/development/project/ (also via github) where you explain what the plan is, what the milestones are, and how you will implement the deliverables. After such a project page, I suggest that the next deliver is a draft tutorial or example script that explains how people will be interacting with the still to-be-developed code through the FieldTrip interface. I can review that draft tutorial/example script and check whether it is in line with the FieldTrip design criteria.

I will close this pull request. You are welcome to reopen if the code is more complete, or start a new one.

robertoostenveld:
this is for now continued in https://github.com/fieldtrip/website/pull/453

PR #1872, opened 2021-09-02
ENH - added FOOOF functionality, using Brainstorm code

schoffelen:
started fooof implementation


schoffelen:
basic functionality is there, need to write a test function still

PR #1873, opened 2021-09-03
ENH: added volume-based registration of electrodes to ft_electroderealign

StolkArjen:
Due to construction work on ft_convert_coordsys, default behavior of the volume-based registration approach as described in the iEEG analysis protocol had changed. Specifically, ft_volumenormalise calls ft_convert_coordsys to perform an initial rigid body transformation for a better initial alignment of the source and target scans. This initial warp, however, is not reflected in the params field spit out by ft_volumenormalise and used in a subsequent call to ft_warp_apply for normalizing iEEG electrodes per the protocol. As a result, electrodes were no longer correctly mapped onto the template brain when that initial transformation was anything else but identity.

This PR solves the above issue while also housing previous functionality where it belongs, in ft_electroderealign. Following this PR, calls to the lower-level ft_warp_apply are no longer necessary and (iEEG) electrodes can be spatially normalized as follows:

cfg                      = [];
cfg.nonlinear      = 'yes';
cfg.spmversion  = 'spm12';
cfg.spmmethod  = 'new';
mri_norm = ft_volumenormalise(cfg, mri);

cfg                      = [];
cfg.elec              = elec;
cfg.method        = 'volume';
cfg.volume         = mri_norm;
cfg.warp             = 'mni';
elec_norm = ft_electroderealign(cfg);

### DIFF
--- a/ft_electroderealign.m
+++ b/ft_electroderealign.m
@@ -120,7 +120,7 @@ function [elec_realigned] = ft_electroderealign(cfg, elec_original)
 %   cfg.moveinward     = number, the distance that the electrode should be moved
 %                        inward (negative numbers result in an outward move)
 %
-% If you want to align electrodes to the freesurfer average brain, you should
+% If you want to align ECoG electrodes to the freesurfer average brain, you should
 % specify the path to your headshape (e.g., lh.pial), and ensure you have the
 % corresponding registration file (e.g., lh.sphere.reg) in the same directory.
 % Moreover, the path to the local freesurfer home is required. Note that, because the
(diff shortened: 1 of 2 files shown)


StolkArjen:
Just for fun, and to depict the issue more graphically, here's a screengrab. Left shows without considering the initial warp, resulting in offset electrode positions. Right shows the new procedure considering both the initial registration performed in ft_convert_coordsys and final normalization in ft_volumenormalise.

Screen Shot 2021-09-01 at 4 24 02 PM

robertoostenveld:
I discussed this with @schoffelen and we are positive about the proposed functionality change. However, to make it more consistent with `ft_prepare_sourcemodel`, we will implement it slightly differently. The call to `ft_volumenormalise` will move to inside the `ft_electroderealign` function.

StolkArjen:
Thanks for discussing this PR, both. To add one detail of complication, ft_electroderealign currently considers just the elecpos field as it is obviously meant to do when adjusting electrode positions. For spatial normalization, however, we want foremost the chanpos field to be adjusted, and perhaps also the elecpos. This strategy requires a deviation from the current approach in the case of alignment to mni or fsavg space.

robertoostenveld:
Looking at the code, I recognize the inconsistency in handling of elects and chanpos. However, that is for all methods, not only for the MNI normalization. I propose we tackle that in a separate PR.

robertoostenveld:
this PR is now replaced by #1875

PR #1875, opened 2021-09-07
added MNI registration of electrodes to ft_electroderealign

robertoostenveld:
See #1873, this is a reimplementation according to the comments there.

- the input cfg to ft_volumenormalise can now also have `cfg.initial`, just like `cfg.spmparams`
- the output of ft_volumenormalise does not have the `final` 4x4 matrix any more, since it is not being used elsewhere in FT


StolkArjen:
The code looks good to me. I might be able to briefly test and update the tutorial accordingly if the PR is pulled in this week.

schoffelen:
Same here!

StolkArjen:
Related follow-up: https://github.com/fieldtrip/fieldtrip/pull/1880

Issue #1874, opened 2021-09-07
improvement of spatial clustering for parcellated data


schoffelen:
**Is your feature request related to a problem? Please describe.**
Working with Rowan on one of his projects, we ran in to the issue of passing a cfg.connectivity matrix to ft_timelockstatistics, for parcellated data. The thing was that the order of the parcels in the data were different from the order of the labels in the atlas that was used for the cfg.connectivity. Consequence: BS results. This was caused by the assumption that the spatial dimension (and order of the items therein) of the cfg.connectivity matrix is the same as the order of the parcels in the data.

**Describe the solution you'd like**
Build in support in ft_statistics_montecarlo to create cfg.connectivity based on an atlas (perhaps better called brainordinate) field in the cfg. This namely allows also for a check of the order of the parcels w.r.t. the order in the data, and would lead to a more robust spatial neighbourhood matrix.

**Describe alternatives you've considered**
n.a.

**Additional context**
While looking into this, we also noticed that the subfunction in ft_checkdata that 'unparcellates' data (for the purpose of plotting it in ft_sourceplot) only works if the order of the labels in the brainordinate field is the same as the order in the data. This is due to an explicit (and too strict) if statement, and not needed, because the unparcellation seems to work robustly enough.

schoffelen:
pinging @rowan-sommers

schoffelen:
This has been addressed by PR 1877

PR #1877, opened 2021-09-07
ENH - code to use brainordinate to generate parcel adjacency matrix

schoffelen:
This addresses #1874


schoffelen:
TO DO:
-test function
-perhaps consider ft_timelockstatistics and ft_freqstatistics to pass on the brainordinate in the output 
-make the ft_checkdata a bit more lenient w.r.t. the 'unparcellation'

robertoostenveld:
so the idea is that the data is defined on parcels (aka virtual channels) and that - based on the full 3D specification of how the parcels are shared - you determine which parcels are to be considered neighbors?

Did you consider implementing this in `ft_prepare_neighbours` rather than implementing this in `ft_statistics_montecarlo`?

Neighbours are not only used for cluster stats. And we try to treat parcellated data as channel-data.

schoffelen:
No, I actually hadn't thought of implementing this in this way. That sounds like an elegant solution. Let me just think of this a bit, whether it is straightforward enough. Right now it feels as if it can only be done in a way that is 'dubbelop'. Stay tuned

PR #1880, opened 2021-09-08
enh: organize output of ft_electroderealign with method-mni

StolkArjen:
See https://github.com/fieldtrip/fieldtrip/pull/1875, this addition handles the output from cfg.method=mni.

As for the handling the coordsys, the coordsys from the templatemri is used, which is 'spm' for the default. Otherwise, 'mni' is chosen, that is the method's name.

### DIFF
--- a/ft_electroderealign.m
+++ b/ft_electroderealign.m
@@ -672,7 +672,7 @@ switch cfg.method
     % remember the transformation
     elec_realigned.homogeneous = norm.m;
     
-  case {'project', 'moveinward'}
+  case {'project', 'moveinward', 'mni'}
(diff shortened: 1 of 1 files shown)


StolkArjen:
For completeness, the output looks good:

Screen Shot 2021-09-08 at 9 14 56 AM

robertoostenveld:
thanks. FYI: `mni` and `spm` are considered the same in fieldtrip, see https://www.fieldtriptoolbox.org/faq/acpc/

StolkArjen:
Thanks, I indeed wasn’t sure if that had changed.

PR #1882, opened 2021-09-10
test function failure due to random behavior of random numbers

schoffelen:


robertoostenveld:
hmm, that is unexpected: random numbers behaving randomly 😁

PR #1887, opened 2021-09-21
add support to ft_redefinetrial to stitch pseudo-continuous segmented data back together

robertoostenveld:
The use case for this is to do `ft_preprocessing` on continuous data, segment it into non-overlapping segments with `ft_redefinetrial`, use `ft_rejectvisual`, and then put it all back together again. 

In case of keeptrials=no in rejectvisual, this will result in multiple (long) segment, one for each continuous part that was reconstructed.

In case of keeptrials-nan  in rejectvisual, this will result in a single continuous segment where parts of the data are replaced by nan.


robertoostenveld:
@schoffelen can you review this?

PR #1888, opened 2021-09-26
Add trialfun for BIDS and improve ft_trialfun_general

robertoostenveld:
Rather than adding more functionality to ft_trialfun_general, I decided to split it in separate functions. If general functionality is desired, it would be better implemented in `ft_definetrial` instead.

- added ft_trialfun_bids
-  added ft_trialfun_gui using existing code (mostly from @vlitvak) from ft_trialfun_general
-  added ft_trialfun_show using existing code from ft_trialfun_general


robertoostenveld:
please note that `ft_definetrial` has not really changed. I updated and reformatted the help considerably, and made a small change to the warning/error handling if `trl==[]`

robertoostenveld:
The internal logic of `ft_trialfun_general` (which was functional, but pretty poor) has been improved. I also added `cfg.trialdef.overlap`for reading continuous segments, in line with `ft_redefinetrial`.

robertoostenveld:
another notable change is that `cfg.trialdef.eventtype='gui'` is now better documented. It was originally implemented by @vlitvak for SPM, but apparently never widely adopted in FieldTrip documentation although it is very handy.

robertoostenveld:
the script `test_tutorial_spike_Neurosim.m` started failing. I renamed it to `failed_tutorial_spike_Neurosim.m` as I believe that the reason for failure is not due to an actual regression, but because of an error in the test script and the demo data that it uses.

Issue #1891, opened 2021-09-29
Reshape in ft_connectivity_wpli does not work if channel combinations are supplied


stefanches7:
Hello, I face a following bug: with the 2-dimensional input (Repetitions x Channelcombinations), that is declared possible in the function's comment, reshape in 
https://github.com/fieldtrip/fieldtrip/blob/7cfc41a851c643df8a29ca49afdee6e542d4847a/connectivity/ft_connectivity_wpli.m#L69

throws a "Size vector must have at least two elements" error.

schoffelen:
From the bug report, it is not clear what the use case is that the associated PR tries to resolve. The test_bug1891 does not provide any insights either.

stefanches7:
@schoffelen Right sorry, that would be the use case. I'm happy to include additional information wherever you see fit. I will comment the use case in the PR/test now

schoffelen:
Gotcha, I was confused by the input of a real-valued square matrix in the test function... the meaning of the dimensions would be observation x channelcombinations in this case, where the squareness (implicitly) suggested chan x chan or so.

stefanches7:
Yes it wasn't the luckiest choice :D now I've changed it to 200*400 and added a comment.

schoffelen:
OK, thanks!

PR #1892, opened 2021-09-29
Wpli connectivity:  Reshape for dimensions == 2 with an additional dimension

stefanches7:
#1891


connectivity/ft_connectivity_wpli.m
```diff
@@ -66,7 +66,11 @@
   else
     wpli     = outsum./outsumW; % estimator of E(Im(X))/E(|Im(X)|)
   end
-  wpli = reshape(wpli,siz(2:end)); % remove the first singular dimension
+  if (length(siz) == 2)
+    wpli = reshape(wpli,[ siz(2:end) 1 ] );
+  else
```
schoffelen, line 71: Rather than using the if-else statement, I'd recommend to change line 56 into siz = [size(input) 1]; 
This will always work.

schoffelen (changes requested):

PR #1894, opened 2021-10-01
print welcome message when executing ft_defaults

viccasand:
&cambios practicados para el tutorial de git%


robertoostenveld (commented):
Sorry for me taking some time to review this. I like the idea, but the implementation would better be done in the `ft_defaults` function.

PR #1901, opened 2021-10-15
ENH - some generic speedups and depth-of-recursion limitation

schoffelen:
As reported by Mikkel @mcvinding ft_sourceinterpolate may spend a lot of time in ft_struct2double, even if it is not strictly needed. This PR limits the recursion depth into ft_struct2double, so that no unnecessary time is spent in an attempt to convert data of non-interest. Also some structural improvements to ft_struct2double lead to faster execution times.


schoffelen:
test_ft_sourceinterpolate seems to run through fine.

PR #1906, opened 2021-10-22
Allow ft_volumenormalize to use a mask and TPM for custom templates + parcellation fix on source data

mcvinding:
* Added the option to use a mask to constrain the normalization with the "old" SPM method in `ft_volumenormalize`. The specified mask is passed to the SPM function.
* Made it possible to use a custom tissue probability map (TPM) for normalization with the "new" SPM method in `ft_volumenormalize`.
* Fix to `ft_datatype_parcellation` to ignore the `pow` field as it kept being interpreted as a probabilistic parcellation when passing a source structure with parcellation info to `ft_virtualchannel`.


mcvinding:
I am a bit unsure why and where the error exactly happened. From ft_virtualchannel and all the downwards functions that it require it seems to call ft_datattype_parccelation twice. It is the second time it run that the error occurred. 

You can find the data here: [https://www.dropbox.com/s/rs151mop4zpal8y/source_org.mat?dl=0]( https://www.dropbox.com/s/rs151mop4zpal8y/source_org.mat?dl=0 )

schoffelen:
OK, I downloaded the mat file, and will look at it asap. You can delete it on your end if you want.

schoffelen:
Ah, I see that indeed you have something with (slightly different from what I wrote above):

source.avg.blabla
source.tissue
source.tissuelabel

the question remains: is this a handcrafted situation, or did your input cfg.sourcemodel into ft_sourceanalysis contain tissue/etc. and was this passed on transparently to the output?

mcvinding:
The label info was first projected to the  template source grid and then added to the source estimate afterwards. The source space used for the source estimate was a warped version of the same  template grid.

````Matlab
%% Add atlas info to sources
atlas = ft_read_atlas(fullfile(ft_path, '/template/atlas/aal/ROI_MNI_V4.nii'));
load(fullfile(ft_path, 'template/sourcemodel/standard_sourcemodel3d6mm'));
sourcemodel = ft_convert_units(sourcemodel, 'cm');

cfg = [];
cfg.interpmethod = 'nearest';
cfg.parameter    = 'tissue';
tmp = ft_sourceinterpolate(cfg, atlas, sourcemodel);
tmp.tissuelabel = atlas.tissuelabel;

atlas_grid = ft_checkdata(tmp, 'datatype', 'source');
atlas_grid.inside = sourcemodel.inside;
````
...then after estimating the source, I just added the atlas info to the source structure:

````Matlab
%% add atlas
source_org.tissue = atlas_grid.tissue;
source_org.tissuelabel = atlas_grid.tissuelabel;
````
Then for the virtual channels:

````Matlab
labs = [1, 2, 13, 14, 57, 58, 77, 78, 97, 98]; % ROI

%% Make virtual channel
cfg = [];
cfg.parcellation = 'tissue';
cfg.parcel       = source_org.tissuelabel(labs);
cfg.method       = 'svd';
vrtchannls_org = ft_virtualchannel(cfg, epo, source_org);
````

**Comment by schoffelen** [comment]
Is there a specific reason why you handcraft the parcellation information into the source structure, rather than using the parcellation as 4th input argument into ft_virtualchannel?  (as per the documentation in the function)

**Comment by mcvinding** [comment]
I guess it is because I have missed that was an option 😄 Will it still work given the difference in `pos` for the grid in template coordinates and the grid in subject coordinates?

**Comment by schoffelen** [comment]
I don't remember how strict the pos consistency check is. If it is too strict, we could consider making it more lenient (i.e. check for the number of vertices, and do a tri-check). Alternatively, the parcellation's pos can be (temporarily) changed into the subject's pos.

File: ft_volumenormalise.m [hunk 1]
```diff
@@ -150,6 +151,13 @@
   ft_error('the template anatomy should be stored in an SPM-compatible file');
 end
 
+if ~isempty(cfg.templatemask)
```
robertoostenveld, line 154: please do this exactly the same as in the previous code block

ft_volumenormalise.m
```diff
@@ -181,20 +189,13 @@
 
 % Ensure that the input MRI has interpretable units and that it is expressed in a
 % coordinate system which is in approximate agreement with the template.
+ft_notice('Doing initial alignment...')
```
robertoostenveld, line 192: is there a reason that you now skip the processing of `cfg.initial` ? That was added only in Sept 2021 to deal with an issue reported by Arjen.

ft_volumenormalise.m
```diff
@@ -388,13 +414,10 @@
...
-% Remember the initial and normalisation parameters in the configuration, this allows
-% redoing the transformations without any computations (e.g. estimating them on a T1
-% and applying them on a T2)
-%
-% they are added only here to prevent them to be removed when doing the trackconfig
-cfg.initial   = initial;
+% remember the normalisation parameters in the configuration
+% maintain this order for the time being to prevent them to be removed when doing the trackconfig
 cfg.spmparams = params;
+cfg.final     = final;
```
robertoostenveld, line 420: `cfg.final` (which is a 4x4 affine transformation, even if a nonlinear transformation was applied) was removed in Sept 2021 in favor of `cfg.initial` (which is an affine transformation) plus `cfg.spmparams` (which can be affine or nonlinear). See https://github.com/fieldtrip/fieldtrip/pull/1875

robertoostenveld (changes requested):
it seems that you inadvertently reverted some changes from https://github.com/fieldtrip/fieldtrip/pull/1875. Please bring those back (see detailed comments). Also: can you use 2-space tabs instead of 4-space tabs?

ft_volumenormalise.m
mcvinding: This should also be fixed. It seems that my changes were not written in the most recent version of ft_volumenormalise.

ft_volumenormalise.m
mcvinding: I am not sure... It is supposed to do the same as above?

schoffelen:
I guess we all lost track on this one @mcvinding what is the status of this? From a quick review of the changes it seems as if most of Robert's comments are addressed. Apart from still some trailing 4-space tabs instead of 2-space tabs.

Also the cfg.templatemask option is not documented in the function's help. I don't know whether that's intended. If not, it should at least be documented as an undocumented option I think.

Also, it's still not clear whether the change to ft_datatype_parcellation is needed for generic code improvement, or whether it is needed for Mikkel's specific use case (where a hand-crafted data structure seemed to have been the cause of the problem that prompted the change in the code).

mcvinding:
I have added the documentation and removed the four-space tabs where I found them.

schoffelen:
ouch, now this PR 'thinks' that you changed 197 files. What happened?

mcvinding:
Good question? I am working on a new machine and got the latest version from upstream. Perhaps it think I have made all changes to the master since my last commit?

mcvinding:
I have no idea what happened - some of the "new" changes listed above are 16 months old? Perhaps, I should close this pull request and open a new one?

schoffelen:
Ah, but now it looks good again! No worries

ft_volumenormalise.m
schoffelen: I am not sure whether this has now been adequately fixed. It seems there is now a new cfg argument, called 'initial', which determines the functional behavior of line 196 onwards. cfg.initial does not seem to be initialized and is also not documented. So far, this option seems a bit obscure, it is not clear why (and when) it is needed.

schoffelen (changes requested):

ft_volumenormalise.m
mcvinding: Looking through the GitHub issues, the `cfg.initial` issue is related to how `ft_volumenormalize` handles iEEG data. The comment from @robertoostenveld was that I had edited an older version of `ft_volumenormalize` and therefore skipped this part that was added in Sept. The block on line 192 and forward should now be doing what it is supposed to. It is just not documented what it is supposed to do. I guess it just has to be added to the documentation what `cfg.initial` is? Have I understood correctly that `cfg.initial` is supposed to be the initial alignment between iEEG electrode coordinates and the MRI volume? It seem that it can also be used if you for some reason have done a coarse initial alignment of your volume to the template MRI and therefore want to skip the initial alignment again--for some reason? If this is correct, I can just add it to the documentation?

ft_volumenormalise.m
```diff
@@ -351,6 +386,9 @@
   normalised = setsubfield(normalised, cfg.parameter{k}, spm_read_vols(Vout(k)));
 end
 
+% determine the affine coordinate transformation from individual head coordinates to template coordinates
+final = VG.mat * inv(params.Affine) * inv(VF(1).mat) * initial;
```
robertoostenveld, line 390: if this is not used, then please do not compute it

ft_volumenormalise.m
```diff
@@ -41,6 +44,7 @@
 %   cfg.spmparams        = you can feed in the parameters from a prior normalisation, for example
 %                          to apply the parameters determined from an aantomical MRI to an
 %                          interpolated source resontruction
+%   cfg.initial          = Initial alignmet between target and template (optional; used for intracranial electrode alignment).
```
robertoostenveld, line 47: please fix typo in "alignmet"

trialfun/ft_trialfun_neuromagSTI016fix.m
```diff
@@ -5,29 +5,35 @@
 % values into a new "STI101" channel and then uses the new channel to define trials.
 %
 % Use this function by calling
-%   [cfg] = ft_definetrial(cfg)
+%  [cfg] = ft_definetrial(cfg)
```
robertoostenveld, line 8: please keep the two spaces

trialfun/ft_trialfun_neuromagSTI016fix.m
```diff
@@ -5,29 +5,35 @@
...
 
+% Undocumentet option:
+%  cfg.checkmaxfilter      = check that MaxFilter has run (default = yes).
+
 % Get only specific event type
 cfg.trialdef.eventtype = ft_getopt(cfg.trialdef, 'eventtype', 'STI101');
 cfg.trialdef.eventvalue = ft_getopt(cfg.trialdef, 'eventvalue', []);
 
+% MaxFilter option
+cfg.checkmaxfilter = ft_getopt(cfg.trialdef, 'checkmaxfilter', 'yes');
```
robertoostenveld, line 26: please do not specify a default at this level

robertoostenveld:
@schoffelen, I just discussed this with @mcvinding and he will make some small changes

ft_volumenormalise.m
robertoostenveld: I updated the help on `cfg.initial`

PR #1911, opened 2021-11-03
oxy5 file support for artinis data

Horschig:
This pull request adds support for oxy5 files in FieldTrip. There are still some rough edges (like warnings being thrown which are unnecessary), but I would like to merge ahead already given current available development time and that it seems to work fine.


robertoostenveld:
Thanks for this update, I reviewed the code (for as far it is accessible) and it looks good. Could you also share some data in oxy5 format (through another channel) and add that to https://github.com/fieldtrip/fieldtrip/blob/master/test/test_artinis_oxy.m ?

Horschig:
> Thanks for this update, I reviewed the code (for as far it is accessible) and it looks good. Could you also share some data in oxy5 format (through another channel) and add that to https://github.com/fieldtrip/fieldtrip/blob/master/test/test_artinis_oxy.m ?

yes, I'll try to remember that when I got time ;) I saw in that file that you commented out the oxy4 files because of an error. That same error will occur with oxy5 files. The reason is that the COM interface to OxySoft is also required for oxy5 (due to the compression/encryption algorithms in those files, which require the OxySoft implementation). Does it then make sense to still share the files or not? The COM interface requires Windows.

robertoostenveld:
>  Does it then make sense to still share the files or not? The COM interface requires Windows.

Yes, I would still like to have these files. We won't run the tests automatically (as that is on Linux) but still allows in specific cases to run it manually, e.g. on the windows laptop of @helenacockx or others.

robertoostenveld:
it would be nice to still receive a test oxy5 file, but merging of this PR does not have to wait for that.

PR #1917, opened 2021-11-10
external/brainstorm: Move internal functions to a private subfolder

ftadel:
Having an old bst_get.m or db_template.m added to the Matlab path by FieldTrip could have bad consequences on Brainstorm.
Keeping in the top-level folder only the functions that need to be exposed to FieldTrip (which wouldn't harm much if duplicated in the path)


schoffelen:
OK, in principle this is fine, but I would consider to 
1) do a quick check whether the 'old' functions are sufficiently up-to-date, or whether they can be updated to a more recent version (while you are at it).
2) we might need to update ft_hastoolbox, because it is now looking for bst_get, which is not exposed anymore. Perhaps it should be changed to process_fooof or so.

ftadel:
> do a quick check whether the 'old' functions are sufficiently up-to-date, or whether they can be updated to a more recent version (while you are at it).

Good idea. Done.

> we might need to update ft_hastoolbox, because it is now looking for bst_get, which is not exposed anymore. Perhaps it should be changed to process_fooof or so.

Done?

schoffelen:
merci, monsieur

PR #1919, opened 2021-11-10
ENH - working on a reported issue w.r.t. ill-defined reference for si…

schoffelen:
…mbio


schoffelen:
Some things have already been discussed with @robertoostenveld. Before this can be merged I want to update the website code, specifically the headmodel_eeg_fem tutorial, and the example script. Both are broken in their current state.

The fixes committed so far allow the simbio (and other) leadfields to be average-ref again, as intended in ft_compute_leadfield.

robertoostenveld (approved):
this looks good, as discussed. As this is not urgent, I propose that we postpone merging this until the current source/volume wrinkles have been ironed out with the next release.

PR #1927, opened 2021-11-17
Fix regressionerrors

schoffelen:
This PR addresses 2 regression errors caused by the improved source2volume conversion.

### DIFF
--- a/test/test_ft_qualitycheck.m
+++ b/test/test_ft_qualitycheck.m
@@ -4,6 +4,8 @@ function test_ft_qualitycheck
 % WALLTIME 00:30:00
 % DEPENDENCY ft_qualitycheck
 
+ft_hastoolbox('obsolete', 1);
+
 cfg           = [];
 cfg.dataset   = dccnpath('/home/common/matlab/fieldtrip/data/test/original/meg/ctf275/A0132_Aud-Obj-Recognition_20051115_02.ds'); 
 cfg.savemat   = 'no';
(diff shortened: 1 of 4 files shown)


test/test_ft_qualitycheck.m
```diff
@@ -4,6 +4,8 @@
 % WALLTIME 00:30:00
 % DEPENDENCY ft_qualitycheck
 
+ft_hastoolbox('obsolete', 1);
```
robertoostenveld, line 7: should this not be `compat/obsolete` just like in test_issue623?

utilities/ft_checkdata.m
```diff
@@ -1456,6 +1456,10 @@
     % reshape the first dimension
     tmp = getsubfield(data, fn{k});
     data.(fn{k}) = reshape(tmp, [data.dim dimsiz(2:end)]);
+  elseif startsWith(dimord, '{pos}')
```
robertoostenveld, line 1459: should we then also plan for `{pos_pos}`, which would be a 6D cell array?

robertoostenveld: btw, reshaping the cell-array should indeed be fine

test/test_ft_qualitycheck.m
schoffelen: yes, that's right. I was too rash

utilities/ft_checkdata.m
schoffelen: We could build in support for '{pos_pos}', let's see what can of worms will be opened

schoffelen: OK, I think it's done.

schoffelen:
I am curious whether the daily test batch runs through, I hope that now only test_data2bids will continue to fail...

PR #1928, opened 2021-11-18
ENH - added test case for union selection on raw data (issue 1926)

schoffelen:
Added test for issue 1926, now still causing an error

### DIFF
--- a/test/test_ft_selectdata.m
+++ b/test/test_ft_selectdata.m
@@ -315,6 +315,10 @@ cfg.select = 'intersect';
 datasel1 = ft_selectdata(cfg, data);
 
 cfg.select = 'union';
-datasel2 = ft_selectdata(cfg, data);
-
-
+ok = true;
+try
+  datasel2 = ft_selectdata(cfg, data);
+catch
+  ok = false;
+end
+assert(~ok, 'ft_selectdata with raw input and union selection should fail');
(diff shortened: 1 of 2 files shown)


schoffelen:
The problem is that with a raw data structure, time axes for selection are a cell-array. This is currently not supported in the code of ft_selectdata

robertoostenveld:
hmm, this sounds more like a job for ft_redefinetrial, where you would use a `trl` to specify the selection.

schoffelen:
yes, I realized that I started this as a headless chicken, without sitting down and properly think about it. It feels a bit weird that ft_selectdata is used on a single subject raw data structure for this purpose. Although it is not really illegal

schoffelen:
> hmm, this sounds more like a job for ft_redefinetrial, where you would use a `trl` to specify the selection.

Although I am not sure whether ft_redefinetrial manages to unionize the required latency, i.e. add NaNs for missing samples

robertoostenveld:
I think it would not, but that might depend on the behaviour of ft_fetch_data (and ft_read_data, i.e., what happens if you try to read outside the available data)

schoffelen:
I think that it is fair to decide to not support cfg.select = 'union' in combination with a raw datastructure in ft_selectdata. As a matter of fact, the help in ft_selectdata already mentioned that cfg.select = 'union'/'intersect' is only for multiple input arguments, not for a single input argument with multiple cells. The latter is the use case described in the corresponding issue. 

Earlier I had a try to transparently build in support for 'union' on raw data structures in ft_selectdata, but this proved quite hard. Hence my decision to not support this, and to build in an explicit check + error in ft_selectdata.

@robertoostenveld agreed?

robertoostenveld:
agreed.

PR #1930, opened 2021-11-24
On-fly CSD computation in ft_connectivity_dwpli

stefanches7:
Here is the first draft of the proposed RAM-saving shortcut for (d)wPLI computation.
I've basically introduced a flag allowing to specify the interest in this alternative. This way, the function accepts fourier spectrum as input. 
I have had thoughts about integrating a similar way of computation in different other measurements (e.g. PLI), as well as doing a more compliant changes. So I'm glad if you suggest a way to do so. 
The whole pipeline (starting with `ft_connectivityanalysis`) was tested with the data of "rpttap_chan_freq" dimension. I am happy to come up with a dataset and add a test function :)

### DIFF
--- a/connectivity/ft_connectivity_wpli.m
+++ b/connectivity/ft_connectivity_wpli.m
@@ -14,6 +14,10 @@ function [wpli, v, n] = ft_connectivity_wpli(input, varargin)
 %   Repetitions x Channel x Channel (x Frequency) (x Time)
 % or
 %   Repetitions x Channelcombination (x Frequency) (x Time)
+% or
+%   Repetitions x Channel (x Frequency) (x Time) 
+% inside a "freq" datatype structure with repetition tapers information for
+% on-fly csd computation.
 %
 % The first dimension should contain repetitions and should not contain an
 % average already. Also, it should not consist of leave-one-out averages.
@@ -26,6 +30,8 @@ function [wpli, v, n] = ft_connectivity_wpli(input, varargin)
 %   'dojack'    = boolean, compute a variance estimate based on leave-one-out
 %   'debias'    = boolean, compute debiased wpli or not
 %   'feedback'  = 'none', 'text', 'textbar', 'dial', 'etf', 'gui' type of feedback showing progress of computation, see FT_PROGRESS
+%   'onflycsd'  = boolean, compute CSD on fly (saves memory when many
+%   trials)
 %
 % See also CONNECTIVITY, FT_CONNECTIVITYANALYSIS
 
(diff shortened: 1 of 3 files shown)


schoffelen:
Hi Stefan, sorry that I only get to this now, and thanks for your contributions so far!

If you are confident that the on-the-fly computation from a fourierspctrm works good enough, I'd like to suggest to not make this an option in ft_connectivityanalysis. The reason for this is that we'd like to keep the number of high level options manageable and clean. So, I was wondering whether it would be possible to tell the high-level code that both a crsspctrm (if the user chooses to submit a data structure with crsspctrm already computed) and fourierspctrm (if a user submits the univariate fourier data) are allowed for the wpli computation, and that the lower-level function deals with both cases.

Let me know what you think

stefanches7:
@schoffelen yes, that sounds even more convenient. However, ft_connectivityanalysis.m:252 reads `inparam='crsspctrm'`, which fixates the input to cross-spectrum. I can imagine putting an else-if clause here to allow for singular spectrum input. 

Autodetection in the lower-level function, however, becomes challenging, as we have optional time and frequency dimensions that can either be or be not present. Currently, channel combinations in single dimensions are allowed too. So the only way out in the lower dim function that I see is to introduce the "CSD-PSD" switch parameter. 

You are welcome to let me know what you think about these solutions!

schoffelen:
Hi Stefan,

I reviewed your suggested code changes in some more detail, and I noticed some severe problems with it. Therefore, I overhauled the code that you submitted for the PR, with the intention to stick to the intended functionality. I think that it now works more generically. Could you please have a look, and perhaps test it a bit yourself? I took out the need to define cfg.onflycsd as a separate option in ft_connectivityanalysis. Now the function checks the input for the presence of a fourierspctrm and always passes this one on to ft_connectivity_wpli, flagging this function to do the on-the-fly csd computation.

Your initial implementation only worked if a single frequency bin was in the input data.
Also, what was not good, is that the lower-level function ft_connectivity_wpli received a data structure in case you wanted to compute the CSD on-the-fly. The idea is to keep the input a bit more consitent, and not change a matrix into a structure depending on the options.

What is still missing from this PR is a piece of test code that demonstrates ideally the equality between the output of 2 calls to ft_connectivityanalysis (with cfg.method = 'wpli', and/or 'wpli_debiased'):

1) a freq-structure that contains freq.fourierspctrm of dimord 'rpttap_chan_freq';
and
2) a freq-structure that has been created by calling ft_checkdata(freq, 'cmbstyle', 'full') on the freq structure used in 1).

One place where this could be added is at the bottom of test/test_ft_connectivityanalysis.m

stefanches7:
Sorry for the late response @schoffelen. The changes look good. I'm glad this functionality is a part of Fieldtrip now, and thank you for the credit!

schoffelen:
credit is well deserved. Thanks for your contribution!

PR #1933, opened 2021-12-01
prevent crash when having read in all samples  & decimal separator locale issue for sample time

Horschig:
The latter actually caused the first, because the sample reading function assumed a sample rate 1/2s instead of 1/0.2s (because str2double converts "0,2" to 2 instead of 0.2). This is fixed locally now.


robertoostenveld (approved):
I cannot review the content, but believe that it might address a recent issue mentioned in the NIRS hackathon by @TommasoGhilardi

PR #1936, opened 2021-12-06
Chan Selection Case for QZFM Gen2

neurofractal:
This modification allows the user to select channels for data from QZFM Generation 2 OPMs.

MEG, MEGREF are supported, plus MEGTAN and MEGRAD (corresponding to sensor's sensitive axis).

For context, sensors are not consistently named (i.e. it is not possible to separate mags on the head from reference mags, by name)

### DIFF
--- a/utilities/ft_channelselection.m
+++ b/utilities/ft_channelselection.m
@@ -363,7 +363,7 @@ switch senstype
     % all itab MEG channels start with MAG
     labelmeg = datachannel(strncmp('MAG', datachannel, length('MAG')));
 
-  case{'QZFM_Gen2'}        
+  case{'qzfm_gen2'}        
       % This is for use with QZFM_Gen2 Optically Pumped Magnetometers
       % manufactured by QuSpin Inc.
       % SPECS: https://quspin.com/qzfm-gen-2-update/
--- a/utilities/ft_channelselection.m
+++ b/utilities/ft_channelselection.m
@@ -362,7 +362,23 @@ switch senstype
   case {'itab153' 'itab28' 'itab28_old'}
     % all itab MEG channels start with MAG
     labelmeg = datachannel(strncmp('MAG', datachannel, length('MAG')));
-    
+
+  case{'QZFM_Gen2'}        
+      % This is for use with QZFM_Gen2 Optically Pumped Magnetometers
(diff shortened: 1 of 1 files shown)


robertoostenveld (commented):
In general I think this is a good suggestion, except that I am not so happy with the label `QZFM_Gen2` for the senstype. For all other types we use lowercase strings so `qzfm_gen2` would align better (also because a lot of code does case-sensitive string comparisons). 

But perhaps some more general considerations need to be made: the senstype (or grad.type) currently refers to 1) a specific arrangement of 2) specific sensors. E.g. the sensors in ctf151 are not really different from those in ctf275, but there are more of them. And in bti248_grad there are just as many sensors, placed identically, but the sensors themselves are different. The specification `qzfm_gen2` puts the focus is on the sensor, not on the arrangement. Should the arrangement, i.e. characteristics of the 3D-printed helmet also be represented?

neurofractal:
Hi Robert, and thanks for the speedy response :) I have changed it to lower case.

RE: second point, I would suggest not specifying anything about the arrangement at the moment. The reason for this is that QuSpin currently sell OPMs on a sensor-by-sensor basis. The user can therefore arrange sensors in flexible arrays, which often change from experiment to experiment, to target specific brain regions. 

When OPM systems are commercially available (e.g. Cerca?) AND the arrays do not change shape (i.e. held in a generic helmet), we can think about implementing this separately. 

I see 'qzfm_gen2' as an edge-case for labs using QuSpin OPMs, arranged in changeable sensor arrays.

robertoostenveld:
agreed. At the moment I don't see use-cases where FieldTrip code would need to know the OPM sensor arrangement.

For the future it would be good to consider how to represent information about the placement, e.g. in BIDS rather than specifically in FieldTrip. I am thinking in the direction of `EEGPlacementScheme` and `HeadCircumference` according to this [sidecar](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#sidecar-json-_eegjson). I can imagine it being relevant to document the specific 3D-printed helmet that is used. This would (or could) overlap with actual 3D magnetometer positions+orientations, but still might be of interest (e.g. due to helmet-specific confounding effects on the recordings).

PR #1938, opened 2021-12-07
Make ft_componentanalysis more robust by determining scaling factor based on first non-all-NaN trial

johalgermissen:
Make ft_componentanalysis more robust by determining scaling factor based on first trial that is not all NaN (which might happen if ft_componentanalysis is run based on data together with an existing unmixing matrix & all data in the first trial has been rejected with ft_rejectartifact).


robertoostenveld (approved):

robertoostenveld:
thanks!

PR #1944, opened 2022-01-05
ENH - attempt to speed things up a bit

schoffelen:
executing ft_defaults upon matlab startup takes a relatively long time (on the order of 8 seconds on matlab >2020b on linux). A substantial amount of time is spent in myaddpath (subfunction of ft_hastoolbox). Things can be sped up substantially with the current PR, where ft_hastoolbox allows for a cell-array of inputs, requiring fewer calls (eventually) to the low level matlab path function.

Of course this falls in the category 'prutsen in de marge', but might improve the overall user experience.

### DIFF
--- a/ft_defaults.m
+++ b/ft_defaults.m
@@ -244,127 +244,73 @@ if ~isdeployed
     end
   end
   
+  % this directory contains various functions that were obtained from elsewere, e.g. MATLAB file exchange
+  addtopath = {'fileexchange'};
+
+  % these directories deal with compatibility with older MATLAB versions and with Octave
+  if ft_platform_supports('matlabversion', -inf, '2008a'), addtopath{end+1} = 'compat/matlablt2008b'; end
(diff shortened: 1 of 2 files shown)


ft_defaults.m
```diff
@@ -244,127 +244,73 @@
...
-    % this contains test scripts
-    ft_hastoolbox('test', 1, 1);
+    ft_hastoolbox(addtopath, 3, 1);
   end
   
-  try
-    % this contains the functions for spike and spike-field analysis
-    ft_hastoolbox('contrib/spike', 1, 1);
-  end
+  addtopath = {'template/layout'     'template/anatomy' 'template/headmodel' 'template/electrode' 'template/neighbours' 'template/sourcemodel'}; % these contains template layouts, neighbour structures, MRIs and cortical meshes
```
robertoostenveld, line 292: I shortly worried about the fieldtrip-lite zip file on the ftp, which excludes the template files. But I checked: the directories are still there, so that should not pose a problem.

PR #1945, opened 2022-01-11
ENH - update of external/signal and started test function

schoffelen:


robertoostenveld (approved):
I looked at all the code, and it looks ok to me. I did not run the actual tests, but that is something the dashboard will take care of.

PR #1950, opened 2022-01-13
create test functions for examples on website

schoffelen:


schoffelen:
there were some glitches with respect to the newly added functions. Most of these have been ironed out. 

There is one remaining, and that pertains to the fact that data2bids uses the function writematrix, which is only available as of version 2019 and younger. @robertoostenveld shall I address this before merging this PR? 

I am a bit reluctant w.r.t. a full-blown compat drop in, because we only use limited functionality of writematrix. So I'd rather fix it locally.

schoffelen:
OK, right now I decided to go ahead, and wrote a compat/matlablt2019a version of writematrix (which can only write text files, as intended in data2bids). @robertoostenveld shall we proceed and merge, knowing well that the compatibility version of writematrix does not write xslx files etc? The alternative could be to downgrade the call to writematrix in data2bids to a call to dlmwrite, yet dlmwrite is disrecommended by mathworks

robertoostenveld (approved):
I made some small changes to the help of write matrix, other than that it all seems fine to me.

PR #1955, opened 2022-01-24
ENH keep list of badchannels out of ft_rejectvisual

dnacombo:
Coming out of ft_rejectivisual, need to keep track of selected bad channels even if keepchannel = 'yes'
This is a suggested way.

### DIFF
--- a/ft_rejectvisual.m
+++ b/ft_rejectvisual.m
@@ -368,6 +368,8 @@ cfg.channel = data.label(chansel);
 tmpcfg = [];
 if strcmp(cfg.keepchannel, 'no')
   tmpcfg.channel = cfg.channel;
+elseif strcmp(cfg.keepchannel, 'yes')
+  cfg.badchannel = data.label(~chansel);
 end
 if strcmp(cfg.keeptrial, 'no')
   tmpcfg.trials = cfg.trials;


robertoostenveld:
Thanks for the suggestion. It is not the first time that we touch upon this, i.e. how to keep track of the rejected trials/channels. An existing discussion is bugzilla [bug 1304](http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=1304) and even the ancient [bug 21](http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=21) relates to this. I am sure that it also has been discussed in github issues; I will try to find those as well and read them carefully so that we don't have to rediscover previously valued arguments and considerations.

robertoostenveld:
I now recall that previous confusion in ft_rejectvisual mainly had to do with whether 1) the input cfg.channel and cfg.trial were used to select channels/trials in advance that were bad, or 2) whether those options were used to specify trials/channels that should not be considered for the rejection.  The implementation is (1).

The `dataout.cfg.badchannel` output from `ft_rejectvisual` as suggested in this PR is consistent with the output of `ft_badchannel`. Furthermore, it is consistent with us already giving bad segments as output. Therefore I consider it a good change. 

I reorganized the code a bit, added an "inspect" script to try out various options interactively, and also made sure that `dataout.cfg.badchannel` is always specified regardless of the cfg.keepchannel option.

PR #1958, opened 2022-01-31
add space entity

JonathanVHoecke:
Dear Fieldtrip developers

I would like to propose to adapt the code, so it includes the space entity for coordinatesystem.tsv and electrodes.tsv
as stated in BIDS 1.6.0. https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#space

Thank you very much in advance
on behalf of the ICN lab, Charité Berlin

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -63,7 +63,6 @@ function cfg = data2bids(cfg, varargin)
 %   cfg.echo                    = string
 %   cfg.proc                    = string
 %   cfg.desc                    = string
-%   cfg.space                   = string
 %
 % When specifying the output directory in cfg.bidsroot, you can also specify
 % additional information to be added as extra columns in the participants.tsv and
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Hi Jonathan, 

Thanks for your PR, tt is a good idea to add this. 

By chance I added the `desc` entity last week, so I very recently looked at this. I am not 100% sure whether your implementation makes full sense, but I'll do some edits directly to your code and let you check.

best
Robert

robertoostenveld:
sorry, my commit comment was truncated early. I meant to write "only add `space-<label>` for some files".

There are some files that would have the `space` entity, but others would not. I am not 100% sure whether I did it right now, but looking at https://bids-specification.readthedocs.io/en/stable/99-appendices/09-entities.html#space it is for EEG/MEG/iEEG and for MRI data. Note that NIRS is upcoming with BEP030 and will be very similar to EEG, so I'll treat that here similarly.

For EEG, iEEG and NIRS the data itself does not change, it is only `electrodes.tsv` and `optodes.tsv` that changes. For MEG the positions are stored in the MEG data and also for MRI the voxel positions are coded in the nii file, so for those two the datafile would have the `space` label.  

Does that make sense?

robertoostenveld:
Oh, and for eeg/ieeg/nirs the coordsystem.json file would complement the electrodes/optodes files, so that is why coordsystem also gets `space`. And I also kept it for motion data (which is BEP029).

JonathanVHoecke:
Dear Robert

Thank you very much for your quick reply.
Your changes are much appreciated.
I see you made and extra if/else for make sure that the space entity is added to specific modalities.
I thank that completely covers the cases.

I test your new implementation and it works perfectly.
Thank you very much
Best wishes

Jonathan
PS: I am about to write an issue about the slash forward and backward, I ll come back to that soon.

robertoostenveld:
You seem to have closed the PR prior to me merging it, so it is not part of FieldTrip yet. Also, I see that you are developing on your `master` branch, which is not recommended. The easy solution for me is to simply copy your changes over, commit and merge them. The better solution would be for you to reimplement them on a clean branch, making sure that your master remains in sync with the fieldtrip:master.

JonathanVHoecke:
Dear @robertoostenveld 
Thank you very much for your reply and your efforts to improve the code.
Concerning the branching, I'll check again what happened. Thank you for your feedback!
Best wishes
Jonathan

robertoostenveld:
thanks! Let me know if you identify further issues with data2bids.

Issue #1957, opened 2022-01-28
data2bids scans.tsv unix support


richardkoehler:
**Describe the bug**
Using data2bids on a windows system writes out "filenames" in the scans.tsv files with a backslash, e.g.:
> filename
> ieeg\sub-001_ses-01_task-Rest_run-01_ieeg.vhdr

This can lead to path issues when sharing data and reading in scans.tsv in a unix system. To my knowledge, the forward slash "/" works on both windows and unix systems, so we could simply always write out filenames with a forward slash, e.g.:
> filename
> ieeg/sub-001_ses-01_task-Rest_run-01_ieeg.vhdr

I replaced the following statement in line 2034:
`  this.filename = {fullfile(datatype2dirname(cfg.datatype), [f x])};`

with:
`  path = fullfile(datatype2dirname(cfg.datatype), [f x]);
  path(strfind(path,'\')) = '/';
  this.filename = {path};`

I apologize in advance because I am everything but a matlab expert, but this solution seems to work for me.
Thanks in advance for your input!

**Environment and versions (please complete the following information):**
 - Operating System: Windows 11
 - MATLAB version R2020b
 - FieldTrip version  "most recent master branch from github"

robertoostenveld:
Some background: the `fullfile` function appends the pieces that comprise the full path and filename and should - depending on the platform - insert the right slash in between (either `/' or '\'). There is also the `filesep` function that can be used in MATLAB. 

Your solution works, but I would rather get to the source of the problem and solve it there. Can you try to find specifically where the wrong slash comes from? Could it be that in `cfg.outputfile` you have specified the wrong slashes yourself?

richardkoehler:
Thanks for the quick reply!

Maybe I wasn't precise enough about my issue. So the `data2bids` (and `fullfile`) function does exactly what is supposed to: I work on a windows computer, so it inserts a `\` into the `filename` of the `scans.tsv`.
However, if I now share the dataset with someone using a unix-based system, they might run into an error when reading the `scans.tsv`, as unix only supports `/` and not `\`.

I actually didn't use `cfg.outputfile`, I specified the directory in `cfg.bidsroot`. I tried switching it from 'C:\Users\..\BIDS_rawdata' to 'C:/Users/../BIDS_rawdata', but that didn't solve the problem. I assume `fullfile` automatically converts the `/` to `\`, since I am on windows. 
Is there anything else I could try?

robertoostenveld:
[...]

PR #1959, opened 2022-01-31
change the backward slash to forward slash

JonathanVHoecke:
Dear Fieldtrip developers
Dear Robert

I noticed that the data2bids function always write backslashes. Working on windows 10 pro, with Matlab R2020b, it completely makes sense to use backslash, as this is what the operating system uses. However, the BIDS specification seems to require forward slashes when internally referring to its own files.

We made an improvement in how we would resolve this of the scans.tsv, but there are more places in which backslash should be converted to a forward slash, that I unfortunately am not sure where one could achieve this in the data2bids.m code. I documented below some of those possible cases (that do not pertain to our current dataset) for your information below.

Thank you very much for your help
Best wishes
Jonathan - on behalf of the ICN Lab Charité Berlin

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/02-magnetoencephalography.html#coordinate-system-json-_coordsystemjson

image

https://bids-specification.readthedocs.io/en/stable/99-appendices/03-hed.html

image

https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/01-magnetic-resonance-imaging-data.html#using-intendedfor-metadata

image

Example of an error in the scans.tsv because of the backslashes:

image

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -2045,6 +2045,9 @@ if ~isempty(cfg.bidsroot)
   this = table();
   [p, f, x] = fileparts(cfg.outputfile);
   this.filename = {fullfile(datatype2dirname(cfg.datatype), [f x])};
+  path = fullfile(datatype2dirname(cfg.datatype), [f x]);
+  path(strfind(path,'\')) = '/';
+  this.filename = {path};
   fn = fieldnames(cfg.scans);
   for i=1:numel(fn)
     % write [] as 'n/a'
--- a/data2bids.m
+++ b/data2bids.m
@@ -63,7 +63,6 @@ function cfg = data2bids(cfg, varargin)
 %   cfg.echo                    = string
 %   cfg.proc                    = string
 %   cfg.desc                    = string
-%   cfg.space                   = string
 %
 % When specifying the output directory in cfg.bidsroot, you can also specify
 % additional information to be added as extra columns in the participants.tsv and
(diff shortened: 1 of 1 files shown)


richardkoehler:
@JonathanVHoecke 
Just fyi, I opened an issue concerning this a couple of days ago and there seems to be a consensus that we should move forward with this: #1957
However, I think it would be even better if you created a separate branch for each PR, as this will help you to cleanly separate the different changes you are making. Also, now the changes from your other PR are getting mixed up with this one, as they are all in your "master" branch (see above: Commits).
Maybe you could close this PR, create a new branch called "fix_channels_slash" or something similar, and simply create a new PR from the new branch. Then, in the PR you can simply cross-reference to the issue I posted above, so that the issue may be automatically closed when your PR is merged.

robertoostenveld:
So I have added the space entity to the data2bids code with #1962 and I have updated the channels_tsv column order with #1961. What remains is to implement the replacement of backward into forward slashes for all filenames that end up in json or in tsv files.

robertoostenveld:
I have just now merged a change to the code that replaces the backward into forward slashes for the filenames in scans_tsv. 

There might be other places where this is needed (e.g. in `IntendedFor` in the coordsys.json), but that is not a field which is explicitly constructed by `data2bids`. Therefore, I am keeping that as a responsibility for the user when specifying the `cfg.coordsystem.IntendedFor` option.

PR #1960, opened 2022-01-31
adapt the channels_tsv order

JonathanVHoecke:
Dear Fieldtrip developers
Dear Robert

For the channels.tsv there is a specific order in which the columns need to be written down, apparently.
For EEG and MEG, this is
name type units
For IEEG this is
name type units low_cutoff high_cutoff

Here is an error of the BIDS Validator, and the issue I wrote earlier today for the Validator was declined:
image

https://github.com/bids-standard/bids-validator/issues/1410

In the improvement I made, I was not able to specifically call in your code the modality,
(if modality = ieeg, then ... elif modality = meg, then ...), I just assume that the column names need to be reordered as such by default, and my piece of code takes care of that.

I really appreciate your efforts,

Best wishes
Jonathan - ICN Lab Charité Berlin

specify the order of the columns for the tsv file https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#channels-description-_channelstsv

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1364,6 +1364,13 @@ if need_channels_tsv
   channels_tsv = hdr2table(hdr);
   channels_tsv = merge_table(channels_tsv, cfg.channels, 'name');
   
+  channels_tsv = movevars(channels_tsv,{'type'},'After',{'name'});
+  channels_tsv = movevars(channels_tsv,{'units'},'After',{'type'});   
+  if ismember({'low_cutoff'},channels_tsv.Properties.VariableNames) %this required is for modality ieeg
+     channels_tsv = movevars(channels_tsv,{'low_cutoff'},'After',{'units'});
+     channels_tsv = movevars(channels_tsv,{'high_cutoff'},'After',{'low_cutoff'});
+  end
+  
   % the default for cfg.channels consists of one row where all values are nan, this needs to be removed
   keep = false(size(channels_tsv.name));
   for i=1:numel(channels_tsv.name)


robertoostenveld:
Good suggestion. But given that the user can specify additional columns in `cfg.channels`, would it not be better to explicitly move the columns `before` all other columns? By moving the known columns `after` each other, you have the risk that the first column happens to be and remains a user-specified one.

robertoostenveld:
Using the 'before` option might not be so clean, but I think this would do it

```
% some columns are required to come first, and in a specific order
first = intersect({'name', 'type', 'units', 'low_cutoff', 'high_cutoff'}, channels_tsv.Properties.VariableNames, 'stable');
subsequent = setdiff(channels_tsv.Properties.VariableNames, first, 'stable');
% determine how the columns should be ordered
[dum, order] = ismember([first subsequent], channels_tsv.Properties.VariableNames);
channels_tsv = channels_tsv(:, order);
```

richardkoehler:
@robertoostenveld
Just wanted to let you know that there actually is an ongoing discussion about this here: https://github.com/bids-standard/bids-validator/issues/1410
Currently, there is no actual statement in the BIDS specification that iEEG channels need to be in a specific order. Of course, it does seem to make sense to always place the required columns before optional columns, which would increase readability between datasets.

robertoostenveld:
I made some additional changes in a separate branch and merged that just now

PR #1946, opened 2022-01-12
ENH: rejectvisual_summary: recall toggled channels/trials in a click

dnacombo:
UI adaptation in order not to have to retype the list of channels/trials to toggle if I want to reset or change selection based on previous ones.

A click on the list of previously selected channels/trials will push that list to the toggle editbox.

PR #1966, opened 2022-02-02
recall toggled channels/trials in a click

robertoostenveld:
This fixes #1946


robertoostenveld:
I tested the interaction and it works smoothly. I can also see the utility of this, hence I will merge. 

However, I do think that it is unintuitive and few people will find it. But that comment applies more general to the whole lower-right quarter of that figure. The selection of metrics would be better implemented as a drop-down list rather than a long list of radio buttons. Reporting (on screen) and/or manipulating the selected channels/trials could probably also have a more intuitive GUI. The fancy log at the bottom of the window could be made optional: it makes sense for a full screen window, but would also make sense in other interactive GUIs and would better be integrated with ft_notice or so (i.e. more fluent between command window and the GUI log window). But these are all things that don't have priority now, so I won't go into them.

Issue #1968, opened 2022-02-03
multitrial ft_databrowser


dnacombo:
**Reviewing multitrial data can be cumbersome**

In cognitive neuroscience experiments, reviewing all data "by hand" is common practice.
For **trial-based experiments**, reviewing single trials is difficult with the current `ft_databrowser`.
In the current `ft_databrowser`, two modes exist for this: continuous, and not continuous.

When `cfg.continuous == 'yes'` the data is shown in the original time frame, with gaps between trials:

![image](https://user-images.githubusercontent.com/4716235/152345687-4f9cc024-89dc-47bd-9aad-6651c7eb6a4a.png)

When `cfg.continuous == 'no'`, one trial only is shown:

![image](https://user-images.githubusercontent.com/4716235/152345926-95ea831b-6919-458b-9691-abfc7557b629.png)

Reviewing hundreds of trials can be really a pain because of this, and because drawing can be slow.

**Alternative solutions**
Displaying more trials on one screen is doable :

1. Tweaking the not continuous mode seems like a real challenge because the single trial data is at the heart of the plotting engine.
2. Tweaking the continuous mode is a workaround which could work, but requires temporarily adapting all time stamps in the dataset.

I went for 2. in this #PR.

**Drawbacks**

- The displayed dataset is shown on a continuous time scale, which does not correspond to trial-based time.
- Added artifact windows may span several trials.
- Consider that sampleinfo is increasing and non overlapping
- Had to comment out ft_warnings in `ft_plot_vector` and `ft_plot_text` probably not acceptable, but a strong slow-down

**Process**

- I wrote a wrapper for `ft_databrowser` called `ft_databrowser_multitrial`
- Added 1 option: `cfg.ntrialswin`: the number of trials on the first screen
- Compute `chunksout` of time to be removed from `sampleinfo`
- First shift all event time samples to account for `chunksout`
- Add blank boundary events at first sample of each trial
- Shift artifacts

dnacombo:
This is how it looks now:

![image](https://user-images.githubusercontent.com/4716235/152355732-42ecb8e7-e38f-4dbd-b45d-0dc4b921e495.png)

robertoostenveld:
hmm, I understand the desire for the functionality, but am quite worried about the side effects that would be created like this. E.g. when people make and mark selections for artifacts, or make selections and right-click to execute a "browsefun".

[...]

PR #1969, opened 2022-02-03
Ft databrowser multitrial

dnacombo:
#1968


robertoostenveld:
my commit just now incorporates your change in 1aa24f07e65020111b144f89a88ef15586dcecdf to speed things up. Other than that, I will not adopt anything from this PR and close it. As discussed in #1968 we will aim for a solution in the documentation rather than in the code.

PR #1970, opened 2022-02-03
Add a viewmode button to ft_databrowser

dnacombo:
Showing a 'viewmode' button in the bottom right corner of data browser to switch between `butterfly` and `vertical` modes.

**_Warning:_** not sure what happens in `component` mode.

### DIFF
--- a/ft_databrowser.m
+++ b/ft_databrowser.m
@@ -775,7 +775,10 @@ else
   
   % identify button - to find label of nearest channel to datapoint
   uicontrol('parent', h, 'units', 'normalized', 'style', 'pushbutton', 'string', 'identify', 'userdata', 'i', 'position', [0.91, 0.83, 0.08, 0.04], 'callback', @keyboard_cb)
-  
+
+  % viewmode button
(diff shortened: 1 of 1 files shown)


robertoostenveld:
are you sure that this always works? There are a few pieces of code prior to `redraw_cb` conditional on cfg.viewmode that will not be re-evaluated when you press the button or hit `m`. It might be that those only apply to viewmode=component, which you are explicitly avoiding here. 

I think that https://github.com/fieldtrip/fieldtrip/blob/1fa0f3ef9f5fe4fbcb902de94740e9697b30587f/ft_databrowser.m#L730
is problematic, as the plotting-upon-resize depends on the initial viewmode. But that can be solved, e.g. by always setting `ReSizeFcn` to `resize_cb`  and having `resize_cb` immediately return in case viewmode=butterfly. 

I will give it a try and see whether some code needs to be moved/updated.

robertoostenveld:
trying it out, I notice that the automatic `ylim` is inconsistent between viewmodes; not due to this PR, this was already the case.

robertoostenveld:
when switching between butterfly and vertical, the color for the channels should remain identical

robertoostenveld:
thanks Max, I updated it a bit and merged.

PR #1975, opened 2022-02-14
Update ft_megplanar.m

johalgermissen:
`fourierspctrm` is the name of the respective field in the output of ft_freqanalysis; but the option in the input is called `output='fourier'`


robertoostenveld:
For reference, this follow up on https://github.com/fieldtrip/website/pull/518

Thanks. It will take a night for the documentation to be updated (automatically) on the website.

PR #1978, opened 2022-02-15
make a function that returns the affine coordinate transformations

### LINKED ISSUE #1976: make a function that returns the affine coordinate transformations, like ras2als and ctf2acpc
Is there a conceivable objection against adding a second optional output to ft_convert_coordsys, returning the (affine-) transformation (if method==0), so that it can be re-used outside de function?

Context: now that this function can be very easily applied to switch between generic coordinate systems (e.g. 'lip' 2 'ras' and vice versa), I would find it useful to lay my hands on the corresponding transformation matrices...

What do you think, @robertoostenveld ?

robertoostenveld:
this fixes #1976

### DIFF
--- a/utilities/ft_affinecoordinates.m
+++ b/utilities/ft_affinecoordinates.m
@@ -101,7 +101,7 @@ for i=1:length(generic)
   for j=1:length(generic)
     xxx = generic{i};
     yyy = generic{j};
-    eval(sprintf('%s2%s = transform_generic(''%s'', ''%s'');', yyy, xxx, xxx, yyy));
+    eval(sprintf('%s2%s = transform_generic(''%s'', ''%s'');', xxx, yyy, xxx, yyy));
   end
 end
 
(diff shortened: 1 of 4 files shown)


robertoostenveld:
how about this @schoffelen ?

utilities/ft_affinecoordinates.m
```diff
@@ -0,0 +1,288 @@
...
+  0.0000    0.0000    1.0000    0.0000
+  0.0000    0.0000    0.0000    1.0000
+  ];
+
+% these are all combinations of 90 degree rotations and/or flips along one of the axes
+for i=1:length(generic)
+  for j=1:length(generic)
+    xxx = generic{i};
+    yyy = generic{j};
+    eval(sprintf('%s2%s = transform_generic(''%s'', ''%s'');', yyy, xxx, xxx, yyy));
```
schoffelen, line 104: I don't fully understand the order: yyy,xxx,xxx,yyy (i.e. the swap for the second pair). This suggests that what is known in transform_generic as 'from' should actually be 'to', or vice versa. Or am I missing something? 

Note that the code behaves properly, so it's just probably some confusion on my end

robertoostenveld: yes, you seem to be right about the to/from swap in `transform_generic`. Let me fix it.

PR #1979, opened 2022-02-15
Double vertices

schoffelen:


schoffelen:
Hopefully this really nails it. The previous commit was faulty, due to an incorrect re-indexing of the triangles.

PR #1982, opened 2022-02-16
ENH - replaced radiobuttons with popup/dropdown menu

schoffelen:


schoffelen:
Just some window dressing. I think it works fine, but I asked for your review in case you'd like to change something in the figure window's real estate.

I decreased the size of the output log a bit, and changed the distribution of the individual ui-items that populate the right lower quadrant of the figure

schoffelen:
I tested on matlab2014b as well, and all seems in order. Backward compatibility seems OK

Issue #1952, opened 2022-01-18
Include the possibility to apply signal space projection vectors to Neuromag data


ualsbombe:
**Is your feature request related to a problem? Please describe.**
Data from neuromag MEG systems have signal space projection (SSP) vectors. These can be applied to data to project out stable noise patterns from the environment

**Describe the solution you'd like**
It would be good if the SSP vectors could be projected our from the data set. Maybe something similar to _ft_denoise_pca_ could be used to facilitate it.

**Additional context**
SSP should not be done after maxfiltering

```%% reading neuromag data 

restoredefaultpath
addpath /home/lau/matlab/fieldtrip/
ft_defaults

path = '/home/lau/mne_data/MNE-sample-data/MEG/sample';
filename = 'sample_audvis_raw.fif'; %% any non-maxfiltered neuromag data will do

hdr = ft_read_header(fullfile(path, filename));

%%
hdr.orig.projs %% has the projections and the data associated with it
```

schoffelen:
That's an awesome idea. Not new, still awesome :).

http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2109

ualsbombe:
Haha :)

schoffelen:
Would you like having a first shot at it?

ualsbombe:
To be honest - not at this moment - I promised to post it for someone who could make use of it.
If it also doesn't work for you now, I may be able to pick it up after teaching finishes in May (long term,I know)

schoffelen:
Tell this person to sign up on github, and do it him/her/etc self

ualsbombe:
Afraid they don't have the skills yet - I'll check how much they need it and return

schoffelen:
For inspiration: fieldtrip/external/mne/mne_make_projector creates a projection matrix and returns the spatial topographies of the ssp's. For consideration, one could also hand craft a comp - like structure (i.e. one that contains topo/mixing etc) that can be used in ft_rejectcomponent. 

Alternatively, one could indeed use the projection matrix right away in ft_denoise_ssp. One thing to consider is to first visualize the topographies, because it is not clear to me whether they should always all be applied, or whether a user needs to investigate that first.

schoffelen:
[...]

PR #1984, opened 2022-02-17
implement Neuromag SSP similar to montage/balancing for CTF synthetic gradients , see issue #1952

ualsbombe:
So here's a first attempt. Check the test script - I can't plot the multiplot- maybe because the units are "unknown"

I have only run it on a timelocked so far

`denoised_timelock.grad.chanunit`


gdemarchi (changes requested):
Just s useless comment :-) Thanks for putting me in, just fix my name (line 26) as long as you are in time, I'm not yet brasilian ... GianpaOlo Demarchi, just that :-)

ft_denoise_ssp.m
```diff
@@ -0,0 +1,163 @@
...
+%   cfg.inputfile   =  ...
+%   cfg.outputfile  =  ...
+% If you specify one of these (or both) the input data will be read from a *.mat
+% file on disk and/or the output data will be written to a *.mat file. These mat
+% files should contain only a single variable, corresponding with the
+% input/output structure.
+%
+% See also FT_PREPROCESSING, FT_DENOISE_PCA
+
+% Copyright (C) 2004-2022, whose names to put, Gianpaulo Demarchi, Lau
```
gdemarchi, line 26: Gianpaolo Demarchi instead

ualsbombe:
So sorry, Gianpaolo - I try my best not to butcher names, but this one slipped. I did not want to imply that you were Brazilian ;-)

schoffelen:
Shoot, I messed it up (again) by pulling the PR into a more recent version of FT than the one from which Lau created this PR, causing more changed files than needed. I will look into this. My intention was to update the test script into a function, and made a half-baked attempt to write in some code that allows access to ftp-data from outside the Donders (@robertoostenveld this is not related to this PR per se, but might be considered to be a trick to allow running of the test functions from outside Donders). This being said, the fif-file I used from the natmeg workshop does not seem to contain SSP-projectors, so it's unusable for now. Ergo, the whole exercise on this rainy Saturday afternoon was for nothing, given that I have messed it up now. Will fix later.

Thanks Jean-Paul and Lau so far!

schoffelen:
Sorry for the mess, I created a new PR #1988 with only the 3 original files changed. I suggest to close this one, and merge the other one, once we agree that all is in order

gdemarchi:
> functions from outside Donders). This being said, the fif-file I used from the natmeg workshop does not seem to contain SSP-projectors, so it's unusable for now. Ergo, the whole exercise on this rainy Saturday afternoon was for nothing, given that I have messed it up now. Will fix later.

Do you want an ethics-less .fif file to test upon, like a small empty room measurement? We're plenty of, and this would suffice given the fact that SSP projectors (or at least the Neuromag/Elekta/MEGIN ones) are indeed built on empty room ... no brain there, but would be enough to see whether a) the function works and b) the projectors make the spectrum cleaner ...  just let me know (so I would feel more useful in this not-so-rainy Saturday of (my own) baby sitting 😄 )

PR #1986, opened 2022-02-18
Freesurfer atlas

schoffelen:


schoffelen:
This PR intends to be able to be more generic in reading Freesurfer based segmentations, defined on the volume. Previously, the list of labels and indices was hard-coded in the function, but this list was incomplete.

PR #1988, opened 2022-02-19
implement Neuromag SSP similar to montage/balancing for CTF synthetic gradients , see issue #1952 and messed up PR #1984

schoffelen:


ualsbombe:
The only issue that I had was that chanunit became unknown meaning that some plots could not be made out of the box (e.g. multiplot)

schoffelen:
> 

But were those problems specific to a situation when you are multiplotting without a defined cfg.layout?

ualsbombe:
Yes, they turned out to be - I've added support for only applying specific projectors - before it was all or none.
In terms of pushing - do I just keep pushing to my branch or should I push somewhere else given that there is a new pull request?

schoffelen:
sorry for the hassle: indeed if you want your changes to be visible here, you should push them here (or send me the file in a different way, so that I can push (ssshhhht don't tell @robertoostenveld )

ualsbombe:
So this is a first for me: how do I push to an existing pull request? When on my branch, I can create new pull requests..., but I cannot see where to push to an existing
billede

schoffelen:
Well, you'd need to pull the PR first, i.e. create a remote to my FT:

git remote add schoffelen https://github.com/schoffelen/fieldtrip

Then, when you are on the branch, do:

git pull schoffelen ssp_fieldtrip_issue_1952

In this case you probably get some conflicts in ft_denoise_ssp, which is annoying, so perhaps it's better to make a temporary copy of ft_denoise_ssp somewhere, get rid of your branch altogether, checkout the branch again (now with no changes yet, but still named ssp_fieldtrip_issue_1952), and only then pull the schoffelen one in (still following?). Then, you can place back the changed ft_denoise_ssp, commit it locally, and then push to schoffelen:

git push schoffelen ssp_fieldtrip_issue_1952. If you have write permissions to my branch (not sure, but probably not), it should end up in the PR. If it does not work, you can make a PR on the PR...

ualsbombe:
So I've done all the steps - correctly, I hope, but I indeed did not have permission to push to yours.
I guess I need to make the Pull Request somehow else - which button do I need to click?

schoffelen:
Hmmm. Can you edit the file in the github environment, and paste your changes there?

ualsbombe:
I am not allowed to edit the file - only view
billede

PR #1995, opened 2022-03-02
ENH - fix for brickwall, to be tested

schoffelen:
... before merging

### DIFF
--- a/preproc/ft_preproc_bandpassfilter.m
+++ b/preproc/ft_preproc_bandpassfilter.m
@@ -56,7 +56,8 @@ function [filt, B, A] = ft_preproc_bandpassfilter(dat, Fs, Fbp, order, type, dir
 %
 % See also PREPROC
 
-% Copyright (c) 2003-2021, Robert Oostenveld, Arjen Stolk, Andreas Widmann
+% Copyright (c) 2003-2022, Robert Oostenveld, Arjen Stolk, Andreas Widmann,
+% Jan-Mathijs Schoffelen
 %
 % This file is part of FieldTrip, see http://www.fieldtriptoolbox.org
 % for the documentation and details.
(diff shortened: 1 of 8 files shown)


schoffelen:
I updated the test functions, so that they now also create a 'diagnostic' figure, that compares the different filtered signals. I am quite surprised that there's overall quite some variability, depending on the options used, particularly for the bandstop/bandpass filters.
Yet, take home message for now is that the brickwall performs comparably to the onepass firws filter (with its default options). So I think that this sufficiently addresses the concern that was raised in the issue that is linked to this PR

cjthomas-opensource:
Thank you for the prompt fix, and for the testing!

The variability is almost certainly a Matlab thing. I've noticed that Matlab's filter-design functions will often modify the filter specifications that you ask for in ways that are unexpected, particularly for filters operating at very low frequencies or for band-pass or band-stop filters that are very narrow.

Some of this is unavoidable (if the quality factor you're asking for is incompatible with the IIR filter order or number of FIR points), but it appears to do this in other situations as well. My best guess is that it's picking default specifications for all of the optional designfilt() arguments (pass- and stop-band flatness, flank steepness, etc), and perturbing all of them during filter design instead of freezing the ones the user explicitly specified.

Long story short, testing Matlab's filters to see what they actually do is almost always a good idea.

Thanks again!

cdale8:
I'm now erroring out using this version of brickwall in ft_preproc_bandpassfilter.m on 36000 samples of resting state data when bandpassing 1 - 7.5 Hz (DeltaTheta).  

Error:
line:
fbin1 = nearest(ax, [min(Fbp)    max(Fbp)]);

Error using nearest (line 60)
The limits you selected are outside the range available in the data

dat is a 246 x 36000 matrix of timeseries of source localized (beamforming) values in 246 Brainnetome Atlas regions.
ax becomes a 1 x 36000 vector of values from 0 to 2159940 in steps of 60 (Fs is 600 Hz)
nearest min(Fbp) value is ax(1)=0   (min Fbp being 1 Hz)
nearest max(Fbp) value is also ax(1)=0 (max Fbp being 7.5 Hz)

In the previous code, it seems like the index for ax was deliberately separated for min and max Fbp, solving this issue:
     ax = linspace(0, Fs, size(dat,2));  % frequency coefficients
    fl = nearest(ax, min(Fbp))-1;       % low cut-off frequency
    fh = nearest(ax, max(Fbp))+1;       % high cut-off frequency

schoffelen:
It's not about separating the call to nearest into 2, it's about ax being wrong.

schoffelen:
PS: note that brickwall filters are not particularly good filters. There are better behaved ones around

PR #1997, opened 2022-03-07
ENH - vectorization leads to twofold speedup, performance now in line…

schoffelen:
… with signal toolbox. + documentation update. This PR was inspired by Roy Cox, who noticed that the matlab implementation is much faster than the overloaded version in external/signal. Not anymore after merging.


robertoostenveld:
Looks good to me.

PR #2000, opened 2022-03-10
FIX - improved robustness if voxels are not isotropic

schoffelen:


schoffelen:
If voxels are highly anisotropic, the determination of the order of the permutation was not sufficiently robust.

schoffelen:
Whooohooo! just claimed issue/PR number 2000

PR #2007, opened 2022-03-25
update to afni code, generic support for afni atlases

schoffelen:
This PR allows for the possibility to read more generically afni atlases.

I updated the code in external/afni with the version from github (pulled March 25,2022) afni/afni/src/matlab

I adjusted some of the functions to allow for decoding the atlas labels from the header information of the files, and filed a PR with the AFNI folks, to keep things in sync.

https://github.com/afni/afni/pull/361


schoffelen:
TBD:
- ft_read_atlas now outputs the TT atlas in double precision, rather than originally in int8

- the labels of the bricks now are prepended with left/right, which is different from before

schoffelen:
https://sscc.nimh.nih.gov/afni/doc/misc/afni_ttatlas/index_html now refers to the file that is in fieldtrip/template. This description indeed lacks L/R indication. The TTatlas from https://afni.nimh.nih.gov/pub/dist/atlases/afni_atlases_dist/ actually contains L/R indication in the labels. Also, the former file does not provide the ATLAS_LABEL_TABLE in the header, so a workaround (hard coded list) is still needed for this file

PR #2013, opened 2022-04-14
modifed ft_chantype to correct an issue where the chantype is mis-ide…

Richter2552:
…ntified as other if only sti102 is enabled, i.e. present in the .fif file. Modified read_trigger to correct a typo that causes the fix for signed integers to fail to be applied.

### DIFF
--- a/fileio/private/read_trigger.m
+++ b/fileio/private/read_trigger.m
@@ -156,7 +156,7 @@ end
 if strncmpi(dataformat, 'neuromag', 8) && ~fixneuromag
   for k = 1:size(dat,1)
     switch hdr.chantype{chanindx(1)}
-      case 'binary trigger'
+      case 'digital trigger'
         if any(dat(k,:)<0)
           dat(k,:) = double(typecast(int16(dat(k,:)), 'uint16'));
         end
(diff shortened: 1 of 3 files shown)


fileio/private/read_trigger.m
```diff
@@ -156,7 +156,7 @@
 if strncmpi(dataformat, 'neuromag', 8) && ~fixneuromag
   for k = 1:size(dat,1)
     switch hdr.chantype{chanindx(1)}
-      case 'binary trigger'
```
schoffelen, line 159: huh? it seems that 'binary trigger' is not supposed to exist at all?

schoffelen (commented):
Hi Craig, it looks good to me, it looks as if fileIO for data from 'old' systems won't be compromised by these changes. Yet: do you have any test code/data handy for documentation/testing?

Richter2552:
Hola JM,

The data I have is not great, since it is based on an experiment that only I really understand. I will make some data where I count through the 16 potential trigger lines in some way so the code can be easily tested.

I agree, it looks like ‘binary trigger’ emerged as a result of a case of the ‘whoopsies’. ;-)

How you doing? All good? I’m trying to get Sanjeev to finish his analysis, but he’s just moved to Germany for a postdoc. I’ll stay on him.

Best,

C.

schoffelen:
OK, thanks, I did not see the doc, but I don't see a reason not to merge this. Thanks!

Richter2552:
Great!!

Richter2552:
Yeah, it shouldn’t mess anything up if you merge it in now.

schoffelen:
hehe, I certainly hope not. Already merged a while ago.

PR #2014, opened 2022-04-14
FIX - fix for issue2012 catching a bug in getdimord for free

schoffelen:
This addresses an issue raised by @rensch13, who wants to create virtual channels using precomputed lcmv spatial filters, and channel level TFR data (fourierspctrm)


schoffelen:
I suggest that René reviews this

PR #2015, opened 2022-04-21
implemented fractional orders for superlets

ludovicbellier:
Currently, superlets (SLT) are implemented in Fieldtrip, based on a [2021 Nat Comm paper](https://www.nature.com/articles/s41467-020-20539-9). They are a sophistication of the classical wavelet transform, where sets of wavelets of increasing order (instead of one wavelet of a given order) are used to estimate power at each point in the time-frequency space. This allows time-frequency (TF) super-resolution and reveals fine TF details.

Adaptive superlets (aSLT) are simply the use of SLT with varying orders across frequencies (e.g., o5 at 1Hz, which means a set of 5 wavelets; ...; o15 at 150Hz). It is akin to multitapers, where one would use an increased amount of tapers with increasing frequencies to allow for more spectral smoothing. Only, in the current implementation, only integer / discrete orders are supported, which leads to "banding" (see left panel of the figure below; I simply generated applied aSLT on a dirac delta at fs=1,200Hz, with `cfg.foi = 1:150;`, `cfg.order = round(linspace(5, 15, length(cfg.foi)));`, and less importantly `cfg.width = 3;` and `cfg.combine = 'additive';`).

The same authors released [another paper in 2021 in IEEE](https://ieeexplore.ieee.org/abstract/document/9287873?casa_token=2D2pqKptGDIAAAAA:dS49r2Q8RXNImmpOT3cz2pReuVGeVaDTNpOEcalKvYw1T04X0v0vpnT-01L-I0KYzG2awSwA), enabling the use of fractional orders (e.g., o4.2), in what would then be called fractional adaptive superlets (faSLT). This effectively prevents banding (see right panel of the figure below; same SLT parameters as for aSLT, only with `cfg.order = linspace(5, 15, length(cfg.foi));`), and is a straightforward improvement of the existing implementation.

image

The purpose of this pull request is simply to provide the community with this updated implementation of faSLT.

About the details of this faSLT implementations, I compared formulas (9) and (14) of their IEEE paper (representing the superlet transform and the fractional superlet transform, respectively), and basically the difference is that for the fractional part we compute an extra wavelet in the set and put it to the power of the fractional exponent. In English, if we want an order 2.23, we would compute 3 wavelets (number of cycles depend on the base wavelet width and the 'additive' or 'multiplicative' parameter, but let's say we have a base wavelet of 5 cycles and an 'additive' set of wavelets, then 3 wavelets of this set would have 5, 6 and 7 cycles), and set the convolution of our signal with the 3rd wavelet (of 7 cycles) to the power of .23.

Looking at the figure my implementation seems to work fine, but I invite colleagues with an actual math / engineering background to double check it =)

Best,
Ludovic

### DIFF
--- a/ft_freqanalysis.m
+++ b/ft_freqanalysis.m
@@ -650,37 +650,41 @@ for itrial = 1:ntrials
       % equivalent one-liners:
       %   multiplicative: cycles = arrayfun(@(order) arrayfun(@(wl_num) cfg.width*wl_num, 1:order), cfg.order,'uni',0)
       %   additive: cycles = arrayfun(@(order) arrayfun(@(wl_num) cfg.width+wl_num-1, 1:order), cfg.order,'uni',0)
-      cycles = cell(length(cfg.foi),1);
+      order_int = ceil(cfg.order);
+      cycles = cell(length(cfg.foi), 1);
       for i_f = 1:length(cfg.foi)
-        frq_cyc = NaN(1,cfg.order(i_f));
+        frq_cyc = NaN(1, order_int(i_f));
         if strcmp(cfg.combine, 'multiplicative')
-          for i_wl = 1:cfg.order(i_f)
-            frq_cyc(i_wl) = cfg.width*i_wl;
+          for i_wl = 1:order_int(i_f)
+            frq_cyc(i_wl) = cfg.width * i_wl;
           end
         elseif strcmp(cfg.combine, 'additive')
-          for i_wl = 1:cfg.order(i_f)
-            frq_cyc(i_wl) = cfg.width+i_wl-1;
+          for i_wl = 1:order_int(i_f)
+            frq_cyc(i_wl) = cfg.width + i_wl - 1;
           end
         end
         cycles{i_f} = frq_cyc;
       end
       
       % compute superlets
-      spectrum = NaN(nchan,length(cfg.foi),length(cfg.toi));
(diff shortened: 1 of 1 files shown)


schoffelen:
The code changes look OK to me, and the figure looks convincing! At some point it would be nice to also have a little bit of documentation on the website, to demonstrate the utility of superlets in practice (just to draw some more attention to the method itself, and to the fact that it's implemented in FieldTrip). Do you have any ideas about this @ludovicbellier

ludovicbellier:
Thanks Jan-Mathijs @schoffelen! Completely agreed, as so far superlets look more like an incremental step up from wavelets. Well, two ideas:

1) On my side I plan to do a side-by-side comparison of wavelets (with adaptive order, to be fair) versus fractional adaptive SuperLeTs in a predictive modeling approach (likely something like "do decoding models trained on faSLT-estimated inputs/features perform better than ones trained on wavelet-estimated inputs?"). If I do expect a slight improvement from faSLT, it might be obscured by the fact that I average frequency bins between 70 and 150Hz to get the HFA, so whether it was spectrally fine-grained or blurred matters less in the end, which leads us to...

2) Oscillatory bursts. This is at least one use case where superlets could really shine, by enabling fine-grained delineation of transient bursts in both time and frequency over the whole spectrum. I don't have immediate plans to explore that, but if someone you know do I'd be happy to connect and if needed help with implementing faSLT (e.g., sharing my insights on the several parameters you can tune).

Best,
Ludovic

PR #2023, opened 2022-05-12
Update ft_prepare_mesh.m

TTD92:
Fix bug where a structure field was referred to but never generated.


schoffelen:
this looks like a minor change, and although the context is undocumented, let's merge.

PR #2028, opened 2022-05-16
ENH - added overloaded functionality for resampling (if signal toolbo…

schoffelen:
…x is absent)

### DIFF
--- a/external/signal/README
+++ b/external/signal/README
@@ -6,6 +6,15 @@ some functions in the Mathworks signal processing toolbox.
 Note that the hilbert and hanning functions are not from the Octave toolbox, 
 these have been taken from SPM (hilbert), or were handcrafted (hanning)
 
-Last update was performed Jan 11, 2022 by Jan-Mathijs, using version
+Updates:
+Jan 11, 2022 by Jan-Mathijs, using version
 [4fcb75] (i.e. the default branch) from 
 https://sourceforge.net/p/octave/signal/ci/default/tree/
+
+May 16, 2022 by Jan-Mathijs, added functionality for resample:
+-upsample
+-downsample
+-upfirdn
+-resample
+-firls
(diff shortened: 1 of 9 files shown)


schoffelen:
This functionality overloads the signal processing toolbox's resample function, in its default application (as used in ft_resampledata). the corresponding test fuction test_external_signal has been adjusted to check for basic compatibility. Numeric relative tolerance is ensured at a level of 0.001

robertoostenveld:
is the anti-aliasing filtering done in the same way?

schoffelen:
Yes, the anti-aliasing is the same (but different from how it is done in the octave version).

I am now making some changes to ft_resampledata (and test_ft_resampledata) to address a concern that was ventilated by Alain de Cheveigne on the discussion list, in that the default anti aliasing might be suboptimal if strong signal components close to new nyquist frequency are present.

My suggestion is to include an optional cfg.lpfilter explicitly in ft_resampledata.

robertoostenveld:
I think it will be a good idea to add an cfg.lpfilter option, but that might keep the situation with the filtering (sometimes it is default built in the low-level code, sometimes not) implicit.

How about making a FAQ or example page with some example code, that shows the effect of aliasing (using a narrow band signal above the new nyquist frequency and using broad-band white spectrum noise) and how to deal with that for each of the cfg.method options?

I think that here - and potentially in more situations - we should add the URL to the new example page in the help of the matlab function.

schoffelen:
hmmm, this leads to a significant amount of extra work...

Note BTW that the aliasing problem does not seem altogether an issue, at least not in the MATLAB 2021b version of resample. 

I already wrote some code (that generates some figures) in the test function for ft_resampledata, and will decorate this code with some text, and promote it to an FAQ

schoffelen:
https://www.fieldtriptoolbox.org/faq/resampling_lowpassfilter/ created FAQ

PR #2032, opened 2022-05-18
Fix dashboard

schoffelen:
This PR aims at fixing the failing dashboard, which was due to some overloaded functions being added to fieldtrip/external/signal. Due to slight incompatibilities between the matlab and octave-derived functions, a lot of test functions started failing. 

- upfirdn did not work yet for matrices, now it does.

- Also, there was some confusion when ft_default.toolbox.signal was set to 'compat', and some to-be-overloaded functions (fir1/fir2) were taken from matlab's signal toolbox. Therefore those functions have been added.

- the fir_filterdcpadded function was a bit too strict on equality testing for symmetric/antisymmetric. I added some leniency because the overloaded functions have some numerical roundoff differences.
 
I think that now most (if not all) should be fixed again. I ran the majority (i.e. the shortest ones) of the previously failing test functions locally, and they run through again.

### DIFF
--- a/bin/synchronize-private.sh
+++ b/bin/synchronize-private.sh
@@ -1477,6 +1477,7 @@ sync ${ARRAY[*]}
 ARRAY=()
 ARRAY+=(test/private/isalmostequal.m)
 ARRAY+=(private/isalmostequal.m)
+ARRAY+=(preproc/private/isalmostequal.m)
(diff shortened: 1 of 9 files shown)


robertoostenveld (approved):
You know best what failed and I guess that is now all resolved. It looks good to me.

PR #2033, opened 2022-05-18
Update read_EDF to fix slow EDF reading in.

Frederik-D-Weber:
The low level function was slow for any reading of EDF files beyond 2 MB when the fast .mex files of reading 16bit could handle up to 2 GB. 
The error was that it was treated as limiting to 2 MB when it should be 2 GB!
1024^2 is clearly a Mega and 1024^3 is a GB
This improves the slow reading in speed of EDFs to a high degree.
This saves a lot of people working with EDF a lot of waiting time (order of magnitude less).


schoffelen:
Thanks, well caught!

PR #2034, opened 2022-05-19
Data2bids updated to follow changes in BEP for Motion.

sjeung:
The BIDS extension proposal for motion data has gone through some changes and this PR is to bring data2bids up-to-date.  A major change has been made to the way motion.json is specified. 

Motion data will be saved in its own folder called "motion".
File names have a new entity specific for motion data, with key "tracksys".
(A tracking system in motion recording shares effective sampling rate and number of samples.)
Different tracking systems are represented as an array in motion.json file.


data2bids.m
```diff
@@ -2047,25 +2110,33 @@
     filename = fullfile(cfg.bidsroot, ['sub-' cfg.sub], ['sub-' cfg.sub '_scans.tsv']);
   end
   
+  % get filename
   this = table();
   [p, f, x] = fileparts(cfg.outputfile);
   this.filename = {fullfile(datatype2dirname(cfg.datatype), [f x])};
+  
+  % get date
+  if isfield(cfg, 'acq_time')
```
robertoostenveld, line 2119: I think that `acq_time` should be in `cfg.scans`, not at the main level

robertoostenveld:
sorry, I see that I accidentally also included some changes to `ft_colormap` that I was working on. I'll keep them in here, although they are unrelated to this PR.

robertoostenveld (approved):
Let's merge this PR. If you have further updates, please start a new branch from "master" and send a new one. Thanks for your contribution!

PR #2036, opened 2022-05-20
draft updates to motion2bids

robertoostenveld:
This is an attempt to make `data2bids` for motion data more consistent with other data types. It needs to be further discussed and aligned with bep029. 

- move all the general user-specified motion settings to the top
- do not loop over multiple TrackingSystems, make one complete "scan" (3 files) per TrackingSystem
- remove the special handling of TrackingSystems as an array of objects
- add tracksys entity to the file name for channels.tsv
- move acq_time from top level cfg to cfg.scans

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -615,7 +615,7 @@ cfg.coordsystem.FiducialsCoordinateSystemDescription            = ft_getopt(cfg.
 cfg.channels.name                       = ft_getopt(cfg.channels, 'name'                        , nan);  % REQUIRED. Channel name (e.g., MRT012, MEG023)
 cfg.channels.type                       = ft_getopt(cfg.channels, 'type'                        , nan);  % REQUIRED. Type of channel; MUST use the channel types listed below.
 cfg.channels.units                      = ft_getopt(cfg.channels, 'units'                       , nan);  % REQUIRED. Physical unit of the data values recorded by this channel in SI (see Appendix V: Units for allowed symbols).
-% specific options for EEG/MEG channels
+% specific options for EEG/MEG/iEEG channels
 cfg.channels.sampling_frequency         = ft_getopt(cfg.channels, 'sampling_frequency'          , nan);  % OPTIONAL. Sampling rate of the channel in Hz.
 cfg.channels.description                = ft_getopt(cfg.channels, 'description'                 , nan);  % OPTIONAL. Brief free-text description of the channel, or other information of interest. See examples below.
 cfg.channels.low_cutoff                 = ft_getopt(cfg.channels, 'low_cutoff'                  , nan);  % OPTIONAL. Frequencies used for the high-pass filter applied to the channel in Hz. If no high-pass filter applied, use n/a.
@@ -633,9 +633,11 @@ cfg.channels.orientation_component      = ft_getopt(cfg.channels, 'orientation_c
 cfg.channels.wavelength_actual          = ft_getopt(cfg.channels, 'wavelength_actual'           , nan);
 cfg.channels.wavelength_emission_actual = ft_getopt(cfg.channels, 'wavelength_emission_actual'  , nan);
 cfg.channels.short_channel              = ft_getopt(cfg.channels, 'short_channel'               , nan);
-
 % specific options for motion channels
-% cfg.channels.tracked_point         = ft_getopt(cfg.channels, 'tracked_point'  , nan);
+cfg.channels.sampling_frequency         = ft_getopt(cfg.channels, 'sampling_frequency'          , nan);
+cfg.channels.component                  = ft_getopt(cfg.channels, 'component'                   , nan);
+cfg.channels.tracked_point              = ft_getopt(cfg.channels, 'tracked_point'               , nan);
+
 %% columns in the electrodes.tsv
 cfg.electrodes.name             = ft_getopt(cfg.electrodes, 'name'             , nan);  % REQUIRED. Name of the electrode
 cfg.electrodes.x                = ft_getopt(cfg.electrodes, 'x'                , nan);  % REQUIRED. Recorded position along the x-axis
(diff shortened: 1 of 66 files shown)


sjeung:
Hi Robert, 
thanks for going over the changes I submitted. I have some questions about the listed suggestions. 

- do not loop over multiple TrackingSystems, make one complete "scan" (3 files) per TrackingSystem
- remove the special handling of TrackingSystems as an array of objects
- add tracksys entity to the file name for channels.tsv

These three changes seem to point to the same issue, which is the concatenation of metadata and channels for multiple tracking systems in a single motion.json or channels.tsv (See [BEP 029](https://drive.google.com/drive/u/0/folders/1M7qyj8nP8aZS_6hLyDGov7YwxVe2UX1i ), open point 5.1.). By 3 files per tracking system I understand the motion.tsv, motion.json, and channels.tsv. The looping or array handling is difficult to avoid unless the spec itself switches back to option 2 in open point 5.1. 

I do notice some hurdles in entering metadata this way using the current structure of data2bids, because it requires the user to have all the information about all of the tracking systems at once. On the other hand, the concatenated option was more preferred by motion people that we have approached. I think it's a question of whether we should tolerate some awkwardness in reading and writing for people to have an easier overview of what's in a data set when they inspect it. We have been thinking that the BEP was more or less final but it is still an extension proposal, and switching the style can be definitely considered. I am tagging @juliuswelzel too because it's going to be a major change in spec if to be implemented. I would also like to hear out his experience on converting his example data set.

data2bids.m
```diff
@@ -633,9 +633,11 @@
 cfg.channels.wavelength_actual          = ft_getopt(cfg.channels, 'wavelength_actual'           , nan);
 cfg.channels.wavelength_emission_actual = ft_getopt(cfg.channels, 'wavelength_emission_actual'  , nan);
 cfg.channels.short_channel              = ft_getopt(cfg.channels, 'short_channel'               , nan);
-
 % specific options for motion channels
-% cfg.channels.tracked_point         = ft_getopt(cfg.channels, 'tracked_point'  , nan);
+cfg.channels.sampling_frequency         = ft_getopt(cfg.channels, 'sampling_frequency'          , nan);
+cfg.channels.component                  = ft_getopt(cfg.channels, 'component'                   , nan);
+cfg.channels.tracked_point              = ft_getopt(cfg.channels, 'tracked_point'               , nan);
+
```
JuliusWelzel, line 640: I would propose adding "tracking_system" as this is a required column specified in the google doc.

cfg.channels.tracking_system              = ft_getopt(cfg.channels, 'tracking_system'               , nan);

data2bids.m
```diff
@@ -1342,16 +1342,21 @@
...
-  motion_json.ORNTChannelCount               = sum(strcmpi(hdr.chantype, 'ORNT'));
-  motion_json.VELChannelCount                = sum(strcmpi(hdr.chantype, 'VEL'));
-  motion_json.ANGVELChannelCount             = sum(strcmpi(hdr.chantype, 'ANGVEL'));
-  motion_json.ACCChannelCount                = sum(strcmpi(hdr.chantype, 'ACC'));
-  motion_json.ANGACCChannelCount             = sum(strcmpi(hdr.chantype, 'ANGACC'));
-  motion_json.MAGNChannelCount               = sum(strcmpi(hdr.chantype, 'MAGN'));
-  motion_json.JNTANGChannelCount             = sum(strcmpi(hdr.chantype, 'JNTANG'));
+  motion_json.SamplingFrequency     = hdr.Fs;
+  motion_json.StartTime             = nan;
+  motion_json.MotionChannelCount    = hdr.nChans;
```
JuliusWelzel, line 1347: This is specified in a motion-specific field in the motion.json and not per tracking system. (See google doc https://docs.google.com/document/d/1iaaLKgWjK5pcISD1MVxHKexB3PZWfE2aAC5HF_pCZWo/edit#bookmark=id.yt4ntz7whu4b)

JuliusWelzel:
Hi Robert,

also from my side many thanks for the suggestions and comments. I agree with Sein, that changing the structure of the motion.json, outputting 3 files per tracking system was something we considered but decided against in the past. . When working with multiple tracking systems (e.g. many IMUs that record data onboard), would result in a ton of files per recording. Additionally, the idea of having one overall motion.json file with all tracking systems has the benefit of having an overview in one file. This approach can also be used in other modalities (e.g. eye-tracking) where similar devices can take slightly different specs but might be stored in one modality.json file. Using a key value to match the "sub" systems in the channels file also felt convenient for us. This approach might seem unnecessary if one has access to out-of-the-box synchronized systems (which I believe you have at the Donders :D).
But happy to discuss switching back to option 2 in open point 5.1. in the google doc.

Best,
Julius

robertoostenveld:
Hi @JuliusWelzel and @sjeung, thanks for jumping on this so quickly. My idea was actually to first have Helena take a look at this (with whom I discussed this Friday, so some background info is missing from the PR) prior to discussing with you. But since you looked at it already, perhaps I should just skip that step and directly discuss with you some aspects w.r.t. the FT `data2bids` implementation, and (linked) some aspects w.r.t. the proposed specification. 

Are you available for a zoom call next week? I will send you the modified test/demo script as well, so that you can see how I would consider the FT  script to ideally look like.

sjeung:
Sure! I am available for a call the 24th (Tuesday) in the morning or Wednesday/Friday.

sjeung (changes requested):
Can we allow representation of subsecond precision here?

for instance by updating this line to
val = datestr(val, 'yyyy-mm-ddTHH:MM:SS.FFF'); % following RFC3339 and BIDS

Here is a related issue and bids-validator seems to support fractional seconds as well.
Because multimodal recordings will rely on scans.tsv for intermodality synch it will be a very useful addition
https://github.com/bids-standard/bids-validator/issues/957

robertoostenveld:
this can be closed, as it is now integrated with #2034

robertoostenveld:
let's continue discussion there

PR #2038, opened 2022-05-22
old fix to known Neuromag trigger bug

mcvinding:


robertoostenveld:
this only adds the checkmaxfilter option

Issue #2035, opened 2022-05-19
Empty room coordsystem_json in data2bids


mcvinding:
I ran in to a problem trying to convert empty room recordings from a Neuromag MEG system. It seems that the whole code block from line 1520 in `data2bids` assumes that the coordinate system is in head coordinates (either CTF or Neuromag, depending on which system). Most of the values seems to be hard-coded in the block and not read from the file. E.g.:

````Matlab
coordsystem_json.MEGCoordinateSystem                 = 'ElektaNeuromag';
coordsystem_json.MEGCoordinateUnits                  = 'm';
coordsystem_json.MEGCoordinateSystemDescription      = 'Neuromag head coordinates, orientation RAS, origin between the ears';
% coordinate system for head localization coils
coordsystem_json.HeadCoilCoordinates                 = []; % see below
coordsystem_json.HeadCoilCoordinateSystem            = 'ElektaNeuromag';
coordsystem_json.HeadCoilCoordinateUnits             = 'm';
coordsystem_json.HeadCoilCoordinateSystemDescription = 'Neuromag head coordinates, orientation RAS, origin between the ears';
````

However, for an empty room recording, there is no head could or coordinates. When reading the data this is made clear that "No device to head transform available in fif file" and "MEG channels will likely have coordinates in device frame, not head frame" (in the low-level function `mne2grad`). A similar check whether devise-to-head transformation exists should probably be implemented for `data2bids` to check whether coordinates are in headc oordinates or device coordinates?

It seems that it can be fixed with a check `isempty(hdr.orig.dev_head_t)` at the beginning of the code block. But then the coordinate system information of the device should be specified similar to the head coordinates.

I am not sure if this is a Neuromag specific issue or if it also applies to CTF files?

robertoostenveld:
I discussed this with Mikkel, he will send a PR to deal with the case of not having digitizer information in the fif file.

mcvinding:
[...]

PR #2043, opened 2022-06-03
improve support for empty-room MEG datasets in data2bids

mcvinding:
Cf.: https://github.com/fieldtrip/fieldtrip/issues/2035

`data2bids.m` had another update while I added the MEG empty room compatibility.


robertoostenveld:
I just discussed this with @mcvinding, he will share a empty room fif file, I will also find an empty room CTF file, and with those I will make a test script to make sure that it works with both.

PR #2046, opened 2022-06-10
FIX - for issue 2026

schoffelen:


robertoostenveld:
It would be good at a certain point to phase out `tokenize` in favor of [split](https://nl.mathworks.com/help/matlab/ref/splitlines.html), which was introduced in R2016b. We would have to add a drop-in replacement in https://github.com/fieldtrip/fieldtrip/tree/master/compat/matlablt2016b

PR #2048, opened 2022-06-11
fix the channel selection for EDF files

roschkoenig:
in the channel selection part of read_EDF, the desired channels are first saved from `chanindx` to `EDF.chansel` and then later the code tries to select the to-be-loaded channels through `EDF.chansel(chanindx)`, which under certain condition throws an error (since `chansel` may be shorter than what is indexed in `chanindx`). 
The proposed change fixed this for my usecase.

### DIFF
--- a/fileio/private/read_edf.m
+++ b/fileio/private/read_edf.m
@@ -342,7 +342,7 @@ elseif needdat || needevt
     useChanindx = true;
   elseif ~isempty(chanindx) &&  isfield(EDF, 'chansel')
     % a subset of channels should been selected from the predefined list
-    chanindx = EDF.chansel(chanindx);
+    chanindx = EDF.chansel;
     useChanindx = true;
   elseif  isempty(chanindx) &&  isfield(EDF, 'chansel')
     % all channels from the predefined list should be selected


schoffelen:
Hi, it is unclear at this point whether the fix, which might work for your usecase, does not break the function for other people, who might have slightly differently formatted EDF data. Therefore, in order for us to be able to evaluate (and be sure) that your fix is sufficiently generic, it would be good to share some example data and test code that allows for quick reproduction of the problem.

jdien07:
I've tangled with this code myself (Bug 2954 bugzilla), although this current report is not related.  It's a mess and could use a rewrite from scratch.  I can't tell if the proposed fix makes sense just by looking at it, but I did notice that the next if then block makes no sense, specifically:

  elseif  isempty(chanindx) &&  isfield(EDF, 'chansel')
    % all channels from the predefined list should be selected
    chanindx = EDF.chansel(chanindx);
    useChanindx = true;

It's saying that if chanindx is empty, then use it to select a subset from EDF.chansel.  Since chanindx is empty, this code simply sets EDF.chansel to [].  Is that the intended behavior?  Not sure but that's not how one would normally write the code if so, and the comment also seems to imply otherwise.  Someone needs to run this edf code through a full set of the possible permutations to test it properly.  I don't normally use edf files myself so I can't help with this.

roschkoenig:
Can't speak to the larger issues here, sorry. 
Have reproduced my error in the google drive folder: https://drive.google.com/drive/folders/1k6jErKb1hm-p1foJc63RECwPhKp-dcIb?usp=sharing with some test data downloaded from here: https://www.teuniz.net/edf_bdf_testfiles/ 

I get an error (`Index exceeds number of array elements`) without my fix, and behaviour as I expect with the fix.

robertoostenveld:
for our own reference: I have copied the test data to `/home/common/matlab/fieldtrip/data/test/issue2048`

robertoostenveld:
I can confirm the error using the data and test script provided by @roschkoenig 

I don't fully understand the current code yet, but see in `ft_read_header` line 851 that 

```
  case 'edf'
    % this reader is largely similar to the bdf reader
    if isempty(chanindx)
      hdr = read_edf(filename);
    else
      hdr = read_edf(filename,[],1);
      if chanindx > hdr.orig.NS
        ft_error('FILEIO:InvalidChanIndx', 'selected channels are not present in the data');
      else
        hdr = read_edf(filename,[],chanindx);
      end
    end
``` 

I do not think that the check `chanindx > hdr.orig.NS`  is correct in general, as `chanindx` is a vector with indices, whereas the `if` statement expects a single boolean. If you pass it a vector of booleans, it behaves different that one might  expect:

```
>> if [1 2 3] > 2, disp(1), else disp(0); end
     0
```
so it only checks the first boolean of the list.

The part I do not understand yet (as it is undocumented) is what is supposed to happen when `read_edf` gets three input arguments.

robertoostenveld:
it seems that the `chanindx` option in `ft_read_header` has not been thought trough properly. It works in case `chanindx=[]` is specified when reading the header, AND if the resulting header is passed to `ft_read_data`.

There are some cases where it works, see code below

```
%%
% this works

cfg = [];
cfg.channel = 'FP2';
cfg.dataset = 'test_generator.edf';
data = ft_preprocessing(cfg);

%%
% this also works


hdr = ft_read_header('test_generator.edf');
idx_to_load = find(ismember(lower(hdr.label), 'fp2'));
dat         = ft_read_data('test_generator.edf', 'header', hdr, 'chanindx', idx_to_load);

%%
% and this also works

hdr = ft_read_header('test_generator.edf', 'chanindx', 4);
dat = ft_read_data('test_generator.edf', 'header', hdr);


%%
% but this fails

hdr         = ft_read_header('test_generator.edf');
idx_to_load = find(ismember(lower(hdr.label), 'fp2')); % this is channel 3 from the selected 10 with 200 Hz (out of 16 channels in the EDF file)
dat         = ft_read_data('test_generator.edf', 'chanindx', idx_to_load);

```

robertoostenveld:
There is other non-standard use of `chanindx` in ft_read_header and ft_read_data in the case of `besa_besa` and in the case of the persistent `cachedata` being used. I think it would be good to look up the original use cases. 

@roschkoenig perhaps you can use the work-around with `ft_preprocessing` in the mean time. That is the recommended way to read and process data in FieldTrip any way.

jdien07:
Correction to my previous comment - the code I noted sets chanindx to empty if chanindx is empty, so even more senseless.  To be honest, while I don't think this part we're talking about was my code, I can't rule it out either as I don't have access to the fieldtrip builds from then and can't remember.  I had to rebuild a big chunk of it to try to make it work.  At any rate, it would be worth referring back to our discussions about Bug 2954 bugzilla in 2015 as it lays out some of the complexities involved in how chanindx interacts with EDF.chansel and how we tried to resolve them.  The comments I added to the file summarize the behavior of read_edf in this regard.  Whatever test files you apply here, make sure you include a case with varying sampling rates.

robertoostenveld:
I looked at it in detail. Although this PR suggested by @ roschkoenig might solve the problem, it does not solve the underlying flaw in the code that surfaced when passing the same `chanindx` from ft_read_data to ft_read_header. In ft_read_header it causes some channels to be hidden from the user, so the chanindx passed to ft_read_data should therefore specify the selection **in** the selection that was given to ft_read_header.

PR #2052 solves it at a higher level, and also moves some of the checks on correct `chanindx` specification to a better location (i.e., to the low-level code).

robertoostenveld:
Tonight the whole batch of regression test scripts will run, and if all pass, the solution will be part of the tagged release (version 20220615) that will then automatically be made.

Issue #2047, opened 2022-06-11
ft_read_event.m fails when reading epoched .fif file where event codes are character strings with different lengths


jdien07:
**Describe the bug**
At line 1714 the code:
events_code = str2num(cell2mat(events_id(:, 2)));
fails when reading an epoched .fif file where the event codes are character strings if the event codes are different lengths due to cell2mat requiring them to be the same length:

Error using cat
Dimensions of arrays being concatenated are not consistent.
Error in cell2mat (line 83)
            m{n} = cat(1,c{:,n});
Error in ft_read_event (line 1714)
        events_code = str2num(cell2mat(events_id(:, 2))); 

**To Reproduce**
Read attached test file.

**Expected behavior**
The data should be read rather than crashing the script.

**Environment and versions (please complete the following information):**
 - Operating System: Pop!_OS 22.04 LTS (Ubuntu distro)
 - MATLAB version R2022a
 - FieldTrip version fieldtrip-20220607
[iCM_003_20181205_081017_Cz-epo.zip](https://github.com/fieldtrip/fieldtrip/files/8883047/iCM_003_20181205_081017_Cz-epo.zip)

PR #2050, opened 2022-06-13
fix for #2047: ft_read_event.m fails when reading epoched .fif file where event codes are character strings with different lengths 

jdien07:
Making jump from SVN to git.  Had to redo changes.  I think I got it right this time.


schoffelen:
Sorry to be so slow in catching up on this @jdien07, Joe! Thanks for the suggested fix. I took the liberty to use a slightly different syntax, I fixed the git conflict, and added a test function. I am pretty sure that this will fix the issue for you. The changes will become available in the release branch right away, and will percolate to the newest release version once all tests run through smoothly.

PR #2051, opened 2022-06-15
FIX #2048 - resolved issue in channel selection for EDF

robertoostenveld:
- in ft_read_data  do not pass chanindx to ft_read_header
- do not read the EDF header  twice to check for a valid chanindx, check that in read_edf() instead
- updated help and comments in the code
- added test script


robertoostenveld:
Sorry, I messed up the PR. I will do another one.

PR #2053, opened 2022-06-16
improve support for empty-room MEG datasets in data2bids

robertoostenveld:
this incorporates #2043


robertoostenveld:
@mcvinding I opted for using `MEGCoordinatesystem=Other` in the `coordinatesystem.json`, plus a description. Furthermore, I extended ft_read_header and thereby also ft_read_sens that they now will also use the information specified in `coordinatesystem.json`. Please check it out. 

I will already merge, so that the test batch can run tonight. If needed, we can make more changes.

PR #2068, opened 2022-07-01
timelockanalysis: use parfor instead of for

AljenU:
With some small adaptations, it is possible to use a parfor loop in
ft_timelockanalysis. This potentially speeds up the calculation, when the
parallel computing toolbox is available.

Tested with https://github.com/AljenU/fieldtrip-parallel-support/tree/additional_tests, however no tests were found that take more than a few seconds on ft_timelockanalysis, making it hard to actually compare performance.


AljenU:
I am looking at a way where, for ~keeptrials, the size of  covsig can stay at  [nchan, nchan], instead of using [nrpt, nchan, nchan]. I found it is possible. For clarity, it should use a small helper function that contains the first half of the parfor loop, that calculates dat and numsmp. Due to how fieldtrip does the preamble, it cannot be a nested function nor a subfuction, but needs to be in its own file.
Where to put such a file, for what i have now called ft_timelockanalysis_covdat() ?

PR #2078, opened 2022-07-13
ENH - test function to check reproducibility of issue

schoffelen:


schoffelen:
wrote test function to check whether reported issue can be reproduced. I tested on matlab 2022a on linux and the function runs through just fine. I wonder about the specifics of the input data (that we don't know in detail), but for now it's a fair assumption that the code works fine.

PR #2088, opened 2022-08-08
setting numerical event values to string events in the order provided as input

arnodelorme:
Using this data file

https://openneuro.org/crn/datasets/ds003061/files/sub-001:eeg:sub-001_task-P300_run-2_eeg.set

```matlab
cfg              = [];
cfg.dataset      = filename;
cfg.trialdef.eventtype  = 'stimulus';
cfg.trialdef.eventvalue = { 'standard' 'oddball_with_reponse' };
cfg.trialdef.prestim        = 1; % in seconds
cfg.trialdef.poststim       = 2; % in seconds
cfg = ft_definetrial(cfg);

% Baseline-correction options
cfg.demean          = 'yes';
cfg.baselinewindow  = [-0.3 0];
data = ft_preprocessing(cfg);
``` 

When I run preprocessing, I lose the data.trialinfo field because the events are not numerical - so I do not know which trials are which. Also, I want to extract trials together (standard and oddball_with_reponse) because I want to preprocess them together. With the attached modification, I can get the trialinfo field and the number (1 and 2) correspond to the other of the event values provided as input ('standard' 'oddball_with_reponse').

```
data = 

  struct with fields:

           hdr: [1×1 struct]
         label: {79×1 cell}
          time: {1×636 cell}
         trial: {1×636 cell}
       fsample: 256
    sampleinfo: [636×2 double]
     trialinfo: [636×1 double]
          elec: [1×1 struct]
           cfg: [1×1 struct]
```

### DIFF
--- a/trialfun/ft_trialfun_general.m
+++ b/trialfun/ft_trialfun_general.m
@@ -194,6 +194,8 @@ else
     elseif ischar(event(i).value) && numel(event(i).value)>1 && (event(i).value(1)=='S'|| event(i).value(1)=='R')
       % on brainvision these are called 'S  1' for stimuli or 'R  1' for responses
       trlval = str2double(event(i).value(2:end));
+    elseif ischar(event(i).value) && numel(event(i).value)>1 && ~isempty(cfg.trialdef.eventvalue)
+        trlval = strmatch(event(i).value, cfg.trialdef.eventvalue);
     else
       trlval = nan;
     end


schoffelen:
Hi Arno,

I haven't looked at the particular dataset, but I am wondering about the following:

these days cfg.trl / trialinfo is allowed to be a table. So that means that there is no strict need anymore to convert eventvalue strings into a numeral. Would a table-based solution also work for you? In that case we could adjust ft_trialfun_general to output a cfg.trl in case the stimulus codes are strings?

schoffelen:
hi @arnodelorme what's your view on the above? I think implementing it as a table would be elegant.

schoffelen:
This is followed up in https://github.com/fieldtrip/fieldtrip/pull/2109

schoffelen:
Yes, I can imagine that some people might get stuck there. We have now an option in ft_definetrial (cfg.representation) that can overrule the tabular format (cfg.representation = 'numeric'). That's one for the noobs.

PR #2094, opened 2022-08-17
ENH - added coordsys to elec upon import from eeglab

schoffelen:


arnodelorme:
Looks good to me. The noze direction is always +X in EEGLAB so I do not think a lot of the code will be necessary. Even though you can set a different noze direction, EEGLAB will automatically rotate it and put back the noze in the +x direction.

schoffelen:
OK, thanks for the feedback. Since we don't know what users are capable of, I suggest to keep the other options in - even if they are probably never used, the are not in the way, right?

arnodelorme:
Yes, that sounds good.

PR #2099, opened 2022-08-30
revamping line attributes management for plotting

schoffelen:


schoffelen:
This PR results from an inventory of the consistency of the linecolor management throughout the plotting. There were some glitches with the browsefuns for ft_databrowser, which are now fixed. Also, the linecolor management has been revamped in general in order to handle channels and conditions in separate dimensions, so that it may become possible to plot each line in its own color. The idea would be to also handle linestyle and linewidth in a similar way (to be done), and to work towards a 'spatial_color' option for butterfly plots. This PR sets the stage for that.

schoffelen:
Let's merge and see what happens. test_tutorial_plotting seems to run through fine, but most likely any bugs introduced cannot be detected automatically because they need visual inspection of the figure output

PR #2101, opened 2022-09-10
FIX padding length

ruiliu-psy:
Let padding length vary along with resampling rate,  so as to keep equal length of resampled segments.


schoffelen:
Hi Rui, would it be possible to describe a bit the issue that this PR aims to solve? Do you have snippet of test code that can be included e.g. in the test_example_irasa function?

ruiliu-psy:
> 

Hi JM, the issue was padding length was computed with the fixed trials length (say 1000 time points), however, irasa resamples segments of the trial (900 time points, then 100 zeros was padded to it, just a hypothetical example) at **line 102,** so that when it was upsampled, longer time series (1100 time points) was plugged into fft() and cut off according to the original length (1000th frequency bin) at **line 199**. This RP fixed the issue by allowing the padding length vary along the resampling rate to ensure that fft() takes an equal length of time series (1000 time points).

schoffelen:
The question behind my question is: should there be something in a test script to demonstrate what the consequences of the changes in the code are?

Specifically:
- when people used the unchanged code, is the output wrong?
- was there an error?
- was the output suboptimal?
- should the users of the code be informed about the changes, e.g. when using the old version the output was suboptimal/wrong/buggy etc?

ruiliu-psy:
The change is minor, so no need for an extra testing script. I don't think the user need to be informed additionally, since the change is minor. Please have a look at the comparison before and after the change proposed by this PR and let me know if you think more action needs to be taken. Thanks!
Screenshot 2022-09-15 at 16 35 55

schoffelen:
looks good to me. Thanks for the information, and I'll merge.

PR #2104, opened 2022-09-13
ENH - make more memory efficient

schoffelen:


schoffelen:
I did a quick comparison of the old implementation vs. the new one, using a ~6 GB data structure (275 channel MEG, 780 trials) on the DCCN's HPC cluster.

Calling 'old-style' ft_timelockanalysis (with cfg.covariance = 'yes') reports:the call to "ft_timelockanalysis" took 89 seconds and required the additional allocation of an estimated 10747 MB

Calling 'new-style' ft_timelockanalysis reports:the call to "ft_timelockanalysis" took 35 seconds and required the additional allocation of an estimated 24 MB

I think it's worth the hassle. I took out the intermediate call to ft_checkdata(data, 'datatype', 'timelock') in case keeptrials is not true to begin with, and put back the original running sum for the computation of the average + var. 

I think that all is correct, but given the central position of this function in the codebase I wouldn't mind a pair of extra eyes...

PR #2105, opened 2022-09-13
issue when using cfg.interplimits = 'mask_individual' in ft_plot_topo when using multiple masks

helenacockx:
- the second time the function ft_plot_topo is called, the interpolation occurs over the different masks instead of within each mask.
- this is due to the reuse of previous_maskimage (line 166) which only contains a boolean version of the maskimage (created line 245) instead of the mask containing a different value for each new mask (created line 167-189).
- I proposed a solution by differentiation between maskimage and maskimage_bool
- should be checked if the proposed solution is also correct when using style 'imsat' (line 279) or style 'colormix' (line 296 & line 299).

### DIFF
--- a/test/test_pull2105.m
+++ b/test/test_pull2105.m
@@ -1,3 +1,9 @@
+function test_pull2105
+
+% WALLTIME 00:20:00
+% MEM 3gb
+% DEPENDENCY ft_topoplotER ft_plot_topo
+
 % create artificial data with 2 channels
 data.time = [-10: 10];
 data.label = {'C3'; 'C4'};


schoffelen:
Hi Helena, would it be a lot of work to also include a snippet of test code to demonstrate what you mean (e.g. in fieldtrip/test/test_pull2105.m), can be with artificial data?

helenacockx:
However, this example shows that my proposed solution has not solved the issue yet...

helenacockx:
Sorry, I forgot to close my masks, that was the problem.

schoffelen:
Thanks Helena! The issue and the test function are clear. I made some stylistic changes. I will also convert the test script into a function before merging.

schoffelen:
merging time!

Issue #2110, opened 2022-09-23
Fix units when reading NeuroOmega files


richardkoehler:
**Description**
When reading in MatLab files from NeuroOmega amplifiers using for example:
`ft_read_data(file, 'header', hdr, 'dataformat', 'neuroomega_mat', 'headerformat', 'neuroomega_mat')`,
the data in units of `µV` is currently calculated by dividing the array of values of each channel, which are an array of integers:

![grafik](https://user-images.githubusercontent.com/70506259/191917151-7c54429f-11eb-4b20-ba56-dd1d6b2ab656.png)

by the `BitResolution` of each channel, which is a `double`:

![grafik](https://user-images.githubusercontent.com/70506259/191917343-956514d5-8c49-492b-92a0-5a65d8d95118.png)

---

**However**, according to communication with the manufacturer, the file also contains  the `Gain` for each channel:

![grafik](https://user-images.githubusercontent.com/70506259/191923334-9ff74ca3-f327-49e2-9bd0-dab9fa59c92c.png)

... which we must also divide by to acquire the actual value in `µV`, like: `data[µV] = rawdata[int] * BitResolution / Gain`.

---

**Two checks** additionally convinced me that this must be true:
1. In my case for an ECoG recording: `BitResolution / Gain => 38.147 / 55 ~ 0.7`, which is the ` Input Bit Resolution` specified in the user manual (see last row):
![grafik](https://user-images.githubusercontent.com/70506259/191924277-52265c04-ef0f-45cf-bf2e-02b4b387614b.png)

2. We performed a recording with a NeuroOmega amplifier and a 2nd amplifier (TMSi Saga) in the same session, and only if we divide by the gain (55), do the power spectra line up:
![grafik](https://user-images.githubusercontent.com/70506259/191927406-600bd46b-019c-401b-88d0-bd8a8cd9ecb9.png)
If we do not divide by the gain the power spectra do not match:
![grafik](https://user-images.githubusercontent.com/70506259/191927337-6ca9d084-2ef7-4357-ba17-ad8506b85ed3.png)

---

Of course, I am open for discussion and I want to stress that I am neither a computer nor data scientist so I might be missing some background information.
I will create a draft pull request though, as a suggestion to fix this issue.

---

**Environment and versions:**
 - Operating System: Windows 11
 - MATLAB version: R2021b
 - FieldTrip version: most recent master branch from github

schoffelen:
[...]

PR #2111, opened 2022-09-23
Fix reading neuroomega mat

richardkoehler:
This pull request is related to #2110 and is aimed at fixing correct conversion of rawdata to the unit 'µV'.
Additionally, support is implemented for reading ECoG data which is acquired from the manufacturers ECoG headbox.

**Edit**:
I would be happy to provide the sample data used for the test. This is empty data that contains no sensitive data.

**Changes/Additions:**
- Add chantype 'ecog' to support reading ECoG channels that start with 'CECOG'
- Add test with example data
- Data conversion to µV takes into account the '_Gain' of the amplifier

### DIFF
--- a/fileio/ft_read_data.m
+++ b/fileio/ft_read_data.m
@@ -1301,6 +1301,7 @@ switch dataformat
     for i=1:hdr.nChans
       v=double(hdr.orig.orig.(hdr.label{i}));
       v=v*hdr.orig.orig.(char(strcat(hdr.label{i},'_BitResolution')));
+      v=v/hdr.orig.orig.(char(strcat(hdr.label{i},'_Gain')));
       dat(i,:)=v(begsample:endsample); %channels sometimes have small differences in samples
     end
     
--- a/fileio/ft_read_header.m
+++ b/fileio/ft_read_header.m
@@ -2079,8 +2079,8 @@ switch headerformat
       chantype={'micro'};
     end
     
-    chantype_dict={'micro','macro',     'analog', 'micro_lfp','macro_lfp','micro_hp','add_analog','emg', 'eeg';...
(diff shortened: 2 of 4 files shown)


schoffelen:
Hi Richard, @richardkoehler thanks for this. I already started typing a reply in the corresponding issue, but missed the PR altogether. It would be great if you could share the data file with us: we don't work with this fileformat ourselves, so have no way of testing.

richardkoehler:
@schoffelen  I would be very happy to share the example data. Where can I upload it? I could also push it to the repository if you point me to the right location.

schoffelen:
I don't know how big the file is, but ideally it is as small as possible :). Sending it to us in any of the ways listed here is fine. Or another way that is convenient for you is fine too...

https://www.fieldtriptoolbox.org/faq/how_should_i_send_example_data_to_the_developers/

richardkoehler:
@schoffelen  Sorry, I totally overlooked that item in the FAQ, thanks for pointing it out!
The file is currently 59 MB, but there is definitely a lot of potential for reducing its size. Should I leave it as is or crop the file before sending it to you - not sure if you think 59 MB is too large?

schoffelen:
It would be nice if it could be cropped a bit. 59MB is not that much, but many files of 59MB is :).

richardkoehler:
@schoffelen  I have cropped it to around 1.5MB now. You should hopefully be able to download it from my GoogleDrive: https://drive.google.com/file/d/11bG39ST5Ts06fCFCxby2maL15_K3FxCh/view?usp=sharing
Let me know if there are any issues!

schoffelen:
Thanks! I managed to grab the data. I stored it on our file system for future reference, and verified that the test function runs through.

richardkoehler:
@schoffelen  Thanks so much for the support!

PR #2115, opened 2022-10-07
FIX - bug in layout plotting and goto topo if viewmode is butterfly

schoffelen:


schoffelen:
This PR fixes an issue with layout plotting (viewmode = 'butterfly') when there are additional channels (on top of the ones that are described in the layout), and it addresses a request of Rao, where (again with viewmode = 'butterfly') the next interactive figure is a topo, rather than a singleplotER

PR #2118, opened 2022-10-11
FIX - trialfun regexp

### LINKED ISSUE #2117: ft_trialfun_general cannot recognise trialvalues with "0"
A recent commit has updated how ft_trialfun_general recognises character trailvalues, and it excludes "0" from valid characters:
https://github.com/fieldtrip/fieldtrip/blob/616a14c5a516967f208f3ad18118e642393c4e79/trialfun/ft_trialfun_general.m#L195
I wonder whether the exclusion of "0" is intentional or the code should be like this:
```matlab
elseif ischar(event(i).value) && ~isempty(regexp(event(i).value, '^[SR]+[\s]*+[0-9]{1,3}$')) 
```

tiborauer:
fixes #2117

### DIFF
--- a/trialfun/ft_trialfun_general.m
+++ b/trialfun/ft_trialfun_general.m
@@ -192,7 +192,7 @@ else
     
     if isnumeric(event(i).value) && ~isempty(event(i).value)
       trlval = event(i).value;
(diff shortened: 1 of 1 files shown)


schoffelen:
Thanks!

Issue #2096, opened 2022-08-18
BIDS with NaN samples


arnodelorme:
This dataset fails to import

https://openneuro.org/datasets/ds002680/versions/1.2.0

Because there are NaNs in the sample column (onset column is OK). It does pass the validator.

https://openneuro.org/datasets/ds002680/versions/1.2.0/file-display/sub-002:ses-01:eeg:sub-002_ses-01_task-gonogo_run-1_events.tsv

arnodelorme:
Aug 8th 2022 was my last pull

schoffelen:
sorry @arnodelorme I was unclear. What snippet of code did you run, causing the 'failure to import' said data?

arnodelorme:
You place the dataset above in 

```
cfg.dataset = xxxx
```

And then call any Fieldtrip function.

schoffelen:
With the current version of the code and downloaded data (using run 1 for subject 2 as referred to above), I cannot reproduce the earlier error: 

>> filename=fullfile(pwd,'sub-002_ses-01_task-gonogo_run-01_eeg.set');
>> event=ft_read_event(filename);
found matching BIDS sidecar '/home/dyncon/jansch/Downloads/ds002680/sub-002/ses-01/eeg/sub-002_ses-01_task-gonogo_run-01_events.tsv'
reading header information from '/home/dyncon/jansch/Downloads/ds002680/sub-002/ses-01/eeg/sub-002_ses-01_task-gonogo_run-01_eeg.json'
reading '/home/dyncon/jansch/Downloads/ds002680/sub-002/ses-01/eeg/sub-002_ses-01_task-gonogo_run-01_eeg.json'
>> cfg.dataset = filename;
>> cfg.trl = [1 1000 0];
>> data=ft_preprocessing(cfg);
found matching BIDS sidecar '/home/dyncon/jansch/Downloads/ds002680/sub-002/ses-01/eeg/sub-002_ses-01_task-gonogo_run-01_eeg.json'
reading '/home/dyncon/jansch/Downloads/ds002680/sub-002/ses-01/eeg/sub-002_ses-01_task-gonogo_run-01_eeg.json'
found matching BIDS sidecar '/home/dyncon/jansch/Downloads/ds002680/sub-002/ses-01/eeg/sub-002_ses-01_task-gonogo_run-01_channels.tsv'
[...]

PR #2119, opened 2022-10-13
the 'n/a' in a BIDS events.tsv file is not replaced by NaN, but remains a string

mcvinding:
I got an error when reading events from a BIDS _events.tsv table. `ft_read_event`read `'n/a'` as char ranter than NAN. I cannot find how FieldTrip read tsv tables (happens from line 2389 and onwards). Since Matlab's `readtables` interpreted 'n/a' as NaN.

I added this code block to read BIDS tsv event files. It should be safe as long as it is BIDS as all event.tsv files ought to follow the same formatting.

Might be related to https://github.com/fieldtrip/fieldtrip/issues/2096 - as it will give an error if you try to import BIDS eeg data with NAN values in the trigger.

The NAN value, in my case, came from a system trigger in a BioSemi eeg dataset (.bdf fileformat) that first had been BIDSified with `data2bids`.


robertoostenveld:
This should not happen, can you share the tsv file that gives you trouble? 

The reading of BIDS tsv files happens on https://github.com/fieldtrip/fieldtrip/blob/25b9beec17c016ac28dd1c3e62272a6576c87c49/fileio/ft_read_event.m#L2401 which causes a function in [fileio/private](https://github.com/fieldtrip/fieldtrip/blob/master/fileio/private/) to be called. Note that BIDS events.tsv files are not guaranteed to have the samples column, but they always do have the offset column (in seconds), hence the header is needed to transform seconds in samples. 

The function in private that is called is either 
- https://github.com/fieldtrip/fieldtrip/blob/master/fileio/private/bids_tsv.m
- https://github.com/fieldtrip/fieldtrip/blob/master/fileio/private/events_tsv.m

right now I cannot recall the rationale for there being two functions, but the code you inserted with your PR duplicates at least one of these.

robertoostenveld:
or rather than sharing the whole file, you can also copy and paste a short snipper here as code

mcvinding:
Ok, I was looking for a function called `event_tsv` but I could not find it - I should have guessed it was in private. 

The eventvalue is event_tsv all the way down here, even if isbids = 1

https://github.com/fieldtrip/fieldtrip/blob/25b9beec17c016ac28dd1c3e62272a6576c87c49/fileio/ft_read_event.m#L2406

So it uses the `event_tsv`, but perhaps it ought to use `bids_tsv` instead?

mcvinding:
Here is an example event.tsv that gave the error:

````
onset	duration	sample	type	value
0	n/a	1	STATUS	65280
0	n/a	1	Epoch	n/a
5.06982421875	n/a	10384	STATUS	65535
7.34814453125	n/a	15050	STATUS	65355
````

**Comment by robertoostenveld** [comment]
I stored it in a events.tsv file and added a corresponding meg.json file, and did this

```
reading header information from 'sub-01_task-words_meg.json'
reading 'sub-01_task-words_meg.json'
event = 
  1×4 struct array with fields:
    type
    value
    sample
    offset
    duration
    timestamp
>> event(1)
ans = 
  struct with fields:

         type: 'STATUS'
        value: 65280
       sample: 1
       offset: []
     duration: NaN
    timestamp: 0
>> event(2)
ans = 
  struct with fields:

         type: 'Epoch'
        value: 'n/a'
       sample: 1
       offset: []
     duration: NaN
    timestamp: 0
>> event(3)
ans = 
  struct with fields:

         type: 'STATUS'
        value: 65535
       sample: 10384
       offset: []
     duration: NaN
    timestamp: 5.0698
>> event(4)
ans = 
  struct with fields:

         type: 'STATUS'
        value: 65355
       sample: 15050
       offset: []
     duration: NaN
    timestamp: 7.3481
```

is it the `value='n/a'` of the second event that bother you?

**Comment by robertoostenveld** [comment]
Regarding using `bids_tsv` or `events_tsv`: I think that it uses the first in case a proper datafile is present and the 2nd when that datafile is not (or might not be ) present. 

So `ft_read_event('sub-01_task-words_meg.fif')` would call bids_tsv`, whereas `ft_read_event('sub-01_task-words_events.tsv')` would call `events_tsv`.

**Comment by mcvinding** [comment]
Yes, in this case, it is the second event that seems to be the cause of errors.

I ran ft_read_events for the data file and the event file, and both runs had `eventformat = 'events_tsv'` when it came to line 2406. For the data file, it is defined at line 259 to use events_tsv.

https://github.com/fieldtrip/fieldtrip/blob/25b9beec17c016ac28dd1c3e62272a6576c87c49/fileio/ft_read_event.m#L259

````Matlab
dat_file = 'sub-002_task-task+off_run-1_eeg.bdf';
eve_file = 'sub-002_task-task+off_run-1_events.tsv';

eve = ft_read_event(dat_file);
eve = ft_read_event(eve_file)
````

Changing eventformat in ft_read_events to bids_tsv at line 259 do however lead to an error in bids_tsv in this code block (line 50-55):

````Matlab
fid = fopen_or_error(headerfile, 'r');
json = fread(fid, [1, inf], 'char=>char');
fclose(fid);
  
% convert the json string into a structure
json = jsondecode(json);
````

````
Error using jsondecode
JSON syntax error at line 1, column 1 (character 1): expected value but
found 'onset'.
````

**Comment by mcvinding** [comment]
I think I might have found the reason. In ft_read_event line 2450-2453 it converts string values to double and updates event.value, but specifically skips this step if the value is NaN - hence it is left as a string. Skipping lines 2451-2453 fixes the issue for me. My question is then why it is there in the first place, and if removing this will cause more harm?

The code in question:
```Matlab
value = str2double(event(i).value);
if ~isnan(value)
  event(i).value = value;
end
```

PR #2123, opened 2022-10-24
Add Harmonic Field Correction (HFC)

neurofractal:
Adds functionality for **Harmonic Field Correction (HFC)** which models interference as a harmonic magnetic field. It is particulaly useful for MEG data with low channel numbers (e.g. OPM data). Homogenous field based on **[Tierney et al. (2021) NIMG, 118484](https://doi.org/10.1016/j.neuroimage.2021.118484)**. Harmonic expansion based on **[Tierney et al. (2022) NIMG, 119338](https://doi.org/10.1016/j.neuroimage.2022.119338)**.

Code from @georgeoneill and myself, adapted from SPM.


ft_denoise_hfc.m
```diff
@@ -0,0 +1,240 @@
...
+cfg = ft_checkconfig(cfg, 'forbidden',  {'trial'}); % prevent accidental typos, see issue 1729
+cfg = ft_checkconfig(cfg, 'forbidden',  {'channels'}); % prevent accidental typos, see issue 1729
+
+% Check if the data has a grad structure
+if ~isfield(data,'grad')
+    error(['Data needs a grad structure']);
+end
+
+% set the defaults
+cfg.L               = ft_getopt(cfg, 'L', 1);
```
schoffelen, line 83: for consideration: may I suggest to change the option cfg.L into cfg.order? This makes the meaning of the option a bit more intuitive to the user.

ft_denoise_hfc.m
```diff
@@ -0,0 +1,240 @@
...
+        ' not present in cfg.channel. These channels will NOT be used for '...
+        'Harmonic Field Correction']);
+end
+
+% generate harmonic basis set
+opt = [];
+opt.li = cfg.L;
+opt.v = data.grad.coilpos(x,:);
+opt.o = data.grad.coilori(x,:);
+N = spm_opm_vslm(opt);
```
schoffelen, line 118: If I understand well, the spm_opm_vslm function is from Tim's toolbox, which is hosted on github, but outside SPM. I would like to suggest that we don't include the function in fieldtrip/private (if anything, it would be supplied in the Fieldtrip distro in the folder external/OPM, to make explicit that it is code that is not our responsibility/fruit-of-our-thoughts). Alternatively, given that Tim's toolbox is a toolbox in its own right, we may want to require the user to 'install' it themselves, and make the availability of this function on the matlab-path  their responsibility. Availability of (external) toolboxes is typically checked by the utility funcition ft_hastoolbox.m 

As a service to the users, fieldtrip for instance includes (parts of various versions of) SPM in external/spm, and also some stuff from eeglab, brainstorm, and loads of other packages. Maintaining this is a bit tricky for actively developing code, so ideally we would start off by *not* including the toolbox there.

schoffelen (changes requested):
Thanks Robert! I have a few small comments for consideration, which are inspired by our general development/maintenance scheme. This pertains to the naming of the option cfg.L (which is a bit nitpicky admittedly), and to the handling of Tim's function (i.e. where should it go in the repo, if at all)

neurofractal:
@schoffelen - This should all done. Sorry it's a bit messy - changed from external/OPM to external/opm (no capitals!)

schoffelen (approved):

schoffelen:
Just out of curiosity: how did you generate the data object? Was it hand crafted, or does FieldTrip already support reading the data format to some extent? 

I noticed data.hdr.fieldori as an exotic field, so I was wondering.

neurofractal:
Excellent thanks :)
We use this hacky script to load our data into FT format (https://github.com/neurofractal/analyse_OPMEG/blob/master/data_functions/ft_opm_create.m). I have been meaning to make this better, and incorporate into FT by building https://www.fieldtriptoolbox.org/getting_started/opm_fil/ . I'll raise an Issue to get this started!

schoffelen:
> Excellent thanks :)
> We use this hacky script to load our data into FT format (https://github.com/neurofractal/analyse_OPMEG/blob/master/data_functions/ft_opm_create.m). I have been meaning to make this better, and incorporate into FT by building on https://www.fieldtriptoolbox.org/getting_started/opm_fil/ . I'll raise an Issue to get this started!

That would be great, thanks Robert!

PR #2125, opened 2022-11-01
ENH - addressed issue 2072

schoffelen:


schoffelen:
To get issue2072 out of the way, I decided to solve it myself. I ran test_ft_selectdata, which ran through fine. Let's see what the rest of the batch of test functions will be doing.

PR #2126, opened 2022-11-01
ENH - added test function and a fix in case reject='complete'

schoffelen:
This PR currently fixes issue 2124 in the case of cfg.artfctdef.reject = 'complete';

@robertoostenveld a generic solution to the issue for partially overlapping trials with different samples in the overlap would be a bit more tricky to implement. I am not sure whether this has ever been allowed in the  past. Should we opt for a generic solution, or rather throw an informative error when ft_fetch_data (called by ft_redefinetrial) gets confused by the user specification?

### DIFF
--- a/ft_rejectartifact.m
+++ b/ft_rejectartifact.m
@@ -228,18 +228,18 @@ end
 
 % ensure the crittoilim input argument is valid
 if ~isempty(cfg.artfctdef.crittoilim)
-  
+
   if (size(cfg.artfctdef.crittoilim,2) ~= 2 ...
       || (size(cfg.artfctdef.crittoilim,1) ~= size(trl,1) ...
       && size(cfg.artfctdef.crittoilim,1) ~= 1))
     ft_error('if specified, cfg.artfctdef.crittoilim should be a 1x2 or Nx2 vector');
   end
-  
+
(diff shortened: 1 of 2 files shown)


robertoostenveld:
I like the ft_selectdata on the trials for complete. 

For partial with overlap, I am now looking over individual trials and use appenddata to put them back together again. How is that?

schoffelen:
Sounds good. The only worry would be a massive speed reduction, but that is the responsibility of the user I guess. Also, I wonder what happens if the trials are not in chronological order: the hasoverlap detection heuristic assumes this. I don't know whether trials can be not-in-order, but if they are, ft_fetch_data will throw an error eventually...

robertoostenveld:
speed will indeed be sub-optimal, but supporting partial and complete rejection with overlapping trials is more important than making that new functionality fast.  

I though that for `hasoverlap` the `sortrows` on the sampleinfo would deal with the situation of non-chronological trials. The actual cutting up of the data is done per trial and there the order should not matter, hence there I kept the trials in identical order as the input. Could you check this once more, as I also considered this as a risk?

If I know look at the code, I actually don't see why `trialinfo` cannot be kept. Do you see the logic for removing it?

schoffelen:
Ah, i missed the sortrows indeed.

PR #2132, opened 2022-11-09
ENH - code clean up of ft_artifact_zvalue

schoffelen:


schoffelen:
I cleaned up the code a bit:
- rewrote the logic a bit w.r.t. cfg.memory 'low'/'high', to avoid duplication of code
- removed commented out sections

I updated the help doc a bit

I added the possibility to not do the z-scoring, to allow for thresholds with physically interpretable quantities (e.g. degrees of visual angle for eyetracker channels)

PR #2137, opened 2022-11-17
reading datetime object

JonathanVHoecke:
Dear Fieldtrip team

I would like to request a small PR to have the tsv reader function not read dates as datetime objects, but as text.
Datetime objects are not able to be updated or merged because datetime objects do not behave as cells:
image

image

Fieldtrip internal datetime object handling does the job, when the tsv column is read as a text.

### DIFF
--- a/fileio/ft_read_tsv.m
+++ b/fileio/ft_read_tsv.m
@@ -28,4 +28,4 @@ function tsv = ft_read_tsv(filename)
 % $Id$
 
 ft_info('reading ''%s''\n', filename);
-tsv = readtable(filename, 'Delimiter', 'tab', 'FileType', 'text', 'TreatAsEmpty', 'n/a', 'ReadVariableNames', true);
+tsv = readtable(filename, 'Delimiter', 'tab', 'FileType', 'text', 'TreatAsEmpty', 'n/a', 'ReadVariableNames', true, 'DatetimeType','text');


JonathanVHoecke:
@robertoostenveld I see what you mean. Well, at least in the data2bids function, the datetime objects need to be handled as cells that are adaptable, because it reads in sessions.tsv or scans.tsv.
I agree that other uses require a normal datetime object read from a tsv file for mathematical purposes.

helenacockx:
Hi Jonathan,

Can you write a small example script of what you exactly mean? I quickly tried to merge two tables containing datetime objects, or even one containing a datetime object and the other containing a cell, but I don't have any issues with this:
```
t1=table(1, datetime);
t2 = table(2, datetime);
t12 = mergetable(t1, t2)

t3 = table(3, {char(datetime)});
t123 = mergetable(t12, t3)
```
In this case, my text is in the same format as my datetime though. So I don't know whether this is also the case for you.

helenacockx:
Anyway, I think it is probably better to fix this issue within `mergetable` (or within `data2bids`) than within `readtable` because of the reason Robert is giving.

JonathanVHoecke:
Hi Robert and Helena

I have been able to write a piece of code that mimics exactly what is going on:
https://github.com/fieldtrip/fieldtrip/pull/2137/commits/e08b6fd3c63ea958593773b93f840137916bb664

I hope that location is the right folder to demonstrate the bug.

I actually found out, that it is not precisely related to updating the datetime object, but in the way how a datatime object is handled when adding or updating other information, in my case updating clinical scores to the sessions.tsv.

robertoostenveld:
I see in your example that 

```
>> sessions_tsv_1
sessions_tsv_1 =
  2×3 table
         session_ids                  date            UPDRS
    ______________________    ____________________    _____
    {'ses-EcogLfpMedOn01'}    18-Nov-2022 05:57:40      7  
    {'ses-EcogLfpMedOn02'}    19-Nov-2022 05:57:40      8  
>> this
this =
  1×2 table
         session_ids          UPDRS
    ______________________    _____
    {'ses-EcogLfpMedOn01'}      9  
```

for the `this` the date is unspecified. You cannot have an empty datetime object, but there apparently does exists a Not-a-Time value:

```
>> datetime([])
Error using datetime
Numeric input data must be a matrix with three or six columns, or
else three, six, or seven separate numeric arrays. You can also
create datetimes from a single numeric array using the 'ConvertFrom'
parameter. 
>> datetime([nan nan nan])
ans = 
  datetime
   NaT
```

PR #2141, opened 2022-11-18
Fix: cfg.funcolorlim-dependent scaling of cloud radius

StolkArjen:
This PR fixes a bug in ft_plot_cloud involving the scaling of the cloud sizes. Before this fix, the radii of the clouds were determined by the range of the data, irrespective of the specifications of the cfg.funcolorlim variable.

PR #2142, opened 2022-11-20
fix: ft_plot_cloud & test_ft_plot_cloud

StolkArjen:
This is a follow-up from https://github.com/fieldtrip/fieldtrip/pull/2141


robertoostenveld:
thanks!

PR #2147, opened 2022-12-01
Adding compiled files for M1 Mac

arnodelorme:
This minor change would avoid a lot of confusion


robertoostenveld:
Looking at the rather straightforward suggested change, the question to ask is: why are ft_definetrial and ft_redefinetrial separate functions? We don't have ft_prreprocessing and ft_repreprocessing either. What is your take on this @schoffelen? Should we perhaps consult the people in our MEG/EEG meeting?

schoffelen:
confusion between ft_definetrial and ft_redefinetrial is indeed a recurring issue. Let's consult the local userbase.

arnodelorme:
Added MAC silicon compiled files

arnodelorme:
@robertoostenveld and @schoffelen based on discussion with Robert, I only kept the changes related to M1 Mac compiled files.

robertoostenveld:
this was replaced by #2270

PR #2156, opened 2022-12-09
first go at reading the lookup table from SimNIBS v4

ualsbombe:
ft_read_atlas has been changed such that it can read the lookuptable from SimNIBS


schoffelen:
Hi Lau, thanks for this! I took the liberty of changing the code logic a bit. Strictly speaking we're not dealing with a mesh and an associated label file here (i.e. the freesurfer surface based aparc/aseg case), but with a volumetric representation and a lookuptable. Therefore I moved the relevant code to be handled similarly to the way that freesurfer_volumes are handled. If that's all OK with you, let me know, and I'll merge

schoffelen:
OK, I will merge soon, perhaps after cleaning up the documentation a bit.

schoffelen:
I got carried away a bit, and overhauled ft_read_atlas to clean up the duplicated code, and to update the doc

Issue #2166, opened 2023-01-19
document the cfg.figure option


robertoostenveld:
@britta-wstnr noticed that the `cfg.figure` option cannot be found in the help.

I believe this to be supported in the following functions (identified with grep):

```
ft_databrowser.m
ft_electrodeplacement.m
ft_layoutplot.m
ft_movieplotTFR.m
ft_multiplotER.m
ft_multiplotTFR.m
ft_neighbourplot.m
ft_singleplotER.m
ft_singleplotTFR.m
ft_sourceplot.m
ft_topoplotCC.m
ft_topoplotER.m
ft_topoplotIC.m
ft_topoplotTFR.m
```

Another option to check in the documentation is `cfg.figurename`. It is not as widely supported, but still should be part of the help.

robertoostenveld:
Britta also mentioned that the figure or window titles are not handled correctly and/or consistently. That might be something else to look into.

schoffelen:
When I do, I get: 

```
jansch@mentat002:~/matlab/fieldtrip
1113 $ grep -R open_figure *.m
ft_clusterplot.m:    open_figure(keepfields(cfg, {'position', 'visible', 'renderer', 'figurename'}));
ft_databrowser.m:h = open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename'}));
ft_electrodeplacement.m:    h = open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
ft_electrodeplacement.m:    h = open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
ft_geometryplot.m:fig = open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
ft_layoutplot.m:h = open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer'}));
ft_movieplotTFR.m:h = open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename'}));
ft_multiplotER.m:open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
ft_multiplotTFR.m:open_figure(keepfields(cfg, {'figure', 'position', 'visible', 'renderer', 'figurename', 'title'}));
[...]

PR #2167, opened 2023-01-19
Attempt at fixing figure name in ft_sourceplot

britta-wstnr:
As mentioned in #2166 `cfg.figurename` in `ft_sourceplot` gets ignored as it gets set regardless of the presence of this field. 
This fixes it - but might be a bit sloppy as `cfg.figurename` does get passed to `open_figure` in the code above my fix - was not sure if it should rather get fixed there? 

CC @robertoostenveld

### DIFF
--- a/ft_sourceplot.m
+++ b/ft_sourceplot.m
@@ -1499,12 +1499,16 @@ else
 end
 
 % set the figure window title
-if ~isempty(dataname)
-  set(gcf, 'Name', sprintf('%d: %s: %s', double(gcf), mfilename, join_str(', ', dataname)));
+if ~isempty(cfg.figurename)
+  set(gcf, 'Name', cfg.figurename);
 else
-  set(gcf, 'Name', sprintf('%d: %s', double(gcf), mfilename));
+    if ~isempty(dataname)
+      set(gcf, 'Name', sprintf('%d: %s: %s', double(gcf), mfilename, join_str(', ', dataname)));
+    else
+      set(gcf, 'Name', sprintf('%d: %s', double(gcf), mfilename));
+    end
+    set(gcf, 'NumberTitle', 'off');
 end
-set(gcf, 'NumberTitle', 'off');
 
 % do the general cleanup and bookkeeping at the end of the function
 ft_postamble debug


robertoostenveld:
would it not be better to have the figure title always handled by `open_figure` ?

Right now it appears from the ft_sourceplot code that - as an afterthought - someone added the lines at the end of the function that, based on dataname, update the figure title.

britta-wstnr:
Indeed, hence why I said this is a "sloppy" fix - that just fixes it exactly where the problem occurs right now

robertoostenveld:
cfg.figurename and cfg.dataname seem two options that aim at solving the same thing. It would be better to merge the functionality.

robertoostenveld:
this is a pointer in the direction of a fix https://github.com/fieldtrip/fieldtrip/commit/ad58d6f8e92d1331313b4bb8c74d7cff39b77793

I see the following options:
1. the user fully specifies cfg.figurename
2. the user does not specify the full name, but the function name (using [mfilename])(https://nl.mathworks.com/help/matlab/ref/mfilename.html) is used AND
  a. part of the name is specified by the user in cfg.dataname
  a. part of the name comes from cfg.inputfile
  c. part of the name comes from the input variable names using [inputname](https://nl.mathworks.com/help/matlab/ref/inputname.html)

Whether a, b, c, should be optional or always used, I don't know

robertoostenveld:
@britta-wstnr

 just for your ft_sourceplot case: if you do this 

```
% construct the figure title
if isempty(cfg.figurename)
  if isfield(cfg, 'dataname') && ~isempty(cfg.dataname)
    cfg.figurename = sprintf('%d: %s: %s', double(gcf), mfilename, join_str(', ', cfg.dataname));
  elseif isfield(cfg, 'inputfile') && ~isempty(cfg.inputfile)
    cfg.figurename = sprintf('%d: %s: %s', double(gcf), mfilename, join_str(', ', cfg.inputfile));
  elseif nargin>1
    cfg.figurename = sprintf('%d: %s: %s', double(gcf), mfilename, join_str(', ', arrayfun(@inputname, 2:nargin, 'UniformOutput', false)));
  else
    cfg.figurename = sprintf('%d: %s', double(gcf), mfilename);
  end
end

```

just prior to `open_figure`, does that work and address your issue?

Btw, this is me typing some code without trying it out. I am not sure about the `set(gcf, 'NumberTitle', 'off')`, should that not be inside `open_figure` and always apply?

robertoostenveld:
Hi @britta-wstnr, have you tried the code suggestion I gave above?

britta-wstnr:
Hi @robertoostenveld no, I have not had the time yet to look at it again!

Issue #2130, opened 2022-11-08
Updated support for FIL (UCL) OPM data format


neurofractal:
Building on #1395 I would like to contribute some code to load in data acquired from our OPM system at UCL. There have been a few cosmetic changes since 2020. Ultimately the user should be able to load in data through ```ft_preprocessing``` followed by noise reduction etc.

I am also happy to update the [tutorial script](https://www.fieldtriptoolbox.org/getting_started/opm_fil/) for the website.

robertoostenveld:
that would be great! Could you start a pull request in draft mode, and first make only a test script that demonstrates with a small data file (not to be added to git but sent using dropbox or so) how the low-level ft_read_xxx functions are called, and that ft_preprocessing works? 

The OPM time series will be simple, have you though about how `ft_read_sens` or so would work? Usually the sensor positions are in the MEG data, but in your case they are more like how electrode positions are separate from the EEG data files.

neurofractal:
A note to myself - need a way of switching between single and double precision. Currently hard coded.

robertoostenveld:
> A note to myself - need a way of switching between single and double precision. Currently hard coded.

Just some tricks that I have implemented over the years in various low-level `fileio` function:

If you would store the number of channels and the number of samples (or time * fsample), you can divide the file size by the total number of channels*samples: that should be 4 (single) or 8 (double).

If you have a timestamp channel in the data (as in the CTF meg4 files), you can look at a short piece of that and see whether it increases linearly or random/jerky with either single oir double.

You could even look at a short piece of data in any channel and compare the standard deviation of the derivative over a short time segment. When you read single precision as double, or the other way around, the data will  be a mess and hence the standard deviation will be huge.

neurofractal:
Referring to #2172 

> How do you refer to the different types of helmets or casts? I guess you have some designs that are printed in different sizes or customized to different subjects? That would be compares to an EEG cap with a 1020 electrode layout in the sizes 56cm, 58cm and 60cm.

We currently don't standardise in this way. Most sensor arrays are customised to each individual subject at the moment. I guess during import we could add grad.type = 'FIL_56_head_custom':

```FIL_56_head_custom``` where,

- FIL = institution 
- 56 = sensor number
- head = what is being scanned (we also image the spine with OPMs)
- custom = to differentiate from generic arrays when we adopt these

Importantly, this information would ideally come from the accompanying ```_meg.json``` file.  **From a BIDS perspective would this be sensible?**

[...]

PR #2172, opened 2023-01-24
Small bug fix for opm_fil subfunction

neurofractal:
Part of #2130

This fixes an error when reading the header - json files are not stored as meg.json, but have a prefix that is the same as the .bin file (e.g. sub-001Y_ses-001_task-encoding_control_run-001_meg.json)

### DIFF
--- a/fileio/private/opm_fil.m
+++ b/fileio/private/opm_fil.m
@@ -116,7 +116,7 @@ if needhdr
     hdr.grad.label = positions.name;
     hdr.grad.coilpos = [positions.Px positions.Py positions.Pz];
     hdr.grad.coilori = [positions.Ox positions.Oy positions.Oz];
-    hdr.grad.type = 'megmag';
+    hdr.grad.type = 'meg';
     hdr.grad.tra  = diag(ones(1,length(positions.name)));
     if ~isempty(coordsys)
       hdr.grad.unit = coordsys.MEGCoordinateUnits;
--- a/fileio/private/opm_fil.m
+++ b/fileio/private/opm_fil.m
@@ -40,11 +40,17 @@ needdat = (nargin==5);
 [p, f, x] = fileparts(filename);
 assert(isfile(filename), sprintf('file "%s" not found', filename));
 
-channelsfile  = fullfile(p, 'channels.tsv');
-coordsysfile  = fullfile(p, 'coordsystem.json');
-datafile      = fullfile(p, 'meg.bin');
-headerfile    = fullfile(p, 'meg.json');
-positionsfile = fullfile(p, 'positions.tsv');
+% Remove the _meg part of the filename
+if strcmp(f(end-3:end),'_meg')
+    f = f(1:end-4);
+end
+
+% Get the BIDS compliant files
+channelsfile  = fullfile(p,[f '_channels.tsv']);
+coordsysfile  = fullfile(p,[f '_coordsystem.json']);
+datafile      = fullfile(p,[f '_meg.bin']);
+headerfile    = fullfile(p,[f '_meg.json']);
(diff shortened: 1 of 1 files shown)


robertoostenveld:
do you already want me to merge this, it will there be more coming in relation to #2130? If so, I suggest you change this PR into "draft" mode for the time being.

robertoostenveld:
Hi Robert, what is the status of this? I'd like to merge and close this PR if possible.

neurofractal:
Hey - sorry still work in progress. Give me another 1-2 weeks :)

robertoostenveld:
Sure!

You might be interested to hear that we are now actively analyzing OPM data recorded with the (for now only 8 channel) FieldLine system at NatMEG in Stockholm. Going through this analysis I anticipate that we will be actively making some changes to FieldTrip code.

neurofractal:
:) Awesome. Keep me in the loop - happy to contribute!

neurofractal:
So I have changed hard-coded precision from double to single. The lab has used this as standard for 3 years and will continue to do so. I didn't want to slow things down by checking for standard dev differences between single/double read-in every time.

robertoostenveld:
Regarding the change in `grad.type`: We should come up with a strategy for the overall `ft_senstype` sensor nomenclature and how to deal with layouts and with template grad arrays with 3D positions. 

I recently added a whole bunch of mat-files with grad structures for https://github.com/fieldtrip/fieldtrip/issues/183, basically to consolidate the squid-based arrays and nomenclature. I also have the layout and 3D positions for the FieldLine "alpha1"** OPM helmet that is now in Stockholm and that was used previously at https://www.cuanschutz.edu. On https://www.fieldtriptoolbox.org/template/layout/ we have a whole bunch of consistently named layouts for the Easycaps, thanks to them consistently naming and documenting them. For other EEG systems it is a bit more messy, and I expect OPM systems to become much more EEG-like in terms of flexibility. That poses problems in documenting (in BIDS) and visualization.  

How many different helmets do you have at the FIL? How do you refer to them, does each have a unique identifying name or type? If you consider the [EEG BIDS specification](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#sidecar-json-_eegjson) how would you document the OPM equivalent of `CapManufacturer` and `CapManufacturersModelName`?

neurofractal:
> We should come up with a strategy for the overall ft_senstype sensor nomenclature

Yes I was thinking about this earlier too... We should definitely add [QZFM](https://quspin.com/products-qzfm/) sensors to this. Is it possible to add grad.type = 'qzfm' to the grad structure during import which could then be picked up by ft_senstype?

> How many different helmets do you have at the FIL? How do you refer to them, does each have a unique identifying name or type? If you consider the EEG BIDS specification how would you document the OPM equivalent of CapManufacturer and CapManufacturersModelName?

Unfortunately we don't have one standard arrangement, as we use custom [3D-printed scannercasts](https://www.sciencedirect.com/science/article/pii/S1053811921011058#fig0001). This should be temporary before we move towards multi-subject helmets. I know the manufacturer Cerca has a generic helmet with standard arrangement for maybe 64 sensors? Unsure of the way forward here from a BIDS perspective...

robertoostenveld:
the [qzfm](https://quspin.com/products-qzfm/) would be the chantype (if needed, but it can also be megmag), the senstype or grad.type is the description of the whole array. For EEG it could be `1020`, for MEG it could be `ctf275`. So in this case it would also refer to the cap or helmet.

How do you refer to the different types of helmets or casts? I guess you have some designs that are printed in different sizes or customized to different subjects? That would be compares to an EEG cap with a 1020 electrode layout in the sizes 56cm, 58cm and 60cm.

PR #2173, opened 2023-01-26
Fix try/catch syntax for older MATLAB versions

gllmflndn:
With MATLAB < R2017a, the line:
https://github.com/fieldtrip/fieldtrip/blob/4792c437305b1013d7204ce64e5121af09d2f67f/fileio/private/read_yorkinstruments_hdf5_meta.m#L61
fails to parse with the MATLAB Compiler:
```
MATLAB code
'fieldtrip/fileio/private/read_yorkinstruments_hdf5_meta.m'
contains the following syntax error(s):
L 61 (C 11): SYNER: Parse error at '.': usage might be invalid MATLAB syntax.
```


schoffelen:
Thanks, Guillaume @gllmflndn !

PR #2175, opened 2023-01-26
ENH - updated code + doc, added circumcenter computation

schoffelen:


schoffelen:
@StolkArjen would you be OK with this PR? Essentially, it uses your 'circumcenter' function (now available online in one of the example scripts) to convert the HLC channels into 6 movement parameters.

StolkArjen:
Of course, and thanks for checking

PR #2176, opened 2023-01-27
improve display and interaction with anatomical MRI

robertoostenveld:
following https://github.com/fieldtrip/website/issues/633

### DIFF
--- a/ft_electrodeplacement.m
+++ b/ft_electrodeplacement.m
@@ -1144,13 +1144,14 @@ else
   ft_plot_mesh(removefields(opt.headshape, 'color'), 'tag', 'headshape', 'facecolor', 'skin', 'material', 'dull', 'edgecolor', 'none', 'facealpha', 1);
 end
 
+% apply uniform light from all angles
 lighting gouraud
-l = lightangle(0,  90); set(l, 'Color', 0.5*[1 1 1])
-l = lightangle(0, -90); set(l, 'Color', 0.5*[1 1 1])
-l = lightangle(  0, 0); set(l, 'Color', 0.5*[1 1 1])
-l = lightangle( 90, 0); set(l, 'Color', 0.5*[1 1 1])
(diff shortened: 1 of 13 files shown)


robertoostenveld:
The plotting of the anatomical labels "L" and "R" on an MRI has moved from ft_sourceplot to ft_plot_slice, which allows other functions to use it as well. The plotting of anatomical labels has been extended to "A", "P", "I", "S". The plotting of anatomical labels now also works for an MRI in which which the `ijk` volume is not aligned with the `xyz` axes. 

ft_sourceplot now has the option `cfg.flip` to determine whether it is automatically flipped (to avoid upside down confusion), the default is 'yes' (as that is what @schoffelen had implemented earlier with `align_ijk2xyz`). It now calls ft_volumereslice with method=flip.

ft_volumerealign now also has the option `cfg.flip`, just like ft_sourceplot. The default is 'no', but I am not yet 100% sure about when to support this and/or when to enable this as default. I think that flipping makes sense for interactive clicking on the fiducials, but not  for any of the other methods. For example when specifying voxel indices as fiducials, those would be wrong if you also flip.

Are there other functions that should also get the option `cfg.flip` ?

robertoostenveld:
Prior to merging this, I think it would be good to demonstrate the new functionality for plotting labels. Based on that, we should decide where and when this is to be used.

In light of https://github.com/fieldtrip/website/issues/633, if you now read the DICOMs and pass them straight into ft_volumerealign with method=interactive, it WILL show the L/R labels. This addresses my initial concern of that function having 50%/50% chance resulting in a correct/incorrect result in case there is no vitamine-E capsule. As it now shows the labels, the chance of flipping L/R will be much smaller.

robertoostenveld:
Let me merge this so that some others can try it out from the master branch. If the regression tests don't fail tonight, it can also be released this evening.

PR #2185, opened 2023-02-03
Fix electrodeplacement

helenacockx:
match handling of cfg.channel & cfg.elec
labels of cfg.channels were saved as a cell in a cell (instead of character in a cell)


schoffelen:
This fix probably addresses #2186 reported by @Irina-Ch, thanks @helenacockx !

robertoostenveld:
I tried the plotting just now, it is still inconsistent. I don't get errors, but pressing the "v" shortcut on the keyboard does not consistently result in redrawing. Also, it starts with a non-shiny head surface, but after "v" it is shiny. And double-clicking removes the position and grays out the label, but does not cause the figure to be redrawn, so the removed electrode is still in the figure until you add another electrode (as that triggers a redraw).

The implementation of this, and possibly other functions with graphical user interaction, needs to be reconsidered.  But for now it works well enough.

PR #2187, opened 2023-02-10
reading and writing JSON/binary JSON based JMesh files

### LINKED ISSUE #1992: add support in ft_read_headshape for the .jmesh format
See http://openjdata.org/ and #1991

As discussed with @fangq

fangq:
here is a patch to allow fieldtrip to read/write JSON (.jmsh) and binary JSON (.bmsh) encoded mesh files.

close #1992

### DIFF
--- a/fileio/ft_read_headshape.m
+++ b/fileio/ft_read_headshape.m
@@ -1344,21 +1344,21 @@ switch fileformat
 
   case {'neurojson_jmesh' 'neurojson_bmesh'}
     if(fileformat == 'neurojson_bmesh')
-        shape.orig = loadbj(filename);
+        jmesh = loadbj(filename);
     else
-        shape.orig = loadjson(filename);
+        jmesh = loadjson(filename);
     end
 
-    % shape.orig metadata
-    if(isfield(shape.orig, encodevarname('_DataInfo_')))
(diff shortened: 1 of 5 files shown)


fileio/ft_read_headshape.m
```diff
@@ -1341,7 +1342,84 @@
     shape.hex = shape.hex(:,1:8);
     shape.hex = shape.hex + 1; % this should be one-offset
 
-    
+  case {'neurojson_jmesh' 'neurojson_bmesh'}
+    if(fileformat == 'neurojson_bmesh')
+        shape.orig = loadbj(filename);
```
robertoostenveld, line 1347: The original `orig` field should not be stored in the output structure. It exposes file-format specific details, which (from experience) result in headaches in the long long-term when our own or others' code starts using it. 

The fieldtrip mesh data structure itself should be as small as possible but as complete/extended as necessary. We can later always add something to make it more complete, but we cannot remove something (as that breaks the API compatibility).

fangq: `shape.orig` is dropped

robertoostenveld:
@schoffelen I have reviewed and made some edits to the code in this PR. Let's talk it over before I merge this (and my changes).

PR #2191, opened 2023-02-20
improve code for dssp algorithm

schoffelen:


schoffelen:
the test script for this function is already in place. I made a minor change, and will merge for now.

PR #2192, opened 2023-02-21
ENH - utilize persistent variables for speedup

schoffelen:


schoffelen:
This is a relatively minor thing, but I came across a substantial relative amount (in my case ~20%) of time spent in these functions when calling a sequence of fieldtrip functions multiple times in a loop. The context then is that a lot of bookkeeping time is spent in the memory profiling related ambles (time of which is multiplied by two when ft-functions start with a call to ft_selectdata -> even if cfg.trials/channel = 'all'), as well as in ft_defaults.

schoffelen:
Hi @robertoostenveld do you foresee any problems if I merge this?

PR #2194, opened 2023-02-22
FIX - dashboard failure due to bmsh writing issue

schoffelen:


schoffelen:
This PR fixes the broken test function test_issue1992, which was failing due to two issues:

1) Matlab2022b requires a fix to a low level jsonlab function, see https://github.com/fangq/jsonlab/commit/0448eb1ead7978cb90d2d251fd25f9f3652858b4. This was indeed already referred to by @fangq in the associated issue https://github.com/fieldtrip/fieldtrip/pull/2187 

2) some of the earlier proposed fieldnames were still used in the test function, but this has become obsoleted (things like trilabel etc are intended for something else)

For now, I just changed the code in external/jsonlab, since it seems that the github repo's master version (2.9.8) of jsonlab is not yet in an official release, hence I suspect that by changing the code locally according to the commit, it's as if I did a git pull

PR #2197, opened 2023-02-22
FIX - made grouping of bipolar referencing scheme more robust

schoffelen:


schoffelen:
TO DO: create test function, using the data that was uploaded in the corresponding issue. 

Thanks for reporting!

PR #2207, opened 2023-03-09
ENH - added option to input deltat

schoffelen:


schoffelen:
I would like to have the lower level function at least have this option. Not sure whether we should think through building in support at the level of preproc/ft_preprocessing. This would require either an extra option, or a break w.r.t. backward compatibility. @robertoostenveld what do you think?

PR #2211, opened 2023-03-14
fix: accommodate a change to in_fopen_nk

StolkArjen:
This PR fixes a change to external/brainstorm/in_open_nk introduced in 2021*. Specifically, the 'samples' field was renamed into 'times' and its values were divided by the sample frequency. As a result, private/read_brainstorm_header would error on the missing samples field when parsing the header information. The below code change fixes this issue as well as converts the times into samples based on the sample frequency.

*https://github.com/fieldtrip/fieldtrip/commit/c1114e1df9942423ebf142e5508fa6284415ef2c
prior:
sFile.epochs(i).samples = [0, hdr.ctl(1).data(i).num_samples - 1];
after:
sFile.epochs(i).times   = sFile.epochs(i).samples ./ hdr.sample_rate;
sFile.epochs(i).times   = [0, hdr.ctl(1).data(i).num_samples - 1] ./ hdr.sample_rate;

### DIFF
--- a/fileio/private/read_brainstorm_header.m
+++ b/fileio/private/read_brainstorm_header.m
@@ -37,7 +37,11 @@ function [hdr] = read_brainstorm_header(filename)
 hdr.orig        = sFile.header;
 hdr.nChans      = sFile.header.num_channels;
 hdr.Fs          = sFile.header.sample_rate;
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks for the clean change to the code.

PR #2212, opened 2023-03-15
Ft2fiff

schoffelen:


robertoostenveld:
this will also be of interest to @arnodelorme

schoffelen:
This PR stays a bit a moving target. I am not entirely happy with it, specifically with respect to some issues I experienced related to coordinate systems etc. but I'd rather merge this now, to see how the basic functionality behaves. 

I will work on the docs next, and further improve fieldtrip2fiff along the way

PR #2213, opened 2023-03-15
Update ft_read_data.m

Jinboasltw:
To ensure the value of b within valid range, change the format of the output of openNSx to double.

Here is the test data: https://www.dropbox.com/s/ojea8u1glmqhbxg/example.zip?dl=0


schoffelen:
thanks @Jinboasltw for uploading the data, unfortunately it is too big to be of practical use. So I have not downloaded the data and could not test your code changes.

For test data we would ideally have minimal size data files, otherwise it does not scale. Given that the suggested code changes are small and I cannot think of any negative side effects, I will merge it for this time (taking the risk that - if this PR breaks someone else's analysis - I need to spend time on fixing it).

PR #2220, opened 2023-03-23
ft_sourceplot info axis

dnacombo:
In an [app](https://github.com/dnacombo/ft_dipole_topo_MEEG), I add topographies on the right side of an enlarged ft_sourceplot figure.
When ft_sourceplot creates this `subplot`, any overlapping axis is deleted (topographies in my case). Not sure why this used not to happen, but it's now preventing my app to work. `axes` does not delete anything.


dnacombo:
argh !
I only wanted to push the last commit!

dnacombo:
ok fixed.

schoffelen:
Hi @dnacombo is it still needed to look into this? If so, please supply some test data and test code so that we can at minimum test on our end that the suggested change does not break anything (in other words, mentioning that your suggested fix works for your 'app' is great, but it shouldn't break anything for the rest of the world). If it is not needed anymore, please feel free to close

PR #2231, opened 2023-04-21
document the units of the inverse operation for eLORETA

robertoostenveld:
- added check that the user did not specify leadfield options in case the input already contains the leadfield
- added check that the user did not specify a regularization parameter in case the input already contains the filter

### DIFF
--- a/inverse/ft_inverse_eloreta.m
+++ b/inverse/ft_inverse_eloreta.m
@@ -13,10 +13,14 @@ function [estimate] = ft_inverse_eloreta(sourcemodel, sens, headmodel, dat, C, v
 % and
 %   estimate    contains the estimated source parameters
 %
+% If the input data "dat" is in V, "cov" is in V^2, and the leadfield is in V/Am,
+% then the the estimated output "mom" is in Am, the output "pow" is (Am)^2, and the
+% output "filter" is in Am/V.
+%
 % Additional input arguments should be specified as key-value pairs and can include
-%   'keepfilter'       = remember the spatial filter,    can be 'yes' or 'no'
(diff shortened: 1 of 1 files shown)


robertoostenveld:
This follows from a conversation with @mcpiastra. The page https://www.fieldtriptoolbox.org/faq/units/ is related, as are the occasional questions on the mailing list such as https://mailman.science.ru.nl/pipermail/fieldtrip/2021-June/040932.html

It would be nice to have the units in general better documented. Documenting them at the lowest accessible level (i.e. `ft_inverse_xxx`)  seems feasible. Documenting them at a higher level is more difficult, since there can be so many choices in the analysis pipeline that affect this.  What might be good to document generally though is that inverse computations are not arbitrary if all data, geometrical objects and models (like the leadfield) are specified in SI units.

robertoostenveld:
@schoffelen can I ask you to review this content-wise? I will make sure to regression-test the checks for the leadfield and filter options being empty before it gets merged.

mcpiastra:
hi @robertoostenveld and @schoffelen ,
there are 10 left ft_inverse_something where I would now add checks for the leadfield. is **one** PR for all functions a good strategy?
thanks,
MC

robertoostenveld:
> is one PR for all functions a good strategy?

Yes, that would be fine. Probably the changes will be very similar to all functions, so that facilitates review.

schoffelen:
I have reviewed this, and updated the documentation about the units, wherever possible (note: for methods like music, sloreta, rv the units drop out of the equations, and/or are not straightforward to determine).

I'll merge for now, and trust that the dashboard will catch the regression errors that might have been introduced

PR #2233, opened 2023-04-24
inverse solutions - add checks in hasfilter

mcpiastra:
- added check that the user did not specify leadfield options in case the input already contains the leadfield
- added check that the user did not specify inverse solution parameters in case the input already contains the filter

This relates to the PR #2231.

### DIFF
--- a/inverse/ft_inverse_dics.m
+++ b/inverse/ft_inverse_dics.m
@@ -174,9 +174,15 @@ if hasmom
 end
 
 if hasfilter
+  % check that the options normalize/reducerank/etc are not specified
+  assert(all(cellfun(@isempty, leadfieldopt(2:2:end))), 'the options for computing the leadfield must all be empty/default');
+  % check that the options for the inversion are not specified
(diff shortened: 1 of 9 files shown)


mcpiastra:
the first commit is about the empty-checks in DICS. 
these checks might be missing in ft_inverse_dipolefit, where there is no hasfilter, no hasleadfield, no leadfieldopt.
all other inverse solutions have been modified. when no invopt was present, I put a check only in the numerical parameter of the inverse solution. not sure about this.

schoffelen:
Hi @mcpiastra is there anything more needed here before this can be merged. I am going over a bunch of issues/PRs which have not shown signs of activity for a long time. It would be a waste not to finish it. I assume that the motivation for this is that users shouldn't specificy options that will not have a downstream effect, so all in all I would say it's fine to merge this, and to wait and see what a run of the test functions is going to give us. Probably not much. 

One consideration could be to add an informative error message if the assertions fail (so that the user understands why the error occurs).

mcpiastra:
hi @schoffelen , I believe this was a "side-effect" of another intervention. I do not recall what the undesired downstream effect was. It does not seem too invasive though, so I think it would be ok to merge it.

PR #2236, opened 2023-04-26
speed up ft_affinecoordinates

schoffelen:


schoffelen:
This overhaul intends to keep the same functionality of the original function, but at the benefit of a much increased speed. Initially, when working on an interactive matlab session (slower machines on our cluster), a single call to ft_affinecoordinates took about 10 seconds. (it turns out that the equivalent call on a faster machine is about 5 times faster). In addition to that, the original implementation of the function uses a lot of eval() statements (bad practice :) ).

The proposed version has the same functionality, but is 50 to 100 times faster.

PR #2239, opened 2023-04-27
Update write_brainvision_eeg.m

LJGz:
A rounding error caused future calculated sampling rate to not be an integer when calculating back to hdr.Fs or importing into EEGLAB srate.


robertoostenveld:
thanks! The sampling rate indeed does not have to be an integer.

PR #2241, opened 2023-05-01
FIX #2240 - rename test files and add Precision field to the JSON

### LINKED ISSUE #2240: regression test shows that OPM_FIL is broken
With PR #2172 some changes were made by @neurofractal in the reading code for the OPM_FIL data. These caused [test_issue1395](https://github.com/fieldtrip/fieldtrip/blob/master/test/test_issue1395.m) to fail. That is a test script based on this data https://github.com/tierneytim/OPM/tree/master/testData

I can see two problems: 

The first is that the filenames are now more strictly BIDS like, with an underscore in their name.  If I replace in the script the test data files version 1 into "v2" (see below), the file names don't seem to be a problem any more.

```
|-- testData_v1
|   |-- T1w.nii
|   |-- channels.tsv
|   |-- coordsystem.json
|   |-- meg.bin
|   |-- meg.dat
|   |-- meg.json
|   |-- meg.mat
|   |-- positions.tsv
|   `-- testCustom.gii
`-- testData_v2
    |-- test_T1w.nii
    |-- test_channels.tsv
    |-- test_coordsystem.json
    |-- test_meg.bin
    |-- test_meg.dat
    |-- test_meg.json
    |-- test_meg.mat
    |-- test_positions.tsv
    `-- test_testCustom.gii
```

However, the second problem is the warning `Warning: trigger channel TRIG1 looks like noise and will be ignored` that results in the events being empty, and the subsequent analysis failing. I think this is because the precision changed from double to single, where double is apparently correct for the example data. 

I think this would be very easy to solve by adding a `Precision: single' or 'double' field to the `test_meg.json` file. I am fine with it being single by default, but al least that allows me to hack this test dataset such that the script again works.

robertoostenveld:


robertoostenveld:
@neurofractal can you review this suggested change to the code and to the test dataset?

neurofractal:
@robertoostenveld aah sorry didn't realise there was a test function already. Everything looks fine - I have tried the updated code with some data from our OPM system and works well.

PR #2242, opened 2023-05-03
Updated support for QZFM_UCL data #2130

neurofractal:
#2130

- So we are now thinking of including HelmetManufacturer and HelmetManufacturersModelName in the json file. The latter would contain a text field with something like: 'slotnumber-56_location-wholehead_fit-custom'

- I have used the grad.type = 'QZFM_UCL' to differentiate ourselves from other facilities using QZFM sensors

All seems to work fine with existing test data.

- Issue: I can't seem to get ft_channelselection working without providing a third input argument (i.e. it's not picking up the senstype automatically for some reason)

### DIFF
--- a/fileio/ft_chantype.m
+++ b/fileio/ft_chantype.m
@@ -236,7 +236,7 @@ elseif isheader && issubfield(input, 'orig.chs.coil_type')
   for sel=find([input.orig.chs.kind]==602)' % Resp
     chantype(sel) = {'respiration'};
   end
-  
+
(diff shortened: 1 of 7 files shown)


schoffelen:
Hi @neurofractal Rob, sorry for having dropped this for so long. I think that the PR is still relevant, no? Would you still be willing to work on this a bit? If so, that would be great, perhaps we can get things on track again. Would you mind first to resolve the current conflict, and that we take it from there?

PR #2246, opened 2023-05-06
ft2mne

schoffelen:
second round of improvements in order to get a smooth interface between fieldtrip data representations and fif-files (and the way that MNE-python expects them)


schoffelen:
This PR seems to successfully implement the correct representation of CTF sensors, such that MNE-python can work with it (this has been checked only to the extent that the topoplot in the MNE butterfly plot looks OK).

PR #2267, opened 2023-07-02
ENH - fix numerical rounding error for zero padding

schoffelen:


schoffelen:
this relates to the issue raised by @matthiasgrabenhorst

PR #2269, opened 2023-07-04
Update resample.m

arnodelorme:
Fix out crash trying to resample Hanson/Wakeman data from 1100 Hz to 100 Hz when signal processing toolbox absent

### DIFF
--- a/external/signal/resample.m
+++ b/external/signal/resample.m
@@ -119,7 +119,7 @@ Ly = ceil(Lx*p/q);
 % pre and postpad filter response
 
 nz_pre = floor(q-mod(L,q));
-hpad = [zeros(nz_pre,1);h];
+hpad = [zeros(nz_pre,1);h'];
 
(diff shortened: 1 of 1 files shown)


robertoostenveld:
@schoffelen this is something I will discuss in more detail with Arno, as we might have to push this upstream to octave, and/or check whether the octave implementation has (in general) already been fixed and we should just update.

external/signal/resample.m
```diff
@@ -119,7 +119,7 @@
 % pre and postpad filter response
 
 nz_pre = floor(q-mod(L,q));
-hpad = [zeros(nz_pre,1);h];
+hpad = [zeros(nz_pre,1);h'];
```
robertoostenveld, line 122: The vector `h` should not be transposed, but should be forced to always be a column vector. That way it works both when people pass h as a row and as a column, in line with the Mathworks implementation.

PR #2276, opened 2023-07-11
Update DEPENDENCY list of test functions by adding a new field in the test description for data.

contsili:
The data is split into three categories NODATA/PUBLICDATA/PRIVATEDATA.

For this the git grep command was used in Git Bash.

Some commits are related with the use of Git LFS. I tried to use Git LFS to fix an error that was given when I was trying to push the data to the remote repository. Git LFS was finally not used since the problem was solved with git fetch and git merge. So, the commits related with the use of Git LFS shouldn't be important.

This pull request solves the issue #2275


robertoostenveld (changes requested):
There are some issues with this PR. First, the commits related to LSF should not be there, we don't want to merge them. Even though you tried to revert it, you can see that there are still remaining issues with `getdimord.m`. 

Furthermore, you made changes to the failed functions in `test`, although those were moved to `test/invalid'. For those to be properly represented, this pull request should have been made  in a branch that itself branched off from  https://github.com/contsili/fieldtrip/tree/invalid_directory. I think that a git rebase could solve this issue.

Finally, and more content oriented: I suggest that you keep the format of the header field similar to the others, so `% DATA no/public/private` (i.e., an uppercase KEY followed by a value) instead of `% NODATA/PUBLICDATA/PRIVATEDATA`

Using `git cherry-pick` and `git rebase`you should be able to fix all of this (and some search-and-replace for the content). However, that might be more hassle than worth it; you can also simply copy all files that you want to keep out pf the directory, delete the branch and PR, create a new branch, copy the files back in and make a new PR.

Issue #2281, opened 2023-07-15
Update dccnpath and add test_dccnpath


contsili:
I suggest the ``dccnpath`` help documentation to be updated as:

`````
DCCNPATH manages the filename and path for test files. It helps to locate and read test file from Linux, Windows or macOS computers both inside and outside the DCCN.

Use as 
  filename = dccnpath(filename) 

The default location for FieldTrip and its test data is '/home/common/matlab/fieldtrip'. This function will search-and-replace this string by the location that applies to your computer. It will replace '/home' by 'H:' and will replace forward by backward slashes. 

In case you have a local copy:

1. You should override the default location by 
global ft_default 
ft_default.dccnpath = '/your/copy'; 

2. Note that most test scripts expect data located at /home/common/matlab/fieldtrip/data/ftp or /home/common/matlab/fieldtrip/data/test, hence you should organize your local copy of the data under /your/copy/data/ftp and /your/copy/data/test.

3. dccnpath will check your local copy to find if the files and folders are downloaded. If not, it asks for your permission to download them. Public data are downloaded from https://download.fieldtriptoolbox.org. Private data are available only to people that have access to the DCCN computer cluster.


Copyright (C) 2012-2022, Donders Centre for Cognitive Neuroimaging, Nijmegen, NL
`````

I have already updated ``dccnpath()`` and created its test script ``test_dccnpath()``. I will make a relevant PR for that.

In short, the updated ``dccnpath()`` has kept its old functionality by using three alternatives (alternative0, alternative1, alternative2). Alternative0 finds the right path in the DCCN cluster, alternative1 finds the right path in the working directory and alternative2 finds the right path in the local copy. When alternative2 is used, ``dccnpath`` checks if the files and folders needed to run the corresponding test script are downloaded. If not, it downloads them automatically.

It is important to mention that in order for contributors to use alternative2 they should have manually predefined the path to their local copy by typing: 
```
global ft_default 
ft_default.dccnpath = '/your/copy'; 
```
This is also mentioned clearly in the suggested ``dccnpath`` help documentation

contsili:
Here I will explain more clearly what we want to achieve with the updated `dccnpath()`:

**Input:** the path to where the test data are downloaded in the DCCN cluster, e.g., /home/common/matlab/fieldtrip/data/ftp/testdata (public data) or /home/common/matlab/fieldtrip/data/test/testdata (private data)

[...]

PR #2282, opened 2023-07-15
Update dccnpath

contsili:
This PR is associated with issue #2281.

However, in ``dccnpath`` still two things need to be done (see: `Todo` notes in the ``dccnpath`` script):
1.  `% Todo: FTP connection of private data:` Download **private data** to the local copy from the DCCN cluster. This will apply only for employees and interns of the DCCN that have access to the DCCN intranet. I did not do that yet since I have no connection to the DCCN intranet. I will have connection after 1st of August. 
2.  `% Check if filename exists. If not, download it (Todo: after permission):` Ask the permission of the user prior to downloading the data.

The ``test_dccnpath`` has been tested in Windows and it works. It should also be tested in Mac/Linux.

The list of DEPENDENCIES in the ``test_dccnpath`` are not accurate. They should be checked again. One question is how we are going to label the DATA DEPENDENCY. For now ``test_dccnpath`` only uses public data, but when   `% Todo: FTP connection of private data` is completed it should also work on private data. My suggestion ``% DATA public private``.

Help documentation still needs to be updated based on issue #2281.

Lastly, in ``dccnpath`` something that needs more attention is ``if ~exist(alternative2, 'file')``. This checks if the file or folder exists. If not, it downloads them automatically. However, in the case of the folder it is important to check what files are missing from that folder and not generally if the folder is missing. Because if the folder exists but it is empty then nothing is downloaded while it should have.


contsili:
With the latest commit I added features to `dccnpath()` and the corresponding tests in` test_dccnpath()`, as discussed in the meeting we had with @robertoostenveld on 17/07/2023.

**Questions**
1. When ft_default.dccnpath is not defined and the data are downloaded in the tempdir, shall I add a `warning('local path was not explicitly defined. Using tempdir to store the test data')`? Already, dccnpath() prints at least one warning each time it runs. Maybe is better to not add another one. Anyhow, the path (tempdir in this case) where the data are downloaded is always printed after dccnpath() runs, so the user can see the path also without the warning.

2. For contributors that are connected in the DCCN intranet and have remote access to the private test data, dccnpath() will use the alternative0 to find the right path. Am I right?

3. For alternative1: I kept the old code:
```
elseif endsWith(pwd, alternative1)
% exist(alternative1, 'file') also returns 7 in case the present working directory name matches alternative1
% this is not the match we want
alternative1 = '';
end
```
   Why are we using this chunk of code? If the contributor's current working directory in MATLAB is the one that the test data are downloaded, then why he should not be able to use alternative1 to find the right path?

4. The error that is printed for **typos** in the path is:

   > Error using matlab.internal.webservices.HTTPConnector/copyContentToByteArray
   >     The server returned the status 404 with message "Not Found" in response to the request to URL
   >      https://download.fieldtriptoolbox.org/test/ctf/Subject0.ds.

    This error is printed automatically by MATLAB `webread `function. This error appears when the data need to be downloaded, but also when the data do not need to be downloaded (since they already exist in the local copy). This happens because when the user does a typo to the path, the function always thinks that it is not downloaded (since the path is wrong) and tries to download. Is this solvable?

5. When ft_default.dccnpath is given by the user, dccnpath() should always select alternative2 **or** should first check alternative0 and alternative1 and then if these do not exist go to alternative2? Right now it does the second.

6. Finally, in dccnpath something that needs more attention is if `~exist(alternative2, 'file') || ~exist(alternative2, 'dir')`. This checks if the file or folder exists. If not, it downloads them automatically. However, in the case of the folder it is important to check what files are missing from that folder and not generally if the folder is missing. Because if the folder exists but it is empty then nothing is downloaded while it should have. (I mentioned this also in my previous comment in this PR, but it is important to see if this might cause problems when trying to find the right path with dccnpath())

contsili:
The `dccnpath` and` test_dccnpath` are ready for review @robertoostenveld 

Two issues I want to highlight: 

1. `dccnpath` cannot find the present working path when we call `dccnpath('/home/common/matlab/fieldtrip/data/ftp/test/ctf/Subject01.ds/BadChannels’)` because `BadChannels` is a file with no file extension, so matlab thinks it is a folder.

2. `dccnpath` & `test_dccnpath` can use `fprintf()` or `ft_notice()` to print more nicely their outputs. Right now `dccnpath()` uses the MATLAB command `warning()`.

PR #2285, opened 2023-08-04
Update ft_crossfrequencyanalysis.m

SyanahWynn:
The result of ft_crossfrequencyanalysis (method 'pac') gave me matrix with only zeros
This seemed to be caused by NaNs being present in my data.
Changing line 305 from:
`rsspctrm = reshape(mean(cfcdata,1),[nchan nlf nhf nbin 1]);`
to
`crsspctrm = reshape(mean(cfcdata,1,'omitnan'),[nchan nlf nhf nbin 1]);`
seems to solve the issue.

### DIFF
--- a/ft_crossfrequencyanalysis.m
+++ b/ft_crossfrequencyanalysis.m
@@ -273,7 +273,7 @@ switch cfg.method
     else
       dimord = 'chan_freqlow_freqhigh' ;
       crsspctrm = zeros(nchan,nlf,nhf);
-      cfcdatamean = reshape(mean(cfcdata,1),[nchan nlf nhf nbin 1]);
+      cfcdatamean = reshape(nanmean(cfcdata,1),[nchan nlf nhf nbin 1]);
       
       for k =1:nchan
         pac = squeeze(cfcdatamean(k,:,:,:));
@@ -301,7 +301,7 @@ switch cfg.method
       
     else
       dimord = 'chan_freqlow_freqhigh_phase' ;
-      crsspctrm = reshape(mean(cfcdata,1,'omitnan'),[nchan nlf nhf nbin 1]);
+      crsspctrm = reshape(nanmean(cfcdata,1),[nchan nlf nhf nbin 1]);
       crsspctrm(isnan(crsspctrm)) = 0;
       
     end % if keeptrials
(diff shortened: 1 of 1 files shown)


schoffelen:
Dear @SyanahWynn,

Thanks for this suggestion. Since a long time we have the nanmean/nanstd/nansum functions in fieldtrip that have the same functionality as the (more recently introduced by mathworks) 'omitnan' functionality for mean/std/sum. I think it may be good for consistency (and code maintainability) to either change your suggested change into a call to nanmean, or to change all occurrences of nanmean/nanstd/nansum throughout the fieldtrip codebase into equivalent functional calls to mean/std/sum. 
Given that nanmean/nanstd/nansum are based on mex-files (i.e. compiled code, which is platform (and version)) dependent, I think it is worth seriously considering a change from using nanmean (which by the way used to be a name of a supported mathworks based matlab function) into mean  etc. However, we first need to get a feeling about how far back this 'omitnan' functionality is supported. FieldTrip aims at backward compatibility for at least five years, but ideally longer. Could you try and identify since which matlab version the 'omitnan' functionality is supported for mean/sum/std?

schoffelen:
hmm, I quickly started a matlab2014b on linux, and there the 'omitnan' does not work.

@robertoostenveld, what do you think? Would it be worth considering to do the nanmean -> mean replacement and have an overloaded version of mean in the appropriate matlabltXXXX directory, once we figure out which directory this should be?

(Side note: I initially tried to start an interactive matlab 2012b on the dccn cluster, and it failed due to some java related error -> needs to be replicated, and if so, followed up).

schoffelen:
feel free to reopen and work on this. for now I will merge and close given the lack of response so far.

SyanahWynn:
Hi Jan-Mathijs,

Perfect, thanks. I did not respond as it seemed like something you were discussing with Robert regarding other functions as well.

Best,
Syanah
________________________________
Van: Jan-Mathijs Schoffelen ***@***.***>
Verzonden: vrijdag 1 september 2023 14:03
Aan: fieldtrip/fieldtrip ***@***.***>
CC: Syanah Wynn ***@***.***>; Mention ***@***.***>
Onderwerp: Re: [fieldtrip/fieldtrip] Update ft_crossfrequencyanalysis.m (PR #2285)


feel free to reopen and work on this. for now I will merge and close given the lack of response so far.

PR #2286, opened 2023-08-07
Support README, LICENSE and CHANGES to be specified in the cfg  

### LINKED ISSUE #2084: default readme in BIDS dataset after conversion
hey

We (BIDS maintainers) are trying to see if we can improve the quality of the average README in BIDS dataset. Those tend to be pretty thin and are missing some information we would like to see there even if it cannot be formalized "yet" into the BIDS specification.

Hoping to leverage the bids converters to help to do.

So I am going around the different BIDS converters repo to see if we can start improving the  default README they put in a the BIDS dataset.

There is template that is provided in the BIDS starter kit: https://github.com/bids-standard/bids-starter-kit/blob/main/templates/README.MD

Some possible options:
1. use this template is
2. use a subset of it
3. add a link to the template README
4. something else (a mix of 2 and 3?)

What do you think? I can try to open a PR if you want.

Also let us know if you think that the template README in the starter kit could / should be improved .

Feel free to disregard if you are already using some version of the template README.

robertoostenveld:
This fixes #2084 as for `cfg.README` it now includes the default template from the starter kit.

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -71,7 +71,7 @@ function [cfg] = data2bids(cfg, varargin)
 %   cfg.desc                    = string
 %
 % If you specify cfg.bidsroot, this function will also write the dataset_description.json
-% file. Among others you can specify the following fields:
+% file. Among others, you can specify the following fields:
 %   cfg.dataset_description.writesidecar        = 'yes' or 'no' (default = 'yes')
 %   cfg.dataset_description.Name                = string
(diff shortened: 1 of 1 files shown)


robertoostenveld:
I also reordered the MATLAB help a little bit. 

@Remi-Gau what do you think of it like this?

robertoostenveld:
This is following https://github.com/sccn/bids-matlab-tools/issues/143

PR #2293, opened 2023-08-11
ENH - more robust detection of MEG for planar_combined combos

schoffelen:


schoffelen:
issue: ft_channelselection returned empty when called with the wildcard 'MEG', but the input channel typ was detected to be ctf<something>_planar_combined

PR #2295, opened 2023-08-17
codespell: config, workflow (fail if typos introduced) and some typos fixed

yarikoptic:
For now, for demo, is done only for `utilities`.  
Can slowly expand the coverage -- what would be the next, but also feel free to just take over or take as is with later PRs doing the expansion.


robertoostenveld:
What kind of magic is this 🤯🤔 ?? 

Thanks, I will investigate what this is and how this works as github action. It certainly caught a lot of typo's already!

.codespellrc
```diff
@@ -0,0 +1,8 @@
+[codespell]
+skip = .git,*.pdf,*.svg,external,src
+ignore-words-list = ans,siz,homogenous,pth,nam,ext,dum,siz,toi,whos,als
+# case sensitive etc as regex
+ignore-regex = \b(ALS|correctT)\b|n\(ot a
```
robertoostenveld, line 5: I cannot make sense of this regexp, it seems syntax-wise wrong.

robertoostenveld:
Let's merge, see what happens and figure how to activate the GH action upon a PR.

.codespellrc
yarikoptic: it is python regex, https://docs.python.org/3/library/re.html what in syntax confuses you?

robertoostenveld: I looked it up. 

I was not familiar with `\b`, thought it was a backspace - now I know.

The `n\(ot a` confused me, I thought the `(` was a meta character at the start of a section, but now I see that it is to match `n(ot a landmark)`.

Thakns, all clear now!

PS I have solved it instead with a few more words to ignore, and with a `.codespellignorelines` exclude file.

yarikoptic:
great to see you "won" over codespell action.  I am not quite certain why github complained

[Error: .github#L1](https://github.com/fieldtrip/fieldtrip/commit/67d791388b934a8d5ba4308356c74f9f0cb3b167#annotation_13447906285)
actions/checkout@v3 and codespell-project/actions-codespell@v2 are not allowed to be used in fieldtrip/fieldtrip. Actions in this workflow must be: within a repository owned by fieldtrip.

even after you have merged it into main codebase. Do you remember?

robertoostenveld:
Oh, that was due to the repo settings. Initially/historically I had actions completely disabled in the repository settings, then I switched it to only allow actions from within the fieldtrip organization, giving this error. After I also allowed actions from elsewhere, it worked fine. 

Screenshot 2023-08-18 at 13 27 26

robertoostenveld:
This morning I have been playing around with https://github.com/fieldtrip/playground/actions. I can already see some other things to implement this way, such as [spellcheck](https://github.com/marketplace/actions/github-spellcheck-action) on the complete fieldtrip website. 

Most of the current [automation](https://github.com/fieldtrip/automation) is (due to historical reasons) still with cron jobs and bash scripts. I might also want to reimplement some of that using github actions.

yarikoptic:
If you are really getting into automation, look also into  https://github.com/pre-commit/pre-commit . Here is a repo where we use it quite extensively and you can find pointers to the projects enabled in that config: https://github.com/dandi/dandi-cli/blob/master/.pre-commit-config.yaml

PR #2298, opened 2023-08-23
update dccnpath and added test script

robertoostenveld:
This closes #2282 and #2281 and includes the following improvements

- better documentation of what `dccnpath` does
- better handling of local copies of test data
- automatically download the public data
- various use cases of `dccnpath` are now tested in its own test script

The credits go to @contsili, I only made some cleanups to his work. See also https://github.com/contsili/fieldtrip/pull/6


robertoostenveld:
I identified some other uses of `dccnpath` in the test scripts that are not supported any more. I have also fixed those.

Issue #2301, opened 2023-08-24
Improve code testing by contributors using GitHub actions


robertoostenveld:
*Description:* Add GitHub actions to automatically run a large subset of the tests on pull requests. Then the contributor will receive a message on the tests that failed and need further inspection.

The test scripts can possibly run in the cloud. If that is not the case, then the contributor will be asked if they want to download the data on their local computer. Note, these test scripts can run only on PUBLICDATA/NODATA.

robertoostenveld:
I implemented a github action based on https://github.com/marketplace/actions/create-or-update-comment, which automatically adds a comment to the pull request. The content of the comment comes from a python script that finds the appropriate test scripts.

vijayiyer05:
I believe you already know, but just in case: this issue could leverage the [open availability of MATLAB on GitHub Actions for public codebases](https://www.mathworks.com/solutions/continuous-integration.html).  
![image](https://github.com/fieldtrip/fieldtrip/assets/23032671/1eb0ea71-018b-4ac0-af6b-643b738e983c)

If any questions or issues arise on this, please don't hesitate to reach out.

robertoostenveld:
I have been trying out this action in #2302, #2303, #2304, #2305 and #2306.

It was only in #2306 that I got it to work.

PR #2302, opened 2023-08-24
small change to the help of ft_preprocessing 

robertoostenveld:
to test the GH action in #2301.


robertoostenveld:
I see that the `suggest_tests.yml` was not executed, probably because it includes actions that were not whitelisted. I changed the repo settings to allow actions from anywhere.

robertoostenveld:
the action to suggest tests ran, but was not able to place a comment. Let's merge this and try again.

PR #2306, opened 2023-08-24
small change to test #2301

robertoostenveld:


robertoostenveld:
whohoo!!! 

It works.

robertoostenveld:
It would be nice to reformat the output a bit, such that it is easier to copy-and-paste the test scripts into MATLAB. Something that would work nicely is 

```
edit test_001 test_002 test_003
```

which allows a quick inspection of the test script prior to clicking the "run" button to execute each of the scripts.

Also, since the comment is formatted in Markdown, we can use a code block to distinguish the part to be evaluated in MATLAB from the explanation.

PR #2308, opened 2023-08-29
code improvements to implement BIDS derivatives pipeline

robertoostenveld:


robertoostenveld:
for reference: I think that the "suggest-test" action of which the result is available under "Checks" failed because the branch from which the PR originated was slightly too far behind on the master branch. The "suggest-test" action on the master branch had in the meantime been updated.

PR #2311, opened 2023-08-30
updated walltime to more realistic value

robertoostenveld:
There were two test scripts requesting a walltime of 10 hours, but only using a few minutes. I updated their walltime to 10 minutes.

### DIFF
--- a/.github/workflows/suggest_tests.yml
+++ b/.github/workflows/suggest_tests.yml
@@ -4,7 +4,7 @@ on:
   pull_request_target:
 
 jobs:
-  suggest-test:
+  suggest-tests:
     runs-on: ubuntu-latest
     permissions:
       pull-requests: write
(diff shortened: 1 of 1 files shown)


robertoostenveld:
The "suggest-tests" action continues to fail, even though I updated the `yml` to make it conditional on whether any tests were found. I don't understand and cannot find online which `yml` is being used in the test, but it seems that it is the one from the master, not the one from the PR branch. That means I first have to merge this before we can check the PR comment being processed correctly.

robertoostenveld:
Ah, I think that is what the `on: pull_request_target` in the YML refers to. The YML comes from the target (in this case "main" on fieldtrip/fieldtrip) and not from the PR branch (in this case "walltime" on robertoostenveld/fieldtrip).

PR #2312, opened 2023-08-30
updated walltime to more realistic time

robertoostenveld:
this is a similar PR as #2311, it updates three more test scripts


robertoostenveld:
The "suggest-tests" action now ran through correctly, showing in the action details that there are no suggestions, and not posting a comment. 

Rather than repeatedly updating the YML, merging into master, and updating the PR, the next time I might beter try to use https://github.com/nektos/act to debug the github action.

PR #2313, opened 2023-08-30
improve representation of events and triggers following fieldtrip2fiff

schoffelen:
This PR intends to improve the writing of fif-files from FieldTrip, such that as much as possible information is retained:
-specifically: if a raw data structure with a single trial is written as a continuous file, and if there's a digital stim channel, the stim channel will be represented in the header such that it is correctly recognized as such upon reading the file into fieldtrip, which is good for ft_read_event
-also it is now possible to write an event(-like) structure into a fif-file, i.e. not as a companion <blabla>-eve.fif file, but directly into the bigger data file. I couldn't any documentation about whether this is not allowed, and it seems less random as compared to more arbitrary file naming conventions.
-note that only the sample index of the event is stored (along with a numeric code, and potentially a mapping that allows for a richer type/value definition). The offset and duration fields of a fieldtrip event structure can not be represented in the fif-file.

To be done:
-rearrange the code flow a bit to make it more generic (e.g. in the detection of stimulus channels), if possible
-add some checks and balance + ft_info (e.g. throw an error if an event structure is passed into the fieldtrip2fiff when the input data is not raw).

### DIFF
--- a/external/mne/fiff_write_events.m
+++ b/external/mne/fiff_write_events.m
@@ -35,7 +35,12 @@ eventlist = reshape(eventlist',numel(eventlist),1);
 %
 %   Start writing...
 %
-fid  = fiff_start_file(filename);
+if ischar(filename)
+  fid = fiff_start_file(filename);
+else
+  % assume the supplied filename to be a file identifier to an open file
+  fid = filename;
+end
 
 fiff_start_block(fid,FIFF.FIFFB_MNE_EVENTS);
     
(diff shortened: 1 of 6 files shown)


schoffelen:
Oh, and to be done too -> test whether MNE-python likes the new files.

robertoostenveld:
> You should test whether your modifications do not break anything.

Hmm, the github bot repeatedly coming with the same suggestion for testing is not that useful. Better would be if it were to post the comment once and then update it if needed.

schoffelen:
I have tested test_fieldtrip2fiff (and added a test case for the issue that sparked this PR). I will merge for now.

PR #2314, opened 2023-08-30
export raw and timelock data to CTF dataset 

robertoostenveld:
This is desired for implementing BIDS derivatives for EEG and MEG, which are to be stored in a similar file format as the raw BIDS datasets.


robertoostenveld:
The suggested tests seem to be due me forking this off from an older version of the master branch. In the mean time, a whole bunch of test scripts were updated in the master, not in the branch that is part of the PR.

PR #2325, opened 2023-09-28
ENH - Replace non-linear orientation optimization in the SAM spinning…

MalteHoel:
… case with an equivalent approach based on eigendecomposition

I think that the "spinning" case in ft_inverse_sam.m uses a non-linear optimization algorithm to solve a problem that can equivalently be solved by eigendecomposition. I will explain below why I think this is the case.

I think it might be good to replace the non-linear algorithm with the one based on eigendecomposition for the following reasons.

1.  It is more in line with the remaining cases in ft_inverse_sam.m (i.e. "robert", "gareth", and "moissev"), which are also based on eigendecomposition.
2. In my test case, it produced slightly better results. I followed the fieldrip beamformer tutorial found under https://www.fieldtriptoolbox.org/tutorial/beamformer_lcmv (with "lcmv" replaced by "sam", and after removing the "leadfield" field from the lf datastructure so that the determination of "meansphereorigin" does not fail). Then, the value of the cost function, defined in SAM_costfun.m, was, up to rounding error, consistently smaller in the eigendecomposition-based approach. This makes sense, since the nonlinear algorithm searches for the minimizer up to some tolerance, while the eigendecomposition numerically computes the exact minimizer.
3.  It is more efficient. After replacing the non-linear algorithm with the eigendecomposition-based algorithm, the call to ft_inverse_sam in the above test scenario was sped up by a factor of x1.7 on my machine.

I want to apologize if this is already well-known to the developers, and there is some reason for preferring the non-linear optimization algorithm that I am not aware of. If this is the case, I think it would be nice to add a comment to the code mentioning the equivalent eigendecomposition approach and briefly explaining the reasoning for not using it.

# Equivalence of non-linear approach and eigendecomposition approach
I will now explain why I think these two approaches are equivalent. If this is already well-known, please ignore the following.

In the non-linear approach, the optimization is performed over some angle $\theta$. This angle is only ever used after a call to settan.m, where the first step is computing the vector 

```math
\eta = \begin{pmatrix}\cos(\theta)\\ \sin(\theta) \end{pmatrix}.
```

Hence the optimization is effectively performed over the unit circle in $\mathbb{R}^2$. Furthermore, ft_inverse_sam at the start of the "spinning" orientation optimization computes two vectors $u$ and $v$, which form an orthonormal basis of the corresponding tangential plane. Denoting by $O = (u, v) \in \mathbb{R}^{3 \times 2}$, we see that settan computes $(O \cdot \eta)^\intercal$. Now denote by $n$ the number of sensors, $L \in \mathbb{R}^{n \times 3}$ the leadfield, $C$ the covariance matrix, $N$ the noise covariance matrix, and $C_{\text{inv}}$ the inverse covariance matrix (I am not writing $C^{-1}$ since the "inverse covariance matrix" is computed using ft_inv, and thus e.g. due to regularization might differ from the exact inverse). Taking a close look at SAM_costfun.m, we see that it computes the value

```math
f(\eta) = \sqrt{\frac{\eta^\intercal \cdot O^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot N \cdot C_{\text{inv}} \cdot L \cdot O \cdot \eta}{\eta^\intercal \cdot O^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot C \cdot C_{\text{inv}} \cdot L \cdot O \cdot \eta}}.
```

This function is then minimized using a non-linear optimization algorithm. Now note that since $f$ is a nonnegative function, we have
```math
\arg \min_{\|\eta\| = 1} f(\eta)
=  \arg \min_{\|\eta\| = 1} f^2(\eta)
= \arg \max_{\|\eta\| = 1} \frac{1}{f^2(\eta)}.
```
But if we set $A = O^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot C \cdot C_{\text{inv}} \cdot L \cdot O$ and $B =  O^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot  N \cdot C_{\text{inv}} \cdot L \cdot O$, we have
```math
\arg \max_{\|\eta\| = 1} \frac{1}{f^2(\eta)} = \arg \max_{\|\eta\| = 1} \frac{\eta^\intercal A \eta}{\eta^\intercal B \eta}.
```
But this is now a generalized Rayleigh quotient, which is maximized in an eigenvector corresponding to the largest generalized eigenvalue of the pair $(A, B)$. I refer to Sekihara et. al., Adaptive Spatial Filters for Electromagnetic Brain Imaging, Chapter 13.3, for an argument on why the optimization problem is equivalent to the eigenproblem (or to the Wikipedia article on Rayleigh quotients if one just wants a reference to the statement). It is now exactly this eigenproblem that I implemented in the commit corresponding to this pull request.

Furthermore, I also added the computation of the pseudoZ value for the "spinning" case via the SAM_costfun function. This was primarily done to compare the cost function values of the eigendecomposition-based approach with the cost function values computed using the non-linear optimization approach. Should you consider this code for merging, this function can either be deleted or implemented more efficiently without reference to SAM_costfun.m.

With kind regards
Malte


schoffelen:
Hi Malte, Thanks for your very clear explanation, and sorry that we missed each other when I visited Muenster a few weeks ago (and you were ill at home I believe). I like your suggestions very much. The SAM beamformer is not so often used as far as I know, and I haven't looked at the code myself for a long time. Your suggested changes to the 'spinning' fixedori method make perfect sense. 

Given all that you have written in your comment, do you think that there is any merit in the Gareth and Robert methods? 

The Gareth method assumes an identity noise matrix, and indeed assumes C to be non-regularised (so that invC * C *invC reduces to C). Also, it does not do the projection on the plane, but I would hope (but not expect) that the eigenvalue decomposition knows how to deal with the ill-conditioned matrices. -> as a side note I think that the dependence on the meansphereorigin can be removed, because tanu/tanv can be computed from the svd-decomposition of the dipole's leadfield, no? 

Anyway, could your suggestion not replace the gareth method altogether, where your suggestion (almost reduces) to Gareth's method, when noise = eye (or not specified by the user)?

Also, I think that the Robert methods makes least sense of all.

The Moiseev method I don't understand, but I think we should keep that in.

MalteHoel:
Hi Jan-Mathijs,
thank you very much for your quick answer! During your visit to Münster, I was at home with Corona, otherwise, I would have loved to discuss these topics in person.

I completely agree with your sentiment about the 'gareth' and 'robert' methods. I think the 'robert' case computes the same orientation and filter as the 'fixedori' + 'unitgain' case in ft_inverse_lcmv.m, so removing it would also remove code duplication, and would not really take anything away from users.

Furthermore, I think the remaining cases ('spinning', 'gareth', and 'moiseev') can all be framed in the same setting. Concretely, if '~isempty(toi)' is false in the 'moiseev' case, this method computes the orientation by maximizing
```math
f(\eta) = \frac{\eta^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot L \cdot \eta}{\eta^\intercal  \cdot L^\intercal \cdot C_{\text{inv}} \cdot N \cdot C_{\text{inv}} \cdot L \cdot \eta}
```
which up to the projection on the tangential plane and the identification of $C_{\text{inv}}$ with $C_{\text{inv}} \cdot C \cdot C_{\text{inv}}$ agrees with the eigendecomposition based approach for the 'spinning' case.

Furthermore, if we assume the noise covariance matrix to be of the form $N = \sigma \cdot I$, where $I$ denotes the identity matrix, the quotient above reduces to 
```math
f(\eta) = \frac{1}{\sigma} \cdot \frac{\eta^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot L \cdot \eta}{\eta^\intercal  \cdot L^\intercal \cdot C_{\text{inv}}^2 \cdot L \cdot \eta}
```
and since the factor $\frac{1}{\sigma}$ does not matter for the computation of the optimizer, we see that the optimization problem is equivalent to the generalized eigenproblem of the 'gareth' approach. Hence the 'gareth' approach can be seen as a special case of the 'moiseev' approach.

(On a side note, in the 'moiseev' case with '~isempty(toi)', one does not use $C_{\text{inv}}$ in the numerator but instead uses $C_{\text{inv}} \cdot \widetilde{C} \cdot C_{\text{inv}}$, where $\widetilde{C}$ is not the original covariance matrix, but a new estimation of it over some custom time interval. Do you know why this option is here? Is there some use case where it is beneficial to compute $C_{\text{inv}}$ and $\widetilde{C}$ from two different time intervals?)

In total, do you think it might be a good idea to go a step further and merge the three approaches 'spinning', 'gareth', and 'moiseev' into a single approach? The code of this new method might then look somewhat similar to the current code of the 'moiseev' approach, with an additional option to project to the tangential plane. This would then unify the 'spinning' and the 'moiseev' cases. If one then uses $\sigma \cdot I$ for the noise covariance matrix, the behavior of the 'gareth' approach is reproduced.

Furthermore, I really like your suggestion of replacing the computation of the tangential plane via some 'meansphereorigin' with a method based on a singular value decomposition of the leadfield. This should remove the restriction of ft_inverse_sam + 'spinning' to forward models where 'meansphereorigin' makes sense, right?

Thank you again for your feedback on this pull request! I already look forward to your thoughts on this!

schoffelen:
Duuhh, I have messed up the PR, probably because I branched off from a more recent FieldTrip version before I pulled in your suggested changes, and then pushed back again. Sorry about that, let me quickly whether this only messes up the history, or otherwise try to undo it.

My intention was to already give some suggestions in the code to move forward, and then to discuss some more details in the conversation.

schoffelen:
I think that the code changes in the other functions will be harmless, but since this PR is now a bit messed up, I would like to propose to close it, and create a new one, unless you know how to easily cherrypick the relevant commit in this PR and discard the rest. Would that be alright with you @MalteHoel? I could do it (noblesse oblige), but perhaps you want to do it, to have a clean digital trace back to you, so that credit can be given where it's due. (not credit for messing up the PR, but credit for the suggested code changes). 
Sorry for the extra work.

My suggested way forward would be to save the version of ft_inverse_sam in this PR, close the PR, and create a new branch with only the changes to ft_inverse_sam (and copy over the discussion from above into the new PR.

MalteHoel:
Hi Jan-Mathijs,
if I understand your last sentence right we should continue the discussion in this pull request? I hope I got this right, otherwise, I will create a new pull request.

In your comment above you suggested merging the 'gareth'-case with the 'spinning'-case, and I suggested also merging the 'moiseev'-case, so that after removing the 'robert'-case there would only be one option for orientation estimation. In the commit above I tried to implement this. Compared to the previous changes, this commit is more invasive, and I would like to hear your opinion on if this goes too far.

I want to briefly describe my thought process for the implementation. 

As a first change, you suggested replacing the estimation of the tangential plane in the 'spinning'-case, which was previously based on fitting a sphere to the head model, with an estimation based on a singular value decomposition of the lead field. If you think we should not do this, you can ignore my last two commits and I will try again :D. Now, as described above, the spinning case computes the generalized eigenvectors of the pair $A = O^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot C \cdot C_{\text{inv}} \cdot L \cdot O$ and $B = O^\intercal \cdot L^\intercal \cdot C_{\text{inv}} \cdot N \cdot C_{\text{inv}} \cdot L \cdot O$. Here, the columns of $O$ are an orthonormal basis of the tangential plane. If now $L = U \cdot S \cdot V^\intercal$ is an SVD of $L$, we would choose $O = V(:, 1:2)$. But now $L \cdot V(:, 1:2)$ is the rank-reduced lead field, and thus the 'spinning'-case with SVD-derived tangential plane is, up to the identification of $C_{\text{inv}} \cdot C \cdot C_{\text{inv}}$ with $C_{\text{inv}}$, equivalent to the 'moiseev'-case with the parameter 'rankreduction'=2 and 'isempty(toi)' evaluating to 'true'.

In total, if we estimate the tangential plane from the lead field, we can remove the 'spinning'- and 'gareth'-cases without losing functionality, since
1. the 'moiseev'-case with a noise covariance matrix of the form $\sigma \cdot I$ reproduces the 'gareth'-case.
2. the 'moiseev'-case with rankreduction set to 2 (and without specifying a 'toi') reproduces the modified 'spinning'-case. 

This is essentially the idea behind my commit. The 'fixedori'-case distinction is removed, and the 'rankreduction' parameter controls whether the whole orientation in $\mathbb{R}^3$ is estimated (for 'rankreduction' = 3), or if only the tangential component of the orientation is estimated (for 'rankreduction' = 2).

The other changes are only minor, I will quickly go over them.
1. I chose to use $C_{\text{inv}}$ over $C_{\text{inv}} \cdot C \cdot C_{\text{inv}}$, since the exact expression would be $C^{-1}$, and thus $C_{\text{inv}}$ makes more sense to me.
2. One difference between the 'spinning'-case and the 'gareth'/'moiseev'-cases with 'rankreduction'=2 is that the 'spinning'-case embeds the orientation back into $\mathbb{R}^3$, while 'gareth'/'moiseev' export only the tangential coordinates of the estimated orientation (i.e. a vector in $\mathbb{R}^2$). I think the behaviour of the 'spinning'-case makes more sense, since without some information about the full leadfield, you cannot reconstruct the 3-dimensional embedding of the 2-dimensional coordinates. My approach in the commit was to always compute the full lead field in $\mathbb{R}^{n \times 3}$ ($n$ number of sensors), and perform the potential rank reduction directly in 'ft_inverse_sam.m' since I then have access to the tangential plane defined by the leadfield and can embed the tangential coordinates of the estimated orientations back into 3d-space.
3. If 'noisecov' is empty, in the previous implementation the 'spinning'-case estimates the noise covariance as $\sigma \cdot I$, while the previous implementation of the 'moiseev'-case threw an error. In the merged approach I chose the behaviour of the 'spinning'-case.
4. In the previous implementation, there was a check if the headmodel is 'singlesphere' (and 'fixedori' is not 'moiseev'). If this was the case, a check was performed if the estimated orientation corresponds to the radial direction. I think this can be avoided if, in the case of a MEG 'singlesphere' head model, we only try to estimate the tangential component of the orientation in the first place. This seems more natural to me. I thus added a check so that in the MEG 'singlesphere' case one always only tries to estimate the tangential part of the orientation.

I hope this explains why I implemented the code in the way I did. I would love to hear your thoughts on the suggested changes.

schoffelen:
Hi Malte,

This really starts looking great. The only thing I don't fully understand is your motivation to call the ori field in case of the tangential computation tangential_ori. Is there a specific reason for this? Apart from this it's pretty much good to go I think.

MalteHoel:
Hi Jan-Mathijs,

my thought was that a user might assume that 'ori' is an estimation of the complete dipole orientation. By using 'tangential_ori', I thought that it might be more transparent that the search space for the orientations is restricted to the tangential plane, and thus any radial component is ignored.

But I now see that this might just add extra confusion and make the handling of the output more cumbersome. Furthermore, the documentation in ft_inverse_sam.m already describes under what conditions only tangential orientations are considered. Do you think it might thus be better to always call the orientation estimate 'ori'?

schoffelen:
Yes, I think ori is just fine. As a side note, if we precompute leadfields with a reducerank = 2, the resulting ori of any other inverse method will always be in the plane as well...

MalteHoel:
I have removed the usage of 'tangential_ori' and replaced it with 'ori'.

schoffelen:
Thanks! almost there. Could I ask you to try and merge the master branch from github.com/fieldtrip/fieldtrip into your PR branch, in a final attempt to clean up the detected file changes here?

MalteHoel:
It seems like it worked, it now only shows 3 changed files.

PR #2327, opened 2023-10-05
FIX #2326 - fixed time range selection for select_topoplotER in ft_multiplotER

### LINKED ISSUE #2326: Time label mismatch in 'ft_multiplotER' and 'ft_topoplotER' integration
**Describe the bug**
When using the 'ft_multiplotER' function to plot MEG time courses and then attempting to generate a topology plot for a specific time window, there is an issue where the time label displayed on the topoplot does not correspond correctly to the manually selected time range.

**To Reproduce**
1. call 'ft_timelockanalysis' to calculate timelocked average ERF
2. call 'ft_multiplotER' to create an interactive figure for event-related fields versus time
3. manually select a specific time window within the interactive plot 
4. click to call 'ft_topoplotER' and generate topology plot by averaging ERF across specified time window

`
%% time lock analysis
cfg=[];
tlc=ft_timelockanalysis(cfg,data);

%% interactive topology plot
figure;
cfg = [];
cfg.layout = 'CTF275_helmet.mat';
cfg.viewmode = 'butterfly';
% cfg.colorgroups = 'sequential';
ft_multiplotER(cfg, tlc);
`

**Expected behavior**
the time label on the left bottom on topoplot should correspond to the time range specified earlier in the interactive plot

**Screenshots**
image

**Environment and versions (please complete the following information):**
 - Operating System: centos
 - MATLAB version R2021b

robertoostenveld:
The problem reported by @RaoXie in #2326 was due to the time range selection assuming that the x-axis of the figure corresponded to the time scale of the data. That is not the case, as the figure is still constructed by plotting many channels on top of each other according to the butterfly layout.  Whenever a layout is involved ft_plot_vector makes use of local pseudo-axes.
 
Furthermore
- renamed viewmode layout into topographic
- do not recursively store the topographic layout in the butterfly layout
- added check and informative error for labelcharN
- make chanindx selection in pointsymbol, pointsize, pointcolor

### DIFF
--- a/ft_multiplotER.m
+++ b/ft_multiplotER.m
@@ -52,17 +52,14 @@ function [cfg] = ft_multiplotER(cfg, varargin)
 %                       can be a single style for all datasets, or a cell-array containing one style for each dataset
 %   cfg.linewidth     = linewidth in points (default = 0.5)
 %   cfg.linecolor     = color(s) used for plotting the dataset(s). The default is defined in LINEATTRIBUTES_COMMON, see
-%                       the help of this function for more information
+%                       the help of this function for more information.
 %   cfg.directionality = '', 'inflow' or 'outflow' specifies for connectivity measures whether the
 %                       inflow into a node, or the outflow from a node is plotted. The (default) behavior
 %                       of this option depends on the dimord of the input data (see below).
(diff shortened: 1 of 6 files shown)


robertoostenveld:
I tested a few of the suggestions but not all of them.

PR #2331, opened 2023-10-13
ENH - add insideonly option

schoffelen:


schoffelen:
Note: I tested the first set of suggested files, apart from the beamformingextended tutorial.

PR #2335, opened 2023-11-03
FIX: t-value to p-value conversion in ft_regressconfound

StolkArjen:
The t-value to p-value conversion was missing an absolute operation, as also suggested here:
https://www.mathworks.com/matlabcentral/answers/175091-how-to-calculate-2-tailed-p-value-using-t-value-and-degree-of-freedom

Thanks to Ben Graul (Dartmouth College) for spotting the issue.


schoffelen:
well spotted (I guess)! So I assume that the intended p-value is a two-sided one. And as a matter of fact, the old computation was incorrect, even for a one-sided one, right?

StolkArjen:
Indeed, it's two-sided. I believe the old definition contained a typo, occasionally yielding negative or otherwise insensible p-values.

PR #2337, opened 2023-11-07
Use single quote

gllmflndn:
For backward compatibility with versions of MATLAB where the `string` class is not implemented.


robertoostenveld:
thanks for catching this!

PR #2338, opened 2023-11-08
Use single quote

gllmflndn:
Other cases spotted after #2337.


robertoostenveld:
I just tried catching them with test_checkcode, but got too many false positives.

Issue #2188, opened 2023-02-10
error with dipole scan + EEG precomputed leadfield 


mcpiastra:
hi, I have been trying to compute a dipole scan with an EEG pre-computed leadfield and timelocked data.
This is the code I used for the dipole scan:
```
input_fem = [];
input_fem.time = 1;
input_fem.label = avg_787.label;
input_fem.elec = elec_787;
input_fem.avg = avg_787.avg;

cfg_dipfit = [];
cfg_dipfit.numdipoles   = 1;                              
cfg_dipfit.headmodel    = [];
cfg_dipfit.sourcemodel  = lf_787;
cfg_dipfit.gridsearch   = 'yes';                           
cfg_dipfit.nonlinear    = 'no';
cfg_dipfit.elec         = elec_787;                           
cfg_dipfit.latency      = 0.01;                               
cfg_dipfit.channel      = 'EEG';
cfg_dipfit.reducerank   = 'no';

dipfit              = ft_dipolefitting(cfg_dipfit,input_fem);
```
where avg,elec,lf_787 are the timelocked data, elec file and leadfield for dataset 787, resp.

The error:
```
Unrecognized field name "cond".

Error in eeg_infinite_dipole (line 50)
cond     = vol.cond;

Error in ft_compute_leadfield (line 488)
      lf = eeg_infinite_dipole(dippos, sens.elecpos, headmodel);

Error in ft_dipolefitting (line 578)
      lf = ft_compute_leadfield(dip.pos, sens, headmodel,
      leadfieldopt{:});

Error in show_results_787 (line 83)
dipfit              = ft_dipolefitting(cfg_dipfit,input_fem);
```
is given by the fact that after identifying the dipole with minimum error (ft_dipolefitting line 497), the leadfield is computed again in ft_dipolefitting line 578, where the headmodel is assigned to be "infinite", instead of retrieving the leadfield that is already computed.

mcpiastra:
after discussing it with @schoffelen, I will make some changes in my local fork and see if I can fix it locally.
The idea would be to copy the corresponding leadfield along with the dipole that gives the minimum error, so that this leadfield can be reused in line 578 for the "compute the model potential distribution and the residual variance".

mcpiastra:
I do not know how to change the label btw

schoffelen:
Indeed, just to add some info:
[...]

PR #2339, opened 2023-11-09
do not recompute lf if already existing

mcpiastra:
this relates to the issue #2188 
after the dipole scan, the leadfield should be recomputed only if not already existing

### DIFF
--- a/ft_dipolefitting.m
+++ b/ft_dipolefitting.m
@@ -574,8 +574,10 @@ end
 switch cfg.model
   case 'regional'
     if success
-      % re-compute the leadfield in order to compute the model potential and dipole moment
-      lf = ft_compute_leadfield(dip.pos, sens, headmodel, leadfieldopt{:});
+      if ~lf
+          % if there is no leadfield, re-compute it in order to compute the model potential and dipole moment
+          lf = ft_compute_leadfield(dip.pos, sens, headmodel, leadfieldopt{:});
+      end
       if isfield(dip, 'mom') && isfield(dip, 'ampl')
         % the orientation and amplitude have already been estimated, this applies to the case of a fixed dipole orientation
         dip.pot = (lf * dip.mom) * dip.ampl;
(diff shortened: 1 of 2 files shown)


schoffelen (commented):
Hmmm, are you sure that this change will work? Also in general?

Specifically, you check for ~lf, but this will throw an error if the variable lf does not exist in the function. This could be the case for instance if someone runs a nonlinear fit only (i.e. without gridsearch).
Also, generically, lf to me seems the forward solution to one of the dipoles in the grid, which is updated in the for-loop across the dipoles in the grid. The lf outside the gridsearch loop is just the leadfield of the last dipole in the grid, and not necessarily the one that gave the optimal solution.

Or am I missing something?

mcpiastra:
you are right, sorry.
I now modified the code so that the corresponding leadfield is copied along the dipole, and there is a check on the field of of dipole.
does it make sense to you now?
thanks.

ft_dipolefitting.m
```diff
@@ -590,8 +596,12 @@
   case 'moving'
     for t=1:ntime
       if success(t)
-        % re-compute the leadfield in order to compute the model potential and dipole moment
-        lf = ft_compute_leadfield(dip(t).pos, sens, headmodel, leadfieldopt{:});
+        if ~isfield(dip, 'lf')  
+            % if there is no leadfield, re-compute it in order to compute the model potential and dipole moment
+            lf = ft_compute_leadfield(dip(t).pos, sens, headmodel, leadfieldopt{:});
+        else
+          lf = dip.lf;
```
schoffelen, line 603: would you need to consider the loop-variable (and also in line 466?)?

schoffelen:
Hi @mcpiastra, have you tried to run the suggested test functions on your end (given the changed code)? Let me know if those functions ran through without error, then I will merge.

mcpiastra:
hi @schoffelen, after the last commit, yes, all the suggested tests run with no error.
thanks!

schoffelen:
Hi @mcpiastra what needs to be done to close/merge this?

mcpiastra:
Hi @schoffelen , actually I was recently having the same question. What shall I do? Write a short script where this is used, with a dummy leadfield?

schoffelen:
Well, indeed it would be good to have a snippet of code (either in an appropriate already existing test function, or in a new test function), that just goes through the movements.
This would mean a call to ft_dipolefitting with a precomputed leadfield, with gridsearch = 'yes', and without the nonlinear optimization step. If it runs through without error, you're done.

PR #2347, opened 2023-11-23
ENH - speed up the reading of well-structured, but clunky fif-files

schoffelen:
…fif-files

### DIFF
--- a/external/mne/fiff_make_dir_tree.m
+++ b/external/mne/fiff_make_dir_tree.m
@@ -31,7 +31,7 @@ function [tree, last] = fiff_make_dir_tree(fid,dir,start,indent)
 
 %
 %  Define the relevant constants here
-%  no need to get the whole fiff_define_costants
+%  no need to get the whole fiff_define_constants
 %
 FIFF_BLOCK_START     = 104;
 FIFF_BLOCK_END       = 105;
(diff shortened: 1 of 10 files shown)


schoffelen:
In working with the first DCCN-recorded fif-files of the OPM system, we noticed a significant slowdown in file handling.

The main cause of this is the fact that the Fieldline data are organized as a one-sample-per-tag fif-file, which results in a massive fileio overhead. This is partly due to the low-level readers (from external/mne), and partly due to the overhead as implemented in ft_read_header (basically requiring the fif directory tree to read several times).

The code changes in this PR intend to speed up the reading (just in a handwavy way: reading on my Macbook pro a ~7.5 minute continuous data file required about 4 to 5 minutes originally), so any speed up will be welcome.

As a side note, the MNE-python also rely on heuristics that are similar to those in the old matlab code, so they will also suffer from slow reading (although overall the reading is faster than in matlab in general)

schoffelen:
tagging @tc-fieldline and @HXKnappe in case they're interested in how this develops :).

britta-wstnr:
Thanks @schoffelen
We have an issue up here: https://github.com/mne-tools/mne-python/issues/12234

For completeness/comparison: MNE-Python takes 77 seconds to read and load a 7:40 minute data set with such sample-per-tag representation on a MacBook Air -- but is also slower than for segments-per-tag data.

schoffelen:
Merging this for now, to keep things moving forward. Sort out with the mne-matlab repo later.

PR #2353, opened 2023-12-04
FIX - updated function to latest version from mne-matlab

schoffelen:


schoffelen:
Also updated in mne-tools/mne-matlab

PR #2356, opened 2023-12-06
FIX #2352 - open file using ASCII-US encoding

### LINKED ISSUE #2352: issue for read Nicolet .e file
**Describe the bug**
When reading my Nicolet .e file, in the script: read_nervus_header.m and the section of function channelInfo = read_nervus_header_ChannelInfo(h, StaticPackets, Index, ITEMNAMESIZE, LABELSIZE)

The nrIdx always very large and cause the loop to run even position already out of the actual file size, and it would take 3-5 minutes to read the file. 
for example: the log channel info nrIdx:  [2293760 4587520]
so i add bewlow code to stop the loop when it read exceed the file size, and it resolved the issue. Reading file could be complete in seconds.
>  if (ftell(h) >= filesize)
>     break;
>   end
full loop is below: 
for i = 1: nrIdx(2) 
  channelInfo(i).sensor = deblank(cast(fread(h, LABELSIZE, 'uint16'),'char')');
  channelInfo(i).samplingRate = fread(h,1,'double');
  channelInfo(i).bOn = logical(fread(h, 1 ,'uint32'));
  channelInfo(i).lInputID = fread(h, 1 ,'uint32');
  channelInfo(i).lInputSettingID = fread(h,1,'uint32');
  channelInfo(i).reserved = fread(h,4,'char');
  fseek(h, 128, 'cof');
  if (ftell(h) >= filesize)
    break;
  end
end

robertoostenveld:
@schoffelen I wonder whether we should do this more explicitly for other file formats as well. We don't encounter this problem very often, but this certainly not the first time. 

At other places in the code we do it explicitly when reading, for example

```
char(fread(fid, 16, 'uint8=>char'))
```

I think that any piece of code that freads a `char` without explicit specification of the character encoding is at risk.

### DIFF
--- a/fileio/private/read_nervus_header.m
+++ b/fileio/private/read_nervus_header.m
@@ -34,7 +34,7 @@ UNITSIZE = 16;
 ITEMNAMESIZE  = 64;
 
 % ---------------- Opening File------------------
-h = fopen_or_error(filename,'rb','ieee-le');
+h = fopen_or_error(filename,'rb','ieee-le','US-ASCII');
 
 nrvHdr = struct();
 nrvHdr.filename = filename;


schoffelen:
Noted. I suggest to merge this now nonetheless, there's currently no bandwidth to look into this.

PR #2358, opened 2023-12-14
add FieldLine beta2 helmet layouts

robertoostenveld:
this adds the beta2 layouts, tagging @schoffelen and @tc-fieldline.


robertoostenveld:
this also fixes #2359 and adds `doublecirclecross` and `helmet` to ft_prepare_layout

robertoostenveld:
I have executed all suggested tests and there were no issues.

robertoostenveld:
There is a separate layout for bx, by, and bz, and a combined one that combines the channels as triplets like for neuromag306all. They are now also shown on https://www.fieldtriptoolbox.org/template/layout/#fieldline-opm-system

Issue #2365, opened 2024-01-10
Document and implement OPM coregistration on basis of only Polhemus 3D tracker.


robertoostenveld:

robertoostenveld:
mentioning @contsili 

This requires that the sensor positions PLUS the fiducial/marker positions are known. The markers are clicked and expressed in polhemus coordinates (relative to the head), which  allows the sensor positions to be computed relative to the head.

This requires a template sensor array, which we already have in fieldtrip/template/gradiometer. 

I suggest that we add fiducial locations to the applicable gradiometer templates, just as we have it for `elec` structures. I believe that ft_transform_geometry will automatically work on those fiducial positions, just as on coilpos, coilori, and chanpos. 

We do have a template for fieldlinealpha1, but don't have a shared template for fieldlinebeta2 (we do have it, but it is not in FieldTrip yet).

robertoostenveld:
Hi @tc-fieldline, 

I presume that there are markers on the alpha1 helmet similar as on the beta2 helmet. Can you give me their coordinates? I will then add them to the `template/gradiometer/fieldlinealpha1.mat` structure for consistency. copy them here, or send me an email with an excel file or so. 

thanks, Robert

contsili:
I also have written code for the co-registration of the helmet with head points using the polhemus and the co-registration using the 3 HPI coils for a 32-sensor recording and a full 144-sensor recording. In both cases, a magnetic phantom was used.

robertoostenveld:
This had been done by now, it is one of the methods described in https://www.fieldtriptoolbox.org/tutorial/coregistration_opm/

PR #2366, opened 2024-01-10
improve templates for FieldLine OPM helmets

robertoostenveld:
- added beta2 helmet, including fiducials
- rotate alpha1 helmet 90 degrees
- update plotting functions to deal with fiducials

Note that `headshape.fid` was already there for reading and plotting, and `mri.fid` is used in reading (but not plotting). Furthermore, `sens.fid` is already used when reading some of the EEG electrode formats.


robertoostenveld:
Oh, I should have mentioned that this contributes to the solution for #2365.

PR #2367, opened 2024-01-12
Update paper.md

matsvanes:
Update funding information (acknowledgements)


robertoostenveld:
is it correct that the funding for JMS is not mentioned any more?

schoffelen:
yes, I think that the relative amount of time that I was funding myself in relation to the duration of the process probably does not justify it being mentioned

PR #2370, opened 2024-01-17
Check multiple elements in string of dataset_description field

henneysq:
The `cfg.dataset_description` fields `'Authors', 'Funding', 'EthicsApprovals', 'ReferencesAndLinks'` should be specified as an array of strings of multiple elements are defined (such as multiple authors).

Currently, the `fix_dataset_description` in `data2bids.m` change this field from a character type to array, but does not check the content.

This PR implements a check of the string content. If multiple elements are given in a single string, it will coerce them into individual array elements. The check assumes naïvely that if semi-colon delimination is not used, then commas are used to separate elements.

@robertoostenveld please let me know what you think. I have not added any warnings or tests (yet).

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1,2706 +1,2709 @@
 function [cfg] = data2bids(cfg, varargin)
 
-% DATA2BIDS is a helper function to convert MRI, MEG, EEG, iEEG or NIRS data to the
-% Brain Imaging Data Structure. The overall idea is that you write a MATLAB script in
-% which you call this function multiple times, once for each individually recorded
-% data file (or data set). It will write the corresponding sidecar JSON and TSV files
-% for each data file.
-%
-% Use as
-%   data2bids(cfg)
-% or as
-%   data2bids(cfg, data)
-%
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Could you resolve the conflicts? Furthermore, could you avoid making whitespace changes as those make it hard to review the actual changes. My default whitespace settings are to use spaces instead of tabs, use 2 spaces as the default indent, and I regularly do ctrl-a (select all) ctrl-i (auto indent). So if you do that, I think the irrelevant whitespace changes will be reverted.

henneysq:
Don't know how I accidentally indented everything. Maybe something automatic in my IDE. Anyways, I removed the indentation, and it resolved the conflict as well.

robertoostenveld:
thanks, I have merged it. 

The next time I recommend you first make a branch and _not_ make the changes on your master branch, as that would make things difficult if your pull request is not merged. Your master would in that case get out of sync of ours, leading to merge problems and eventually conflicts on your side. Typically you would always keep your master "clean".

PR #2375, opened 2024-01-19
Defacemesh

schoffelen:
use a plane for exclusion region


schoffelen:
Yes, all good suggestions. Will incorporate those. Probably a useful feature for volumes too indeed, will document.

robertoostenveld:
thanks! 

FYI: I looked through the website documentation, it only seems to use ft_defacemesh with the default (which is now the box), so no changes are needed in the documentation.

PR #2379, opened 2024-01-27
Data format MEF 2.1, 3.0 and MED 1.0 importer

jiecui:
This is an update for importing neural data format MEF 2.1, 3.0 and [MED 1.0](http://darkhorseneuro.com/) into FieldTrip.


schoffelen (changes requested):
ft_hastoolbox has different changes at different locations (i.e. in utlities/ft_hastoolbox, and  in fileio/private/ft_hastoolbox). Obviously, these changes should be the same. Note that there are mechanisms in place that synchronize different copies of the same function across the codebase, so the change needs to be made only on one of the files.

jiecui:
Can you please let me know the "mechanism" (perhaps a link to the corresponding website) that can synchronize the different copies of the same function? Thanks!

schoffelen:
There's a shell script: bin/synchronize_private.sh that takes care that changes in one version of a function end up everywhere. This will be deployed automatically once a change has been merged. You don't need to do anything, just ensure that only one copy of the files contains the 'union' of all changes that now seem to be distributed among the two different versions.

jiecui:
OK. Thanks! I'll double-check function ft_hastoolbox to make sure only one copy has the requested changes.

jiecui:
Hi, I have checked the code to make sure "utlities/ft_hastoolbox" and "fileio/private/ft_hastoolbox" are identical. Please let me know whether you can merge the changes now. Thanks!

schoffelen:
OK, thanks. Yet, from a conceptual point of view, I am wondering how we should proceed here. Perhaps @robertoostenveld can chime in. As I see it, this PR actually wants to merge code into FieldTrip that is currently in another repo: https://github.com/jiecui/mef_reader_fieldtrip/tree/master. The given repo, besides containing an external folder, which is the beef of the current PR here, also has some overloaded copies of fileio functions from fieldtrip ft_read_data/event/header, and ft_preprocessing, which I think also are included in this PR. Now, the question is who is going to take responsibility for keeping the repos in sync, and minimizing the risk of users ending up with ill-specified or incompatible sub-parts of the code on their paths. I see a risk in you (or somebody else) upgrading/improving sub-functionality of the mef_read_fieldtrip code, but these changes might not percolate automatically to fieldtrip/external/mayo_mef. The installation instructions are detailed, but also a bit confusing, because they aim at accommodating various potential FieldTrip version specific issues, and the installation instruction includes some 'band aids' that suggest to copy over FieldTrip files from your repo into fieldtrip root, and also to move around the mayo_mef folders. I think that this is unnecessarily complicated (and error prone).

To me, it looks that the external folder is well designed, yet best supported at its origin (thus, not as an external folder in the FieldTrip repo), and that the PR here should focus on providing the glue code in the relevant fieldtrip functions (i.e. ft_read_data/event/header, and ft_hastoolbox) + documentation w.r.t. what is needed to get started with this type of data. 

This would then obviously require instructions to include an installation of (an adjusted version of your) mef_reader_fieldtrip repo + the other stuff that you have already documented in your installation instructions. Given that a user already needs to download the MED reader from an external source, I don't think that this will lead to much additional overhead (and it would greatly simplify steps 2/4/5 in your installation instructions). 

So, all in all, my suggestion for moving forward would be to:
-simplify the PR, by just including suggested changes to fileio/ft_read_* utilities/ft_hastoolbox and the test function (and perhaps ft_preprocessing, not sure about that)
-simplify the mef_reader_fieldtrip repo, by removing the fileio utilities and ft_preprocessing code -> because that will be in FieldTrip eventually. Perhaps also move the mayo_mef folder a few levels up and/or rename it
-update the installation installation instructions in your repo, and provide documentation about getting started with MEF data on https://www.fieldtriptoolbox.org/getting_started/

This would I think give the best balance in code maintainability in the future, and also gives sufficient credit to @jiecui 

I am curious about what the two of you think about this.

PR #2381, opened 2024-01-31
Improve JOSS formatting

matsvanes:
- Reformatted the code snippets to span multiple lines when too long
- removed some of MVE's funding (i.e., they were irrelevant)
- cleaned up .bib file

PR #2382, opened 2024-01-31
Formatting and funding updates

matsvanes:
-     Reformatted the code snippets to span multiple lines when too long
-     removed some of MVE's funding (i.e., they were irrelevant)
-     cleaned up .bib file

### DIFF
--- a/JOSS/paper.bib
+++ b/JOSS/paper.bib
@@ -237,10 +237,6 @@
   file = {/Volumes/T5_OHBA/Literature/Zotero/storage/JCYPKSY2/Szucs and Ioannidis - 2017 - Empirical assessment of published effect sizes and.pdf}
 }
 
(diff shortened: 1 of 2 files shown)


robertoostenveld:
following the incorrect PR in #2381, this now actually looks better; it only has some small changes to the existing files.

robertoostenveld:
Let's hope that the merged pull request also synchronizes properly with https://bitbucket.org/fieldtriptoolbox/fieldtrip and https://gitlab.com/fieldtrip/fieldtrip; that is done with an automatic script that is executed upon a merge on github.

robertoostenveld:
Oh, actually nothing should happen, as bitbucket and gitlab only have the master branch synched, not the other branches like the JOSS one.

PR #2383, opened 2024-02-01
ENH - make some of the code more robust for string input

schoffelen:


schoffelen:
@robertoostenveld this fixes an old issue. I was wondering whether we should build an option into ft_checkconfig (by default to be 'no', which converts any strings into chars). This could be part of ft_default (e.g. ft_default.checkchar). I think that the overhead in ft_checkconfig would be minimal, and at least it captures all occurrences of "" in the cfg ' s

robertoostenveld:
do we have any use cases where we expect people to specify strings? I don't think so.

Why would the default for cfg.checkchar be 'no', would we not in general want to convert strings to chars?

And should it be `checkchar` or `checkstring`?

schoffelen:
> do we have any use cases where we expect people to specify strings? I don't think so.

My expectations are not too high here. I think that @mjcourte used a string in the original issue report. Not sure whether current generations of Matlab users use strings and char intermixed. 

> 
> Why would the default for cfg.checkchar be 'no', would we not in general want to convert strings to chars?

Well, I was thinking that it would create a little bit of extra overhead, but I don't mind defaulting to 'yes'

> 
> And should it be `checkchar` or `checkstring`?

Yes, well that's a good point. checkstring would be more appropriate

robertoostenveld:
Regarding enabling it by default: if it is not enabled by default, chances are very high that people will encounter the problem (when they specify a string as value) and not find the solution themselves. The documentation in ft_defaults is quite sparse, and when people call ft_whatever it is unlikely that they'll even consider the relevance of ft_defaults.

robertoostenveld:
Yes, pleas go ahead!

PR #2385, opened 2024-02-05
ENH - allow ft_channelrepair to deal with neighboring bad channels

robertoostenveld:
@NataschaRoos and I came across the situation that we want to identify some bad channels in some trials, i.e. in a sparse way, where certain channels in certain trials are replaced by nans. Subsequently we want to interpolate them.

The interpolation goes like this:

```
for i=1:numel(data_bad.trial)
  cfg = [];
  cfg.trials = i;
  cfg.badchannel = []; % this will be autodetected based on nans
  data_repaired_singletrial{i} = ft_channelrepair(cfg, data_bad);
end

data_repaired = ft_appenddata([], data_repaired_singletrial{:});
```

However, we came across the situation that bad channels are neighbours. This pull request updates ft_channelrepair such that bad and missing channels are removed from the `cfg.neighbours` definition prior to interpolating.

### DIFF
--- a/ft_channelrepair.m
+++ b/ft_channelrepair.m
@@ -19,8 +19,9 @@ function [data] = ft_channelrepair(cfg, data)
 %   cfg.order          = order of the polynomial interpolation (default = 4 for methods 'spline' and 'slap')
 %   cfg.senstype       = string, which type of data to repair. Can be 'meg', 'eeg' or 'nirs' (default is automatic)
 %
-% The weighted neighbour approach cannot be used reliably to repair multiple
-% bad channels that lie next to each other.
+% The weighted and average method are less reliably in case multiple bad channels lie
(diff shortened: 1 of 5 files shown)


robertoostenveld:
@schoffelen could you please review this?

schoffelen:
BTW quick side thought here, without having looked at the code: with multiple trials, it could be that this strategy really blows up the cfg.previous. Is this something to keep an eye on?

robertoostenveld:
good point, `data_repaired.cfg.previous` would indeed be very large. Better would be to do

```
cfg = [];
cfg.keepprevious = 'no';
data_repaired = ft_appenddata(cfg, data_repaired_singletrial{:});
```

which is using an option that I recently added, also to `ft_defaults`

schoffelen (approved):
Looks good to me. I caught a typo in the docstring

PR #2387, opened 2024-02-07
Joss ref

matsvanes:
Update Matlab reference in bibtex file (as software rather than misc)


matsvanes:
@robertoostenveld This commit hopefully solves this issue.
https://github.com/openjournals/joss-reviews/issues/5566#issuecomment-1929931831

PR #2391, opened 2024-02-13
Bep020

henneysq:
This PR updates the `data2bids` to more closely resemble the behaviour intended by (BEP020)[https://docs.google.com/document/d/1eggzTCzSHG3AEKhtnEDbcdk-2avXN6I94X8aUPEBVsw/edit#heading=h.9tphvz6ot0j1].

Currently, the only changes made are:
1. suffix updated from `'eyetracker'` to `'eyetrack'`
2. `'eyetrack'` dir added as destination for `'eyetrack'` files

This PR does not (yet) address the specifications of sidecar content.

### DIFF
--- a/data2bids.m
+++ b/data2bids.m
@@ -1127,7 +1127,7 @@ stim_settings = keepfields(cfg.stim, fn);
 % make the relevant selection, all json fields start with a capital letter
 fn = fieldnames(cfg.eyetrack);
 fn = fn(~cellfun(@isempty, regexp(fn, '^[A-Z].*')));
-eyetracker_settings = keepfields(cfg.eyetrack, fn);
+eyetrack_settings = keepfields(cfg.eyetrack, fn);
 
 % make the relevant selection, all json fields start with a capital letter
 fn = fieldnames(cfg.motion);
@@ -1390,14 +1390,14 @@ end
 
 %% need_eyetrack_json
 if need_eyetrack_json
-  eyetracker_json.SamplingFrequency = hdr.Fs;
-  eyetracker_json.StartTime = nan;
-  eyetracker_json.Columns = hdr.label;
+  eyetrack_json.SamplingFrequency = hdr.Fs;
+  eyetrack_json.StartTime = nan;
+  eyetrack_json.Columns = hdr.label;
 
   % merge the information specified by the user with that from the data
   % in case fields appear in both, the first input overrules the second
-  eyetracker_json = mergestruct(eyetracker_settings,  eyetracker_json, false);
-  eyetracker_json = mergestruct(generic_settings,     eyetracker_json, false);
+  eyetrack_json = mergestruct(eyetrack_settings,  eyetrack_json, false);
+  eyetrack_json = mergestruct(generic_settings,     eyetrack_json, false);
 end
 
 %% need_motion_json
--- a/data2bids.m
+++ b/data2bids.m
@@ -2603,7 +2603,7 @@ switch typ
   case {'emg'} % this is being discussed at https://github.com/bids-standard/bids-specification/issues/1371
     dir = 'emg';
   case {'eyetrack'} % this is specified in BEP020: https://docs.google.com/document/d/1eggzTCzSHG3AEKhtnEDbcdk-2avXN6I94X8aUPEBVsw/edit#heading=h.64qz71yd6qpa
-    dir = 'eyetrack'
+    dir = 'eyetrack';
 
   otherwise
     ft_error('unrecognized data type ''%s''', typ);
--- a/data2bids.m
+++ b/data2bids.m
@@ -54,7 +54,7 @@ function [cfg] = data2bids(cfg, varargin)
 %   cfg.ses                     = string, optional session name
 %   cfg.run                     = number, optional
 %   cfg.task                    = string, task name is required for functional data
-%   cfg.suffix                  = string, can be any of 'FLAIR', 'FLASH', 'PD', 'PDT2', 'PDmap', 'T1map', 'T1rho', 'T1w', 'T2map', 'T2star', 'T2w', 'angio', 'audio', 'bold', 'bval', 'bvec', 'channels', 'coordsystem', 'defacemask', 'dwi', 'eeg', 'emg', 'epi', 'events', 'eyetracker', 'fieldmap', 'headshape', 'ieeg', 'inplaneT1', 'inplaneT2', 'magnitude', 'magnitude1', 'magnitude2', 'meg', 'motion', 'nirs', 'phase1', 'phase2', 'phasediff', 'photo', 'physio', 'sbref', 'stim', 'video'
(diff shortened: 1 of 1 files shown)


data2bids.m
```diff
@@ -310,7 +310,7 @@
 cfg.nirs          = ft_getopt(cfg, 'nirs');
 cfg.audio         = ft_getopt(cfg, 'audio');
 cfg.video         = ft_getopt(cfg, 'video');
-cfg.eyetracker    = ft_getopt(cfg, 'eyetracker');
+cfg.eyetrack      = ft_getopt(cfg, 'eyetrack');
```
robertoostenveld, line 313: Please add prior to this a line

    cfg = ft_checkconfig('renamedval', {'suffix', 'eyetracker', 'eye track'});

it can go before setting all defaults. This allows old scripts to keep working, and prints a warning. Otherwise people would probably get an error further down in the code.

data2bids.m
```diff
@@ -593,11 +593,13 @@
...
-%% eyetracker is not part of the official BIDS specification
-% this follows https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continuous-recordings.html
-cfg.eyetracker.Columns                    = ft_getopt(cfg.eyetracker, 'Columns'               );
-cfg.eyetracker.StartTime                  = ft_getopt(cfg.eyetracker, 'StartTime'             );
-cfg.eyetracker.SamplingFrequency          = ft_getopt(cfg.eyetracker, 'SamplingFrequency'     );
+%% eyetracker is not part of the official BIDS specification, but is included in proposal BEP020:
+% https://docs.google.com/document/d/1eggzTCzSHG3AEKhtnEDbcdk-2avXN6I94X8aUPEBVsw/edit#heading=h.9tphvz6ot0j1
+% The current implementation follows:
+% https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continuous-recordings.html
+cfg.eyetrack.Columns                      = ft_getopt(cfg.eyetrack, 'Columns'               );
```
robertoostenveld, line 600: please also add somewhere at the top

    cfg = ft_checkconfig('renamed', {'eyetracker', 'eye track'});

to rename this field in case the user supplied it with the old name

robertoostenveld:
you write that you changed it to 

> 'eyetrack' dir added as destination for 'eyetrack' files

but that is not the case as far as I can see. On line 2589 it seems to go in the `beh` folder.  In https://github.com/bids-standard/bids-specification/pull/1128 I read 

>  Eye-tracking data MUST be stored in the main data recording modality or
>  `<datatype>` directory (for example `func` when combined with fMRI, or
> `beh` when combined with behavioral measures).

I believe this also applies to the remainder of `{'events' 'stim' 'physio' 'audio' 'video'}`. Should we add an option `cfg.datatypedir` which by default is empty and auto-determined using the `datatype2dirname()` function, but which allows it to be user-specified for eyetrack and such? There are now two calls to `datatype2dirname()`, those should then use the cfg.datatypedir field (which itself then needs to be set somewhere at line 725 or so).

henneysq:
> https://docs.google.com/document/d/1eggzTCzSHG3AEKhtnEDbcdk-2avXN6I94X8aUPEBVsw/edit#heading=h.9tphvz6ot0j1

Hmm, I was referring to the google doc, which shows the following convention:
```
sub-<participant_label>/
    eyetrack/
        sub-<participant_label>[_ses-<label>]_task-<task_label>[_acq-<label>]_eyetrack.tsv
        [sub-<participant_label>[_ses-<label>]_task-<task_label>[_acq-<label>]_eyetrack.json]
        [sub-<participant_label>[_ses-<label>]_task-<task_label>[_acq-<label>]_events.tsv]
```

Is that out of date? And should I refer to https://github.com/bids-standard/bids-specification/pull/1128 instead?

henneysq:
Perhaps, `cfg.datatypedir` should default for empty but be mandatory for `{'events' 'stim' 'physio' 'audio' 'video' 'eyetrack'}` to reflect the [sentiment of the proposal](https://arc.net/l/quote/pfyfxvyo):

"Eye-tracking data MUST be stored in the main data recording modality or `<datatype>` directory"

robertoostenveld:
> ... google doc ...
> Is that out of date? And should I refer to https://github.com/bids-standard/bids-specification/pull/1128 instead?

Yes, that is out of date. The discussion and development continued on the bep020 pull request, which is 
https://github.com/bids-standard/bids-specification/pull/1128. And that PR points to https://github.com/mszinte/bids-specification/tree/bep020 and that is where I think most recent developments have happened. 

> Perhaps, cfg.datatypedir should default to empty but be mandatory for {'events' 'stim' 'physio' 'audio' 'video' 'eyetrack'} to reflect the [sentiment of the proposal](https://arc.net/l/quote/pfyfxvyo):
> 
> "Eye-tracking data MUST be stored in the main data recording modality or <datatype> directory"

good idea,  it does not hurt to make it required for all of those.

henneysq:
I realise today that when allocating `eyetrack` files to the functional imaging dir (e.g. `meg`), they compete for the `sub-<participant_label>[_ses-<label>]_task-<task_label>[_acq-<label>]_events.tsv` filename. So the latter of `meg` and `eyetrack` data to be converted overwrites the prior.

@robertoostenveld, does this imply that the two should be merged? In that process, one would need to ensure their allignment in time.

henneysq:
Or is it in fact also an outdated requirement from the google doc that `eyetrack` should have an `_events.tsv`?

henneysq:
In commit `efb818e`, I inverted the logic of the `need_eyetrack_json` when checking whether to write the `_events.tsv`. This resolves the issue, so that the MEG file gets to dictate the content of `_events.tsv`.

robertoostenveld:
> Or is it in fact also an outdated requirement from the google doc that eyetrack should have an _events.tsv?

Yes, I read in one of the comments on the commit that detected saccades (which are timestamped "events") are to be interpreted as derivative and not stored along with the raw data. 

Do you have other events in your eye tracking recording, besides the saccades? Like triggers? In your ["raw1" version](https://github.com/henneysq/meg-ahat/blob/main/documentation/DATA_MANAGEMENT.stub) you would still have the eyetrack.tsv and the corresponding events.tsv in the beh folder, right? I assume that, because in that version it is not yet aligned with the MEG.

In your "raw2" version you would align them along the same time axis and merge them, so events from the MEG and from the eyetracker could be fused in a single events.tsv (you would actually use those shared events for the alignment). So after time shifting also saccades from the raw1/beh/events.tsv could still be represented in the raw2/meg/events.tsv.

henneysq:
Hmm I actually moved the `eyetrack` to `meg` already, but I probably should keep it in `beh` for `raw1` as they are not alligned yet.

I have actually been working on how to figure out the allignment of `eyetrack` and `meg`. It appears that quite a bit of information is "lost" in the current implementation. I will try to explain it, but we should probably have a talk about it to make sure we understand eachother.

Before commit `efb818e` when `eyetrack` wrote an `_events.tsv` file, this contained `FIX` and `SACC` events, while `_eyetrack.tsv` contained the timeseries of x-/y-coordinates. If I understand it correctly, these are extracted from the source `.asc` file where the two are intermingled/stacked in time. However, after commit `efb818e`, the events are no longer saved anywere and get lost in the transition.

So what I will have to do is
1. In `raw1`, keep writing the `_events.tsv` file, but to `beh`
2. In `raw2`, align the time axes of `meg` and `eyetrack` and merge events into on `_events.tsv` file

To achieve 1., I will revert the change in commit `efb818e` and set `datatypedir` to `beh`.

I can achieve 2. locally, but I'm not sure I would be able to logically implement it into `data2bids`. Unless it can be implemented with a merge of the two that *assumes* that events are on the same timeline already. Let me know what you think.

PR #2414, opened 2024-05-13
Updating MFF code to import data epochs

arnodelorme:
Fix an important bug in calculating event latency for data epochs.


ft_prepare_leadfield.m
```diff
@@ -185,6 +185,7 @@
 % check whether units are equal (NOTE: this was previously not required,
 % this check can be removed if the underlying bug is resolved. See
 % http://bugzilla.fieldtriptoolbox.org/show_bug.cgi?id=2387
+sourcemodel.unit = 'mm';
```
schoffelen, line 188: what is the intended fix for this line? I suggest to remove it, because the units should not be silently added here, and there is no reason to assume that they are 'mm'.

schoffelen (changes requested):

ft_prepare_leadfield.m
arnodelorme: Good point. A local modification I forgot to remove.

arnodelorme:
I have removed the change. However, I know @robertoostenveld would prefer for it not to be there at all, so as not to clutter the history. That requires rebasing and editing the list of commits and ignoring some of them, I think (I am not even sure this is possible as the original change was part of another commit).

schoffelen:
The intended code updates have been merged in a fresh PR. Note: the additional mexmaca64 version of read24bit seems to already be available, so I kept it out of the new PR

PR #2415, opened 2024-05-18
support for neuropixel data recorded with SpikeGLX 

robertoostenveld:
[Marco Numi](https://www.neuro.bio.lmu.de/members/sensory-neuroscience/marco-numi/index.html) asked me over email whether we could implement support for Neuropixel data recorded with SpikeGLX software. I found this https://github.com/jenniferColonell/SpikeGLX_Datafile_Tools/tree/main/MATLAB, and using the example added support to the low-level fieldtrip reading functions. That means that high-level functions like ft_preprocessing and ft_databrowser now also support it. 

For example
```
cfg = [];
cfg.dataset = 'myRun1_4_7_2022_mag_stim_comnination_9_10_150uT_2sec_10reps_g0_t0.imec0.lf.bin';
ft_databrowser(cfg); 
```


robertoostenveld:
I believe that there is still some useful metadata in the original header (now in `hdr.orig`) regarding the arrangement of the pixels. And there is apparently also a default `.imro` file with mapping information between contacts and channels.  That could possibly be used to make an [elec](https://www.fieldtriptoolbox.org/faq/how_are_electrodes_magnetometers_or_gradiometers_described/) structure. The elec structure could then be used to make a 2D channel layout, but also for 3D plotting. 

An alternative for the 2D channel layout would be to follow this https://www.fieldtriptoolbox.org/tutorial/layout/#making-a-layout-for-lfpspike-electrodes. It would be good to get some input from Marco to decide whether to add more code to fieldtrip, or whether it is better to add some documentation that explains how a researcher can do it themself.


In the technical brochures on https://www.neuropixels.org/probes-np1-0 I found for example this

> The probes feature 960 low-impedance TiN recording sites densely tiled
> along a thin, 10 mm-long, straight shank. The 384 parallel, configurable,
> low-noise recording channels ...

The example data has 384 plus one binary channel, so might match. However, I don't understand yet how the 960 electrode contacts map onto the 384 channels. The meta header suggests that there an offset in the channels/electrodes. Getting the channel-to-electrode mapping right would be relevant for visualization.

marconumi:
Thank you so much for you preciious help and I am extremely sorry for the late response. 
The reader works and I can use low-level funtion like ft_preprocessing, ft_timelock and ft_freqanalysis. 
Concerning your doubt, the probe has 960 electrodes but it's possible to employ "just" 384 of these for each recording session and these can be chosen dynamically from the 960 total ones (for example, we use the first 384).

robertoostenveld:
thanks for the feedback, then this issue can be closed.

robertoostenveld:
oh no, sorry: I first have to merge and then close

PR #2416, opened 2024-05-23
adopt code to support SIESTA usecase 2.3

robertoostenveld:
This is needed to deal with the https://github.com/robertoostenveld/Wakeman-and-Henson-2015 pipeline to be used under https://github.com/SIESTA-eu/wp15 on the 1.0.6 version of the dataset https://doi.org/10.18112/openneuro.ds000117.v1.0.6 in BIDS format


schoffelen:
Running the siesta code using this PR's branch works fine. merging...

PR #2419, opened 2024-05-31
ENH - added missing mex files

schoffelen:


schoffelen:
I want to also still check the overall freshness of the code in external/spm12 against the version of the code in spm/spm
I think we also would want to update spm_platform etc.

PR #2422, opened 2024-06-05
FIX - to deal with sub2ind not supporting nans anymore in new matlab …

schoffelen:
…version


schoffelen:
grep -l sub2ind *.m
ft_meshrealign.m
ft_prepare_sourcemodel.m
ft_sourceplot.m
ft_volumelookup.m
ft_volumerealign.m

I will have to check the listed functions as well, these may suffer from the same potential issue

PR #2432, opened 2024-07-04
Revert "speed up user experience in ft_defaults / multiple calls to ft_hastoolbox"

schoffelen:
Reverts fieldtrip/fieldtrip#2425


schoffelen:
This PR caused more trouble in the dashboard than anticipated. Since it does not solve a big issue, I revert for now, because I don't have the bandwidth to look into it in detail

PR #2438, opened 2024-08-27
ENH - first basic working implementation of amm and sss

schoffelen:


schoffelen:
This is a first shot for amm-support in fieldtrip. Things to still do are:
- write a test function
- consider storing the per segment adaptive matrices
- harmonize this implementation with dssp as far as possible
- implement sss (probably will be in a different PR)

PR #2442, opened 2024-09-09
Update read_brainvision_eeg.m

coxroy:
pre-allocate 'dat' for binary-vectorized format. reading much faster (similar to binary-multiplexed)

### DIFF
--- a/fileio/private/read_brainvision_eeg.m
+++ b/fileio/private/read_brainvision_eeg.m
@@ -101,6 +101,7 @@ elseif strcmpi(hdr.DataFormat, 'binary') && strcmpi(hdr.DataOrientation, 'vector
   fseek(fid, 0, 'bof');
 
   numsamples = (endsample-begsample+1);
+  dat = zeros(hdr.NumberOfChannels,numsamples);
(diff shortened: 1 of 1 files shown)


schoffelen:
Thanks Roy! Typically we would also ask a contributor to write a little snippet of test code (to be added to our test code base), but this fix is a bit of a no-brainer so good for now. I'll merge. Feel free to communicate on the discussion e-mail list that you have fixed it.

What will happen next, is that our overnight test battery will use the codebase from the master branch to run a (large) series of tests. If this runs through fine without error, the idea would be that a new tagged release version of the code will be made available as a zip-file, and that the latest version of the code on the master branch will be pushed to the release branch. I noticed - however - that this automatic release version creation has somehow stopped since somewhere in July. For the time being your changes will be visible only in the master branch. Note to self and to @robertoostenveld we need to check what's going on here, I haven't received PRs from the website repo - which typically flag the new releases - in  a while

robertoostenveld:
Please note that I also just pushed 781db124c0587fbdca25469d92be68fdff7ae1bf to fix an issue I encountered with a test datafile from pybv.

PR #2444, opened 2024-09-16
ENH - support for windows

schoffelen:


schoffelen:
I can only test this in a limited way: the changes in the current PR allow to run test_headmodel_dipoli on a windows PC in the DCCN trainee room. Not much more to be done for now, apart from merging, and hoping that other people pick up on this.

schoffelen:
Added note: the current compiled executables for the different platforms seem to anecdotally require a different approach for them to work (for JM): for MacOS/Linux the emulated interactive use works fine (and explicitly defining the boundaries a -g arguments doesn't), whereas for windows it's the other way around.

PR #2445, opened 2024-09-16
Fix GUI fiducial selection on head surface

contsili:
Minor fix that allows to use the GUI to click the fiducials on the extracted head surface.

You can do that by running the code:
cfg = [];
cfg.method = 'interactive';
cfg.viewmode = 'surface';
[realign] = ft_volumerealign(cfg, mri)


robertoostenveld:
thanks!

PR #2449, opened 2024-09-19
FIX - units conversion in ft_volumereslice

contsili:
Small fix for correct physical units conversion (e.g., mm -> m) in `ft_volumereslice.`

I also checked other fieldtrip functions that use `ft_scalingfactor` and all the units conversions were fine.

Good news regarding testing: I ran the test procedure locally on Windows (without DCCN intranet access) using the steps from [FieldTrip testing webpage](https://www.fieldtriptoolbox.org/development/testing/). `dccnpath `automatically downloaded all necessary test data. All I had to do is just copy and paste the code from the testing webpage. Great for our external contributors! (oh by the way, all the tests passed :) )

### DIFF
--- a/ft_volumereslice.m
+++ b/ft_volumereslice.m
@@ -103,13 +103,13 @@ else
     % with respect to the origin of the coordinate system
     switch mri.coordsys
       case {'ctf', '4d', 'bti', 'eeglab'}
-        xshift = 30.*cfg.resolution;
+        xshift = 30.*ft_scalingfactor('mm', mri.unit);
         yshift = 0;
-        zshift = 40.*cfg.resolution;
+        zshift = 40.*ft_scalingfactor('mm', mri.unit);
       case {'neuromag', 'itab'}
         xshift = 0;
-        yshift = 30.*cfg.resolution;
-        zshift = 40.*cfg.resolution;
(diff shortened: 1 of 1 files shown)


contsili:
By the way, _test_bug2986_ and _test_ft_volumerealign_ use private data. I do not know why it is suggested by GH actions to run when outside the DCCN. This needs to be fixed.

robertoostenveld:
I don't understand why cfg.resolution is used in the computation of the shifts. Should it not be `xshift = 30 * ft_scalingfactor('mm', mri.unit)` or so? The bounding box around the coordinate system origin does not depend on the resolution of the reslicing. If the MRI is in mm, the shift should be 30. If the MRI is in cm, the shift should be 3, if the MRI is in m, the shift should be 0.03.

contsili:
I agree with you.

It all starts from line 95: cfg.resolution = ft_getopt(cfg, 'resolution', 1 * ft_scalingfactor('mm', mri.unit)). The ft_scalingfactor in this line was added because of this bug: https://bugzilla.fieldtriptoolbox.org/bug/2906.html

I will have a quick look and let you know if we need to fix more things in the code.

contsili:
I fixed it so that the bounding box does not depend on the resolution of the reslicing.

PR #2450, opened 2024-09-20
ENH - Inside the local copy, the directories should have a specific hierarchy.

contsili:
Improved help documentation of `dccpath` to explain to external contributors (that have manually downloaded test data in their computer) how to structure the folders inside their local copy.


contsili:
The edits look good to me. We could merge it.

PR #2452, opened 2024-10-01
Stop cfg.README from being used in data sidecar json files

jens-k:
Currently, the content of cfg.README is written into each data file sidecar json. This violates BIDS format (according to the OpenNeuro validator). The cfg.README field should only be used for creating the README file, not as a property for the each data file sidecar json.


schoffelen:
Sorry @jens-k for joing the party so late. Is this still something that should in principle be merged? If so, could you resolve the (current) conflicted file?

schoffelen:
gotcha! Thanks for getting back about this

PR #2453, opened 2024-10-03
Prevent MATLAB crash on Windows due to sparse matrix in read_trigger function

### LINKED ISSUE #2451: MATLAB crash on Windows due to sparse matrix in read_trigger function
**Describe the Bug**
MATLAB crashes at `read_trigger` (line 100) on the following code:

`if skewness(abs(diff(dat(i,:)))) > 5`

This issue occurs on Windows devices. The crash has been reproduced on two separate machines: a personal laptop and a Donders PC station.

**To Reproduce**
```
cfg = [];
cfg.dataset = '<OPM data>';  % Download from https://we.tl/t-U190lcH16V
cfg.trialdef.eventtype = 'ai61';
cfg.trialdef.prestim = 0.2;  % ISI varied from 0.8 to 1.2 seconds
cfg.trialdef.poststim = 0.4;
cfg = ft_definetrial(cfg);
```

**Expected Behavior**
MATLAB crashes on Windows devices but does not crash when run on the Donders Cluster or on macOS.

**Environment and Versions** 

- Operating System: Windows 10
- MATLAB Versions: 2020B, 2021b, 2022a, 2023b
- FieldTrip Version: 74c5a8e18

**Additional Context**
Upon further investigation, it was identified that the issue arises from `dat ` variable being a sparse double matrix. When converted to a full matrix, MATLAB no longer crashes. The issue seems to originate from the low-level function `fiff_read_raw_segment()` at the following lines (121-126):

```
if ~isempty(cal)
    cal = sparse(cal);
end
if ~isempty(mult)
    mult = sparse(mult);
end
```

contsili:
To prevent crash of MATLAB on Windows, we force `dat` to be a full matrix instead of a sparse double.

As an external contributor, I could only run one test for `read_trigger` and none for `ft_read_data`. The one test passed.

Also, when we add my OPM data in the FT download server, I can write a test  `test_issue2451`.

This closes #2451.

### DIFF
--- a/fileio/ft_read_data.m
+++ b/fileio/ft_read_data.m
@@ -1253,9 +1253,7 @@ switch dataformat
     ft_hastoolbox('mne', 1);
     if (hdr.orig.iscontinuous)
       dat = fiff_read_raw_segment(hdr.orig.raw,begsample+hdr.orig.raw.first_samp-1,endsample+hdr.orig.raw.first_samp-1,chanindx);
-      if ispc
-        dat = full(dat); % prevent MATLAB crash on Windows due to 'dat' being a sparse matrix, see issue 2451
-      end
+      dat = full(dat); % prevent MATLAB crash on Windows due to 'dat' being a sparse matrix, see issue 2451
       dimord = 'chans_samples';
     elseif (hdr.orig.isepoched)
       % permutation of the data matrix is time consuming, and offsets the time gained by not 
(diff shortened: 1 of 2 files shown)


schoffelen:
> > When outside the DCCN, please consider testing: test_yorkinstruments, test_bug1262, test_pull731, test_bug1404, test_issue1585
> 
> All these tests have private data and cannot run outside the DCCN. This is the 2nd time I notice this problem with the GH actions. Shall we make an issue about it and fix it?

But does that mean that these test function by accident got a "% DATA no" statement in their header? It would be good to do a sanity check, but how would you do that in an automatized way?

contsili:
> When outside the DCCN, please consider testing: test_yorkinstruments, test_bug1262, test_pull731, test_bug1404, test_issue1585

From all these tests only test_bug1404 has "% DATA no". 

So as I see it, there are two problems:
1. Some functions are labelled as "% DATA no" while they should be "% DATA private". I can try to fix that with ''grep'' command and a bit of coding in the bash to edit files automatically.
2. GH actions identifies the "% DATA private" as runnable outside the DCCN. So GH actions should be modified/corrected. I believe @robertoostenveld worked on the GH actions, so maybe he could chime in on this issue.

PR #2461, opened 2024-11-06
implemented function to estimate head circumference

robertoostenveld:
as discussed with @chrisNatMEG


chrisNatMEG:
@Robert Just tried it on some of my own data (manually marked the fiducials with ft_electrodeplacement on a scalp mesh) and it worked without problems. The value fits with the cap-size we used for the subject I tried it on.
Also like the figure it generates. It works as a good sanity check.

Issue #2460, opened 2024-11-06
ft_badsegement does not use preproc correctly


chrisNatMEG:
**Describe the bug**
ft_badsegment applies preproc options when calculating std and mean for zvalue parameters but not when actually calculating the level (line 171). The level calculation uses the unprocessed data.

Issue can be fixed by replacing line 171 with:
dat = preproc(data.trial{trl}, data.label, data.time{trl}, cfg.preproc);
level = artifact_level(dat, cfg.metric, mval, sd, connectivity);

schoffelen:
Hi Christoph @chrisNatMEG , thanks for catching this. It indeed does not seem to make sense NOT to apply the same preprocessing options at the (approximate) location that you mention. BTW, in the master branch of FT the call to ```artifact_level``` takes place in line 176 so it seems that you use a slightly outdated version of the code. Would you be fine with making a PR for this? this will give you the full digital credits in the git-history of the file

robertoostenveld:
I just looked at this with Christoph, it also affects ft_badchannel and ft_baddata. The rejectvisual_summary function also does the z-scoring, but that is fine. @chrisNatMEG will make a PR.

robertoostenveld:
FYI I shared the [git and GitHub tutorial slides](https://docs.google.com/presentation/d/1xaJdMa--5sovtNZI_BunG1ZbSM8qOwFJCExh8OK7mg0/edit?usp=sharing) from last week with Christoph, and also pointed him to the tutorial at https://www.fieldtriptoolbox.org/development/git/

schoffelen:
https://xkcd.com/1597/

PR #2462, opened 2024-11-06
fixed missing preproc in ft_badsegment, ft_badchannel and ft_baddata

chrisNatMEG:
Added missing preproc before level is calculated in ft_badsegment, ft_badchannel and ft_baddata. Also changed name of the index variable in the loop calculating std and mean for zvalue to better match its use.


robertoostenveld:
thanks, looks good! I will merge. 

After merging it will immediately be in the master branch of `fieldtrip:fieldtrip`, and after testing (which happens automatically in the evening) it should also become part of the release branch and distributed in the daily "YYYYMMDD" version.

robertoostenveld:
this closes #2460

PR #2468, opened 2024-11-26
fix too strict option checking in the ft_inverse_ functions

schoffelen:


schoffelen:
I think that this should do it, I checked the majority of the failing test functions (apart from the ones that only crashed after > 2 minutes.

robertoostenveld:
Seems to have worked, this morning the dashboard did not report any errors. Thanks!

PR #2476, opened 2025-02-05
Added TMSi 32- and 64-channel infinity gel headcap

contsili:


schoffelen:
Hi @contsili , have you checked whether these files generate topoplots etc without glitches? The reason I ask, is because the other *.lay files also contain a (dummy) COMNT and SCALE label (as the last 2 channels). Perhaps you want to consider adding those as well. If the electrode positions are normalised to [-0.5 0.5] I think you can just copy over the same X/Y coordinates as in corresponding lay files.

contsili:
Hi @schoffelen , yes, running 
cfg    	    = [];
cfg.layout = 'EEGChannels32TMSi.lay';
ft_layoutplot(cfg)
 
cfg    	    = [];
cfg.layout = 'EEGChannels64TMSi.lay';
ft_layoutplot(cfg)

I get:
Screenshot (5)

So COMNT and SCALE look fine.

robertoostenveld:
The COMNT and SCALE labels (and positions) will be added by ft_prepare_payout when needed.

PR #2478, opened 2025-02-06
update ft_realtime_micromedproxy.m and .p

bvandyck:
replace tcp_udp_ip with Instrument Control Toolbox, replace matlabversion with verLessThan


robertoostenveld:
Thanks! Please send me the updated m-file by email, so that others can also request that from me when desired.

bvandyck:
OK, should be in your mailbox!

PR #2483, opened 2025-02-25
ENH - allow for uncoupling of computation from interaction, …

schoffelen:
…by optional output of intermediate results


schoffelen:
The use case here is a situation, where I want to run an interactive ft_artifact_zvalue - 4 times per dataset - for about 200 datasets. Rather than doing this interactively while having to wait for the (lengthy) computations to run, this PR allows for running the computations asynchronously (and storing them), after which the multiple interactive function calls are somewhat more enjoyable.

PR #2485, opened 2025-03-09
Extend codespell setup

yarikoptic:
Continuation to
- #2295 

I was about to remove fieldtrip folder from the laptop since "I was done -- `git grep codespell` returned presense", but running it brought many hits!  Only then I saw that I was not ambitious enough and just did a "demo" ;)  So decided to improve 

TODOs
- [x] after verifying that action does test two folders now, fix typos in that `plotting/`


yarikoptic:
very nice -- codespell action found typos, and annotator pointed to them e.g.

image

pushing the fixes now ;)

PR #2486, opened 2025-03-09
Just a few more "manual" typo fixes

yarikoptic:
Compliment to
- #2485 

Just could not see quietly starting at the output of `❯ codespellit list-hits-sorted` so decided to do half measure solution and fix just a few... Individual commit messages have "command" and what was changed to what


schoffelen (changes requested):
Thanks for the suggestions! I would gladly merge all changes to the files that are NOT in the external (sub)folders. This code is not maintained/developed by us, and is ideally in sync with a specified version that is 'officially' hosted elsewhere. The typos should be fixed at the official location, rather than locally in the FieldTrip code base. It would be great if you could revert the changes to the external files

robertoostenveld:
Hi Jan-Mathijs, I think this could be merged using [this strategy](https://stackoverflow.com/questions/15232000/git-ignore-files-during-merge) where we'd ignore `fieldtrip/external/*`. Do you want me to give that a try?

schoffelen:
> Hi Jan-Mathijs, I think this could be merged using [this strategy](https://stackoverflow.com/questions/15232000/git-ignore-files-during-merge) where we'd ignore `fieldtrip/external/*`. Do you want me to give that a try?

sure, if not too much hassle. Thanks! and thanks to @yarikoptic for the contribution

schoffelen:
I pulled in the branch locally, and did a ```git checkout upstream/master on the files in external.

PR #2487, opened 2025-03-09
copying info from header also when unknown channel

mcpiastra:
This follows a conversation with @robertoostenveld


robertoostenveld:
Maria-Carla and I reviewed and tested this already on her computer with an appropriate EDF file, so I think further testing is not needed prior to merging.

PR #2494, opened 2025-03-28
set channels type from bids

dnacombo:
Not sure why, but channel types found in bids sidecar _channels.tsv isn't ported into hdr.
If that was a mistake, then this should fix it...

### DIFF
--- a/fileio/ft_read_header.m
+++ b/fileio/ft_read_header.m
@@ -2875,7 +2875,8 @@ if (strcmp(readbids, 'yes') || strcmp(readbids, 'ifmakessense')) && isbids
       assert(length(channels_tsv.type)  == hdr.nChans, 'number of channels is not consistent with the BIDS channels.tsv');
       assert(length(channels_tsv.units) == hdr.nChans, 'number of channels is not consistent with the BIDS channels.tsv');
       hdr.label    = channels_tsv.name;
-      hdr.chantype = repmat({'unknown'}, [length(hdr.label), 1]);
+      hdr.chantype    = channels_tsv.type;
+      % hdr.chantype = repmat({'unknown'}, [length(hdr.label), 1]);
       hdr.chantype(contains(channels_tsv.type, 'NIRS')) = {'nirs'};
       hdr.chanunit = channels_tsv.units;
     end
--- a/fileio/ft_read_header.m
+++ b/fileio/ft_read_header.m
@@ -2875,9 +2875,7 @@ if (strcmp(readbids, 'yes') || strcmp(readbids, 'ifmakessense')) && isbids
       assert(length(channels_tsv.type)  == hdr.nChans, 'number of channels is not consistent with the BIDS channels.tsv');
       assert(length(channels_tsv.units) == hdr.nChans, 'number of channels is not consistent with the BIDS channels.tsv');
       hdr.label    = channels_tsv.name;
-      hdr.chantype    = channels_tsv.type;
-      % hdr.chantype = repmat({'unknown'}, [length(hdr.label), 1]);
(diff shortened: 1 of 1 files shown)


robertoostenveld:
Hi Max,

The `hdr` field is actually never updated, also not when you make a selection of channels or so. I am actually not even sure whether it should be in the data structure that comes out of ft_preprocessing. In principle the structure shoudl be self-describing and generic, whereas the `hdr` field makes it data type specific. What made you look into the header in the first place?

dnacombo:
Hi Robert,
I was looking for channel types to select only EEG or EKG channels.

robertoostenveld:
Ah, that makes sense. FieldTrip already knows about certain channel types (see `ft_chantype.m`), and always uses lowercase.  Could you make `hdr.chantype` all lower case - that makes a lot of them comparable to the old/existing fieldtrip standard and will help functions 

I also suggest to remove the special `contains(channels_tsv.type, 'NIRS')` handling and have all identifiers remain according to BIDS. In the longer run it would be better to adopt and/or move to the BIDS naming schemes elsewhere in FieldTrip, rather than to replace all BIDS identifiers here by those now used in FieldTrip.

dnacombo:
Nice.
Now I see the the chantype isn't carried along to the elec field after redefinetrial. This seems strange.

I read data from a [bids dataset](https://openneuro.org/datasets/ds003690/versions/1.0.0) with lots of info in the sidecars.
with these commands:

`    cfg = [];
    cfg.dataset = fs(1).name;

    cfg.hpfilter = 'yes';
    cfg.hpfreq = .1;
    cfg.hpfilttype = 'firws';
    cfg.lpfilter = 'yes';
    cfg.lpfilter = 35;
    cfg.lpfilttype = 'firws';

    data = ft_preprocessing(cfg);

    cfg.trialdef.eventtype = 'cue';
    cfg.trialdef.prestim = 0.2;
    cfg.trialdef.poststim = 6;
    cfg = ft_definetrial(cfg);

    cfg.demean = 'yes';
    cfg.baselinewindow = [-.2 0];
    data = ft_redefinetrial(cfg,data);
`
the data.hdr variable contains

```
>> data.hdr.chantype(60:end)
ans =
  7×1 cell array
    {'eeg'  }
    {'eeg'  }
    {'veog' }
    {'heog' }
    {'ecg'  }
    {'pupil'}
    {'pupil'}
```

but after redefinetrial, the data.elec.chantype field contains
```
data.elec.chantype(60:end)
ans =
  7×1 cell array
    {'eeg'}
    {'eeg'}
    {'eeg'}
    {'eeg'}
    {'eeg'}
    {'eeg'}
    {'eeg'}
```

dnacombo:
Cannot run tests. I get the following error for all tests.
```
>> test_yorkinstruments
using a temporary directory
Index exceeds array bounds.
Error in dccnpath (line 188)
    datadir = datadir{1};
              ^^^^^^^^^^
Error in test_yorkinstruments (line 12)
filename = dccnpath('/project/3031000.02/test/original/meg/yorkinstruments/yi.meghdf5');
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ 

```

robertoostenveld:
Hmm, that is weird. The [test_yorkinstruments.m](https://github.com/fieldtrip/fieldtrip/blob/master/test/test_yorkinstruments.m) script specifies "DATA private" (which is also the case: it is not on the download server as tutorial data or so, hence you don't have access to it) and therefore it should not have suggested this test for you to run. The other tests suggested for testing outside the DCCN are also not executable by you.

Let's just merge and see what happens. Tonight the tests will then be executed on our compute cluster, with all In-house) data.

PR #2495, opened 2025-04-01
ENH - allow for multiple inputs to be plotted as subplots

schoffelen:


robertoostenveld:
subplots should already be possible if you do it like this

```
for i=1:4
  cfg.figure = subplot(2,2,i)
  ft_topoplotER(cfg, timelock{i})
end
```

robertoostenveld:
ah, that use case makes sense. Should we consider subplots also for other plotting functions? For example, topoplotIC does it by default.

schoffelen:
Yes, I looked at ft_topoplotIC as well. It calls topoplot_common in a for-loop (where in each iteration a subplot is created). For symmetry sake, we could make the option to plot one topo per figure. this PR also intends to do the subplots for ft_singleplotTFR (and obviously ft_topoplotTFR). I briefly looked at ft_clusterplot, but that one is a historical nightmare (and does subplots by default)

schoffelen:
let's see how this flies

PR #2505, opened 2025-06-03
Fix unreachable code blocks for collecting qsub jobs

marcelzwiers:
The logfile and error file were specified with wildcards and checked for existence, without ever expanding the wildcards. Potentially (presumable), this is causing qsubcellfun to sometimes wait forever for jobs (a decade-old bug that rendered it pretty useless)

### DIFF
--- a/qsub/qsublist.m
+++ b/qsub/qsublist.m
@@ -194,7 +194,7 @@ switch cmd
           % there is no way polling the batch execution system
           retval = 1;
       end
-    elseif isempty(logout) && isempty(logerr) && isempty(pbsid)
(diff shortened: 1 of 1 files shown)


marcelzwiers:
I think that, even though it was correct, my patch uncovered an issue elsewhere in the code, because now none of the jobs seem to be collected anymore...

robertoostenveld:
If you two have both thought about it and tested it, I don't see a reason not to merge. So please go ahead.

marcelzwiers:
Unfortunately, it seems that qsubcellfun is still waiting forever (for a few percent of the jobs)

PR #2515, opened 2025-07-14
Figure name handling by open_figure

schoffelen:
This is a restart of #2167, which has been around for a while, and which has never been finished due to low priority.

### DIFF
--- a/ft_movieplotER.m
+++ b/ft_movieplotER.m
@@ -101,25 +101,6 @@ end
 tmpcfg = removefields(cfg, {'baseline', 'baselinetype'});
 tmpcfg = ft_movieplotTFR(tmpcfg, data);
 
-% this is needed for the figure title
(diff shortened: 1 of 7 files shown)


schoffelen:
Quick recap:

- Across the code base, many (but not all) functions that create figures, use open_figure
- open_figure has an figurename input argument, so it would make sense to add the title to the figure in open_figure
- currently still many functions add the figure name in the main function, rather than creating a cfg.figurename on the fly, and passing this into open_figure
- also, cfg.figurename handling is a bit obscure, it partly relies also on the definition of (undocumented) (cfg.)dataname (which is used to pass around names of variables between the different interactive plotting functions) 

This WIP PR is meant to make diverging code again a bit more consistent.

britta-wstnr:
Thanks @schoffelen ! 🤗

PR #2521, opened 2025-08-19
Fixes an offset in ft_electrodeplacement plotting

StolkArjen:
Hi Robert/JM,

It seems a bug has crept into ft_electrodeplacement, causing part of the button and controls to be pushed outside the plotting window, making them unreachable. I've recalibrated and adjusted this offset in the function.

best,
Arjen

Before fix:
Screenshot 2025-08-19 at 11 57 17 AM

After fix:
Screenshot 2025-08-19 at 11 56 14 AM


schoffelen:
OK, thanks.

PR #2529, opened 2025-08-29
Added reference data option to ft_denoise_ssp

chrisNatMEG:
Added an option to ft_denoise_ssp for computing the SSP projector from the principal components of a reference data (for example, an empty room recording). The number of principal components to project out of the data was added as a cfg option. Added an option to select the channels that can be used, e.g., for applying to different sensor types or axes separately. Default is to use all channels. Also added a line (188) removing the desired balance before applying the montage to grad if the updatesens is true. Otherwise it ends up with two ssp balances containing the same information.

### DIFF
--- a/ft_denoise_ssp.m
+++ b/ft_denoise_ssp.m
@@ -1,23 +1,33 @@
-function [data] = ft_denoise_ssp(cfg, data)
+function [data] = ft_denoise_ssp(cfg, varargin)
 
 % FT_DENOISE_SSP projects out topographies based on ambient noise on
 % Neuromag/Elekta/MEGIN systems. These topographies are estimated during maintenance
-% visits from the engineers of MEGIN
+% visits from the engineers of MEGIN.
+% Alternatively, computes projectors from reference data (e.g., empty room) if it 
+% is given as an additional input. For best results, make sure to preprocess
+% the reference data the same as the data to denoise. 
 %
 % Use as
-%   [data] = ft_denoise_ssp(cfg, data)
+%   [data] = ft_denoise_ssp(cfg, data) OR
+%   [data] = ft_denoise_ssp(cfg, data, refdata) 
 % where data should come from FT_PREPROCESSING and the configuration
 % should contain
 %   cfg.ssp        = 'all' or a cell array of SSP names to apply (default = 'all')
 %   cfg.trials     = 'all' or a selection given as a 1xN vector (default = 'all')
 %   cfg.updatesens = 'no' or 'yes' (default = 'yes')
+%   cfg.channel      = Nx1 cell-array with selection of channels from the refdata 
+%                      (default = 'all'), see FT_CHANNELSELECTION for details
 %
-% To facilitate data-handling and distributed computing you can use
-%   cfg.inputfile   =  ...
+% If refdata is specified the configuration should also contain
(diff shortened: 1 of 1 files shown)


ft_denoise_ssp.m
```diff
@@ -1,23 +1,33 @@
...
+%   cfg.numcomponent = number of principal components to project out of the data 
+%                      (default = 3)
+%
+% To facilitate data-handling and distributed cmputing you can use
+%   cfg.inputfile   =  ...  
 %   cfg.outputfile  =  ...
 % If you specify one of these (or both) the input data will be read from a *.mat
 % file on disk and/or the output data will be written to a *.mat file. These mat
-% files should contain only a single variable, corresponding with the
+% files should contain only a sinegle variable, corresponding with the
```
schoffelen, line 30: typo introduced

ft_denoise_ssp.m
```diff
@@ -78,16 +100,45 @@
...
+  ft_error('SSP without reference data can only be applied to neuromag data');
 end
 
 % select trials of interest
-tmpcfg = keepfields(cfg, {'trials', 'showcallinfo', 'trackcallinfo', 'trackusage', 'trackdatainfo', 'trackmeminfo', 'tracktimeinfo', 'checksize'});
+tmpcfg = keepfields(cfg, {'channel', 'trials', 'showcallinfo', 'trackcallinfo', 'trackusage', 'trackdatainfo', 'trackmeminfo', 'tracktimeinfo', 'checksize'});
 data   = ft_selectdata(tmpcfg, data);
+
+if ~isempty(refdata)
+    datachs = ft_channelselection('MEG',data.label); % pick MEG channels from data
```
schoffelen, line 112: is there a reason why this should be hard coded MEG?

ft_denoise_ssp.m
```diff
@@ -78,16 +100,45 @@
...
+    end
+    cfg.channel = datachs;  
+    tmpcfg = keepfields(cfg, {'channel', 'showcallinfo', 'trackcallinfo', 'trackusage', 'trackdatainfo', 'trackmeminfo', 'tracktimeinfo', 'checksize'});
+    refdata = ft_selectdata(tmpcfg, refdata);
+end
+
 % restore the provenance information
 [cfg, data] = rollback_provenance(cfg, data);
 
+if ~isempty(refdata)
```
schoffelen, line 127: should we consider to have ft_componentanalysis to do the heavy lifting here? One motivation could be that ft_componentanalysis is NaN aware, not sure about pca.

ft_denoise_ssp.m
robertoostenveld: you don't want trigger or EEG channels to be included, right?

chrisNatMEG: > you don't want trigger or EEG channels to be included, right?

That was the thought. We want to take all the MEG channels but nothing else. I considered making this an optional config input since I found that if you have different sensor types (or axes in OPMs with different noise profile) it can be better to treat them separately, e.g., apply SSP to X, Y and Z separately. But I came to the conclusion that it's better to do the channel selection for data before (where you could select '*bz') and then select all the MEG channels that are left from refdata. Not sure if this is the best way to do it.

PR #2532, opened 2025-09-10
update the representation of a sequence of montages or linear projections

robertoostenveld:
This replaces (or builds upon) #2529 in which @chrisNatMEG implemented an option to pass reference and compute an SSP projector inside ft_denoise_ssp, which is subsequently applied. Prior to the application, previous projections are undone (note to self: I am not even sure that is correct). The existing code for undoing the projections for ft_denoise_ssp and ft_denoise_synthetic turned out to be broken, since they assumed that only a single projector could have been applied. 

I redefined grad.balance.current to be a cell-array; previously it was a string that was supplemented with the grad.balance.previous cell-array. Having all montages/projections in a single cell array allows treating them more consistently in the code.

This change in the bookkeeping of the montages/projections required that the handling of "inverse" montages and especially the bookkeeping of applying montages to grad structures (or elec/topo) to be revised. Previously that could be done with the "inverse" and "balancename" options in ft_apply_montage. I disabled those options and made the inverse computation and the bookkeeping explicit everywhere.

### DIFF
--- a/test/test_bug3229.m
+++ b/test/test_bug3229.m
@@ -51,7 +51,7 @@ for i=1:5
   data.trial{i} = randn(4,1000);
   data.time{i}  = (1:1000)/1000;
 end
-data.elec = elec;
+data.elec = fixbalance(elec);
 data = ft_checkdata(data);
 
 %%


robertoostenveld:
I am now at the level that some of the reading/importing functions still need to be updated, like `neuromag2grad`. These are relevant, as they feed a grad structure into the FIeldTrip code without it going past `ft_datatype_grad` (which updates the old to the new structure).

There are still some challenges with `undobalancing`, which seems to be needed for `channelposition`, which in turn is called in `ft_datatype_grad`, causing (in some cases) an infinite recursion loop.

I think that I should first test this with the full regression batch, see where it breaks, and use those specific test scripts to further refine the transition from the old to the new representation of the `balance.current` field.

Issue #1323, opened 2020-03-03
Blackrock New File Format (3.0)


NickHalper:
Blackrock is expecting a change to our file format in an upcoming release (I think at least a couple of months away), so I wanted to reach out to you to make sure it is on your radar for incorporation into Fieldtrip. As always, we will allow people access to our previous file formats, and the adoption will be slow, so this would be supported in addition to the current ones that you load. 

The primary change is a move to 64 bit timestamps. The other change is the inclusion of an additional event type in our Neural Events File: an indicator for the recording pausing, resuming, and stopping. 

We can provide the file spec as soon as it is all signed off on our side, which should be in just a couple of weeks. 

Is this something that would be possible for you to incorporate? Besides example of files that uses this new format, is there anything else you would need from us?

schoffelen:
Hi Nick,
Thanks for reaching out to us, and for the notification. I think it would be great (if possible) to notify us whether (and if so, when) you will make file-readers available for matlab. Currently, the way FieldTrip is interfacing with the blackrock file format is through matlab functions from your NPMK toolbox (openNev, openNsx etc), which users are required to fetch from your github repository. If there are any changes to the functionality of those functions, given some proper documentation and some example data, it should be relatively straightforward to incorporate support for the new file format, and ensure backward compatibility.
Once the new file format is 'live' and matlab readers are supported, we can then broadcast this to the subset of Fieldtrip users that work with blackrock data, so that with our guidance, support for the fileformat can be realized.

robertoostenveld:
for reference, it would be good (in the future) to document the different versions of the file formats on http://www.fieldtriptoolbox.org/getting_started/blackrock/

It would also be good to implement a test script for the different blackrock file formats

```
mac036> cd matlab/fieldtrip/test
mac036> grep blackrock *.m
  < returns empty>
```

schoffelen:
Hi @NHalper,
 just to clarify, we don't provide a service in the sense that once a new fileformat becomes available, we directly incorporate it into the toolbox. We simply don't have the resources to do all of this ourselves. If our own scientific work does not directly benefit from it, we rely on other people to provide us with (low/intermediate/high)-level code in order to build in the functionality that is wished for. This means that either technically skilled end users need to help us out here (where we wouldn't know who's going to be up for the current challenge), or the commercial party that indirectly might benefit from the supported functionality in the toolbox. Therefore, I think it would be great to see how we can make this into a win-win situation, with the workload properly balanced.

[...]

PR #2534, opened 2025-09-17
updated support for blackrock_nsx

StolkArjen:
Our new blackrock system stores Channel IDs under _orig.ElectrodesInfo.ChannelID_  rather than _orig.ElectrodesInfo.ElectrodeID_, which ft_reader_header currently expects.

This PR adds support for this presumably new field while preserving backwards-compatibility.

Relatedly, ft_hastoolbox was referencing an old github repository location (also updated).

Best,
Arjen


robertoostenveld:
thanks, looks good!

schoffelen:
Just quickly, @StolkArjen at some point (about 5 years ago) the blackrock people reached out to us, regarding their 'new' fileformat, and the associated updated matlab readers. Perhaps this may be relevant to you, I have just reopened #1323 as a reminder. Feel free to close again if you think this is out of scope

StolkArjen:
In short, @schoffelen, I'm afraid I can't say much about how the different NMPK toolbox versions handle the older file formats. We've just started using a blackrock system, and from an initial look it appears that loading data with the latest NMPK on the path works smoothly; orders of magnitude faster than using the default version in /external. I'd be inclined to simply update the latter if I were confident that it wouldn't break backward compatibility, but that's not something I can systematically test here.

I've read through the issue you referenced, as well as related issues, and I'll keep an eye on those threads as we move forward with analysis and run into any problems.

StolkArjen:
@robertoostenveld: I'll see if we can collect a mini dataset in our next testing at the hospital, e.g., a 30-sec recording.

PR #2559, opened 2026-02-07
Add IAC connectivity method to ft_connectivityanalysis (cfg.method = iac)

kikiluvbrains:
This PR adds an optional instantaneous amplitude coupling (IAC)
connectivity method to ft_connectivityanalysis.

The implementation is fully self-contained and follows the methodology
described in Wei et al. (2021) without introducing additional assumptions
or modifications. IAC is motivated by amplitude envelope correlation:
when two signals exhibit high amplitudes at the same time, their
amplitude envelopes covary. A time-resolved (instantaneous) measure can be formed from the Hadamard product of the two amplitude-envelope time series, followed by averaging over time/trials to obtain a scalar coupling estimate.

The method is only used when cfg.method = 'iac' is explicitly selected.
Existing functionality and defaults are unaffected.

### DIFF
--- a/ft_connectivityanalysis.m
+++ b/ft_connectivityanalysis.m
@@ -950,10 +950,8 @@ switch cfg.method
       for j = 1:ntime
         % loop over all frequency bins
         for i = 1:nfreq
-          dat       = data.fourierspctrm(:,:,i,j).';
-          % powcorr_ortho will be computed per taper, tapers are assumed to
-          % be non time dependent with each other hence this should be
-          % allowed
+          dat = data.fourierspctrm(:,:,i,j).';
+          % powcorr_ortho will be computed per taper
           taper_order = repmat(1:tapers, 1, ceil(nrpttap/tapers));
           taper_order = taper_order(1:nrpttap);
           for t = 1:tapers


schoffelen:
Thanks for filing the PR. Yet, can you explain what the difference(s) are in your implementation in comparison to ft_connectivity_powcorr_ortho, which is based on Hipp's original work? Do we need 2 different implementations of the same metric?

schoffelen:
I see. I'd rather extend the existing code to support the time dimension, instead of having  a separate function for that. This is better for long term maintainability, documentation, and consistency.

schoffelen:
I don't why we would need a separate 'option' cfg.iac. The metric that you refer to is derived from Hipp's work, and I haven't seen it convincingly applied in the literature? Also I cannot find figure 16 in the reference you put in your comment.

In general, subtracting a mean of a complex variable from another complex variable, and then doing some downstream mathematical operations on those variables (e.g. conjugate multiplication, taking the imaginary part, or taking the magnitude): this subtraction is not a baseline subtractions, it's an ad hoc operation which should be mathematically well motivated, and ideally should be motivated in the context of comparing it with earlier described methods (e.g. Hipp) that don't do it. If not, I don't thing we should implement it in FieldTrip, because then there's no solid documentation/literature to point to. People invent connectivity metrics all the time, and not all of them make sense.

kikiluvbrains:
I see what you mean. While I’m waiting for a response and reviewing the appendix for more detail on the motivation behind the ad-hoc operations, I can continue integrating the existing code to support the time dimension and clean it up. I can also test it on some existing datasets of mine, as well as any fieldtrip native datasets you’d recommend.

kikiluvbrains:
I’ve pushed the current changes of the time dim inclusion to this PR. If you would prefer that I open a separate PR, that works for me as well. I’m very much on board with doing clean-up of powcorr_ortho. I was also thinking of addressing the ntap > 1 case. Could you let me know if there is already a working reference or implementation for this? I wasn’t able to locate one. For now I’m using a single taper per trial (Hanning), but I can also test with multiple tapers (like with DPSS). One idea would be to compute the connectivity metric separately for each taper and then combine the estimates at the end.

kikiluvbrains:
Ok, thank you. I mistakenly made my changes to an earlier version of that file.

PR #2563, opened 2026-02-17
FIX - preserve Matlab path order in qsubfeval > getcustompath

marcelzwiers:
This line in getcustompath erroneously reordered the Matlab path into alphabetical order:
```matlab
setdiff(p, f)
```
Proof of concept:
```matlab
p={'ccc','bb','a'};
f='bb';
>> setdiff(p, f)

ans =

  1×2 cell array

    {'a'}    {'ccc'}
```

There are two simple solutions:

1. Use 'stable' option that was introduced in R2012
```matlab
>> setdiff(p, f, 'stable')

ans =

  1×2 cell array

    {'ccc'}    {'a'}
```

2. Use:
```matlab
>> p(~strcmp(p, f))

ans =

  1×2 cell array

    {'ccc'}    {'a'}
```
I have chosen option 2, but either one is good to me...

### DIFF
--- a/qsub/private/getcustompath.m
+++ b/qsub/private/getcustompath.m
@@ -38,6 +38,7 @@ if ispc
 else
   p = tokenize(path, ':');
 end
+
 % remove the matlab specific directories
 if ispc
   s = false(size(p));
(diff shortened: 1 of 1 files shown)


robertoostenveld:
thanks, good catch!

PR #2582, opened 2026-04-13
Bvrf file support

### LINKED ISSUE #2565: We should add support to the BVRF format for BrainProducts EEG
https://www.brainproducts.com/support-resources/brainvision-recording-format/
https://github.com/brain-products/bvrf_reader

schoffelen:


robertoostenveld:
should `fileio/private/dataset2files.m`also be updated to deal with the triplet and make it more robust for the user specifying either one?

schoffelen:
good point, will do

PR #2584, opened 2026-04-22
enhancement - make sure chantype is lower case

vlitvak:
ft_apply montage uses isequal and that fails if case is different


robertoostenveld:
as far as I can tell, fieldtrip has implicitly always worked with lower case chantypes, so making that explicit should not break anything.

PR #2587, opened 2026-05-06
bugfix in read_neuralynx_ncs

### LINKED ISSUE #2583: Handling of NumValidSamp when marking invalid samples in read_neuralynx_ncs
**Describe the bug**
While using the neuralynx .ncs reader, I noticed a potential issue in how invalid samples are marked using NumValidSamp.

The current implementation contains a line (Line 220):
`Samp((NumValidSamp+1):end, k) = nan;`

Here, `NumValidSamp` is a vector defined by (Line 209)
`NumValidSamp = zeros(1,numrecord);`

but it is used without indexing by k.

In newer versions of Matlab (my version is 2024b), MATLAB issues a warning and falls back to using only the first element.
`Warning: Colon operands must be real scalars. This warning will become an error in a future release.`

This leads to unintended behavior.

**To Reproduce**
These two results differ when NumValidSamp varies across records.
```
NumValidSamp = [512 400 512];
Samp = zeros(512,3);

% Current behavior
Samp_bug = Samp;
for k = 1:3
    Samp_bug((NumValidSamp+1):end, k) = nan;
end

% Expected behavior
Samp_correct = Samp;
for k = 1:3
    Samp_correct((NumValidSamp(k)+1):end, k) = nan;
end
```

**Expected behavior**
Each record should use its own number of valid samples:
`Samp((NumValidSamp(k)+1):end, k) = nan;`

**Environment and versions:**
 - Operating System: Windows 11
 - MATLAB: R2024b
 - FieldTrip: most recent master branch from github

StolkArjen:
This PR fixes a bug where NumValidSamp was incorrectly used as a scalar index. Since NumValidSamp is a vector, it needs to be indexed before being used.

### DIFF
--- a/fileio/private/read_neuralynx_ncs.m
+++ b/fileio/private/read_neuralynx_ncs.m
@@ -217,7 +217,7 @@ if begrecord>=1 && endrecord>=begrecord
       NumValidSamp(k) = fread(fid,   1, 'int32');
       Samp(:,k)       = fread(fid, 512, 'int16');
       % mark the invalid samples
-      Samp((NumValidSamp+1):end,k) = nan;
+      Samp((NumValidSamp(k)+1):end,k) = nan;
     end
     
   end % if isMexv6


robertoostenveld:
can you share a (hopefully small) test file with us? It would not surprise me if the file format has changed over the years and what once was a scalar is now a vector.

StolkArjen:
That's possible. I've been loading the same file for years without issue, but that might be because it was using the neuralynx toolbox instead of this native matlab code. A recent matlab update might have rendered the mex-file from this toolbox ineffective, prompting read_neuralynx_ncs to use the code between lines 198-220. In other words, this bug went unnoticed because that section was likely rarely used by people having an install of the neuralynx toolbox. The good news is that the current PR addresses this particular issue.

Zooming in; when numrecord > 1, NumValidSamp(k) grows in size, meaning it's no longer a scalar index and therefore should _always_ yield an error (highlights are my PR: **(k)**): 

    for k=1:numrecord
      % read a single continuous data record
      TimeStamp(k)    = fread(fid,   1, 'uint64=>uint64');
      ChanNumber(k)   = fread(fid,   1, 'int32');
      SampFreq(k)     = fread(fid,   1, 'int32');
      NumValidSamp(k) = fread(fid,   1, 'int32');
      Samp(:,k)       = fread(fid, 512, 'int16');
      % mark the invalid samples
      Samp((NumValidSamp**(k)**+1):end,k) = nan;
    end

I don't know how often num_recorded > 1. It happened for two of my datasets (out of a dozen), which are about a decade old (too large and potentially too sensitive to share). num_recorded is defined upstream as: (endrecord-begrecord+1); and was 458 for one of my dataset producing the issue. Moreover, each element of it had the same value (512). If that behavior is expected to be true for all datasets, then an alternative solution would be to not let NumValidSamp grow in size with each iteration of k (but I'm not sure about that and it might be safer to let it grow and index it).

robertoostenveld:
the previous GH action bot comment suggested to look into test_bug1770 and test_bug1998, which I did.

In `test_bug1770` and MATLAB R2025b on my mac I also run into the error that you report. However, our nightly regression test does not detect the bug (otherwise we would know). When I manually start R2024b on a linux compute node,  `test_bug1770` does _not_ give an error, but a warning:

> Warning: Colon operands must be real scalars. This warning will become an error in a future release.  

Apparently the time for it to become an error came with 2025b (or perhaps already with 2025a). So the error in the code was already there previously, but was still dealt with by MATLAB. 

 Looking at the code, your fix seems to be correct for all situations, also for the mex files. Each record in the ncs file has a NumValidSamp, so it is always to be treated as a vector. Thanks for finding this bug, which probably dates back to 2005 and which remained innocent until recently.

schoffelen:
Hier spuit elf: We should keep an eye out for this, because I recall this change of warning-to-error behavior for this particular also in the context of e.g. the low-level ctf reading functions (in the context of a previous toolkit, where user used matlab2025b)...

StolkArjen:
"Apparently the time for it to become an error came with 2025b (or perhaps already with 2025a). So the error in the code was already there previously, but was still dealt with by MATLAB."

Ah, that makes sense, and consistent with this, I had also updated to 2025b (from 2023b). Thanks for catching that link.

"Hier spuit elf: We should keep an eye out for this, because I recall this change of warning-to-error behavior for this particular also in the context of e.g. the low-level ctf reading functions (in the context of a previous toolkit, where user used matlab2025b)..."

Good point.

PR #2592, opened 2026-06-01
enh: ft_electrodeplacement compatibility with newer matlab versions

StolkArjen:
Past versions of ft_electrodeplacement would use html-coded font color to indicate whether an electrode had been localized or not. However, newer versions (presumably 9.5 and up) no longer allow per-item coloring in listboxes. To make behavior more robust across matlab versions, this PR keeps the original html-based display where supported, but falls back to plain text status markers (`○` / `●`) when HTML rendering is unreliable.

I also made additional small GUI layout adjustments to improve the placement of the listbox.

Original: 

Bug:
Screenshot 2026-05-31 at 10 28 38 PM

Fixed using status markers:
Screenshot 2026-05-31 at 10 27 09 PM


ft_electrodeplacement.m
```diff
@@ -227,11 +227,15 @@
 chanstring = {};
 markerlab = {}; 
 markerpos = {};
+use_html = verLessThan('matlab','25.1');  % version 2025a and newer no longer support html
```
schoffelen: can I ask to change this into: 

use_html = ~ft_platform_supports('matlabversion', '2025a', inf); 

for code consistency w.r.t. the rest of the code base?

schoffelen (changes requested):

ft_electrodeplacement.m
StolkArjen: Done -- thanks for the suggestion. The code runs fine on both platforms (2024b left, 2026a right):

Screenshot 2026-06-18 at 10 50 23 AM

PR #2593, opened 2026-06-03
Exact TFCE (eTFCE) in tfcestat: remove the nsteps discretisation

devon7y:
## Summary

This evaluates the TFCE integral in `private/tfcestat.m` **exactly** instead of approximating it with a fixed number of height thresholds (`cfg.tfce_nsteps`). The cluster extent `e(h)` is piecewise constant in the height `h` (it changes only when `h` crosses a data value), so the integral

```
TFCE(v) = ∫_{h0}^{h_v} e(h)^E · h^H dh
```

has a closed form on each interval and is obtained in a single pass with a disjoint-set (union-find) cluster-retrieval forest, over the **same spatial connectivity that `findcluster` uses**. This removes the discretisation (and its bias), is typically much faster than the threshold loop, and removes the SPM dependency (it no longer calls `findcluster`/`spm_bwlabel`).

This was suggested as a follow-up to an exact-TFCE PR for the LIMO toolbox.

### Attribution

The exact-TFCE (eTFCE) algorithm is **not my work** — it is the method of:

> Chen, X., Weeda, W. D., Nichols, T. E., & Goeman, J. J. (2026). *eTFCE: Exact Threshold-Free Cluster Enhancement via Fast Cluster Retrieval.* arXiv:2603.03004 — https://arxiv.org/abs/2603.03004

This PR is only a MATLAB implementation of that published method for FieldTrip. The function header credits it.

## What changed

Only `private/tfcestat.m`. The signature, `cfg` fields (`tfce_H`, `tfce_E`, `tfce_h0`, `tail`, `inside`, `dim`, `connectivity`) and the positive/negative handling are unchanged; `cfg.height` is still computed and returned so the cfg round-trip across randomisations is unchanged. `cfg.tfce_nsteps` is accepted but no longer used (there is no step size in the exact method). Both the sensor layout (channel × freq × time with `connmat`) and the `spacereshapeable` source layout (3-D grid) are handled.

## Validation

- **Exactness** — against an independent `graph`/`conncomp` integrator: ~1e-15 for the sensor case (tails −1/0/+1) and the source 3-D grid case.
- **Connectivity diffed against the real `findcluster`** — the per-voxel cluster **extent** from FieldTrip's actual `findcluster` equals the extent from this implementation's edge list at **every** threshold tested (max|err| = 0), for both sensor and source layouts.
- **Convergence** — the existing discretised `tfcestat` (calling the real `findcluster`) converges to this exact result as `nsteps`→∞ (O(1/nsteps)), confirming eTFCE is the exact limit of the current method.

## Notes / things to decide

- **Removes the SPM/`findcluster` dependency** for TFCE (the exact method builds its own adjacency from `connmat` + the non-spatial grid, matching `findcluster`).
- eTFCE computes the **mathematically-correct connected components**, equal to `combineClusters2`. The default mex `combineClusters` has a documented isolated-pixel quirk (it "occasionally labels isolated pixels as separate clusters"); on those rare cases results may differ — arguably a fix.
- Pure MATLAB; for very large source grids the union-find loop is interpreted, but it replaces `nsteps` full `findcluster` passes with a single sweep.
- It's a `private/` function, so happy to discuss the approach. If preferred, the original discretised path can be retained behind a `cfg` option rather than replaced.

### DIFF
--- a/private/tfcestat.m
+++ b/private/tfcestat.m
@@ -1,12 +1,31 @@
 function [stattfce, cfg] = tfcestat(cfg, statobs)
 
-% TFCESTAT computes threshold-free cluster statistic multidimensional channel-freq-time or
-% volumetric source data
+% TFCESTAT computes threshold-free cluster statistic for multidimensional
+% channel-freq-time or volumetric source data.
+%
+% This implementation evaluates the TFCE integral EXACTLY (eTFCE) rather than
+% approximating it with a fixed number of discrete height thresholds. Because
+% the cluster extent e(h) is piecewise constant in the height h (it changes
+% only when h crosses a data value), the integral
+%
+%     TFCE(v) = integral_{h0}^{h_v} e(h)^E * h^H dh
+%
+% has a closed form on each interval and is obtained in a single pass with a
+% disjoint-set (union-find) cluster-retrieval forest, using the same spatial
+% connectivity as FINDCLUSTER. This removes the cfg.tfce_nsteps discretisation
+% (and its bias), is typically much faster than the threshold loop, and needs
+% no SPM/Image-Processing clustering routine.
+%
+% The exact-TFCE (eTFCE) algorithm is the method of Chen, Weeda, Nichols &
+% Goeman (2026), "eTFCE: Exact Threshold-Free Cluster Enhancement via Fast
+% Cluster Retrieval", arXiv:2603.03004; this is a MATLAB implementation of it
+% for FieldTrip and was not developed by the FieldTrip team.
 %
 % See also CLUSTERSTAT, FINDCLUSTER
 
-% Copyright (C) 2021, Jan-Mathijs Schoffelen
-%  
+% Copyright (C) 2021, Jan-Mathijs Schoffelen (original discretised version)
+% Copyright (C) 2026, exact (eTFCE) reimplementation
+%
 % This file is part of FieldTrip, see http://www.fieldtriptoolbox.org
 % for the documentation and details.
 %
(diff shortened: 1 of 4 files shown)


private/tfcestat.m
```diff
@@ -1,12 +1,31 @@
...
+% Goeman (2026), "eTFCE: Exact Threshold-Free Cluster Enhancement via Fast
+% Cluster Retrieval", arXiv:2603.03004; this is a MATLAB implementation of it
+% for FieldTrip and was not developed by the FieldTrip team.
 %
 % See also CLUSTERSTAT, FINDCLUSTER
 
-% Copyright (C) 2021, Jan-Mathijs Schoffelen
-%  
+% Copyright (C) 2021, Jan-Mathijs Schoffelen (original discretised version)
+% Copyright (C) 2026, exact (eTFCE) reimplementation
```
robertoostenveld: please add your name, not what was changed

robertoostenveld:
Can you please provide a test script (with real or simulated data)? It should be starting with a fixed seed for the random number generator, and then computes the TFCE distribution for the various options. The test should then evaluate the outcome and give an error if it does not match the expected outcome (which can be experimentally determined by running it once). It should be a high-level test script, calling `ft_sourcestatistics` or `ft_freqstatistics` and not the private function.

schoffelen:
@devon7y that was quick. Also: for backward compatbility purposes, I would appreciate that the old functionality is also still available, i.e. the user can choose between the 'old' and 'new' implementations. This also facilitates the testing.

Testcode may be implemented along the lines of the code in fieldtrip/test/test_example_tfce<somedate>.m

private/tfcestat.m
devon7y: Done in b2c5591 — the copyright/attribution lines now credit the eTFCE authors (Xu Chen, Wouter D. Weeda, Thomas E. Nichols and Jelle J. Goeman, arXiv:2603.03004) as the creators of the method, and list me as the FieldTrip implementer, making clear I did not develop it.

devon7y:
Thanks both for the quick review — pushed in b2c5591.

**Backward compatibility (@schoffelen):** both implementations are kept and selected with `cfg.tfce_method`:
- `'exact'` (default) — the exact eTFCE evaluation
- `'discrete'` — the original `cfg.tfce_nsteps` height-threshold loop via `findcluster`, restored verbatim from master

Default `'discrete'` behaviour is unchanged, with one fix: the discrete path previously read `ft_getopt(cfg, 'nsteps', 100)` — a field the caller never sets — so the documented `cfg.tfce_nsteps` was silently ignored and 100 steps were always used. It now honours `cfg.tfce_nsteps` (default still 100). Happy to flip the default to `'discrete'` if you'd prefer the most conservative option for backward compatibility.

**Test (@robertoostenveld):** added `test/test_pull2593.m`, a high-level test through `ft_freqstatistics` (not the private function) on simulated data with a fixed random seed. It runs `'exact'` and `'discrete'` (100 and 2000 steps) and errors unless:
- the exact result matches the many-step discrete limit, and is closer to it than the 100-step default (exact removes the discretisation bias);
- the observed TFCE reproduces stored reference values;
- the simulated effect is detected; and
- an unsupported `cfg.tfce_method` raises an error.

Locally (R2025a) the exact-vs-2000-step relative max difference is 6.7e-4 and exact-vs-100-step is 8.5e-3, and the test passes.

**Attribution (@robertoostenveld):** the copyright/attribution lines now credit the eTFCE authors (Chen, Weeda, Nichols & Goeman, arXiv:2603.03004) as the creators of the method, with me listed as the FieldTrip implementer and a note that I did not develop it.

schoffelen:
OK, this looks quite good. I will mege this soon. The test runs through fine. To be done is the adjustment of the example page (upload a new figure, provide reference to paper + explicitly mention the exact method).

Issue #2594, opened 2026-06-17
ft_prepare_sourcemodel fails if symmetry field set


jdien07:
**Describe the bug**
The ft_prepare_sourcemodel has an optional cfg field named .symmetry that according to its header:
cfg.symmetry        = 'x', 'y' or 'z' symmetry for two dipoles, can be empty (default = [])
If one does so, setting it to 'x', then it produces the error:

Error using istrue (line 36)
cannot determine whether "x" should be interpreted as true or false

Error in ft_prepare_sourcemodel (line 469)
    if isempty(cfg.symmetry) || ~istrue(cfg.symmetry)
                                 ^^^^^^^^^^^^^^^^^^^^
**To Reproduce**
cfg.symmetry= 'x';
ft_prepare_sourcemodel(cfg);

**Expected behavior**
A source model is produced

**Screenshots**
n/a

**Environment and versions (please complete the following information):**
 - Operating System: Pop!_OS 24.04
 - MATLAB version R2026aR2
 - FieldTrip version fieldtrip-20260601

**Additional context**

The bug is pretty straightforward.  It should not be using the istrue function for the symmetry field, since the expected values are 'x', 'y', and 'z', not 'true' or 'false' or the like.

It appears to have been introduced in the 9/25/25 commit.  My advice is to simply delete the istrue statement as the intent was apparently to check whether any symmetry was being specified and the only allowable value is empty if the specification is no symmetry, which was the original code before this commit.

schoffelen:
confirmed. Thanks for the catch and for the report. I'll fix it.

PR #2595, opened 2026-06-17
Symmetry

schoffelen:
see #2594


schoffelen:
Hmm, this PR accidentally has another changed file which apparently I failed to push, for the record I mention it here, but life's a bit too short to disentangle all this. I'll leave it in.

PR #2597, opened 2026-06-25
ENH - use neuromag logic for creation of coordsystem.json fieldline data

schoffelen:
Im an working on data2bids conversion of fieldline data, where those data have been 'postprocessed' by the HEDScan software, and thus coregistered to HPI-coils. Thus the fif-file contains digitization points + is defined in neuromag coordinates. This PR adds some logic so that data2bids tries to automagically create the <>_coordsystem.json for FieldLine fif-data.


schoffelen:
NOTE: as discussed, it may be that the current MEG BIDS standard is insufficient for OPM stuff, e.g. w.r.t. coordinate systems etc.

Issue #1379, opened 2020-04-18
fileio\private\ft_getopt MEX file much slower than .m file


NirOfir:
Hi,
I'm using fieldtrip version 07/04/2020 on Windows 10, i7 2.5 GHz processor with 8 GB RAM.
I used ft_preprocessing to rereference and demean EEG data from the hard drive, using this code:
```
cfg = [];
cfg.dataset = FILE_PATH;
cfg.channel = 'EEG';
cfg.trl = trl{:, 1:3};
cfg.demean = 'yes';
cfg.reref = 'yes';
cfg.refchannel = 'all';
data = ft_preprocessing(cfg);
```
This took about 245 seconds for 420 trials, each on average 3 seconds long, with 64 channels and 512 Hz sampling rate. I thought this was slower than expected so I ran the same code (after using clear all to make sure there is nothing that will affect run time) with the profiler and got this result:
![image](https://user-images.githubusercontent.com/43925388/79632037-69818f80-8165-11ea-81b1-f9160ee1d048.png)
The function that takes most time (~85%) is the MEX of ft_getopt, so I tried removing the MEX file from fileio/private to force fieldtrip to use the m file. This run only took 9 seconds (about 27 times faster), and these are the profiler results:
![image](https://user-images.githubusercontent.com/43925388/79632161-283daf80-8166-11ea-822b-03f54a98d454.png)
Note that the m version of fileio/private/ft_getopt takes less than 0.3 seconds in total here. There is another, different, ft_getopt MEX at the end of this part of the list which sits in utilities, but this one is zippy.

I tried the same thing on a much stronger computer (100 GB RAM, 16 core processor), running centos 7 and got basically the same results. 

I also tried loading the entire dataset to memory and rereferencing all of it as one trial, and this took less than 15 seconds, so I'm pretty sure the bottleneck is the ft_getopt MEX. 

For now I've deleted the MEX files from my fieldtrip version, but I wonder if there's something in the code that can be fixed to make it faster, or that the MEX file should be removed? Are there situations in which the MEX provides a speedup?

Thanks!

robertoostenveld:
that is surprising! The idea of the mex files is that (in most cases) they should be faster rather than slower compared to the original m-file implementation.

I wonder whether this is specific to your combination of OS and MATLAB version. 

Could you try this:

```
>> addpath your_fieldtrip_path/test
>> tic; for i=1:280; test_ft_getopt; end; toc
[...]

PR #2602, opened 2026-07-20
ENH - further improvement for speed

schoffelen:


schoffelen:
This also relates to the discussion in #1379, which has resurfaced while I was working on the eyelink asc support.
Long story short: ft_getopt.c cannot do without making a deep copy of the variables, which causes a substantial slowdown of large data objects (nested structures with 'cached' data) are passed through the function. Mathworks does not provide an easy, robust, documented and version-agnostic workaround for this. Therefore, for the current dataformat I decided to work around this, by caching the data in the low level reading function. 
Another (generic) solution would be to discard the mex-file of ft_getopt from fileio/private in case this slowdown issue is encountered, as per the reported (but by now quite old) bug report. This could be documented as a FAQ (albeit it is not really a frequent question it seems).  Given that it now is not an issue (anymore) for *.asc files, and given that I also have found another solution for epoched *.fif files, the only other known dataformat suffering would be gtec_mat. Is it worth the effort of writing a FAQ for this?

robertoostenveld:
I think that speedups for specific data formats should be documented in the manufacturer-specific getting started. If it were documented there for multiple formats, we could have a generic faq on "how to speed up reading of data" that then points to the manufacturer specific ones, and that holds some general recommendations (like using an SSD, and not reading the data over a network connection).

schoffelen:
OK, I'll merge it here then, because the linked issue does not relate to this filetype anymore.

Issue #2605, opened 2026-08-19
ft_defaults calls toggle external/signal on and off, leaving butter() undefined


korbinian90:
**Summary**

If Matlab is started without the Signal Processing toolbox, ft provides the shim via `external/signal`.
This fails when the license available, but not installed. Then `ft_defaults`'s re-initialisation flips the state on every other call.
This causes issues in SPM, PR on ft to follow.

**Describe the bug**

`ft_defaults` decides whether to put `external/signal` on the path using `ft_hastoolbox('signal')`. That check is satisfied by FieldTrip's own fallback implementations, so once the directory is on the path the check reports the MathWorks Signal Processing Toolbox as present and the `elseif` branch removes the directory again. Repeated `ft_defaults` calls therefore toggle it on and off, and on the off cycles `butter`, `filtfilt` and `fir1` are undefined.

In `ft_hastoolbox`:

```matlab
case 'SIGNAL'
  dependency = {has_license('signal_toolbox'), 'window', 'hanning'};
```

and `external/signal` ships both `window.m` and `hanning.m`.

This shows up on any machine that is *licensed* for the Signal Processing Toolbox but does not have it *installed*, because `has_license` calls `license('checkout', ...)`, which tests the entitlement rather than the installation. That is the standard configuration on hosted CI runners: MathWorks-hosted GitHub Actions grant the license while `mpm` installs only the products you ask for.

`signal` is the only fallback affected. `STATS` probes `betacdf`, `raylcdf` and `unidcdf` while `external/stats` ships only `betacdf`; `IMAGES` probes `imerode` and `imdilate` while `external/images` ships neither. Both need all their sentinels, so neither self-satisfies.

**To Reproduce**
On a machine licensed for but without the Signal Processing Toolbox (or simulate it by removing `matlabroot/toolbox/signal` from the path, which leaves the license intact):

```matlab
extsig = fullfile(fileparts(which('ft_defaults')), 'external', 'signal');
for k = 1:4
  clear ft_defaults
  ft_defaults;
  fprintf('call %d: on path=%d  butter="%s"\n', k, contains(path, extsig), which('butter'));
end
ft_preproc_highpassfilter(randn(2,1000), 200, 0.5, 6, 'but', 'twopass');
```

**Expected behavior**
`external/signal` stays on the path across repeated calls, and `butter` remains available.

## Actual behaviour

```
call 1: on path=1  butter=.../external/signal/butter.m
call 2: on path=0  butter=""
[...]

PR #2606, opened 2026-08-19
FIX - make ft_hastoolbox('signal') test robust

korbinian90:
## Summary

Issue #2605 explains the problem

The solution was to make the check for availability of the signal toolbox more robust by checking for functions that are not supplied in `external/signal`.

Change ft_hastoolbox('signal') to test for functions that are not provided in external/signal. It tested for window() and hanning(), both of which external/signal provides itself. Once ft_defaults had added that directory the check reported the MathWorks toolbox as present, so the next re-initialisation removed it again and butter, filtfilt and fir1 became undefined. has_license tests the entitlement rather than the installation, so the license half of the check passed even where the product was absent, which is the norm on hosted CI runners.

The same problem might also occur for other toolboxes and the same solution could apply

## Checklist

- [x] I added unit tests and integration tests for each feature (if applicable).
- [x] No error nor warning in the console.
- [ ] I joined a screenshot of the app before and after the fix.

### DIFF
--- a/test/test_ft_defaults_signal.m
+++ b/test/test_ft_defaults_signal.m
@@ -1,0 +1,32 @@
+function test_ft_defaults_signal
+
+% MEM 1gb
+% WALLTIME 00:03:00
+% DEPENDENCY ft_hastoolbox ft_defaults
+% DATA no
+
+% external/signal must stay on the path when the Signal Processing Toolbox is
+% licensed but not installed, across repeated re-initialisation of ft_defaults
+
+origpath = path;
+cleanup  = onCleanup(@() path(origpath)); %#ok<NASGU>
+
+% simulate a toolbox that is licensed but not installed, as on CI runners
+sigroot = fullfile(matlabroot, 'toolbox', 'signal');
+p = strsplit(path, pathsep);
+p = p(~strncmpi(p, sigroot, numel(sigroot)));
+path(strjoin(p, pathsep));
+
+external_signal = fullfile(fileparts(which('ft_defaults')), 'external', 'signal');
+
+for k = 1:4
+  clear ft_defaults
+  ft_defaults;
+
+  if ~contains(path, external_signal)
+    error('external/signal is not on the path after %d ft_defaults call(s)', k);
+  end
+  if isempty(which('butter'))
+    error('butter is undefined after %d ft_defaults call(s)', k);
+  end
+end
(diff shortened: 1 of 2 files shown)


korbinian90:
Codespell fail is unrelated:
Error: plotting/private/outline_shape.m:21: eary ==> early, easy, eerie

robertoostenveld:
I fail to understand the proposed fix. In a random FT function, ft_hastoolbox('signal') should return true if either the mathworks signal toolbox is present (license+installed) and on the path, or the drop-in replacement is present on the path. By replacing the list of dependencies by some that are in the mathworks signal toolbox but not in `external/signal`, it would always return false, even if `external/signal` is on the path. 

I think the solution confuses the detection itself with the auto-add functionality.

Initially I see these options: the license is present yes/no, the matlab signal directory is installed and on the path yes/no. If you imagine that as a 2x2 matrix, there are 4 options. The option license=yes, directory=no was so far not considered, and that is a bug.  Adding the `external/signal` factor with levels yes/no increases it to an 2x2x2 possibility space with in total 8 options, where some of them are not meaningful (like having both matlab/signal and external/signal on the path).

robertoostenveld:
Looking at your test (and executing it by copying and pasting), I see that you are testing ft_defaults, not ft_hastoolbox.

robertoostenveld:
With the following 
```
function test_ft_default_signal
check_signal()
remove_matlab_signal()
check_signal()
add_matlab_signal()
check_signal()
end

function check_signal()
clear ft_hastoolbox
if ~ft_hastoolbox('signal')
  fprintf('signal is not present\n')
else
  fprintf('signal is present\n')
end
end

function remove_matlab_signal()
sigroot = genpath(fullfile(matlabroot, 'toolbox', 'signal'));
s = strsplit(sigroot, pathsep);
p = strsplit(path, pathsep);
p = setdiff(p, s);
path(strjoin(p, pathsep));
end

function add_matlab_signal()
sigroot = genpath(fullfile(matlabroot, 'toolbox', 'signal'));
s = strsplit(sigroot, pathsep);
p = strsplit(path, pathsep);
p = [p, s];
path(strjoin(p, pathsep));
end
```
I get 
```
>> test_ft_default_signal
signal is present
signal is not present
signal is present
```
which is what I expect and consider to be correct.

also when I use the following in the main function, I get what I expect:

```
check_signal()
remove_matlab_signal()
check_signal()
add_matlab_signal()
check_signal()
remove_matlab_signal()
ft_hastoolbox('signal', 1)
check_signal()
```

robertoostenveld:
ah, I now understand the relevance of comparing SIGNAL to STATS and IMAGES: they should all work the same. And once one of them has been auto-added, it is the internal caching (i.e. the persistent variable) that deals with subsequent calls.

And I realize that my specific use of your test function failed to reveal the problem on my computer, because I had (through my `startup.m`) the global `ft_default.toolbox.signal = 'matlab'` and your script assumes that it is set to `compat`

PR #2610, opened 2026-09-01
ft_connectivity_psi: exclude the last frequency bin from the slope window

### LINKED ISSUE #2609: ft_connectivity_psi: the last nbin+1 frequency bins include the raw top-bin coherency, and are -Inf with normalize='yes'
**Description of the problem**

In `ft_connectivity_psi`'s `phaseslope` subfunction:

    x(1:end-1,:,:,:,:) = conj(x(1:end-1,:,:,:,:)).*x(2:end,:,:,:,:);
    ...
    y(k,:,:,:,:) = imag(nansum(x(begindx:endindx,:,:,:,:),1));   % window k-n .. k+n

`x(1:end-1)` becomes the products conj(C(f))·C(f+δf), but `x(end)` keeps the raw coherency at the highest frequency. Every window that reaches the top bin — the last `nbin+1` output bins — therefore adds `imag(C(fmax))` to the phase slope index. With a 20-bin synthetic coherency and `nbin=3`, bins 17–20 differ from the Nolte et al. (2008) definition by exactly `imag(C(20))`; bins 1–16 match to machine precision.

With `normalize='yes'` it is worse: `coh(end)` is never assigned (stays 0), so `x(end)./coh(end)` is infinite and the last `nbin+1` bins of the output are ±Inf:

    normalize=yes: PSI at bins 16..20 = 0.733345 -Inf -Inf -Inf -Inf

The code is unchanged since the file's first commit (July 2010).

**Suggested fix**

Zero the last element after forming the products (and set its normalizer to 1), so windows reaching the top bin simply contain one fewer product — which is the behaviour of the definition at the edge. PR attached; verified to make the edge bins agree with the reference definition and be finite when normalized.

**Environment:** FieldTrip master @ 2e14f72, Octave 8.4. Script: https://github.com/cindykrafft/research-software-audit/blob/main/audits/fieldtrip/verify/ft11_psi_edge.m

---
_Generated by [Claude Code](https://claude.ai/code)_

cindykrafft:
Fixes #2609 .

`phaseslope` overwrites `x(1:end-1)` with the products conj(C(f))·C(f+δf) but leaves `x(end)` as the raw coherency at the highest frequency, so the last `nbin+1` output bins add `imag(C(fmax))` to the phase slope index; with `normalize='yes'` `coh(end)` stays 0 and those bins are ±Inf.

This zeroes the last element after forming the products and sets its normalizer to 1. Verified (20-bin synthetic coherency, `nbin=3`): the edge bins now match the Nolte et al. (2008) definition to machine precision, and the normalized output is finite everywhere; bins below the edge window are unchanged.

https://claude.ai/code/session_012d2sJAD5EUp4GqAStqoBX7


cindykrafft:
Tests added and run.

**New test:** `test/test_pull2610.m` (`DATA no`) — compares `ft_connectivity_psi` against an independent implementation of the Nolte et al. (2008) definition at every frequency bin, including the bins whose window reaches the top of the range, and checks that the normalized variant is finite and matches its reference. Passes on this branch; fails on master (max abs diff 0.69 at the edge bins).

**Existing test:** `test_ft_connectivity_psi` initially failed on this branch, and the reason is instructive: its input has dimord `rpt_chan_chan`, i.e. no frequency axis, so the "slope across frequency" is computed over a single bin. On master the only nonzero output for that input was the raw imaginary coherency of the single bin — exactly the contamination this PR removes — so on this branch a one-bin PSI is 0 for any `nbin` and the assertion "different `nbin` give different results" cannot hold. I gave the test a 10-bin frequency axis (`rpt_chan_chan_freq`); the updated test passes on both master and this branch, and with a real frequency axis master and branch give identical results away from the edge bins (checked: max |PSI| 0.8652 on both).

`test_ft_connectivityanalysis` cannot run here on either master or the branch: `ft_specest_mtmfft` needs the two-output `dpss` from the MATLAB Signal Processing Toolbox, which neither Octave's signal package nor `external/signal/dpss_hack` provides. If someone with MATLAB can run it on the branch, that would be welcome; the only code changed is the `phaseslope` subfunction of `ft_connectivity_psi`.

---
_Generated by [Claude Code](https://claude.ai/code)_

schoffelen:
Waldorf: Well, Statler, they've finally discovered that the last frequency bin isn't connected to anything.
Statler: Took them long enough. I haven't seen a boundary condition this lonely since the edge of a MATLAB array.

Waldorf: To be fair, the fix in ft_connectivity_psi is quite tidy. The products conj(C(f)).*C(f+df) only exist through fmax-1, so zeroing the final element makes the representation match what the PSI calculation actually means.

Statler: And it fixes the particularly charming consequence of leaving the raw coherency in there: the last nbin+1 bins get an unsolicited contribution from imag(C(fmax)).

Waldorf: Which is bad enough.

Statler: But with normalization enabled, coh(end) remains zero, so we get ±Inf.

Waldorf: Ah yes. Nothing says “phase slope index” quite like infinity at the edge of your spectrum.

Statler: At least they've added a regression test.

Waldorf: A fairly good one, too. It constructs a synthetic cross-spectral density, computes an independent reference implementation of the Nolte et al. definition, and checks every frequency bin—including the troublesome edge bins. It also checks that the normalized result remains finite.

Statler: And, importantly, it fails on master with a substantial edge discrepancy while passing on the branch.

Waldorf: That's the sort of regression test I like. It doesn't merely assert that the new code produces some number; it asserts that the number agrees with an independently constructed definition.

Statler: Don't get sentimental. There is still the small matter of test_ft_connectivity_psi.

Waldorf: Ah, yes. The original test was feeding PSI a rpt_chan_chan input—one frequency bin—and expecting different nbin values to produce different results.

Statler: Which is rather like testing a slope calculator with one point and complaining that the slope doesn't change when you widen the window.

Waldorf: The PR correctly changes that test to rpt_chan_chan_freq with ten frequency bins. That's not merely making the test pass; it's fixing an assumption in the test that was inconsistent with what PSI actually computes.

Statler: Although I'd like to see the rationale made a little more explicit in the test itself. Someone maintaining this five years from now might otherwise wonder why a seemingly innocent dimension was suddenly added.

Waldorf: The comment does explain it, so perhaps we're being unnecessarily difficult.

Statler: That's our job.

Waldorf: There is one other thing. The new regression test reproduces the existing +1 normalization behaviour in its independent reference implementation.

Statler: Yes. That's reasonable for this PR, since the stated change is the edge-bin handling rather than redefining normalization.

Waldorf: But the test comment essentially says, “the +1 is inherited from the original implementation.” That is a useful little archaeological warning.

Statler: Which means nobody should mistake this test for proof that the normalization formula itself is mathematically canonical.

Waldorf: Quite. It proves that the implementation agrees with the reference as constructed, and specifically that the erroneous final raw coherency contribution has disappeared.

Statler: And the test report says the broader test_ft_connectivityanalysis can't be run in the current environment because of the dpss dependency.

Waldorf: That's acceptable, provided somebody with MATLAB gives the integration test a spin. The PR author has explicitly called that out rather than pretending the unavailable test passed.

Statler: Refreshingly honest. Suspicious, even.

Waldorf: So, verdict?

Statler: The bug is real, the fix is small and appropriately scoped, and the new regression test actually exercises the pathological edge case.

Waldorf: In other words—

Statler: I hate to say it—

Waldorf: —I'd approve.

Statler: Don't worry. I'm sure the next frequency bin will be along shortly to ruin everything.

schoffelen:
TBH I would be less concerned about the edge frequency bin misbehaving than about the fact that I don't understand the normalize option in the function. Why is there a +1, and why is the 'coh' variable only computed after x(1:end-1 etc has already been replaced by its conjugate product...?

schoffelen:
as an added note: when ft_connectivity_psi is called through ft_connectivityanalysis, the normalize flag is by default 'no'. Also, regarding the remark above about 'dpss' -> this can be avoided with 'hanning' tapering.

cindykrafft:
Statler: Well, Waldorf, the maintainer says he doesn't care about the edge bin.
Waldorf: He cares about the normalize option. Sensible man. He wants to know what the +1 is for, and why coh is computed after x has already been turned into products.

Statler: So they went and ported phaseslope line for line and ran it.
Waldorf: Of course they did. Script and output are in the audit repository, `audits/fieldtrip/verify/ft11c_psi_normalize_meaning.py`.

Statler: And the verdict?
Waldorf: Because coh is built after x holds the products, the denominator isn't |C(f)|·|C(f+1)| at all. It's |C(f)|·|C(f+1)|²·|C(f+2)| + 1.
Statler: Four coherency magnitudes and a one. That's not a normalization, that's a raffle.

Waldorf: In practice the option multiplies the unnormalized PSI by 1/(s⁴+1) at coherence magnitude s. At s = 0.05 that's 1.000. At 0.5 it's 0.941. At 0.95 it's 0.551.
Statler: So at low coherence it does nothing, and at high coherence it does roughly half of something.
Waldorf: And nobody can say what the something is. It isn't a phase-only slope, which would need the raw magnitudes and no +1. And it isn't Nolte's normalized PSI, which is Ψ/std(Ψ) by jackknife.
Statler: Which the function already has the ingredients for, in its own `v` output!
Waldorf: The +1 looks like a guard against dividing by zero that was never revisited. The FIXME next to it has been asking the same question since 2010.

Statler: So what do they propose?
Waldorf: Restraint, for once. `ft_connectivityanalysis` never sets the option, so the default path is unaffected, and this PR stays scoped to the edge bin. What to do with normalize is the maintainers' call: document it as it is, redefine it as Ψ/sqrt(v), or drop it. They'll prepare whichever as a separate PR.
Statler: A contributor who doesn't widen their own pull request. I need to sit down.
Waldorf: You are sitting down.

Statler: Anything else?
Waldorf: Two small things. The rationale for the frequency axis is now written inside test_ft_connectivity_psi.m itself, commit 9d99bac. And with `cfgf.taper = 'hanning'` in place of `cfgf.tapsmofrq = 2`, as suggested, test_ft_connectivityanalysis passes on this branch and on master under Octave 8.4.
Statler: Under Octave? How?
Waldorf: Three trivial shims for MATLAB functions Octave lacks. Pass-throughs. Not part of the PR.
Statler: So the integration test passes, the unit tests pass, and the only thing left is a fifteen-year-old plus one.
Waldorf: Same as always.
Statler: Same as always.

---
_Generated by [Claude Code](https://claude.ai/code)_

schoffelen:
OK, touché.

Although I still have the feeling that the whole exercise is addressing a bit of a non-issue in the grand scheme of things of, you/Claude were indeed correct in identifying the computation issue related to the edge-frequency bin. Yet, I needed to point him to the issue related to the denominator, for which no solution was provided. I have committed some more changes into ft_connectivity_psi in order to correct the denominator for normalization in the way it might have been originally intended. Also, I extended the number of tests in test_ft_connectivity_psi. I also suggest to remote the test_pull2610, because this one is a bit circular (and currently will fail due to the fact that I changed the computation of the normalizing denominator). 

Also, related to the flagged issue of the dpss, my suggestion would be to create an issue out of this, because it indicates that people who don't have the signal processing toolbox licence, and for whom the dpss_hack was intended may choke in ft_specest_mtmfft. One potential fix here would be to add a second output argument in the dpss_hack route. How to address this in Octave, I wouldn't know.

cindykrafft:
Thanks for taking the normalization in hand.

- `test_pull2610` is removed (9af1fc4), as suggested.
- Your extended `test_ft_connectivity_psi` passes on the branch head under Octave 8.4 (with a stand-in for `sum(..., 'omitnan')`, which this Octave lacks), and `test_ft_connectivityanalysis` with hanning tapers passes on it as well.
- The dpss point is filed as #2614  with a reproduction. `ft_specest_mtmfft` asks `double_dpss` for two outputs and `external/signal/dpss_hack/dpss.m` returns one, so every `taper='dpss'` call through the hack stops with "Too many output arguments" (Octave: "dpss: function called with too many outputs"). `ft_specest_mtmconvol` and `ft_specest_irasa` ask for one output and are unaffected. The issue proposes returning the concentrations as a second output of the hack, precomputed alongside the tapers; since the hack is the only dpss route Octave has, that would cover Octave too.

---
_Generated by [Claude Code](https://claude.ai/code)_

schoffelen:
OK, we're almost ready to merge this one I think. I was thinking a bit more about the normalisation option, and I wonder whether the original intention was actually to incoporporate some kind of 'bandwidth'-parameter normalisation. This could explain the '+1'. If it's indeed a bandwidth correction term, then I would think that the sum should be taken for the denominator and numerator separately, rather than having the sum across the ratio. What do @cindykrafft and Claude think about this?

cindykrafft:
I think the reading holds together, and it does make a difference which way the sums go. With the denominator summed separately, the `+1` per product adds up to the number of products in the window, so the option becomes

    Ψ_norm(f) = Im Σ_W conj(C)·C' / ( |W| + Σ_W |C||C'| )

which is bandwidth-free, whereas the per-term ratio still grows with `nbin` like the raw PSI. Measured on two channels with a 10-sample delay (`ft11d_psi_normalize_bandwidth.py` in my Mytochondria repository), value at the centre bin, ratio of `nbin=8` to `nbin=2`:

| definition | high coherence (|C| ≈ 0.98) | moderate (|C| ≈ 0.5) |
|---|---|---|
| raw sum (normalize='no') | 4.0 | 2.0 |
| per-term ratio (current branch) | 4.0 | 2.1 |
| ratio of sums with the +1 (your reading) | 1.0 | 0.5 |
| ratio of sums without the +1 | 1.0 | 0.6 |
| raw sum / number of products | 1.0 | 0.5 |

So separate sums are the right move if the option is meant to remove the bandwidth dependence. What the `+1` then contributes is only a change of scale: at full coherence the denominator is `2|W|` instead of `|W|`, so the result is half of the magnitude-weighted mean sine of the phase step; at low coherence `Σ|C||C'|` is small and the result is close to the raw sum divided by the number of products. Two cleaner forms give the same invariance with an interpretation attached:

- without the `+1`: `Im Σ conj(C)C' / Σ |C||C'|`, the magnitude-weighted mean sine of the phase step, bounded in [−1, 1];
- raw sum divided by the number of products: the mean imaginary product, i.e. the PSI per bin.

I have no evidence for what the 2010 author meant, so this is a design choice rather than a bug report. Happy to implement whichever you pick on this branch, or leave the option as it is on the branch and do it as a follow-up so this PR can merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #2614, opened 2026-09-05
ft_specest_mtmfft: taper='dpss' fails through external/signal/dpss_hack (one-output dpss), so the no-toolbox route and Octave cannot use multitapers


cindykrafft:
**Description of the problem**

`external/signal/dpss_hack` is the route for users without the Signal Processing Toolbox (its README: made for the 2013 Toronto workshop). `ft_specest_mtmfft` asks the taper function for two outputs:

    [tap, w] = double_dpss(ndatsample, ndatsample*(tapsmofrq./fsample));   % ft_specest_mtmfft.m:172
    ...
    function [tap, w] = double_dpss(a, b, varargin)                         % :429
    [tap, w] = dpss(double(a), double(b), varargin{:});

and `dpss_hack/dpss.m` returns one (`function dps_seq = dpss(seq_length, time_halfbandwidth)`), so every `taper='dpss'` call through the hack stops before any spectrum is computed:

    addpath(fullfile(fileparts(which('ft_defaults')), 'external', 'signal', 'dpss_hack'));
    dat = randn(2, 1000);
    ft_specest_mtmfft(dat, (1:1000)/1000, 'taper', 'dpss', 'tapsmofrq', 2, 'output', 'pow')

    MATLAB (no Signal Processing Toolbox):  Too many output arguments.
    Octave 8.4:                             error: dpss: function called with too many outputs

The same call with `'taper', 'hanning'` runs. `ft_specest_mtmconvol` (:196, :441) and `ft_specest_irasa` (:185, :195) ask `dpss` for one output and are not affected, so through `ft_freqanalysis` it is `method='mtmfft'` with the default `taper='dpss'` that fails, which is what the tutorials use. Octave has no `dpss` at all (the `signal` package does not provide it), so the hack is the only multitaper route there.

The second output is used by `adaptspec_dpss` on every dpss call (`:306`, `:358`), not only for `weightopt='eig'/'adapt'`: with the default `'mean'` it still receives `w` and the `se` output is derived from it.

**Suggested fix**

Return the concentrations from the hack as well. They depend on the time-bandwidth product and the taper index, not on the interpolation to `seq_length`, so they can be precomputed next to the tapers:

    % precompute_dpss.m
    [w{i}, e{i}] = dpss(n, n*s(i)/1000);
    % dpss.m
    function [dps_seq, e] = dpss(seq_length, time_halfbandwidth)
    ...
    e = e{i};

[...]

PR #2622, opened 2026-09-23
Fix/octave compat survey

### LINKED ISSUE #2621: `external/stats` is no longer added to the path since 55ee593: `nanmean` etc. undefined without the Statistics Toolbox (and under Octave); `test_external_stats` fails on master
Since 55ee593 ("no need to exclude mfiles for stats, those are nowadays in stats/private", 2026-09-15) `ft_platform_supports('stats')` raises

```
error: 'exclude_mfiles' undefined near line 137, column 57
error: called from
    ft_platform_supports at line 137 column 10
```

because the commit removed the definition of `exclude_mfiles` but kept its use:

```matlab
  case 'stats'
    root_dir = fileparts(which('ft_defaults'));
    if ~isempty(root_dir)
      external_stats_dir = fullfile(root_dir, 'external', 'stats');
      tf = has_all_functions_in_dir(external_stats_dir, exclude_mfiles);   % <- undefined
```

`ft_defaults` calls it inside a `try` block (the one that decides whether to add `external/stats`), so the error is swallowed and the block ends before `addpath(external_stats)`. The consequence is that `external/stats` is never added to the path. With the MathWorks Statistics Toolbox installed nothing is visible, since the toolbox provides `nanmean` and friends anyway; without it (MATLAB without the toolbox, or GNU Octave, whose statistics package no longer implements `nanmean`) every function that uses `nanmean`, `nanstd`, `nansum`, `nanvar`, `nanmax` or `nanmin` fails with `'nanmean' undefined`. `test_external_stats` exercises exactly this (`restoredefaultpath`, `ft_defaults`, then `exist('nanmean')`) and fails on master:

```
>> test_external_stats
error: assert (exist (filelist {k}, 'file') == 2 || exist (filelist {k}, 'file') == 3) failed
```

To see it directly, on master with no Statistics Toolbox (or in Octave):

```matlab
restoredefaultpath; addpath('/path/to/fieldtrip'); ft_defaults
ft_platform_supports('stats')     % error: 'exclude_mfiles' undefined
which nanmean                     % empty
nanmean([1 NaN 3])                % 'nanmean' undefined
```

Expected: `ft_platform_supports('stats')` returns true, `external/stats` is on the path when the toolbox is not, and `nanmean([1 NaN 3])` is 2, as before 55ee593.

The fix is to pass an empty exclusion list, as the `images` and `signal` cases already do:

```matlab
      tf = has_all_functions_in_dir(external_stats_dir, {});
```

In the Octave survey you asked for on #2614 (every `DATA no` test function on Octave 8.4, results there shortly), this one line accounts for 81 of the 280 failures. PR follows, together with two small Octave-only fixes from the same survey (`compat/octave/startsWith.m`, `endsWith.m` and `contains.m` with a cell array of patterns; `ft_fetch_data` with `addParameter`).

---
_Generated by [Claude Code](https://claude.ai/code)_

cindykrafft:
Fixes #2621.

Three changes found by running every `DATA no` test function under GNU Octave 8.4 (the survey from #2614; results posted there):

1. **`ft_platform_supports('stats')` errors since 55ee593.** The commit removed the definition of `exclude_mfiles` but kept its use in the `'stats'` case, so the function raises `'exclude_mfiles' undefined`. `ft_defaults` calls it inside a `try`, the error is swallowed, and `external/stats` is never added to the path: without the Statistics Toolbox (and under Octave) `nanmean`, `nanstd`, `nansum` and the rest are undefined, and `test_external_stats` fails on master. Pass an empty exclusion list, as the `images` and `signal` cases do. This alone accounts for 81 of the 280 Octave failures.

2. **`compat/octave/startsWith.m`, `endsWith.m` and `contains.m` with a cell array of patterns.** They passed the pattern straight to `strncmp`/`strfind`, which under Octave errors with `nonconformant cell arrays` / `cell arguments must have matching sizes` when the pattern cell has a different size from the string cell, and returns one value per pattern instead of one per string when the sizes happen to agree. `ft_senstype` (`startsWith(sens.label, {'L', 'R'}) & endsWith(sens.label, {'bx', 'by', 'bz'})`), `ft_checkdata` and others call them this way. Loop over the patterns and OR the results, which is what MATLAB's functions do; a char or a cell pattern, the `'IgnoreCase'` option and a char or cell first argument all keep their previous behaviour. This accounts for 77 of the 280 failures.

3. **`ft_fetch_data` declares its six name-value options with `addOptional`.** Octave's `inputParser` assigns `addOptional` arguments positionally, so `'header'` lands in `header`, the header struct in `begsample`, and the string `'endsample'` reaches `istrue(allowoverlap)`, which stops with "cannot determine whether "endsample" should be interpreted as true or false". `addParameter` is the declaration for name-value options in MATLAB and Octave alike; `utilities/ft_fetch_data.m` and its identical copy `fileio/private/ft_fetch_data.m` are changed together. This accounts for 10 failures that only show once 1 and 2 are fixed.

Tested under Octave 8.4.0 on master cfdad9b: with the branch, `ft_platform_supports('stats')` is true, `external/stats` is on the path, `nanmean([1 NaN 3])` is 2, `startsWith({'abc','xbc','zz'}, {'x','ab'})` is `[1 1 0]`, and of the 168 `DATA no` tests that failed for these three causes, 95 now pass and the other 73 stop at an environmental cause (a MEX file not compiled for Octave, a MATLAB-only function, graphics; per-test list in the #2614 comment). Passing tests went from 191 to 286 of 471. Under MATLAB the first change restores the behaviour before 55ee593, the third is a no-op (name-value pairs parse the same way), and the second is not on the path.


cindykrafft:
The three suggested tests that need no data, run on this branch under GNU Octave 8.4 (no MATLAB here):

| test | master | this branch |
|---|---|---|
| `test_bug2269` | fails in `compat/octave/startsWith.m` (cell pattern) | passes |
| `test_ft_fetch_data` | fails, `NaN: dimensions must be scalars` in `test/private/ft_fetch_data.m` | passes |
| `test_issue1292` | same | passes |

The last two failed for the reason the PR fixes: `test/private/ft_fetch_data.m` is a third copy of `ft_fetch_data` with the same `addOptional` declarations, so the header struct landed in `begsample` and `nan()` got a struct as a dimension. I pushed 99b4c3c, which applies the `addParameter` change to that copy as well; the three copies are identical again. The DCCN tests (`test_bug3379`, `test_pull810`, `test_bug1667`) need private data, so I could not run them.

The failing "Check for spelling errors" check is not from this PR: with the repository's `.codespellrc`, the only hit on `plotting utilities` is `plotting/private/outline_shape.m:21`, the variable `eary` (the y-coordinates of the ear outline next to `earx`), and it is the same on current master. If you want the check green, one line in `.codespellrc` does it; I have left it out of this PR since it is unrelated:

```
ignore-words-list = indx,...,datas,earx,eary
```

